PR #9, opened 2019-01-12
Implement only running either authenticator or proxy

### LINKED ISSUE #6: Allow running authenticator and proxy separately as standalone processes
I found this while following the buzzfeed SSO. Looking at it, the code looks clean and seems to cleanup a lot of the configuration messiness from Buzzfeed and bit.ly's SSO. Most importantly for my organization, it adds Okta SSO support.

For my organization, the double OAuth flow is pretty important to prevent having to configure separate applications within Okta for every single internal app we create. The code seems architected for this, but it looks like you've hard coded in cmd/main.go that you instantiate both an Authenticator and Proxy object every time, and that all environment settings must be valid for both simultaneously. There doesn't seem to be some way using environment variables to disable one or the other.

pomerium needs to either split the command line into 2 apps (I know you are intentionally trying to merge them though), add flags/subcommands to only enable one, or provide environment variable settings/combinations to selectively disable one or the other.

mbarrien:
This PR adds the implementation of the SERVICES environment variable to only enable authentication or proxy or both.

I got this running successfully in a test Kubernetes cluster.

Open issues:
* I was forced to put in a "/ping" so that Kubernetes/load balancing liveness checks would work in authenticator mode even when the virtual hostname isn't provided. This is over and above the /ping in each of the authenticator and proxy. Perhaps there's a cleaner way of organizing the code.

### DIFF
--- a/cmd/pomerium/main.go
+++ b/cmd/pomerium/main.go
@@ -35,33 +35,51 @@ func main() {
 		os.Exit(0)
 	}
 	log.Debug().Str("version", version.FullVersion()).Str("user-agent", version.UserAgent()).Msg("cmd/pomerium")
-	authOpts, err := authenticate.OptionsFromEnvConfig()
-	if err != nil {
-		log.Fatal().Err(err).Msg("cmd/pomerium : failed to parse authenticator settings")
-	}
-	emailValidator := func(p *authenticate.Authenticator) error {
-		p.Validator = options.NewEmailValidator(authOpts.AllowedDomains)
-		return nil
+
+	var authenticator *authenticate.Authenticator
+	var authHost string
+	if mainOpts.Services == "all" || mainOpts.Services == "authenticator" {
+		authOpts, err := authenticate.OptionsFromEnvConfig()
+		if err != nil {
+			log.Fatal().Err(err).Msg("cmd/pomerium : failed to parse authenticator settings")
+		}
+		emailValidator := func(p *authenticate.Authenticator) error {
+			p.Validator = options.NewEmailValidator(authOpts.AllowedDomains)
+			return nil
+		}
+
+		authenticator, err = authenticate.NewAuthenticator(authOpts, emailValidator)
+		if err != nil {
+			log.Fatal().Err(err).Msg("cmd/pomerium : failed to create authenticator")
+		}
+		authHost = authOpts.RedirectURL.Host
 	}
 
-	authenticator, err := authenticate.NewAuthenticator(authOpts, emailValidator)
-	if err != nil {
-		log.Fatal().Err(err).Msg("cmd/pomerium : failed to create authenticator")
-	}
+	var p *proxy.Proxy
+	if mainOpts.Services == "all" || mainOpts.Services == "proxy" {
+		proxyOpts, err := proxy.OptionsFromEnvConfig()
+		if err != nil {
+			log.Fatal().Err(err).Msg("cmd/pomerium : failed to parse proxy settings")
+		}
 
-	proxyOpts, err := proxy.OptionsFromEnvConfig()
-	if err != nil {
-		log.Fatal().Err(err).Msg("cmd/pomerium : failed to parse proxy settings")
-	}
-
-	p, err := proxy.NewProxy(proxyOpts)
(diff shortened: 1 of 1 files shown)


desimone:
Hi @mbarrien, thank you for the pull request fixing the docker image and for implementing the service mode flag. Happy to hear you were able to get it working on a Kubernetes cluster!

**RE: Service modes.**

I thought about the service mode feature more, and I agree that supporting it as an environmental var is a cleaner choice. Sorry to have sent you down the wrong direction. It's a little tricky because right now environmental variables are parsed by and are specific to the `authentication`/`proxy` packages.  Higher-level option parsing would be required. Perhaps in  `cmd/pomerium/main` we could have something like:

```go
// Options are the global environmental flags used to set up pomerium's services.
type Options struct {
    // Debug runs all services in debug mode. Logger is set to standard out,
    // formatter is human/console friendly, and logger is set to verbose.
    Debug bool `envconfig:"DEBUG"`
    // Services is a slice of strings representing the service mode.
    // Available options are "all","authetnicate","proxy".
    Services string `envconfig:"SERVICES"`
}

var defaultOptions = &Options{
    Debug:    true,
    Services: string,
}
```

I can take a stab at that in another pull request later today. 

 **RE: service mode selection.** 
Right now, I agree it doesn't make a whole lot of sense to allow multiple modes since there are only two! And on further reflection, even when additional modules being added (authorization, trust inference), in all likelihood an end user will either be running pomerium in "all in one" mode (probably testing it out/debugging) or each component separately. 

**RE: ping endpoint.**

I've been meaning to take a look at health-checks endpoints because there are several throughout pomerium right now. It makes sense to have a single top-level endpoint as you've done here, regardless of what combination of services has been selected. I briefly looked at https://github.com/InVisionApp/go-health for ideas, but would like to dive a little deeper into what an ideal health-check endpoint would look like. I think your top-level ping endpoint is great for now. I'm going to create an issue for what an ideal health-check endpoint would include. 

**RE: Docker fix**

LGTM!


It probably makes sense to break this into multiple PRs.  I'll be following up with a service mode PR and would welcome your feedback.  What do you think?

Thanks again for the PR!

desimone:
I did a little more experimenting with the `/ping` endpoint, and I was able to hit all `ping` even when hitting the IP directly (granted, without TLS verification).   Could you tell me more about the issue you had with K8s registering the health-check endpoint? 
For example, all these endpoints worked for me.

```bash
# directly
$ http --verify=no https://localhost/ping
HTTP/1.1 200 OK
Content-Length: 2
Content-Type: text/plain; charset=utf-8
Date: Tue, 15 Jan 2019 07:52:10 GMT

OK

# authenticate endpoint
$ http https://sso-auth.corp.beyondperimeter.com/ping
HTTP/1.1 200 OK
Content-Length: 2
Content-Security-Policy: default-src 'none'; style-src 'self' 'sha256-pSTVzZsFAqd2U3QYu+BoBDtuJWaPM/+qMy/dBRrhb5Y='; img-src 'self';
Content-Type: text/plain; charset=utf-8
Date: Tue, 15 Jan 2019 07:54:31 GMT
Referrer-Policy: Same-origin
Strict-Transport-Security: max-age=31536000
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-Xss-Protection: 1; mode=block

OK

# proxied endpoint
$  http https://http.corp.beyondperimeter.com/ping
HTTP/1.1 200 OK
Content-Length: 2
Content-Type: text/plain; charset=utf-8
Date: Tue, 15 Jan 2019 07:54:48 GMT

OK

```

Where you getting a TLS error or was the connection being refused outright?

mbarrien:
/ping failed with authenticator mode in that the pod never registered as healthy. As soon as I add the override without a hostname, it works. Note that K8s is either using localhost or 127.0.0.1 for the hostname or even possibly a K8s internal name.

mbarrien:
This PR has now been reduced to just the implementation reading from the SERVICES environment variable; the other changes all work properly, and I got this new version also working in a test environment using the helm package from #10 and the dockerfile fix from #14

desimone:
LGTM. Thank you @mbarrien .

PR #14, opened 2019-01-16
Fix dockerfile, add CMD

mbarrien:
Change the Docker container to copy the pomerium binary instead of the base binaries from the Golang base container. Also makes the pomerium binary be run as the default command. This splits this functionality off from #9


desimone:
Thank you!

PR #21, opened 2019-01-22
authenticate: skip provider login screen

mbarrien:
This PR adds a SKIP_PROVIDER_BUTTON environment variable that skips the Pomerium sign-in page and redirects the user directly to the OAuth provider (Google, Okta, etc) if the user is not yet authenticated with that provider. It replicates the functionality of "-skip-provider-button" flag from https://github.com/bitly/oauth2_proxy. It also acts as an implementation of https://github.com/buzzfeed/sso/issues/132. My organization does use this feature in our current oauth2_proxy deployments, so it would be nice to have it here.

Internally, the implementation is hacky; if the flag is set while in SignIn(), it will call a separate helper whose only job is to extract the auth redirect URL from the authenticator instance, which proceeds to validate it as if it were user input. (The normal code path would extract the auth redirect URL from the URL query string that was provided on the sign in page, which we've also broken out into a helper function.) If there's a better way to do this, I'm open to it. It's also missing unit tests.

### DIFF
--- a/authenticate/authenticate.go
+++ b/authenticate/authenticate.go
@@ -48,11 +48,12 @@ type Options struct {
 	SessionLifetimeTTL time.Duration `envconfig:"SESSION_LIFETIME_TTL"`
 
 	// Authentication provider configuration vars
-	ClientID     string   `envconfig:"IDP_CLIENT_ID"`     // IdP ClientID
-	ClientSecret string   `envconfig:"IDP_CLIENT_SECRET"` // IdP Secret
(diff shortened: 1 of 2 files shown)


desimone:
Thank you @mbarrien . I am beginning to feel like this should be the **default** behavior.  What are your thoughts?

> It's also missing unit tests.

A man after my own heart 😄. No worries.  Most of the handlers need some unit-testing. I was hoping to carve out a few days this week devoted exclusively to improving coverage. I'll backfill with another PR.

PR #26, opened 2019-01-23
Gitlab auth provider

### LINKED ISSUE #20: authenticate/providers : GitLab support
GitLab can serve as an OAuth provider, it would be nice if we could use it together with Pomerium. GitLab.com as well as self-hosted GitLab installations should be supported.

References
- [GitLab OAuth Provider Documentation](https://docs.gitlab.com/11.6/ce/integration/oauth_provider.html)
- [oauth2_proxy GitLab Provider](https://github.com/bitly/oauth2_proxy/blob/v2.2/providers/gitlab.go)

mbarrien:
This adds a completely untested Gitlab provider. I don't have a Gitlab.com account or self hosted Gitlab deployment myself, so haven't tested. This closes #20.

Based off of documentation, examining the bit.ly implementation, and looking at https://gitlab.com/.well-known/openid-configuration, I believe Gitlab to be a normal OIDC provider, with the one quirk that it uses a scope called `read_user` instead of `profile` and `email`. The code here should provide that. For a self-hosted Gitlab, you should be able to set IDP_PROVIDER_URL to your own deployment's URL.

This would be configured using IDP_PROVIDER=gitlab in the environment, along with the standard IDP_CLIENT_ID and IDP_CLIENT_SECRET.

Can someone with Gitlab test?

### DIFF
--- a/authenticate/providers/gitlab.go
+++ b/authenticate/providers/gitlab.go
@@ -1,0 +1,36 @@
+package providers // import "github.com/pomerium/pomerium/internal/providers"
+
+import (
+	"context"
+
+	oidc "github.com/pomerium/go-oidc"
+	"golang.org/x/oauth2"
+)
+
+const defaultGitlabProviderURL = "https://gitlab.com"
+
+// GitlabProvider is an implementation of the Provider interface
+type GitlabProvider struct {
+	*ProviderData
+}
+
+// NewGitlabProvider creates a new instance of a Gitlab provider.
+func NewGitlabProvider(p *ProviderData) (*GitlabProvider, error) {
+	ctx := context.Background()
+	if p.ProviderURL == "" {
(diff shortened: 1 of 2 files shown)


desimone:
@mbarrien thank you for the pull request! I think we had the same idea we have almost identical implementations but I was working on docs as I went. I've temporarily added gitlab [to the docs](https://www.pomerium.io/guide/identity-providers.html#gitlab).

Unfortunately I wasn't able to get it working. Were you able to get yours against the hosted Gitlab? 

@fnkr would you be able to test this or #28 ?

fnkr:
Can't get Pomerium to compile right now.

```
../../authenticate/authenticate.go:14:2: code in directory /root/go/src/github.com/pomerium/pomerium/authenticate/providers expects import "github.com/pomerium/pomerium/internal/providers"
```

PR #37, opened 2019-01-31
cmd/pomerium : restore /ping without hostname

mbarrien:
Re-adding the hostless /ping for the authenticate service. Without these 4 lines, Pomerium fails to stay running when using the helm chart that's checked into the repo; the Kubernetes liveness/readiness probes don't see the pod as successfully running, and Kubernetes goes into an endless restart loop.

I have tried the following:
* Adding `httpHeaders: {name: "Host", value: "insert_redirect_uri_hostname_here"}` to both probes to try to force it to recognize. Pomerium fails to stay running.
* Removing the probes from the deployment definition. Pomerium successfully stays running and responds correctly. Thus it's certain that it's one of the probes failing.
* Modifying the code here to log the request.host golang reports. It reports "10.x.x.x:443". (Actual pod internal IP address for the 10.x.x.x). That won't match the authHost.

 
**Checklist**:
- [ ] documentation updated
- [ ] unit tests added
- [ ] ready for review

### DIFF
--- a/cmd/pomerium/main.go
+++ b/cmd/pomerium/main.go
@@ -68,6 +68,11 @@ func main() {
 
 	topMux := http.NewServeMux()
 	if authenticateService != nil {
+		// Need to handle ping without host lookup for LB
+		topMux.HandleFunc("/ping", func(rw http.ResponseWriter, _ *http.Request) {
+			rw.WriteHeader(http.StatusOK)
+			fmt.Fprintf(rw, "OK")
+		})
 		topMux.Handle(authHost+"/", authenticateService.Handler())
 	}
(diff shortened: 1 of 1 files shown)


desimone:
Apologize for the regression. 

I think you are absolutely on the right track about as to why the request.host is not picking up the ping handler.  It's really curious I wasn't able to duplicate with GKE following [the kubernetes quick start guide](https://www.pomerium.io/guide/kubernetes.html) but it's probably because I used hostnames instead of ips? Another small ping handler is no big deal in the meantime.  

Thank you for the fix @mbarrien!

PR #39, opened 2019-02-03
authenticate: use gRPC for service endpoints

### LINKED ISSUE #4: proxy/authenticator : replace REST with gRPC
Placeholder.

desimone:
**authenticate: use gRPC for service calls**

Service calls that were previously handled by HTTP now use gRPC. Maintaining consistency between server and client implementations which was becoming a pain point. gRPC ensures client and server stay in lock-step.

Adopting gRPC now was a look forward in hopes that gRPC it will ease maintenance of further changes going forward for authenticate and other more complicated services.  gRPC gives us many features for “free” like performance, request cancelling, bi-directional connectivity, and streaming. 

To support internally routed services, two new two new env_vars were introduced `OverideCertificateName` and  `AuthenticateInternalURL`.

**Other changes**

- authenticate: set cookie secure as default.
- authenticate: remove single flight provider.
- authenticate/providers: Rename “ProviderData” to “IdentityProvider” 
- authenticate/providers: Fixed an issue where scopes were not being overwritten 
- proxy/authenticate : http client code removed.
- proxy: standardized session variable names between services.
- docs: change basic docker-config to be an “all-in-one” example with no nginx load.
- docs:  nginx balanced docker compose example with intra-ingress settings.
- license:  attribution for adaptation of goji’s middleware pattern. 

**Known issues**
- Pomerium serves multiple protocols (gRPC & HTTPS) on the same port (443), this provides a challenge  as most load balancers don't support external gRPC well. Internal communication is still fine, and probably preferable. However, the issue is that our outward facing certificate (`*.corp.beyondperimeter.com`) no longer matches the internally route name of the service causing a TLS error. For example, `AUTHENTICATE_INTERNAL_URL=pomerium-authenticate-service.pomerium.svc.cluster.local:443` and `OVERIDE_CERTIFICATE_NAME="*.corp.beyondperimeter.com`.
- Alternatively, if you are using nginx ingress you can use , the `nginx.ingress.kubernetes.io/ssl-passthrough` annotation to enable passthrough TLS connections.  
- Shared-token based mutual authentication is still used. TLS is still established and verified by the client using the system’s certificate storage mechanism. In the future we may want to support mTLS with custom CAs. 

References
- Closes #4 
- https://github.com/kolide/fleet/issues/1881
- [Argo project with same ingress](https://github.com/argoproj/argo-cd/blob/master/docs/ingress.md)

**Checklist**:
- [x] documentation updated
- [x] unit tests added
- [x] tested from source
- [x] tested docker-compose
- [x] tested on [GKE] with nginx-ingress
- [x] add example config for [GKE] with nginx-ingress
- [x] tested on [GKE] with gce-ingress
- [x] add example config for [GKE] with gce-ingress
- [x] tested on AWS using ALB
- [x] ready for review

[GKE]: https://cloud.google.com/kubernetes-engine/docs/


desimone:
At first I thought having HTTPS/gRPC being served on the same port would be a nice convenience but I'm finding this method breaks gRPC when behind many load balancers.

@mbarrien are you guys terminating ssl connections at ingress? Would requiring [ssl-passthrough](https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/annotations.md#ssl-passthrough) break your use-case?

mbarrien:
Our current setup has SSL terminating at an AWS ALB acting as an ingress.

desimone:
This was a bigger PR than I had anticipated. There are some potentially breaking changes for those behind a load balancer or ingress that does not support inbound gRPC. However, I found those issue mitigated by setting the new configuration options which enable internal gRPC routing.

I'll let this sit for a bit, and would welcome any feedback.

/cc @mbarrien

PR #47, opened 2019-02-14
docs: remove port from internal service url

bweston92:
As you can see below an error occurs when you have the port. There is another option for port which screws it up.

```
189a15d0e56","error":"rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp: address pomerium-authenticate-service.pomerium.svc.cluster.local:443:443: too many colons in address\"","time":"2019-02-14T16:27:07Z","message":"proxy: error redeeming authorization code"}
```

 
**Checklist**:
- [ ] documentation updated
- [ ] unit tests added
- [ ] ready for review


desimone:
Good catch. Thanks for the fix!

PR #49, opened 2019-02-17
proxy: support certificate authority to verify server

mbarrien:
This PR adds support for a certificate authority for the proxy to allow connections to the authenticate server when using your own certificate authority.

An unfortunate result of the conversion to using gRPC on the backend is that since AWS Application Load Balancer's [only talk HTTP/1.1](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html#listener-configuration]) from the LB to the application (they can terminate HTTP/2.0 from the client), gRPC is unable to traverse through the load balancer. This forces us to use an internal authenticator URL to talk from the proxy to the authenticator. However, this also means that we can no longer rely on the publicly signed certificates provided by AWS's certificate manager that the ALB uses. Since my organization doesn't have a central certificate authority set up, we were using some self signed certificates.

Assigning a self signed certificate to the authenticate server results in the following error in the proxy logs when the proxy tries to hit the authenticate gRPC endpoint:
```
{"level":"error","fwd_ip":"x.x.x.x","ip":"10.x.x.x","user_agent":"...","req_id":"...","error":"rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: authentication handshake failed: x509: certificate signed by unknown authority\"","time":"2019-02-17T02:19:34Z","message":"proxy: error redeeming authorization code"}
```

Thus we are forced to use a CA to sign the authenticator certificate. However for our purposes, we still don't want a central CA. Instead, we create a CA that itself is self-signed. But that then runs into the problem of the proxy trusting the self-signed CA.

This PR adds environment variables CERTIFICATE_AUTHORITY for base64 encoded contents or CERTIFICATE_AUTHORITY_FILE for a file located on the local disk. If present, it replaces the system certificate pool with one consisting of just the CA cert provided. Note that I'm open to renaming these environment variables if desired (maybe CA/CA_FILE?). It also updates the Helm charts to support passing certificate authority.

While we're here, we update the Helm chart with a more up to date list of possible environment variables, and remove the default routes; when someone overrides the proxy.routes, it actually merged the lists together, so anyone using the chart gets a useless route to httpbin.org in addition to the actual route they had; this removes that.

*bdd edit*
Additional context: 
- #4 
- #39 

**Checklist**:
- [ ] documentation updated
- [ ] unit tests added
- [x] ready for review

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -37,6 +37,8 @@ type Options struct {
 	AuthenticateInternalAddr string   `envconfig:"AUTHENTICATE_INTERNAL_URL"`
 	OverrideCertificateName  string   `envconfig:"OVERRIDE_CERTIFICATE_NAME"`
 	AuthenticatePort         int      `envconfig:"AUTHENTICATE_SERVICE_PORT"`
+	CA                       string   `envconfig:"CERTIFICATE_AUTHORITY"`
+	CAFile                   string   `envconfig:"CERTIFICATE_AUTHORITY_FILE"`
 
 	// SigningKey is a base64 encoded private key used to add a JWT-signature to proxied requests.
 	// See : https://www.pomerium.io/guide/signed-headers.html
@@ -207,6 +209,8 @@ func New(opts *Options) (*Proxy, error) {
 			OverrideCertificateName: opts.OverrideCertificateName,
 			SharedSecret:            opts.SharedKey,
 			Port:                    opts.AuthenticatePort,
+			CA:                      opts.CA,
+			CAFile:                  opts.CAFile,
 		})
 	return p, nil
 }
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -74,6 +74,13 @@ spec:
               name: {{ template "pomerium.fullname" . }}
               key: certificate-key
 {{- end }}
+{{- if .Values.config.ca }}
+        - name: CERTIFICATE_AUTHORITY
+          valueFrom:
+            secretKeyRef:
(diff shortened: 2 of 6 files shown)


proxy/proxy.go
```diff
@@ -37,6 +37,8 @@ type Options struct {
 	AuthenticateInternalAddr string   `envconfig:"AUTHENTICATE_INTERNAL_URL"`
 	OverideCertificateName   string   `envconfig:"OVERIDE_CERTIFICATE_NAME"`
 	AuthenticatePort         int      `envconfig:"AUTHENTICATE_SERVICE_PORT"`
+	CA                       string   `envconfig:"CERTIFICATE_AUTHORITY"`
```
desimone: I think `CERTIFICATE_AUTHORITY` and `CERTIFICATE_AUTHORITY_FILE` are fine names for the setting.

desimone:
Hi @mbarrien,

I was surprised to see how few load-balancers support HTTP2/gRPC (#39).  I did my best to test out various ingresses and deployment options, but I completely missed this one. I’m sorry about that!

Support for custom and self-signed CA’s is great feature in any case.  And as a happy consequence, this change should also allow pomerium to be run on windows (`SystemCertPool` is unsupported on windows), though I haven’t tested it myself. 

Thank you again for the PR. Do you mind fixing the merge conflict and repush’ing?

mbarrien:
Merge conflict fixed.

desimone:
Thank you, @mbarrien !

PR #62, opened 2019-03-11
authorize: support kubernetes

desimone:
- cmd/pomerium: made ping a top level route for all services
- docs: updated kubernetes script and corresponding configuration yaml's
- internal/options: removed email validators


scripts/kubernetes_gke.sh
```diff
@@ -1,64 +1,47 @@
...
-# load  TLS to ingress
+echo "=> load TLS to ingress"
 kubectl create secret tls -n pomerium pomerium-tls --key privkey.pem --cert cert.pem
 
-#                       !!! IMPORTANT !!!
-#          YOU MUST CHANGE THE Identity Provider Client Secret
-#                       !!! IMPORTANT !!!
-# kubectl create secret generic -n pomerium idp-client-secret --from-literal=REPLACE_ME
+echo "=> initiliaze a configmap setting for POLICY from $(policy.example.yaml)"
+kubectl create configmap -n pomerium policy --from-literal=policy=$(cat policy.example.yaml | base64)
```
victornoel: @desimone why should the policy be base64 encrypted? In general in Kubernetes ConfigMap you put directly raw data so that ops can easily check their content with lots of manipulations.
Secret are base64 encrypted to avoid showing them inadvertently.

victornoel: so you would write here: `--from-file=policy=policy.example.yaml`

desimone: It's not so much encrypting but rather base64 encoding is used here to preserve special and formatting characters characters from being mangled. 

My understanding is that while we could load a file in with [ConfigMap](https://cloud.google.com/kubernetes-engine/docs/concepts/configmap#from_files) [2](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-the-key-to-use-when-creating-a-configmap-from-a-file) K8s really expects those files  (`--from-file`) to be in simple K/V format. 

I did a [little bit more digging](https://stackoverflow.com/questions/39420102/how-can-i-store-a-binary-file-in-a-kubernetes-configmap) and it looks like k8s added support for `binaryData` from configMaps in 1.10. However, it looks like they are just doing base64 encoding too.  For example, I found this example in the [openshift docs](https://docs.openshift.com/container-platform/3.11/dev_guide/configmaps.html#configmaps-creating-from-files)

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  creationTimestamp: 2016-02-18T19:14:38Z
  name: example-config
  namespace: default
data: 
  example.property.1: hello
  example.property.2: world
  example.property.file: |-
    property.1=value-1
    property.2=value-2
    property.3=value-3
binaryData:
  bar: L3Jvb3QvMTAw 
```

```bash
$ echo L3Jvb3QvMTAw | base64 -D
/root/100
```

Based on the fact that the underlying implementation is basically the same (base64 encoding binary data), and that the feature is relatively new to kubernetes, and is [not documented](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-the-key-to-use-when-creating-a-configmap-from-a-file) directly, I think it makes sense to keep it as it is. It also keeps things consistent with other deployment options. 

Thoughts? 

See:
- https://github.com/kubernetes/kubernetes/pull/57938
- https://github.com/kubernetes/kubernetes/pull/33549
- https://github.com/kubernetes/kubernetes/issues/32432

victornoel: but do we expect the policy file to be binary file?

I have been putting python files in `ConfigMap` without any problem. I don't think there are problems with any type of files in `ConfigMap`, as long as the indentation is correct in the yaml file (and that's what `--from-file` does), it is not reserved to key/value pairs.

See also for example [this](https://github.com/helm/charts/blob/master/stable/postgresql/templates/configmap.yaml) and [this](https://github.com/helm/charts/blob/master/stable/postgresql/templates/initialization-configmap.yaml) where a postgresql configuration files, sql files and sh files are put inside a `ConfigMap` without any specific encoding.

desimone: I believe in the examples you are showing though they directly loading the files in ConfigMap they then also mount the files, and then postgres loads those config files as file handlers.  To my knowledge, they aren't passing the contents of the configuration file directly as bytes as we are doing in pomerium with `POLICY`. 

https://github.com/helm/charts/blob/master/stable/postgresql/templates/statefulset.yaml#L214

Which is a totally fine option and would work with pomerium too. You'd just use `POLICY_FILE` instead. I supposed I could update the example to use a volume mount but I personally find the stateless approach more attractive. But that's subjective of course. 

 I agree that it would be ideal to be able to pass the contents of the policy file in directly to pomerium without having to wrap it in base64 encoding. The trick is that (though Kubernetes may handle it gracefully) other environments and shells are prone to mangle multi-line, special character encoded files.

desimone: > In that case, I agree with you, it's a limitation of using environment variables, sorry for the noise then :)

Not at all. Was a good discussion that resulted in me learning more about configmap. I appreciate the review. 

>  Still, I know I will be using POLICY_FILE (mostly for ease of introspection and because it feels more natural to use a file to represent the content of a… file :P), and as long as you don't need to base64 encode its content, it should be alright :)

Totally understandable and that should be an option. 

At this point, I'm going to merge and marked as resolved. If you have any issues please let me know.

PR #74, opened 2019-03-28
internal/log: fix setting log level

### LINKED ISSUE #70: There are debug logs when POMERIUM_DEBUG is not set
**Describe the bug**

`POMERIUM_DEBUG` is not set but I see logs of the following form for example in proxy:
```
{"level":"debug","cert-override-name":"xxx.com","addr":"pomerium-authenticate:443","time":"2019-03-25T14:25:13Z","message":"proxy/clients: grpc connection"}
{"level":"debug","cert-override-name":"xxx.com","addr":"pomerium-authorize:443","time":"2019-03-25T14:25:13Z","message":"proxy/clients: grpc connection"}
```

(Actually, I think those should be info also)

**Environment:**

- Pomerium version (retrieve with `pomerium --version`): v0.0.2+45e6a8d

desimone:
- POMERIUM_DEBUG no longer sets log level.
- LOG_LEVEL now responsible for setting global log level.

Fixes #70 

Thank you @victornoel for reporting the issue.

 
**Checklist**:
- [x] documentation updated
- [x] unit tests added
- [x] ready for review

### DIFF
--- a/cmd/pomerium/options.go
+++ b/cmd/pomerium/options.go
@@ -12,9 +12,12 @@ import (
 // or if a file location is not provided, the server will attempt to find a matching keypair
 // in the local directory as `./cert.pem` and `./privkey.pem` respectively.
 type Options struct {
-	// Debug enables more verbose logging, and outputs human-readable logs to Stdout.
-	// Set with POMERIUM_DEBUG
+	// Debug outputs human-readable logs to Stdout.
 	Debug bool `envconfig:"POMERIUM_DEBUG"`
+
+	// LogLevel sets the global override for log level. All Loggers will use at least this value.
+	// Possible options are "info","warn", and "error". Defaults to "debug".
+	LogLevel string `envconfig:"LOG_LEVEL"`
 
 	// SharedKey is the shared secret authorization key used to mutually authenticate
 	// requests between services.
@@ -39,6 +42,7 @@ type Options struct {
 
 var defaultOptions = &Options{
 	Debug:    false,
+	LogLevel: "debug",
 	Services: "all",
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@
 - **WebSocket Support** : With [Go 1.12](https://golang.org/doc/go1.12#net/http/httputil) pomerium automatically proxies WebSocket requests.
 
 **CHANGED**:
-
+- Add `LOG_LEVEL` config setting that allows for setting the desired minimum log level for an event to be logged. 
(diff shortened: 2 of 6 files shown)


cmd/pomerium/options.go
```diff
@@ -39,6 +42,7 @@ type Options struct {
 
 var defaultOptions = &Options{
 	Debug:    false,
+	LogLevel: "debug",
```
victornoel: @desimone should the default be debug? I would have expected a default to info or equivalent, no?

desimone: I could go either way on this which is part of why I wanted to break it out into a user configurable field. Here's my thinking. 

Right now, the DEBUG is basically a switch for people who want to view individual requests details which includes not only the typical request details (time to serve, request id, HTTP verb, url, etc) but also the attached user's details.  I personally find the details especially useful for auditing purposes and was tempted to move it up to INFO but knew others wouldn't want those details. 

I also think it's a good default option because for new users having any issues, I imagine it would be fairly challenging to investigate any problems without the DEBUG switch which adds another step.  When users are ready to deploy pomerium in production, it's probably safe to assume they've become comfortable enough to set the log level they feel best fits their usage but at first, if they hit some issue, I wouldn't want the added frustration of finding the verbose switch.

Thoughts?

victornoel: Interesting, glad I raised the question, because personally, I don't really care to have all the http details (I already have nginx-ingress for that) BUT I am mainly interested in the auditing info such as user details.
I would even go with the user details in INFO: it is about the main business of pomerium (and I guess in the long term serious auditing could be one of its feature :).

Also I think the argument about setting info while in production and having debug by default is totally sound yes!

desimone: The trick is associating those user details with a given request so it's hard to separate the two (even though, like you said, the many users will be put pomerium behind a 3rd party ingress). Without the additional HTTP context(`JoeUser with admins@company.com access /intranet/hr/payroll at 3:20 from IP x.x.x.x`) having user details isn't very useful for auditing purposes which absolutely is a primary usecase for the product. (Fun fact and a tangible example, if you read the Uber vs Waymo lawsuit,  BeyondCorp played a major role in helping to audit the alleged IP theft...)

desimone: I'm going to merge but I like the idea of having audit information demarcated differently, maybe by log level, maybe by another mechanism.

victornoel: > The trick is associating those user details with a given request so it's hard to separate the two

Yes, I guess you still need them

> I'm going to merge but I like the idea of having audit information demarcated differently, maybe by log level, maybe by another mechanism.

This will be the subject of another, more security/audit-oriented release maybe :)

Issue #80, opened 2019-04-08
Google: Missing refresh token during re-authorization after clearing cookies


chemhack:
In an edge case, session state can end up without a refresh token, which causing 500 in pomerium proxy due to unable to fresh oauth token.

1. Login to a pomerium proxied app with Google account.
2. Clear all pomermium cookies
3. Login with Google again. If you check session state at this point, refresh_token is already empy.
4. Wait a few hours until oauth token expire
5. Go to pomerium proxied app.
6. 500 Error

**Expected behavior**

Pomerium should always be able to get a fresh token, since it's "persisted" in cookies, and the chance of losing it is very high.

**Environment:**

- Pomerium version: 0.0.3
- Server Operating System/Architecture/Cloud: kubernetes 1.12

Logs(s):

```proxy: refresh failed [31merror=[0m[31m"missing refresh token"...```

**Additional context**

PR to fix will follow.

desimone:
Closed by #81

PR #81, opened 2019-04-08
Force requesting refresh_token from Google

chemhack:
Fixes Issue #80 

Google only provide refresh_token on the first authorization from the user. If user clears cookies, re-authorization will not bring back refresh_token. A work around to this is to add prompt=consent to the OAuth redirect URL and will always return a refresh_token.

**Checklist**:

- [x] documentation updated
- [ ] unit tests added
- [X] ready for review


desimone:
@chemhack Thank you for the bug report and fix! I was able to reproduce and test the fix.

PR #84, opened 2019-04-12
proxy: Do not override upstream security headers

chemhack:
**Checklist**:
- [ ] updated docs
- [ ] unit tests added
- [ ] ready for review

### DIFF
--- a/internal/middleware/middleware.go
+++ b/internal/middleware/middleware.go
@@ -19,7 +19,9 @@ func SetHeaders(securityHeaders map[string]string) func(next http.Handler) http.
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			for key, val := range securityHeaders {
-				w.Header().Set(key, val)
+				if w.Header().Get(key) == "" {
+					w.Header().Set(key, val)
+				}
 			}
 			next.ServeHTTP(w, r)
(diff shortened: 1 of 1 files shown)


desimone:
Hi @chemhack are you still working on this? Anything I can do to help?

chemhack:
hi @desimone, sorry i didn't have time test this yet. would you have time to take over this?

desimone:
@chemhack sure -- can you help be better understand the use case. As of now, the security headers are pretty conservative. Are you experiencing an issue with a downstream app because of them?

chemhack:
@desimone we have some pages which suppose to live in an iframe. it should be possible to override these settings from upstream.

desimone:
@chemhack I see what you are saying. I think it makes sense to add this as a policy specific setting.  

What do you think about something along the lines of...

```yaml
- from: iframe.corp.example.com 
  to: http://iframe
  allowed_domains:
      - example.com
  respect_upstream_headers: true
```

chemhack:
> @chemhack I see what you are saying. I think it makes sense to add this as a policy specific setting.
> 
> What do you think about something along the lines of...
> 
> ```yaml
> - from: iframe.corp.example.com 
>   to: http://iframe
>   allowed_domains:
>       - example.com
>   respect_upstream_headers: true
> ```

this is even better setup, closing this one.

PR #101, opened 2019-05-02
deployment: use distroless builds

### LINKED ISSUE #97: deployment: docker image does not contain nsswitch.conf
**Describe the bug** 
Pomerium's base image inherits from alpine linux which does not currently have `nsswitch.conf`. This is problematic because Go uses nsswitch.conf by default.  

As a result, Pomerium will not resolve hosts via /etc/hosts but instead always uses DNS. 

**Expected behavior**

To be completely frank, I'm not sure if this is our problem to fix, Go's or Alpine's. 

**Potential fix?**

Could we simply add a minimal `nsswitch.conf` to our `Dockerfile`?

```yml
RUN [ ! -e /etc/nsswitch.conf ] && echo 'hosts: files dns' > /etc/nsswitch.conf
```

**Additional context**

> As @tianon mentions and as commented here: docker-library/docker#82 (comment) it is Go that is hardcoded to behave as glibc (dns first and then use hosts if it fails) if there is no /etc/nsswitch.conf. musl libc does not use this file at all since it does not implement NSS. I'd say that this is a bug in Go which assumes that linux always is glibc.
- https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-354316460

See also: 

- Pomerium #88
- Alpine https://github.com/gliderlabs/docker-alpine/issues/367
- Go golang/go#22846
- Kubernetes https://github.com/kubernetes/kubernetes/issues/71082

/cc @yegle

desimone:
Replaces the current alpine based Dockerfile with [distroless](https://github.com/GoogleContainerTools/distroless). Improvements include:
- Minimal surface area, ideal for static builds like pomerium.
- Includes `ca-certificates`
- Includes`nsswitch`

Closes #97 .

**See also**

- [best practices for building containers](https://cloud.google.com/solutions/best-practices-for-building-containers#use_the_smallest_base_image_possible)
- [Life of a happy developer in a container world by David Gageot
 ](https://www.youtube.com/watch?v=lviLZFciDv4)
- [Distroless Docker: Containerizing Apps, not VMs](https://www.youtube.com/watch?v=lviLZFciDv4)

**Checklist**:
- [x] ready for review

Test image is available on dockerhub as `pomerium/pomerium:distroless`

/cc @yegle @eriksw

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,15 @@
-FROM golang:alpine as build
-RUN apk --update --no-cache add ca-certificates git make
+FROM golang:latest as build
+WORKDIR /go/src/github.com/pomerium/pomerium
 ENV CGO_ENABLED=0
 ENV GO111MODULE=on
-
-WORKDIR /go/src/github.com/pomerium/pomerium
-
-COPY go.mod .
-COPY go.sum .
+# cache depedency downloads
+COPY go.mod go.sum ./
 RUN go mod download
-
 COPY . .
+# build 
+RUN make build
 
-RUN make
-
-FROM scratch
-COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+FROM gcr.io/distroless/static
 WORKDIR /pomerium
 COPY --from=build /go/src/github.com/pomerium/pomerium/bin/* /bin/
 CMD ["/bin/pomerium"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,20 +1,13 @@
-FROM golang:alpine as build
-RUN apk --update --no-cache add ca-certificates git make
-ENV CGO_ENABLED=0
```
eriksw: If you meant to enable cgo, distroless/base would be the way to go instead of distroless/static.

Issue #116, opened 2019-05-14
all: wrap global mux with standard middleware


desimone:
**Describe the bug** 

Several middleware apply to all services and should wrap the global mux. Additionally, non-routed URLs, or services without handlers (authenticate) will bypass middleware causing unexpected behavior like missing CSP headers or missed health-checks.

**Expected behavior**

The global http mux should be wrapped with a standard set of middleware. Such as:

- healthcheck
- referrer
- request id
- ip
- fwd_ip
- etc

**Additional context**

- https://github.com/pomerium/pomerium/issues/90#issuecomment-492057963

desimone:
Closed by #117

PR #117, opened 2019-05-14
cmd/pomerium: wrap global mux with standard middleware

desimone:
These changes abstract the shared middleware to the global serve mux. Headers, request id, loggers, and health checks (ping) middleware are now applied on all routes including 4xx and 5xx responses. 

See
- #116 

**Checklist**:
- [x] unit tests added
- [x] ready for review

/cc @yegle

### DIFF
--- a/cmd/pomerium/main_test.go
+++ b/cmd/pomerium/main_test.go
@@ -1,10 +1,11 @@
 package main
 
 import (
+	"fmt"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"os"
-	"reflect"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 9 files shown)


cmd/pomerium/main_test.go
```diff
@@ -166,34 +167,33 @@ func Test_newProxyeService(t *testing.T) {
...
-		})
+func Test_wrapMiddleware(t *testing.T) {
+	o := &Options{
+		Services: "all",
+		Headers: map[string]string{
+			"X-Content-Type-Options":    "nosniff",
+			"X-Frame-Options":           "SAMEORIGIN",
+			"X-XSS-Protection":          "1; mode=block",
+			"Strict-Transport-Security": "max-age=31536000; includeSubDomains; preload",
+			"Content-Security-Policy":   "default-src 'none'; style-src 'self' 'sha256-pSTVzZsFAqd2U3QYu+BoBDtuJWaPM/+qMy/dBRrhb5Y='; img-src 'self';",
```
yegle: Looks like this Content-Security-Policy header is breaking loading javascripts:

```
Refused to load the script '<URL>' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'script-src-elem' was not explicitly set, so 'default-src' is used as a fallback.

XXX.apkay.com/:83 Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'sha256-pSTVzZsFAqd2U3QYu+BoBDtuJWaPM/+qMy/dBRrhb5Y='". Either the 'unsafe-inline' keyword, a hash ('sha256-wycEadbqyap1lzI8fe5whjcPll4StPOU2pDRk+cjffU='), or a nonce ('nonce-...') is required to enable inline execution.
```

desimone: Thanks @yegle . The CSP/Referrer-Policy should only apply to authenticate service's http handlers. Good catch. Fix incoming.

Issue #115, opened 2019-05-13
refactor configuration handling


travisgroth:
As initially discussed in #89 - I'd like to propose refactoring configuration handling from [pomerium/envconfig](/pomerium/envconfig) to [spf13/viper](/spf13/viper)  I see this as a precursor to #89, since reloading of policy naturally leads into how run time configuration is brought into the proxy and authz services.

- Brings in support for a variety of additional config storage formats
- Has vetted handling of live configuration notifications to support #89 
- Should reduce code surface area in pomerium

### Changes

* Non-breaking: Add support for a new configuration file located at /etc/pomerium/config.[yaml|toml|json]
  * This file can contain _all_ configuration options, including inline policy
* Non-breaking: Configuration file location can be overridden with an env_var
* Breaking: Deprecate POLICY_FILE and require policy to be defined as either a base64 encoded env var or as part of config file policy key

### Config file example

```yaml
authenticate_service_url: https://authenticate.corp.example.com
authorize_service_url: https://authorize.corp.example.com
shared_secret: 9wiTZq4qvmS/plYQyvzGKWPlH/UBy0DMYMA2x/zngrM= 
cookie_secret: uPGHo1ujND/k3B9V6yr52Gweq3RRYfFho98jxDG5Br8=
idp_provider: "google"
idp_client_id: "REPLACE-ME.googleusercontent.com"
idp_client_secret: "REPLACEME"
policy:
    - from: httpbin.corp.beyondperimeter.com
    to: http://httpbin
    allowed_domains:
        - pomerium.io
    cors_allow_preflight: true
    timeout: 30s
    - from: external-httpbin.corp.beyondperimeter.com
    to: httpbin.org
    allowed_domains:
        - gmail.com
    - from: weirdlyssl.corp.beyondperimeter.com
    to: http://neverssl.com
    allowed_users:
        - bdd@pomerium.io
    allowed_groups:
        - admins
        - developers
    - from: hello.corp.beyondperimeter.com
    to: http://hello:8080
    allowed_groups:
        - admins
```

## Internal Spec

The plan is to continue using a struct unmarshalled from the configuration to represent options. 

At a high level, there are two approaches to the refactor:

1. Extend the top level Options struct to support all options and pass this along to each service.  

    * Basic validation and loading can be done in a single location
    * Individual services can still validate the fields they're interested in
    * Makes the largest reduction in maintained/duplicate Service level code
    * Options shared across services don't need to be added to multiple service scope structs

[...]

PR #118, opened 2019-05-15
Refactor and centralize options handling

travisgroth:
Preliminary work for #115 

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] ready for review


travisgroth:
Some commentary - I realized it would be cleanest to first create the central structure we want, and then swap out how it gets populated and validated.  

I've got two tests in TestProxy_Proxy failing that I'm not totally clear on.  Will try to resolve but if you have an idea on what's going on, I wouldn't mind being pointed in the right direction.  I'm hoping it's just a test setup artifact.  I'll dig in later this evening or tomorrow morning.

desimone:
@travisgroth I'm going to temporarily mark this as "ready for review" so that the CI gets run so I can see the error.

proxy/proxy_test.go
```diff
@@ -112,23 +79,23 @@ func TestNewReverseProxyHandler(t *testing.T) {
 	}
 }
 
-func testOptions() *Options {
+func testOptions() *config.Options {
 	authenticateService, _ := url.Parse("https://authenticate.corp.beyondperimeter.com")
 	authorizeService, _ := url.Parse("https://authorize.corp.beyondperimeter.com")
 	configBlob := `[{"from":"corp.example.com","to":"example.com"}]` //valid yaml
```
desimone: My guess is that this is no longer a valid route and is returning a 404 when attempted to be proxied.

desimone:
Ah! I figured out what's going on here.  The tests are reporting the right thing. The "quick" fix is to change  the destination URLs for those two tests from

```
https://corp.example.com/test
``` 

to 

```
https://corp.example.com/
```

These tests should have never passed, and it looks like whatever changes you made fixed the issue. At the time I wrote these tests I believe I was on a plane and thought a `503` error was reasonable given that we didn't control the domain and I  assumed example.com was not live which of course it is! What's concerning is that these tests are passing on master and shouldn't be.

The "real fix" is to be mocking the destination server 

```go
	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintln(w, "Hello, client")
	}))
	defer ts.Close()
```

and passing `ts.URL` to the `testOptions()` function to build 

```go
	configBlob := fmt.Sprintf(`[{"from":"corp.example.com","to":"%s"}]`, to) //valid yaml
```

I can create an issue and take that up after this PR is merged if that sounds good to you.

desimone:
> Some commentary - I realized it would be cleanest to first create the central structure we want, and then swap out how it gets populated and validated.

Looks good so far to me 👍

travisgroth:
Hah.  The comment on the test case and the one following had me thinking of CORS and the domain checking that was previously done with ROOT_DOMAIN.  I couldn't see why it _wouldn't_ pass after going through the code, though.  That site being live was a surprise to me too, but I thought we were testing something more sophisticated.

I'm guessing maybe dockerhub blocks that site?  I can update the domain in this PR for the time being so we get a clean test run.

travisgroth:
Please review again as I had to do a bit of a rebase to catch up with master.  

I'm inclined to say we merge this PR in to get the new structure in place, and I'll open a follow up with the actual viper changes.  I want to get the restructuring in place so we can make more changes without having to rebase this branch again.  Now that all the options logic is in one place, it should be pretty easy to iterate without too many conflicts.

I don't think we need any doc or test changes yet.  LMK if you spot anything.

travisgroth:
Oh one more thing - I commented out a few test cases that I had trouble supporting.  I think we can just rely on the type system to fail correctly in the real world but I'll make sure that's the case.  They're marked with // TODO so we can find them later.

cmd/pomerium/main.go
```diff
@@ -102,56 +103,44 @@ func startRedirectServer(addr string) (*http.Server, error) {
...
 	pbAuthenticate.RegisterAuthenticatorServer(rpc, service)
-	mux.Handle(urlutil.StripPort(opts.AuthenticateURL.Host)+"/", service.Handler())
+	mux.Handle(urlutil.StripPort(opt.AuthenticateURL.Host)+"/", service.Handler())
 	return service, nil
 }
 
-func newAuthorizeService(s string, rpc *grpc.Server) (*authorize.Authorize, error) {
-	if !isAuthorize(s) {
+func newAuthorizeService(opt *config.Options, rpc *grpc.Server) (*authorize.Authorize, error) {
+	if !config.IsAuthorize(opt.Services) {
```
desimone: Could panic if opt is nil?

cmd/pomerium/main.go
```diff
@@ -102,56 +103,44 @@ func startRedirectServer(addr string) (*http.Server, error) {
...
 		return nil, err
 	}
 	pbAuthorize.RegisterAuthorizerServer(rpc, service)
 	return service, nil
 }
 
-func newProxyService(s string, mux *http.ServeMux) (*proxy.Proxy, error) {
-	if !isProxy(s) {
+func newProxyService(opt *config.Options, mux *http.ServeMux) (*proxy.Proxy, error) {
+	if !config.IsProxy(opt.Services) {
```
desimone: Could panic if opt is nil?

cmd/pomerium/main_test.go
```diff
@@ -131,34 +139,35 @@ func Test_newAuthorizeService(t *testing.T) {
...
 	}{
 		{"wrong service", "authenticate", "", "", false},
-		{"bad option parsing", "proxy", "SHARED_SECRET", "false", true},
-		{"bad env", "proxy", "POLICY", "error!", true},
-		{"bad encoding for envar", "proxy", "COOKIE_REFRESH", "error!", true},
-		{"good", "proxy", "SHARED_SECRET", "YixWi1MYh77NMECGGIJQevoonYtVF+ZPRkQZrrmeRqM=", false},
+		{"bad option parsing", "proxy", "SharedKey", "false", true},
+		{"bad env", "proxy", "Policy", "error!", true},
+		// TODO I don't think we should be testing for type violations
+		//{"bad encoding for envar", "proxy", "CookieRefresh", "error!", true},
```
desimone: 👍 can just delete this test.

internal/config/options_test.go
```diff
@@ -1,4 +1,4 @@
-package main // import "github.com/pomerium/pomerium/cmd/pomerium"
+package config // import "github.com/pomerium/pomerium/cmd/pomerium"
```
desimone: s/ `//import "github.com/pomerium/pomerium/internal/config"`

internal/config/options.go
```diff
@@ -0,0 +1,226 @@
+package config // import "github.com/pomerium/pomerium/cmd/pomerium"
```
desimone: Import path should be `//import "github.com/pomerium/pomerium/internal/config"`

internal/config/options.go
```diff
@@ -0,0 +1,226 @@
...
+		},
+		Addr:              ":https",
+		CertFile:          filepath.Join(findPwd(), "cert.pem"),
+		KeyFile:           filepath.Join(findPwd(), "privkey.pem"),
+		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      0, // support streaming by default
+		IdleTimeout:       5 * time.Minute,
+		AuthenticateURL:   new(url.URL),
+		AuthorizeURL:      new(url.URL),
```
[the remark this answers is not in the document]
desimone: TIL! https://play.golang.org/p/lh4NjKWDSC8

desimone:
Thanks @travisgroth !

> I'm inclined to say we merge this PR in to get the new structure in place, and I'll open a follow up with the actual viper changes. I want to get the restructuring in place so we can make more changes without having to rebase this branch again. Now that all the options logic is in one place, it should be pretty easy to iterate without too many conflicts.

I agree. 

Also, once you update, do you mind squashing your changes then I'll do a merge? Github weirdly mangles commit history if I do a squash on my end.

cmd/pomerium/main_test.go
```diff
@@ -99,27 +103,31 @@ func Test_newAuthorizeService(t *testing.T) {
 	grpcServer := grpc.NewServer(grpcOpts...)
 
 	tests := []struct {
-		name     string
-		s        string
-		envKey   string
-		envValue string
+		name  string
+		s     string
+		Field string
```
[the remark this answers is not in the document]
travisgroth: Just matching the new reality of what they were manipulating.  Had to be exported for reflection.

internal/config/options.go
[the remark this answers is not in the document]
travisgroth: Yeah, that drove me nuts for a while.

cmd/pomerium/main.go
travisgroth: Yes.  Argument to be made that we've really screw up in main() if that happens.  I'll add a check in, though.

desimone:
@travisgroth Thanks! LGTM

Issue #114, opened 2019-05-13
docs: add upgrade notes to documentation and release notes


desimone:
**Describe the bug**

Even with versioned documentation, migrating an existing installation between releases would not be clear. This can be confusing and frustrating for users looking to use the latest version of Pomerium. 

**Expected behavior**

Pomerium should provide upgrade documentation covering how to migrate and upgrade between versions. 

**See**
- https://github.com/pomerium/pomerium/issues/107#issuecomment-491938531
- #109 
- https://www.vaultproject.io/docs/upgrading/index.html
- https://semver.org/

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

/cc with thanks and apologies to @jaumann

desimone:
Closed by #137

PR #65, opened 2019-03-21
deployment: update and migrate helm chart to official repository

desimone:
Update helm for the new authorization modules, groups, and try to auto-fill as many values as possible (secrets, dummy ca, certs, etc). 

If/when pomerium gets merged into helm stable, it may makes sense to remove helm files from this repo in favor of maintaining the public repo.

See also:
- https://github.com/helm/charts/pull/12415
- #64 

**Checklist**:
- [ ] get pomerium chart merged to helm/charts stable 
- [x] remove helm from this repo
- [x] update documentation
- [x] related issues referenced

PR #122, opened 2019-05-15
Add Helm chart fixes

tarokkk:
The Helm chart had several issues regarding these variable changes:

> - Removed extraneous AUTHORIZE_INTERNAL_URL config option since authorization has no publica http handlers, only a gRPC service endpoint. [GH-93]
> - Removed PROXY_ROOT_DOMAIN config option which is now inferred from AUTHENTICATE_SERVICE_URL. Only callback requests originating from a URL on the same sub-domain are permitted. [GH-83]

We removed the *.{{rootDomain}} ingress rule as Kubernetes does not support such configurations. The chart now uses the policy configuration as single source of ingress rules.

It works as intended however still lacks documentation of the `omnibusMode` which would be nice.

I would suggest updating the pending Helm repo request https://github.com/helm/charts/pull/12415 or use own chart repository. There is already a repo as a service if you don't want to bother setting it up here: https://charts.banzaicloud.io/

I would also suggest to use versioned images in the chart (not latest) as every configuration variable change will break the chart as well which is frustrating if you are debugging :)

**Checklist**:
- [x] updated docs
- [x] ready for review

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: pomerium
-version: 1.0.0
-appVersion: 0.0.3
+version: 1.1.0
+appVersion: 0.0.5
 home: http://www.pomerium.io/
 icon: https://www.pomerium.io/logo.svg
 description: Pomerium is an identity-aware access proxy.
--- a/helm/templates/authenticate-deployment.yaml
+++ b/helm/templates/authenticate-deployment.yaml
@@ -22,6 +22,7 @@ spec:
     metadata:
 {{- if .Values.podAnnotations }}
       annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
 {{ toYaml .Values.podAnnotations | indent 8 }}
 {{- end }}
       labels:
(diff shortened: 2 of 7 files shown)


desimone:
Hi @tarokkk  -- Thank you for the PR!

Sorry for any headaches... Unfortunately, the Helm directory in this repo has gone stale waiting for the official chart to get merged and has missed out on a bunch of updates (including, it looks like some of your changes) and versioning.

> It works as intended however still lacks documentation of the omnibusMode which would be nice.

Omnibus no longer exists in the latest helm chart https://github.com/helm/charts/pull/12415/files and was deprecated in favor of a full distributed install with split services which is probably the preferred way to install pomerium in production. I'm definitely open to adding omnibus back, but don't currently have the time at the moment. 

> I would also suggest to use versioned images in the chart (not latest) as every configuration variable change will break the chart as well which is frustrating if you are debugging :)

I think this is in the latest (no pun intended) [version of the official chart PR](https://github.com/helm/charts/pull/12415/files#diff-974b10675131f5b6839e8888c7f65edeR135).  
 Apologies for any headaches this caused. Smoother upgrade migration is something I really want to improve. #114 

> I would suggest updating the pending Helm repo request helm/charts#12415 or use own chart repository. There is already a repo as a service if you don't want to bother setting it up here: https://charts.banzaicloud.io/

This could work as the official charts PR has been rotting on the vine.  In any case, having  [1](https://github.com/pomerium/pomerium/tree/master/helm) [2](https://github.com/pomerium/pomerium/pull/65) [3](https://github.com/helm/charts/pull/12415) different helm sources is not sustainable.

I haven't looked at banzai compare to something like chart hub.

tarokkk:
https://charts.banzaicloud.io/ is a Helm chart repository service (backed by [chartmusem](https://chartmuseum.com/) ). The `stable` helm chart repository is slow and troublesome to maintain that's why we are using it for our own charts as well. It's completely free and works like the official repository. If you have any question or trouble I would be glad to help you. We set up pomerium in our sandbox and really liked the concept that's why we try to help with some contribution. If you decided which repo you store the upstream chart I'm glad to move the PR there

desimone:
@tarokkk That sounds perfect. Let's chat offline about how to get this set up.

 In the meantime, do you mind taking this PR to the new repo? https://github.com/pomerium/pomerium-helm

tarokkk:
Closed prior to this PR https://github.com/pomerium/pomerium-helm/pull/1/files

Issue #110, opened 2019-05-08
authenticate : add impersonate user / login as support


desimone:
**Is your feature request related to a problem? Please describe.**

It would be very useful, in certain circumstances, for an administrative user to impersonate or login as another user.  Some example scenarios.  

**Describe the solution you'd like**

User impersonation should allow you to temporarily sign in as a different user through pomerium. Users with impersonation permissions can impersonate all other users in their network and take any action, regardless of the impersonating user's own permission level. 

- UI flow TBD.
- Certain users or groups to impersonate other users. 
- Impersonators must appear as themselves in the change history.

**Explain any additional use-cases**

- To help another user troubleshoot an issue. If your user roles are configured differently, it's possible that your UI will look different from theirs and you'll need to impersonate the other user to be able to see what they see.
- You want to make changes on behalf of another user (for example, the other user is away on vacation and you want to manage their orders or run a report).
- You're an administrator who's setting up user roles, and you want to preview what other users will be able to see depending on the permissions you grant them.

**Prior art**

- https://help.salesforce.com/articleView?id=logging_in_as_another_user.htm&type=5
- https://confluence.atlassian.com/cloud/log-in-as-another-user-776994835.html
- https://support.google.com/admanager/answer/1241070?hl=en

PR #130, opened 2019-05-20
deployment: switch to viper for config

travisgroth:
Further progress on #115 

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] ready for review


desimone:
FYI there are a few lint issues I missed earlier 
```
internal/config/options.go:183:1: comment on exported function IsValidService should be of the form "IsValidService ..."
internal/config/options.go:196:1: exported function IsAuthenticate should have comment or be unexported
internal/config/options.go:206:1: exported function IsAuthorize should have comment or be unexported
internal/config/options.go:216:1: exported function IsProxy should have comment or be unexported
internal/config/options.go:183:1: comment on exported function IsValidService should be of the form "IsValidService ..."
internal/config/options.go:196:1: exported function IsAuthenticate should have comment or be unexported
internal/config/options.go:206:1: exported function IsAuthorize should have comment or be unexported
internal/config/options.go:216:1: exported function IsProxy should have comment or be unexported
```

travisgroth:
>FYI there are a few lint issues I missed earlier

Yup I noticed that while working on this earlier.  Will address.  Some of the structure here is a bit in flux still.

travisgroth:
@desimone I think that should wrap up the functionality switch.  I must admit retaining backwards compat on a few things got more complicated than I'd hoped.  That said, I'm feeling pretty good about the structure now.

There's a few TODOs to add new tests or port ones that weren't easy to fix in place, so I wanted to get any feedback on architecture and patterns before I dive into fixing those.

There's also some TODOs for validation methods that need to be relocated/updated.  My thought there is have Options do type/parsing validation and each Service just make sure that the required options are present.  Just want to make sure the direction looks good before I roll up my sleeves on that.

I think the only major deviation from the spec is that the config file location configurable, and optional, which should be easier to test and use.  If you specify a file, it will fail if it is missing but should mesh env_vars with the file contents.  If you don't specify a file, everything must come from the environment or file defaults (in the case of ssl certs).

authorize/authorize_test.go
```diff
@@ -25,28 +25,29 @@ func TestNew(t *testing.T) {
...
-		{"bad policy file, directory", "gXK6ggrlIW2HyKyUF9rUO4azrDgxhDPWqw9y+lJU7B8=", "", "./testdata/", true},
-		{"good policy", "gXK6ggrlIW2HyKyUF9rUO4azrDgxhDPWqw9y+lJU7B8=", "WwogIHsKICAgICJyb3V0ZXMiOiAiaHR0cDovL3BvbWVyaXVtLmlvIgogIH0KXQ==", "", false},
-		{"good file", "gXK6ggrlIW2HyKyUF9rUO4azrDgxhDPWqw9y+lJU7B8=", "", tmpfile.Name(), false},
-		{"validation error, short secret", "AZA85podM73CjLCjViDNz1EUvvejKpWp7Hysr0knXA==", "", "", true},
-		{"nil options", "", "", "", true}, // special case
+		{"good", "gXK6ggrlIW2HyKyUF9rUO4azrDgxhDPWqw9y+lJU7B8=", "WwogIHsKICAgICJyb3V0ZXMiOiAiaHR0cDovL3BvbWVyaXVtLmlvIgogIH0KXQ==", false},
+		{"bad shared secret", "AZA85podM73CjLCjViDNz1EUvvejKpWp7Hysr0knXA==", "WwogIHsKICAgICJyb3V0ZXMiOiAiaHR0cDovL3BvbWVyaXVtLmlvIgogIH0KXQ==", true},
+		{"really bad shared secret", "sup", "WwogIHsKICAgICJyb3V0ZXMiOiAiaHR0cDovL3BvbWVyaXVtLmlvIgogIH0KXQ==", true},
+		// TODO
+		// I don't think these make sense anymore
```
travisgroth: @desimone thoughts here?  My feeling was most of this gets caught as simpler error cases higher in the stack.

internal/config/options.go
```diff
@@ -148,6 +155,125 @@ func NewOptions() *Options {
...
+	o.bindEnvs()
+	if configFile != "" {
+		log.Info().Msgf("Loading config from '%s'", configFile)
+		fh, err := os.Open(configFile)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to open config file '%s': %s", configFile, err)
+		}
+		err = viper.ReadConfig(fh)
+		if err != nil {
+			log.Fatal().Msgf("Failed to read config: %s", err)
```
desimone: An error is returned if the file cannot be opened but panics if it can't be read by viper. 
Curious, why throw a panic here vs an error above; can we just return an error? 
So for example, I can imagine a scenario where a policy file is updated and for-whatever reason the config file can't be read, I'd rather pomerium keep running with the outdated policy than panic (and recover?).

internal/config/options.go
```diff
@@ -148,6 +155,125 @@ func NewOptions() *Options {
...
+	}
+
+	// Load and initialize policy
+	err = o.parsePolicy()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse Policy: %s", err)
+	}
+
+	// TODO probably should move this
+	// Validation
```
desimone: I agree ... SharedKey and IsValidService checks make more sense in Validation.

internal/config/options.go
```diff
@@ -148,6 +155,125 @@ func NewOptions() *Options {
...
+		err := (&policies[i]).Validate()
+		if err != nil {
+			return err
+		}
+	}
+	o.Policies = policies
+	return nil
+}
+
+// parseURLs parses URL strings into actual URL pointers
```
desimone: 👍

cmd/pomerium/main.go
```diff
@@ -24,14 +24,15 @@ import (
 )
 
 var versionFlag = flag.Bool("version", false, "prints the version")
+var configFile = flag.String("config", "", "Specify configuration file location")
```
desimone: 👍

internal/config/options.go
travisgroth: Yes; agreed.  This should be consistent.  Main can figure out what to do.

travisgroth:
I think the functionality is done here.  Will take a pass at documentation now.

travisgroth:
In the interest of not falling too far behind again, flagging for merge.

Remaining items off the top of my head:

- More thorough documentation on the drop of POLICY_FILE support
- Update Dockerfile so that either
  - a config is always present but empty to be mounted over by the user (I favor this one)
  - the user can pass parameters at run time to a location that they've mounted their config
- Helm updates (we should prefer the config file going forward for anything non-secret
- Update examples

I'll reference this list in the tracking issue.

desimone:
> In the interest of not falling too far behind again, flagging for merge.

Sounds good to me. I've got a rather larger-than-I-planned PR that's about ready (#110) that will likely shortly follow this one. 

> a config is always present but empty to be mounted over by the user (I favor this one)
> the user can pass parameters at run time to a location that they've mounted their config

I'm not sure I understand. Put another way, do you mean that, we either have the dockerfile always mount a policy file which can / cannot be empty OR we don't make that assumption and let the user pass the config file location at run-time?


> Helm updates (we should prefer the config file going forward for anything non-secret

Agree.

travisgroth:
>I'm not sure I understand. Put another way, do you mean that, we either have the dockerfile always mount a policy file which can / cannot be empty OR we don't make that assumption and let the user pass the config file location at run-time?

Yes exactly.  On the first method - we currently fail if we are given a file that doesn't exist...but won't complain about an empty one.  So to give a specific location to users of "just docker" we'd need to embed an empty file for them to mount over.

desimone:
> Yes exactly. On the first method - we currently fail if we are given a file that doesn't exist...but won't complain about an empty one. So to give a specific location to users of "just docker" we'd need to embed an empty file for them to mount over.

Then I agree, I think the first option makes sense.

config.example.yaml
```diff
@@ -0,0 +1,89 @@
+# Main configuration flags
+address: ":8443"     # optional, default is 443
```
desimone: 👍 I like it.

internal/config/options.go
```diff
@@ -146,6 +153,137 @@ func NewOptions() *Options {
...
+	if !IsValidService(o.Services) {
+		return fmt.Errorf("%s is an invalid service type", o.Services)
+	}
+
+	if o.SharedKey == "" {
+		return errors.New("shared-key cannot be empty")
+	}
+
+	if len(o.Routes) != 0 {
+		return errors.New("routes setting is deprecated, use policy instead")
```
desimone: 👍

travisgroth:
>sans those failing tests
Yep working on those and some refactoring.  Down to one test so hopefully wrapping this morning.

> do you mind squashing

Yep no problem.  I kept the commits small so you can review the changes in context.  Once it's final and passing, I'll squash.

I'm getting merge conflicts with #134 so could we tag @nareddyt for review as well?  It seems easy to follow, but if he can take a gander it would be a good double check.  I need to move a few things around since we validate in different places now.

travisgroth:
Don't merge yet. Need one last to do cleaned up. That should fix coverage.

desimone:
\cc @nareddyt if you have a moment to review, would love your feedback on this PR.

travisgroth:
Should be good to go now. 🎉

desimone:
Pulled this down and everything works as expected. Thanks again @travisgroth

nareddyt:
Hey sorry, I'm traveling for the long weekend so I'm a little busy. I took a quick look and LGTM, seems like a great feature for enterprises!

I'll take a closer look on Tuesday. Feel free to cc me on any other requests and I'll look at those too after the weekend.

nareddyt:
Sorry for the delay. LGTM. Moving the public policy tests from `proxy_test` to `policy_test` makes sense :)

travisgroth:
Glad to get a confirmation.  Thanks!

Issue #132, opened 2019-05-21
proxy: Proxy configuration file preprocessed incorrectly, creating incorrect routes


nareddyt:
## Bug Description

When configuring a proxy locally with a `policy.yaml` file, some of the routes created are incorrect. This leads to the proxy routing requests incorrectly.

### Environment 

- Running the source locally. Latest commit to the `master` branch
- Windows 10 with Windows Subsystem for Linux

### Impact

Minimal: I cannot test my fix to #129 with this bug in place. However, I believe I have a fix for this bug (below).

### Root Cause

Specifically, I noticed that [Proxy::routeConfigs](https://github.com/pomerium/pomerium/blob/3eff6cce13e51f97e47cd2c4aacbc82fe9646bcb/proxy/proxy.go#L117) is incorrectly initialized. In the image below, notice that two routes (with the configuration file given below) have the same memory address for the `policy` - highlighted in red. This should not occur, as each policy should be unique.

![image](https://user-images.githubusercontent.com/11142171/58129623-25e67480-7be8-11e9-9ad7-caa6eb115627.png)

After some further digging, I noticed issues with memory addresses [in this section of code](https://github.com/pomerium/pomerium/blob/3eff6cce13e51f97e47cd2c4aacbc82fe9646bcb/proxy/proxy.go#L180). `p.Handle(route.Source.Host, handler, &route)`, where the last argument is a pointer to route. However, the proxy must store **a copy of the route** for later requests outside of this function.

### Proposed Fix

I will make a PR with the fix. The main modification would be [here](https://github.com/pomerium/pomerium/blob/3eff6cce13e51f97e47cd2c4aacbc82fe9646bcb/proxy/proxy.go#L122) to store policies as a copy of the data, instead of pointers that do not exist after the main function exits.

### Policy Configuration

```yaml
- from: hello1.homelab.tejunareddy.com
  to: http://localhost:8081
  allowed_users:
    - tejunareddy@gmail.com
- from: hello2.homelab.tejunareddy.com
  to: http://localhost:8082
  allowed_users:
    - not_a_user@yahoo.com
- from: hello3.homelab.tejunareddy.com
  to: http://localhost:8083
  allow_public_unauthenticated_access: true
```

### Repro Steps

1. Start the server locally with this file
[...]

PR #134, opened 2019-05-21
proxy: support for public unauthenticated routes

### LINKED ISSUE #129: Proxy: Support for unauthenticated routes

Routes from pomerium to different backend services can be configured with the pomerium policy file. These routes can be configured with whitelisted users, groups, and domains.

However, the documentation describes no way to configure a public route. I understand that the point of Pomerium is to serve as a IAP, but some services behind the IAP may not require authentication.

I'd like a way to create unauthenticated routes in the policy file. Something like:

```yaml
- from: hello.corp.beyondperimeter.com
  to: http://hello:8080
  allowed_public: true
```

Then any requests to https://hello.corp.beyondpermimeter.com would be proxied directly to http://hello:8080 without needing to authenticate with an identity provider.

A work-around would be configure a reverse proxy (like nginx) in front of pomerium. This reverse proxy would proxy all public traffic and only send authenticated routes to pomerium.

This workaround increases the complexity of the architecture because 2 reverse proxies must be managed. Additionally, it's not clear how I would setup SSL termination on the initial reverse proxy and disable it on pomerium.

**Additional context**

I am configuring pomerium with docker-compose.

Please let me know if I misunderstood the architecture of pomerium or if there is an easier workaround. If this is a planned feature, I'd love help and contribute to pomerium!

nareddyt:
Fixes #129

### Implementation Description

- Implemented a new access control option that allows public, unauthenticated access to specific targets.
- Added bypass check to `Proxy::shouldSkipAuthentication`

### Testing Done

- Ensured all unit tests pass in `proxy/`
- Ran the following manual integration test

#### Policy Configuration

```yaml
- from: hello1.homelab.tejunareddy.com
  to: http://localhost:8081
  allowed_users:
    - tejunareddy@gmail.com
- from: hello2.homelab.tejunareddy.com
  to: http://localhost:8082
  allowed_users:
    - not_a_user@yahoo.com
- from: hello3.homelab.tejunareddy.com
  to: http://localhost:8083
```

#### Steps

1. Sign out of pomerium with `/.pomerium/signout`
2. Ensure 200 OK for hello3.homelab.tejunareddy.com, but 403 Forbidden for hello1 and hello2.
3. Sign into tejunareddy@gmail.com
4. Ensure 400 OK for hello3 and hello1, but 403 Forbidden for hello2.

### Other Notes

In order for this to work correctly, #132 must be fixed. I locally used the code in #133 to fix this issue, and tested this PR on top of that.

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] ready for review

### DIFF
--- a/docs/reference/readme.md
+++ b/docs/reference/readme.md
@@ -185,6 +185,18 @@ Allowed domains is a collection of whitelisted domains to authorize for a given 
 
 Allow unauthenticated HTTP OPTIONS requests as [per the CORS spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests).
 
+#### Public Access
+
+- `yaml`/`json` setting: `allow_public_unauthenticated_access`
+- Type: `bool`
+- Optional
+- Default: `false`
+
+**Use with caution:** Allow all requests for a given route, bypassing authentication and authorization. 
+Suitable for publicly exposed web services.
+
+If this setting is enabled, no whitelists (e.g. Allowed Users) should be provided in this route.
+
 #### Route Timeout
 
 - `yaml`/`json` setting: `timeout`
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -1,6 +1,7 @@
 package policy // import "github.com/pomerium/pomerium/internal/policy"
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/url"
@@ -29,6 +30,9 @@ type Policy struct {
 	// https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests
 	CORSAllowPreflight bool `yaml:"cors_allow_preflight"`
 
+	// Allow any public request to access this route. **Bypasses authentication**
+	AllowPublicUnauthenticatedAccess bool `yaml:"allow_public_unauthenticated_access"`
+
 	// UpstreamTimeout is the route specific timeout. Must be less than the global
 	// timeout. If unset,  route will fallback to the proxy's DefaultUpstreamTimeout.
 	UpstreamTimeout time.Duration `yaml:"timeout"`
@@ -39,10 +43,17 @@ func (p *Policy) validate() (err error) {
 	if err != nil {
 		return err
 	}
+
 	p.Destination, err = urlParse(p.To)
 	if err != nil {
 		return err
 	}
+
+	// Only allow public access if no other whitelists are in place
+	if p.AllowPublicUnauthenticatedAccess && (p.AllowedDomains != nil || p.AllowedGroups != nil || p.AllowedEmails != nil) {
+		return errors.New("route marked as public but contains whitelists")
+	}
+
 	return nil
 }
 
@@ -56,7 +67,7 @@ func FromConfig(confBytes []byte) ([]Policy, error) {
 	// build source and destination urls
 	for i := range f {
 		if err := (&f[i]).validate(); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("route at index %d: %v", i, err)
 		}
 	}
 	return f, nil
--- a/proxy/handlers.go
+++ b/proxy/handlers.go
@@ -180,8 +180,14 @@ func (p *Proxy) OAuthCallback(w http.ResponseWriter, r *http.Request) {
 // Conditions should be few in number and have strong justifications.
 func (p *Proxy) shouldSkipAuthentication(r *http.Request) bool {
 	pol, foundPolicy := p.policy(r)
+
 	if isCORSPreflight(r) && foundPolicy && pol.CORSAllowPreflight {
 		log.FromRequest(r).Debug().Msg("proxy: skipping authentication for valid CORS preflight request")
+		return true
+	}
+
+	if foundPolicy && pol.AllowPublicUnauthenticatedAccess {
+		log.FromRequest(r).Debug().Msg("proxy: skipping authentication for public route")
 		return true
 	}
 
--- a/policy.example.yaml
+++ b/policy.example.yaml
@@ -19,3 +19,6 @@
   to: http://hello:8080
   allowed_groups:
     - admins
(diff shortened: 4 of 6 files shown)


docs/reference/readme.md
```diff
@@ -185,6 +185,19 @@ Allowed domains is a collection of whitelisted domains to authorize for a given
 
 Allow unauthenticated HTTP OPTIONS requests as [per the CORS spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests).
 
+#### Public Access
+
+- `yaml`/`json` setting: `allow_public_unauthenticated_access`
```
nareddyt: Let me know if this name is OK, or if you'd prefer something along the lines of `disable_all_access_control`

proxy/handlers.go
```diff
@@ -185,6 +185,18 @@ func (p *Proxy) shouldSkipAuthentication(r *http.Request) bool {
...
+	if foundPolicy && pol.AllowPublicUnauthenticatedAccess {
+
+		// Sanity Check: Only allow public access if no other whitelists are in place
+		if pol.AllowedDomains == nil && pol.AllowedGroups == nil && pol.AllowedEmails == nil {
+			log.FromRequest(r).Debug().Msg("proxy: skipping authentication for public route")
+			return true
+		}
+
+		log.FromRequest(r).Error().Msg("proxy: route marked as public contains whitelists, " +
+			"disabling public access and using whitelists")
```
nareddyt: Alternatively, this check can be placed in [internal\policy\policy.go::validate](https://github.com/pomerium/pomerium/blob/409f44a9be1d8bbdf768df2b508b7d427d279f72/internal/policy/policy.go#L37). Would it make more sense to do this validation when the server is started and just completely disallow both options from being enabled? I admit it does seem weird running this check in the dataplane when we could have ran it at startup in the controlplane

desimone: Yeah, I think it would more sense we should pop an error when loading the policy instead. Saves resources, and avoids any user confusion.

Plus, I was thinking that at some point it would be nice to be able to abstract out a the policy validate logic as a stand-alone linter so admins can check their policy prior to any updates.

docs/reference/readme.md
```diff
@@ -185,6 +185,19 @@ Allowed domains is a collection of whitelisted domains to authorize for a given
...
+- Type: `bool`
+- Optional
+- Default: `false`
+
+**Use with caution:** Allow all requests for a given route, bypassing authentication.
```
desimone: Sort of obvious but (how do you authorize without authenticating first), but we might as well be explicit that this bypasses all authentication **and** authorization.

docs/reference/readme.md
desimone: I can't think of anything better. `disable_all_access_control` **is** descriptive. 

I'll try to see what other products call their "turn-off-all-access-control" flags just to see but I think this is fine.

nareddyt:
Moved the validation from the data plane to the control plane. The following error gets printed when a policy contains both a whitelist and public access:

```
FTL cmd/pomerium: authorize error="authorize: `POLICY_FILE` route at index 2: route marked as public but contains whitelists"
```

Additional changes:

- Updated docs
- Updated unit tests
- Ran manual integration tests (described above)

Still haven't decided on the name, let me know if that needs to be changed to `disable_all_access_control` :)

internal/policy/policy.go
```diff
@@ -39,10 +43,17 @@ func (p *Policy) validate() (err error) {
...
 		return err
 	}
+
 	p.Destination, err = urlParse(p.To)
 	if err != nil {
 		return err
 	}
+
+	// Only allow public access if no other whitelists are in place
+	if p.AllowPublicUnauthenticatedAccess && (p.AllowedDomains != nil || p.AllowedGroups != nil || p.AllowedEmails != nil) {
```
desimone: Could we add a test or two for this?

docs/reference/readme.md
desimone: You pick, I'm open to either. 👍

internal/policy/policy.go
nareddyt: I added two tests to `proxy/handlers_test.go` that should test for this. Code coverage report shows this branch is tested. Is there a more in-depth test you'd like to see?

desimone: Maybe I've got this wrong but this but it shows as uncovered for me (this would be for policy, not handlers right?)

https://codecov.io/gh/pomerium/pomerium/compare/c53c07c274f23d97f24e56ab1e75e8a5a288b92c...0952a1de46ba5faf1828567c228d536c8a810775/src/internal/policy/policy.go#L54

nareddyt: Sorry i meant `proxy/proxy_test.go`

desimone: Ah got it, yeah it's weird that codecov isn't showing this as covered as 

```go
{"policy public and whitelist bad", publicWithWhitelistPolicy, true},
``` 

should be doing this right?

nareddyt: Yeah, that line should be checking that. That's really weird. I see that branch covered after running the `TestOptions_Validate` function locally.

image

nareddyt:
By the way, I'm having a lot of trouble squashing my commits. It might be because I merged `master` into the branch a few times. My bad on that, I'll remember to always rebase onto master for any future contributions.

Do you want to try squashing them? I tried and pushed the squashed commit to [this branch on my fork](https://github.com/nareddyt/pomerium/tree/unauthenticated-routes-squash), but it shows the diff for all the merges to master as well...

internal/policy/policy.go
desimone: LGTM then 👍

desimone:
I can squash your commits no problem, the only issue is the git commit history will look like I made the commits and not you. 

Does `git rebase -i master` work?

nareddyt:
Fixed! I did do `git rebase -i master`, but I forgot to fetch from upstream into my `master` branch. The diff looks good now

desimone:
🎉 thank you for the contribution!

PR #139, opened 2019-05-26
deployment: expose config file interface in docker images

travisgroth:
Part of #115

- Existing env-only users should be unaffected.  
- Adds a default yaml location for users to mount a config over 
- Moves the config option to CMD so that pomerium args can be specified at runtime when desired 
  - Example use case is if someone wanted to pass their config as toml or json for some reason
  ```shell
  08:23 $ docker run -it pomerium -help
  Usage of /bin/pomerium:
    -config string
    	  Specify configuration file location
    -version
    	  prints the version
  ```

- I slurped in a test coverage fix that took my last PR down 0.1% and was annoying me 😄 

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] ready for review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,11 @@ RUN go mod download
 COPY . .
 # build 
 RUN make
+RUN touch /config.yaml
 
 FROM gcr.io/distroless/static
 WORKDIR /pomerium
 COPY --from=build /go/src/github.com/pomerium/pomerium/bin/* /bin/
-CMD ["/bin/pomerium"]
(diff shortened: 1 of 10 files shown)


travisgroth:
Hah yep shame is a powerful tool

desimone:
If you don't mind squashing, this is good to merge. 👍

travisgroth:
Squashed to two (seems like they should be separate).  Let me know if you really want it down to one.

desimone:
Two makes sense here. 👍

PR #142, opened 2019-05-27
all: fix incorrect http status codes #135

### LINKED ISSUE #135: all: use of incorrect http status codes
**Describe the bug**

Currently, our handlers return a grab bag of http status code errors that do not aptly describe the underlying error. 

**To Reproduce** Steps to reproduce the behavior:
A few examples
- [Should be 401](https://github.com/pomerium/pomerium/blob/master/proxy/handlers.go#L226-L230)
- [Should be 400](https://github.com/pomerium/pomerium/blob/702cc30b77c1f4251ef270fc24e0ea99e1dba685/proxy/handlers.go#L112)
- [should be 400](https://github.com/pomerium/pomerium/blob/702cc30b77c1f4251ef270fc24e0ea99e1dba685/proxy/handlers.go#L155)

And so on. 

**Expected behavior**

We should be good HTTP citizens and use the right status codes. The biggest offenders right now are around rejecting access (4XX).

**Additional context**

- [Mozilla HTTP Status Codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)

u5surf:
Fixes #135 
I fix proxy and authenticate hander returns correct status code.

**Checklist**:
- [o] updated docs
Nothing?
- [o] unit tests added
Fix some unit test following status code changing
- [o] related issues referenced
#135
- [o] ready for review


authenticate/handlers.go
```diff
@@ -97,7 +97,13 @@ func (a *Authenticate) SignIn(w http.ResponseWriter, r *http.Request) {
...
 	state := r.Form.Get("state")
 	if state == "" {
-		httputil.ErrorResponse(w, r, "no state parameter supplied", http.StatusForbidden)
+		httputil.ErrorResponse(w, r, "no state parameter supplied", http.StatusBadRequest)
+		return
+	}
+
+	redirectURI := r.Form.Get("redirect_uri")
+	if redirectURI == "" {
+		httputil.ErrorResponse(w, r, "no redirect_uri parameter supplied", http.StatusBadRequest)
```
desimone: Could you add a unit test for this case?

authenticate/handlers.go
```diff
@@ -97,7 +97,13 @@ func (a *Authenticate) SignIn(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
```
desimone: I think we can avoid the duplicate `r.Form.Get` call on line 110 and use `redirectURI`

```go
	redirectURL, err := url.Parse(redirectURI)
```

authenticate/handlers.go
u5surf: @desimone  Oh.., sorry, I think I mis-rebased upstream master...It is correct that this diff is not necessary. 
https://github.com/pomerium/pomerium/commit/66b4c2d3cd368b26fdb1ce905d4935f0cf3e940b#diff-d011e1d3c12517c5bd8581ef91d69fa3L107

desimone: No worries! Thank you 👍

desimone:
Thank you @u5surf !

PR #150, opened 2019-05-31
Feature/hot policy

### LINKED ISSUE #89: proxy: allow hot-reloading of policy on change
Right now when you edit `policy.yaml`, you have to restart Pomerium (at least for users who run Pomerium in docker).

It would be great if Pomerium can monitor this file and trigger a reload upon file change (or with a given interval).

Ideally Pomerium should only reload when the semantic of the policy file is changed. E.g. when someone add a comment line, it should not cause a reload. But if it's too hard I'm fine with any change (e.g. compare the hash with an interval).

travisgroth:
Closes #89 

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] ready for review

### DIFF
--- a/authorize/authorize.go
+++ b/authorize/authorize.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/pomerium/pomerium/internal/log"
+
 	"github.com/pomerium/pomerium/internal/config"
 
 	"github.com/pomerium/pomerium/internal/policy"
@@ -62,4 +64,11 @@ func NewIdentityWhitelist(policies []policy.Policy, admins []string) IdentityVal
 // ValidIdentity returns if an identity is authorized to access a route resource.
 func (a *Authorize) ValidIdentity(route string, identity *Identity) bool {
 	return a.identityAccess.Valid(route, identity)
+}
+
+// UpdateOptions updates internal structres based on config.Options
+func (a *Authorize) UpdateOptions(o *config.Options) error {
+	log.Info().Msg("authorize: updating options")
+	a.identityAccess = NewIdentityWhitelist(o.Policies, o.Administrators)
+	return nil
 }
--- a/cmd/pomerium/main.go
+++ b/cmd/pomerium/main.go
@@ -8,8 +8,7 @@ import (
 	"os"
 	"time"
 
-	"google.golang.org/grpc"
-
+	"github.com/fsnotify/fsnotify"
 	"github.com/pomerium/pomerium/authenticate"
 	"github.com/pomerium/pomerium/authorize"
 	"github.com/pomerium/pomerium/internal/config"
@@ -21,6 +20,8 @@ import (
 	pbAuthenticate "github.com/pomerium/pomerium/proto/authenticate"
 	pbAuthorize "github.com/pomerium/pomerium/proto/authorize"
 	"github.com/pomerium/pomerium/proxy"
+	"github.com/spf13/viper"
+	"google.golang.org/grpc"
 )
 
 var versionFlag = flag.Bool("version", false, "prints the version")
@@ -48,15 +49,24 @@ func main() {
 		log.Fatal().Err(err).Msg("cmd/pomerium: authenticate")
 	}
 
-	_, err = newAuthorizeService(opt, grpcServer)
+	authz, err := newAuthorizeService(opt, grpcServer)
 	if err != nil {
 		log.Fatal().Err(err).Msg("cmd/pomerium: authorize")
 	}
 
-	_, err = newProxyService(opt, mux)
+	proxy, err := newProxyService(opt, mux)
 	if err != nil {
 		log.Fatal().Err(err).Msg("cmd/pomerium: proxy")
 	}
+	go viper.WatchConfig()
+
+	viper.OnConfigChange(func(e fsnotify.Event) {
+		log.Info().
+			Str("file", e.Name).
+			Msg("cmd/pomerium: configuration file changed")
+
+		opt = handleConfigUpdate(opt, []config.OptionsUpdater{authz, proxy})
+	})
 	// defer statements ignored anyway :  https://stackoverflow.com/a/17888654
 	// defer proxyService.AuthenticateClient.Close()
 	// defer proxyService.AuthorizeClient.Close()
@@ -180,4 +190,41 @@ func parseOptions(configFile string) (*config.Options, error) {
 		log.SetLevel(o.LogLevel)
 	}
 	return o, nil
+}
+
+func handleConfigUpdate(opt *config.Options, services []config.OptionsUpdater) *config.Options {
+	newOpt, err := parseOptions(*configFile)
+	optChecksum := opt.Checksum()
+	newOptChecksum := newOpt.Checksum()
+
+	log.Debug().
+		Str("old-checksum", optChecksum).
+		Str("new-checksum", newOptChecksum).
+		Msg("cmd/pomerium: configuration file changed")
+
+	if newOptChecksum == optChecksum {
+		log.Debug().Msg("cmd/pomerium: loaded configuration has not changed")
+		return opt
+	}
+
+	if err != nil {
+		log.Error().
+			Err(err).
+			Msg("cmd/pomerium: could not reload configuration")
+		return opt
+	}
+
+	log.Info().
+		Str("config-checksum", newOptChecksum).
+		Msg("cmd/pomerium: running configuration has changed")
+	for _, service := range services {
+		err := service.UpdateOptions(newOpt)
+		if err != nil {
+			log.Error().
+				Err(err).
+				Msg("cmd/pomerium: could not update options")
+		}
+	}
+
+	return newOpt
 }
--- a/docs/reference/readme.md
+++ b/docs/reference/readme.md
@@ -15,6 +15,8 @@ If you are coming from a kubernetes or docker background this should feel famili
 
 In general, any setting specified by environment variable can also be present in the optional config file as the same name but lower cased. Environment variables take precedence.
 
+Pomerium will automatically reload the configuration file if it is changed.  At this time, only policy is re-configured when this reload occurs, but additional options may be added in the future.  It is suggested that your policy is stored in a configuration file so that you can take advantage of this feature.
+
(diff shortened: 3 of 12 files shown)


travisgroth:
FYI this glues together the logic to pick up any changes from the config file.  Services can implement whatever they want/can as far as reflecting changes to their internal data structures.

authorize/authorize.go
```diff
@@ -63,3 +65,10 @@ func NewIdentityWhitelist(policies []policy.Policy, admins []string) IdentityVal
 func (a *Authorize) ValidIdentity(route string, identity *Identity) bool {
 	return a.identityAccess.Valid(route, identity)
 }
+
+// UpdateOptions updates internal structres based on config.Options
+func (a *Authorize) UpdateOptions(o *config.Options) error {
```
desimone: Should this be returning an error ever?

cmd/pomerium/main.go
```diff
@@ -8,6 +8,9 @@ import (
 	"os"
 	"time"
 
+	"github.com/fsnotify/fsnotify"
+	"github.com/spf13/viper"
+
```
desimone: nit: remove newline so imports are grouped and sorted

cmd/pomerium/main.go
```diff
@@ -181,3 +190,25 @@ func parseOptions(configFile string) (*config.Options, error) {
 	}
 	return o, nil
 }
+
+func handleConfigUpdate(opt *config.Options, services []config.OptionsUpdater) *config.Options {
+	newOpt, err := parseOptions(*configFile)
+	log.Debug().Msgf("cmd/pomerium: old config = %s, new config = %s", opt.Checksum(), newOpt.Checksum())
```
desimone: nit: thoughts on using something like the following log invocation? ( I believe this is 0 allocations too)
```go
log.Debug().Str("old-config", opt.Checksum()).Str("new-config", newOpt.Checksum().Msg("cmd/pomerium: config update")
```

docs/reference/readme.md
```diff
@@ -15,6 +15,8 @@ If you are coming from a kubernetes or docker background this should feel famili
 
 In general, any setting specified by environment variable can also be present in the optional config file as the same name but lower cased. Environment variables take precedence.
 
+Pomerium will automatically reload the configuration file if it is changed.  At this time, only policy is re-configured when this reload occurs, but additional options may be added in the future.  It is suggested that your policy is stored in a configuration file so that you can take advantage of this feature.d
```
desimone: 👍 love it. Small typo at the end of the line (feature.d)

internal/config/options.go
```diff
@@ -208,6 +210,7 @@ func OptionsFromViper(configFile string) (*Options, error) {
 		return nil, err
 	}
 
+	log.Debug().Msgf("Read configuration with checksum: %s", o.Checksum())
```
desimone: See above comment about preferring structured logging vs Msgf

go.mod
```diff
@@ -3,6 +3,8 @@ module github.com/pomerium/pomerium
 go 1.12
 
 require (
+	github.com/cnf/structhash v0.0.0-20180104161610-62a607eb0224
```
desimone: I'm a little bit concerned about cnf/structhash just because it looks unmaintained. 

Maybe you already considered it, but I'm curious what you think of using  https://github.com/mitchellh/hashstructure which (I believe) is identical in functionality and is a dependency/funded by hashicorp products.

authorize/authorize.go
travisgroth: Not in authorize as written, but I believe proxy can.  It kind of depends on what the services are doing when you tell them to update.  Since they're in charge of option validation and then doing something with it, it is possible that they will have an issue and want to signal it back up the call stack.  I'm open to changing the interface spec if you don't think it adds value, though.

go.mod
travisgroth: Ah.  Yes I'm open to switching back if hashicorp uses hashstructure.  I had some issues with getting consistent hashes from it originally and tried swapping libraries just to isolate the cause.  

I got it working after the switch and didn't consider going back.  Easy change IIRC.

travisgroth:
Any idea how the coverage on main() dropped?  I'm mildly annoyed by that but I'm having trouble figuring out how that code was considered covered before.

cmd/pomerium/main.go
travisgroth: I like it.  Wasn't familiar with zerolog.  This makes sense after looking at their docs.  👍

internal/config/options.go
```diff
@@ -349,3 +355,18 @@ func IsProxy(s string) bool {
...
+// OptionsUpdater updates local state based on an Options struct
+type OptionsUpdater interface {
+	UpdateOptions(*Options) error
+}
+
+// Checksum returns the checksum of the current options struct
+func (o *Options) Checksum() string {
+	hash, err := hashstructure.Hash(o, nil)
+
+	if err != nil {
```
travisgroth: I debated a bit about what to do here since there is somehow a possibility of an error.  I think the safest thing to do is nothing but another option might be to return a value that will change once a minute or something along those lines.  

The only current use case would be running this _if_ the file changes but thinking further down the road (prometheus returning this, for instance) that might be a bad assumption.  Open to ideas.

internal/config/options.go
travisgroth: Should be addressed.  See if it looks good now.

cmd/pomerium/main.go
travisgroth: Updated.  Please review the new logging approach.

authorize/authorize.go
desimone: No, I think that makes sense.

cmd/pomerium/main.go
```diff
@@ -181,3 +191,38 @@ func parseOptions(configFile string) (*config.Options, error) {
...
+		Msg("cmd/pomerium: configuration file changed")
+
+	if newOptChecksum == optChecksum {
+		log.Debug().Msg("cmd/pomerium: loaded configuration has not changed")
+		return opt
+	}
+
+	if err != nil {
+		log.Error().
+			Str("error", err.Error()).
```
desimone: .Err(err) is a shorthand :)

internal/config/options.go
desimone: That's a really good point. I thought about this a bit as well and all the alternative methods (e.g. clock based or on some create date field) I can come up with feels like a path to unexpected edge cases.

I think you are right that the best thing to do is nothing; and if it turns out there are cases we didn't anticipate where the hashstruct fails, we can address it then. 

Maybe the one thing we could do is throw a warn log message.

cmd/pomerium/main.go
desimone: The result is essentially identical but rs log has a little wrapper for `Err(err)`. https://github.com/rs/zerolog#advanced-fields

cmd/pomerium/main.go
```diff
@@ -181,3 +191,38 @@ func parseOptions(configFile string) (*config.Options, error) {
...
+		return opt
+	}
+
+	log.Info().
+		Str("config-checksum", newOptChecksum).
+		Msg("cmd/pomerium: running configuration has changed")
+	for _, service := range services {
+		err := service.UpdateOptions(newOpt)
+		if err != nil {
+			log.Error().Str("error", err.Error()).Msg("cmd/pomerium: could not update options")
```
desimone: `log.Error().Err(err).`

go.mod
```diff
@@ -3,9 +3,11 @@ module github.com/pomerium/pomerium
 go 1.12
 
 require (
+	github.com/fsnotify/fsnotify v1.4.7
 	github.com/golang/mock v1.2.0
 	github.com/golang/protobuf v1.3.1
 	github.com/magiconair/properties v1.8.1 // indirect
+	github.com/mitchellh/hashstructure v1.0.0
```
desimone: 👍

desimone:
> Any idea how the coverage on main() dropped? I'm mildly annoyed by that but I'm having trouble figuring out how that code was considered covered before.
> 
> ~Edit: nevermind? It's up now. Thanks codecov??~ Overall coverage is up but main shows a drop still. 🤷‍♂

Yeah, this can drive me insane as well. My _best_ guess for this particular commit is that the [following lines are uncovered](https://codecov.io/gh/pomerium/pomerium/compare/1791e3d3f7b909ca6d837d94b9871130e0836053...6b25e46bb3c5d2711d137cc0e172ca3d8ab2fc51/src/cmd/pomerium/main.go#L210).

FWIW, I think the test coverage looks good 😄

internal/config/options.go
travisgroth: >Maybe the one thing we could do is throw a warn log message.

 Right; good call.  I look forward to figuring out the corner case that triggers it.

cmd/pomerium/main.go
travisgroth: 👍

internal/config/options.go
```diff
@@ -70,8 +70,8 @@ type Options struct {
 
 	// AuthenticateURL represents the externally accessible http endpoints
 	// used for authentication requests and callbacks
-	AuthenticateURLString string `mapstructure:"authenticate_service_url"`
-	AuthenticateURL       *url.URL
+	AuthenticateURLString string   `mapstructure:"authenticate_service_url"`
+	AuthenticateURL       *url.URL `hash:"ignore"`
```
travisgroth: Going to open an issue on this.  The workaround here is fine but I believe something is mutating this at startup.  Unclear if that's expected.

desimone: Could it be the empty initialization? https://github.com/pomerium/pomerium/blob/1a76a3ad8601aa40aed905204c47c0ee573a6760/internal/config/options.go#L156

travisgroth:
> is that the following lines are uncovered.

Yeah...but I don't see how they weren't like that already.  Anyway; yeah I'm fine with the coverage too, just looks weird.

internal/config/options.go
travisgroth: I don't think so.  It only impacts one of the two URLs and it only happens the first pass through services at start.  I'll follow up researching more thoroughly later that so we can get this merged.

PR #151, opened 2019-05-31
Enable websocket support

### LINKED ISSUE #77: Demonstrating websocket support?
**Describe the bug**

I am unable to confirm that websocket support works in 0.0.3.

**To Reproduce**

1. Deployed Pomerium 0.0.3
2. Tried to connect to otherwise-working proxied apps where websocket-based features failed
3. Websockets were unable to connect
4. Added a proxy from https://websocket.corp.my-redacted-domain.com to https://websockets.org
5. Demo was unable to connect to wss://websocket.corp.my-redacted-domain.com/
6, Deployed Kaazing gateway at http://kaazing.corp.my-redacted-domain.com:8000
7. Successfully demonstrated insecure websockets on port 8000
8. Configured a policy to point at local gateway
9. Local gateway cannot connect to websocket (this *might* be a Kaazing issue; it's expecting a particular host value, I think.)

**Expected behavior**

Websocket support works.

**Environment:**

- Pomerium version (retrieve with `pomerium --version`): v0.0.3+41c42f5
- Server Operating System/Architecture/Cloud: Synology x86 NAS running docker

Configuration file(s):

policy.yaml:
```yaml
- from: kaazing.corp.my-redacted-domain.com
  to: http://kaazing.corp.my-redacted-domain.com:8000
  allowed_users:
  - <me>
- from: httpbin.corp.my-redacted-domain.com
  to: https://httpbin.org
  allowed_users:
  - <me>
- from: websocket.corp.my-redacted-domain.com
  to: https://websocket.org/
  allowed_users:
  - <me>
```

Logs(s):

There appear to be no logs associated with websocket connection attempts.

**Additional context**

Pomerium is bound directly to port 443.

Happy to add any Pomerium folks who want to try these routes to my allowed users (and share with them the actual URLs involved.) This is not a production system so I'm happy to make sweeping changes.

Connecting to wss://httpbin.corp.my-redacted-domain.com/anything yields an error 200; curiously, nothing is in the logs for this request. `POMERIUM_DEBUG` is true and `LOG_LEVEL` is default; l I am seeing DBG logs.

nareddyt:
Fixes #77 

## Description

Allows proxy to handle websocket connections. 

### Major changes:

- Fix hardcoded HTTP protocol version. Hardcoded to HTTP2, which does not support websockets (at least not well)
- Prevent websocket connections from being proxied through a `timeoutHandler`, as websockets are long-lived connections and should not have timeouts enforced on them.
- Added configuration variable to opt-in to websocket support.

### Other Notes

This is an opt-in feature because it can be potentially abused (see discussion in this PR). By opting-in, the user acknowledges this risk.

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] ready for review

### DIFF
--- a/internal/log/middleware.go
+++ b/internal/log/middleware.go
@@ -164,7 +164,7 @@ func AccessHandler(f func(r *http.Request, status, size int, duration time.Durat
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			start := time.Now()
-			lw := NewWrapResponseWriter(w, 2)
+			lw := NewWrapResponseWriter(w, r.ProtoMajor)
 			next.ServeHTTP(lw, r)
 			f(r, lw.Status(), lw.BytesWritten(), time.Since(start))
 		})
--- a/proxy/handlers.go
+++ b/proxy/handlers.go
@@ -3,6 +3,7 @@ package proxy // import "github.com/pomerium/pomerium/proxy"
 import (
 	"encoding/base64"
 	"fmt"
+	"github.com/pomerium/pomerium/internal/config"
 	"net/http"
 	"net/url"
 	"strings"
@@ -492,4 +493,28 @@ func (p *Proxy) GetSignOutURL(authenticateURL, redirectURL *url.URL) *url.URL {
 
 func extendDeadline(ttl time.Duration) time.Time {
 	return time.Now().Add(ttl).Truncate(time.Second)
+}
+
+// websocketHandlerFunc splits request serving with timeouts depending on the protocol
+func websocketHandlerFunc(baseHandler http.Handler, timeoutHandler http.Handler, o *config.Options) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		// Do not use timeouts for websockets because they are long-lived connections.
+		if r.ProtoMajor == 1 &&
+			strings.EqualFold(r.Header.Get("Connection"), "upgrade") &&
+			strings.EqualFold(r.Header.Get("Upgrade"), "websocket") {
+
+			if o.AllowWebsockets {
+				baseHandler.ServeHTTP(w, r)
+				return
+			}
+
+			log.FromRequest(r).Warn().Msg("proxy: attempt to proxy a websocket connection, but websocket support is disabled in the configuration")
+			httputil.ErrorResponse(w, r, "websockets not supported by proxy", http.StatusBadRequest)
+			return
+		}
+
+		// All other non-websocket requests are served with timeouts to prevent abuse
+		timeoutHandler.ServeHTTP(w, r)
+	})
 }
--- a/docs/reference/readme.md
+++ b/docs/reference/readme.md
@@ -143,6 +143,19 @@ Timeouts set the global server timeouts. For route-specific timeouts, see [polic
 
 If set, the HTTP Redirect Address specifies the host and port to redirect http to https traffic on. If unset, no redirect server is started.
 
+### Websocket Connections
+
+- Environmental Variable: `ALLOW_WEBSOCKETS`
+- Config File Key: `allow_websockets`
+- Type: `bool`
+- Default: `false`
+
+If set, enables proxying of websocket connections. 
+Otherwise the proxy responds with `400 Bad Request` to all websocket connections.
+
+**Use with caution:** By definition, websockets are long-lived connections, so [global timeouts](#global-timeouts) are not enforced.
+Allowing websocket connections to the proxy could result in abuse via DOS attacks. 
+
 ### Policy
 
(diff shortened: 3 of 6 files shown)


desimone:
My first thought was to do something like [this](https://github.com/pomerium/pomerium/blob/7487de94df7c0c585ea8123bf7c1358eaa98fde3/internal/https/https.go#L196-L207) and force downgrade (or use the appropriate fancywriter) as middleware. 

[Relevant comment](https://github.com/golang/go/issues/14797#issuecomment-196103814) (I first thought we could just implement a hijacker for the http2FancyWriter) about hijacker and http2 being incompatible.

internal/log/middleware.go
```diff
@@ -164,7 +164,7 @@ func AccessHandler(f func(r *http.Request, status, size int, duration time.Durat
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			start := time.Now()
-			lw := NewWrapResponseWriter(w, 2)
+			lw := NewWrapResponseWriter(w, 1)
```
desimone:
```go 
// The protocol version for incoming server requests.
//
// For client requests, these fields are ignored. The HTTP
// client code always uses either HTTP/1.1 or HTTP/2.
// See the docs on Transport for details.
Proto      string // "HTTP/1.0"
ProtoMajor int    // 1
ProtoMinor int    // 0
```
```go
lw := NewWrapResponseWriter(w, r.ProtoMajor)
````

🤷‍♂ might work? Thoughts?

Comment (nareddyt) [reply internal/log/middleware.go]: Oh maybe, let me try that. I'll also play around with that you said in the above comment.

**Comment by nareddyt** [comment]
Your suggestion works! I tested it with Home Assistant and Jupyter backends, taking note of the protocol versions in chrome dev tools. All http2 requests are being handled correctly, and all websocket requests are being hijacked as http1.1 requests.

I also added simple switch like you suggested. Now the proxy does not use the `TimeoutHandler` for websocket connections, but defaults to using it for all other requests.

I updated the PR with this code. Could you take a look? Let me know if I need to re-factor anything as well. I will work on adding tests later.

**Comment by desimone** [comment]
Glad it worked! Code looks good. 

The only thing I can think about right now is that withou timeout handler on websocket requests we could run into a familiar issue as #40 . It's not a hair on fire concern as that assumes very specific attack conditions but it's something we should try to guard against as much as possible with these long-lived non-timeoutable requests. I'll think about mitigations more -- both for this PR and in general... but if you can think of any solutions that would be very welcome.

File: proxy/handlers.go [hunk 2]
```diff
@@ -501,3 +501,18 @@ func (p *Proxy) GetSignOutURL(authenticateURL, redirectURL *url.URL) *url.URL {
...
 	return time.Now().Add(ttl).Truncate(time.Second)
 }
+
+// websocketHandlerFunc splits request serving with timeouts depending on the protocol
+func websocketHandlerFunc(baseHandler http.Handler, timeoutHandler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		cn := r.Header.Get("Connection")
+		up := r.Header.Get("Upgrade")
+
+		if r.ProtoMajor == 1 && strings.Contains(cn, "Upgrade") && strings.Contains(up, "websocket") {
```
desimone: nit: Although the line gets longer, you can probably inline the r.Get()'s which saves the call if r.Proto is not 1.

desimone: Should we match for exact string for websocket and upgrade vs contains?

nareddyt:
Yeah, we can't add a timeout to websockets because they are long-lived connections by definition. This is a fundamental issue with websockets and hard to solve...
- https://security.stackexchange.com/questions/48378/anti-dos-websockets-best-practices
- https://samsaffron.com/archive/2015/12/29/websockets-caution-required

Here are some possible mitigations that are not timeout based and implementable in pomerium (in my opinion):
- **Opt-in Websocket Support Config Flag:** Make websocket support opt-in via the configuration variables. Users must explicitly acknowledge the possibility of abuse to enable websocket support.
- **Limit Number of Websocket Connections:** Set hard limits on the number of active websocket connections. Make threshold editable via the configuration. Limits can be enforced per source IP (does not prevent against distributed attacks) or globally across all source IPs (incorrect threshold can cause availability issues).

proxy/handlers.go
nareddyt: Good points, I can make both changes tomorrow.

desimone:
Option 1 feels like it strikes the right balance to me. Thoughts?

nareddyt:
I can implement option 1 in this PR. We can look into the possibility of rate limiting techniques later 👍

nareddyt:
@desimone ready for review!

desimone:
@nareddyt Looks great, thank you!

PR #158, opened 2019-06-03
internal/middleware: validate TLD plus one

### LINKED ISSUE #152: "Invalid redirect parameter" error with v0.0.4 and latest
**Describe the bug**

When I try to connect to pomerium protected service, I get redirected to the authenticate service public url and the following error is shown:
image

**To Reproduce**

I used the helm chart from pomerium/pomerium-helm@1b999f445c8f5e7f361345e310ee5be28abf9119 but applied the few changes needed for v0.0.4 to work (see https://github.com/pomerium/pomerium/releases/tag/v0.0.4).

Edit: I also tried with latest chart and helm, cf comments below.

I'm using the Google IdP and a service account.

I think the configuration is the same as when it was working before with v0.0.3.

**Expected behavior**

I would expect at least to understand what is going wrong, I guess maybe I made an error somewhere but there is zero feedback.

**Environment:**

- Pomerium version (retrieve with `pomerium --version`): v0.0.4+4bdb917, and also latest
- Server Operating System/Architecture/Cloud: AKS/K8S/Helm

Configuration file(s):

```text
    - from: app.xxx.com
      to: http://my-app.my-app.svc.cluster.local:80
      allowed_domains:
      - xxx.com
      timeout: 10m
```

Logs(s):

```text
:56PM DBG authenticate: request duration=0.435802 fwd_ip=10.240.0.35 ip=10.240.0.61 method=GET req_id=a84d3895-836b-9d2a-95ca-af51844c6722 size=1950 status=400 url=/sign_in?redirect_uri=https%3A%2F%2Fapp.xxx.com%2F.pomerium%2Fcallback&response_type=code&shared_secret=ZGhZZ0XXXXXKMEMrbTAwSDZ5TVBNNFRkZko5dnc%3D&sig=2ePoPCSkUgh5pFnGU-zXvevHjKHYB7w5jHf8Kt5dbUk%3D&state=pl3hQJZwffs5lmcpqdVkYljWgDkyrK0F2d5X6oUErWcFofvytP8yhESXB8YqhGJB-CgBLNbeoHBRGKexxLqSWp21BPgGo1y45uXf3FPHYfKTNKfT5XAN88-eQT88b557ho8yCyIcd7qX2HrWohuWZHJSDn8KiPJ3Hh_z3vruDwwIQsYnVHc0m88cYGKPtVjxFTkfjrTH8w&ts=1559318162 user_agent="Mozilla/5.0 (X11; Linux x86_64; rv:67.0) Gecko/20100101 Firefox/67.0"
```

desimone:
Refactor `ValidateRedirectURI` to check if two urls are coming from the same top level domain vs matching for exact subdomain and depth. 

Fixes #152 

See: 
- [Why it's trickier than it sounds](https://godoc.org/golang.org/x/net/publicsuffix#EffectiveTLDPlusOne)

**Checklist**:
- [x] unit tests added
- [x] ready for review

### DIFF
--- a/internal/middleware/middleware_test.go
+++ b/internal/middleware/middleware_test.go
@@ -10,7 +10,7 @@ import (
 	"time"
 )
 
-func Test_SameSubdomain(t *testing.T) {
+func Test_SameDomain(t *testing.T) {
 
 	tests := []struct {
 		name        string
@@ -19,8 +19,10 @@ func Test_SameSubdomain(t *testing.T) {
 		want        bool
 	}{
 		{"good url redirect", "https://example.com/redirect", "https://example.com", true},
+		{"good multilevel", "https://httpbin.a.corp.example.com", "https://auth.b.corp.example.com", true},
+		{"good complex tld", "https://httpbin.a.corp.example.co.uk", "https://auth.b.corp.example.co.uk", true},
+		{"bad complex tld", "https://httpbin.a.corp.notexample.co.uk", "https://auth.b.corp.example.co.uk", false},
 		{"simple sub", "https://auth.example.com", "https://test.example.com", true},
-		{"mismatched lengths", "https://auth.auth.example.com", "https://test.example.com", false},
 		{"bad domain", "https://auth.example.com/redirect", "https://test.notexample.com", false},
 		{"malformed url", "^example.com/redirect", "https://notexample.com", false},
 		{"empty domain list", "https://example.com/redirect", ".com", false},
@@ -31,8 +33,8 @@ func Test_SameSubdomain(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			u, _ := url.Parse(tt.uri)
 			j, _ := url.Parse(tt.rootDomains)
-			if got := SameSubdomain(u, j); got != tt.want {
-				t.Errorf("SameSubdomain() = %v, want %v", got, tt.want)
+			if got := SameDomain(u, j); got != tt.want {
+				t.Errorf("SameDomain() = %v, want %v", got, tt.want)
 			}
 		})
 	}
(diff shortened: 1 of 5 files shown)


internal/middleware/middleware_test.go
```diff
@@ -155,7 +159,7 @@ func TestValidateRedirectURI(t *testing.T) {
 			handler.ServeHTTP(rr, req)
 			if rr.Code != tt.status {
 				t.Errorf("Status code differs. got %d want %d", rr.Code, tt.status)
-				t.Errorf("%s", rr.Body)
+				// t.Errorf("%s", rr.Body)
```
victornoel: @desimone is this commenting on purpose?

victornoel:
@desimone seems good to me, at least the tests :)

internal/middleware/middleware_test.go
desimone: oops!

yegle:
I think I was tripped by this change.

I reverse proxy a website that doesn't match the domain used by Pomerium. I got an error: redirect uri and root domain differ

This is the error when accessing login.foo.com/.pomerium/sign_in?redirect_uri=example.bar.com

This definitely worked before but was broken since a recent change. I was yet to be able to ping point the commit (I don't think this commit introduced the regression).

(Reverse proxy a different root domain means I would get a SSL error and I'm fine with it. I think I'm ok if you don't consider this a regression.)

desimone:
Hi @yegle , sorry to hear that. Can you give me a rough idea of what the domain hierarchy is of your setup? e.g.

`app-failing.int.example.com`
`httpbin.int.example.com`
`proxy.int.example.com`
`authn.int.example.com`
`authz.int.example.com`

yegle:
I have:

1. In `docker-compose.yaml`: `AUTHENTICATE_SERVICE_URL=https://nuc.example.com`
2. In `pomerium.yaml`:

```
  - from: https://www.ojrq.net/                                                 
    to: http://affiliate-links                                                  
```
3. I use DNS to resolve `www.ojrq.net` to `nuc.example.com`

A curl looks like this on my end:
```
$ curl -v -k https://www.ojrq.net/
...
< location: https://nuc.example.com/.pomerium/sign_in?redirect_uri=https%3A%2F%2Fwww.ojrq.net%2F&sig=XXXX-XXXX%3D&ts=XXXX
```

And then I get the `redirect uri and root domain differ` error.

desimone:
Hi @yegle 

I've been working on a draft that describes our security model (#233 )... But just for some background the basic idea is that pomerium's security boundary is tightly coupled to a subdomain. By asserting a certain subdomain is managed by pomerium, we can maintain a single state cookie amongst all managed apps, and enforce that any user session cookie is stripped (amongst other things).

That all said, I'm still not totally clear what you are trying to do (a public route, which resolves to the authenticate endpoint, which would redirect back to an unmanaged app?) Can you tell me a little bit about the intended use-case ? 

Also, what version where you on and what version are you trying now? 

Thanks!

yegle:
Yes I'm aware my use case is unusual.

I was trying to intercept common affiliate links (ojrq.net being one) to my own web server and do the redirect myself. Given I already have a server running docker with Pomerium as a reverse proxy, I just used pomerium to reverse proxy the request to a webserver I wrote (https://github.com/yegle/affiliate-links) running in Docker.

Basically I was using Pomerium as a general purpose reverse proxy. It might not be a valid use case though.

PR #160, opened 2019-06-04
Change Options from pointers to values

### LINKED ISSUE #155: [discussion] Change Options and URL fields to values
This came up in #150 while working on Checksum().  Related to #115 as call/struct signatures were retained as much as possible.

It appears that on service creation, Authenticate.New() tacks on a path to the AuthenticateURL from Options and then uses that to initialize further data structures.  It isn't 100% obvious in the code but, as-written, it actually makes the change to the url field in the Options structure.  

https://github.com/pomerium/pomerium/blob/e982e72146b23914ee0a03dec75cffd279f80faa/authenticate/authenticate.go#L81-L82

While possibly less efficient, my instinct is to protect Options from accidental changes and dereference as much as possible.  I think just changing the field types would make it apparent when Options was being mutated, but not passing options as a pointer at all should make it impossible.

This would touch a few things, so I wanted to discuss before I try to tackle the changes.  I think this makes sense (and wouldn't break anything...) but I'm open to input or correction.

travisgroth:
Closes #155 

**Checklist**:
- [x] unit tests added
- [x] ready for review

### DIFF
--- a/internal/config/options_test.go
+++ b/internal/config/options_test.go
@@ -16,7 +16,7 @@ import (
 
 func Test_validate(t *testing.T) {
 
-	testOptions := func() *Options {
+	testOptions := func() Options {
 		o := NewOptions()
 		o.SharedKey = "test"
 		o.Services = "all"
@@ -34,7 +34,7 @@ func Test_validate(t *testing.T) {
 
 	tests := []struct {
 		name     string
-		testOpts *Options
+		testOpts Options
 		wantErr  bool
 	}{
 		{"good default with no env settings", good, false},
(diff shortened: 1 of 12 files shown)


internal/config/options_test.go
```diff
@@ -194,10 +194,10 @@ func Test_parseURLs(t *testing.T) {
 		if (err != nil) != test.wantErr {
 			t.Errorf("Failed to parse URLs %v: %s", test, err)
 		}
-		if o.AuthenticateURL != nil && o.AuthenticateURL.String() != test.authenticateURL {
+		if o.AuthenticateURL != (url.URL{}) && o.AuthenticateURL.String() != test.authenticateURL {
```
travisgroth: I converted these straight to default values from nil but I think just get rid of nil check?

internal/config/options_test.go
```diff
@@ -194,10 +194,10 @@ func Test_parseURLs(t *testing.T) {
...
 			t.Errorf("Failed to update AuthenticateURL: %v", test)
 		}
-		if o.AuthorizeURL != nil && o.AuthorizeURL.String() != test.authorizeURL {
+		if o.AuthorizeURL != (url.URL{}) && o.AuthorizeURL.String() != test.authorizeURL {
```
travisgroth: I converted these straight to default values from nil but I think just get rid of nil check?

desimone: Yeah, I think it can go. Also, I love that the go playground now supports outside imports!
https://play.golang.org/p/3DW10L34lGJ

travisgroth: Nice.  I saw that today but wasn't sure how new it was.

desimone:
Looks great. Sorry, for the delay... I thought I merged this already!

PR #183, opened 2019-06-13
proxy: fix bug in sign_out handler

### LINKED ISSUE #170: proxied /.pomerium/sign_out doesn't do anything
**Describe the bug**

- Given an app protected behind `https://app.xxx.com`.
- If a user goes to `https://app.xxx.com/.pomerium/sign_out` (via GET method)
- The proxy redirects to the service but does not sign the user out

**Expected behavior**

I believe in the past it was working, but I don't know if it is expected or not.

My use case is that I want to have a "logout" button on my applications and I was making the assumption it was better to redirect to `https://app.xxx.com/.pomerium/sign_out` when clicking on it instead of having a logical dependency on the authenticate service url.

If it's expected, I think it can be acceptable to have my applications know (via configuration) that the sign out url is https://auth.xxx.com/.pomerium/sign_out.

But any case, it should be accessible via a GET request (and be redirected to the referrer I guess?), not only a POST.

**Environment:**

- Pomerium version (retrieve with `pomerium --version`): v0.0.5-dirty+5525408
- Server Operating System/Architecture/Cloud: AKS

Configuration file(s): see #152

desimone:
Fixes #170 

**Checklist**:
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review

/cc @victornoel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### CHANGED
 
+- Proxy's sign out handler `{}/.pomerium/sign_out` now accepts an optional `redirect_uri` parameter which can be used to specify a custom redirect page, so long as it is under the same top-level domain. [GH-183]
+
 ### FIXED
 
 ## v0.0.5
--- a/authenticate/handlers_test.go
+++ b/authenticate/handlers_test.go
@@ -136,7 +136,7 @@ func TestAuthenticate_SignIn(t *testing.T) {
 			"state=example",
 			"redirect_uri=some.example",
 			&sessions.MockSessionStore{
-				LoadError: errors.New("unexpeted"),
+				LoadError: errors.New("error"),
 				Session: &sessions.SessionState{
 					AccessToken:     "AccessToken",
 					RefreshToken:    "RefreshToken",
--- a/proxy/handlers.go
+++ b/proxy/handlers.go
@@ -37,7 +37,7 @@ func (p *Proxy) Handler() http.Handler {
 	mux.HandleFunc("/robots.txt", p.RobotsTxt)
 	mux.HandleFunc("/.pomerium", p.UserDashboard)
 	mux.HandleFunc("/.pomerium/impersonate", p.Impersonate) // POST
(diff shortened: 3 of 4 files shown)


victornoel:
@desimone so this will redirect a call to /.pomerium/sign_out with GET or POST to the authenticate service which will sign out the user?

At the end, where is the user redirected? It's not clear to me actually ^^

desimone:
@desimone that's right. It redirects the user to authenticate's sign out handler which revokes the access token for that user and then redirects the user back the root of the proxied domains (e.g. `https://httpbin.corp.example.com` which likely then redirects the user to start the auth flow again).

We could pretty easily make the redirect parameter configurable that way you can set where you'd like the user to be redirected to on logout. 

so you could do `https://httpbin.corp.example.com/.pomerium/sign_out?redirect_url?https://www.google.com`  Thoughts?

victornoel:
@desimone yep, that seem a good idea, I think I will be using that new possibility (even though I didn't really realise I could need it before ;)

victornoel:
@desimone I was thinking: what about a default `redirect_url` configured in the authenticate service?

desimone:
@victornoel default as in a site-wide (config.yaml) redirect address? 

Right now there is a default address but it is just the root of the proxied domain. So if you logged out of `https://httpbin.corp.your.example/.pomerium/sign_out` you'd be redirected back to  `https://httpbin.corp.your.example` unless you set the `https://httpbin.corp.your.example/.pomerium/sign_out?redirect_uri?https://your.example`

victornoel:
@desimone I didn't know that there already was something available :)

So w.r.t. our discussion about what is the root domain (i.e. that if we have `auth.corp.your.example`, the root domain is `corp.your.example`, right?), then it means I can already set `redirect_uri` to `corp.your.example`?

With this PR I will be able to set it also to `your.example` or to `whatever.your.example`?

desimone:
@victornoel that's right. Anything at the TLD+1 you can set it to.

Issue #35, opened 2019-01-30
add service metrics / instrumentation / telemetry


desimone:
**Is your feature request related to a problem? Please describe.**

Code instrumentation is essential for observability into a distributed system. 

**Describe the solution you'd like**

Add support for common instrumentation capabilities like counters, gauges, and histograms, and provides adapters to popular metrics packages, like expvar, StatsD, and Prometheus.

**Additional context**

- [Site Reliability Engineering How Google Runs Production Systems](http://shop.oreilly.com/product/0636920041528.do)
- https://godoc.org/github.com/go-kit/kit/metrics 
- https://prometheus.io/
- https://peter.bourgon.org/blog/2016/02/07/logging-v-instrumentation.html

travisgroth:
Short writeup on technology/libraries/approach

### Target technology

My take on the current ecosystem is that most new applications are trending towards prometheus as an exposition format.  It is simple, human friendly, and the team behind Prometheus is working on making it a standard.

Additionally, most other monitoring stacks have adapters for prometheus.  Influx, datadog, elastic, all appear to cover it.

Thus, I believe whatever approach is taken should take care to support prometheus as a format as a priority.  If any other metrics formats/strategies can be supported simultaneously, that is a bonus.

### Library/framework support

With the goal of prometheus format, the prometheus client is an obvious choice.  It is widely used and heavily tested.  It doesn't easily support other formats, however.  Some other choices in the ecosystem:

- https://godoc.org/github.com/go-kit/kit/metrics.  Their interface abstraction keeps your code from needing to understand the different concrete metric formats, but every integration and metrics set must be maintained separately.  By their own admission, if you know your target stack, you should use it directly.  

- https://github.com/armon/go-metrics.  Low verbosity and the in memory dump could be interesting during debugging.  It only supports summary type metrics, however, which is somewhat limiting; they also don't seem to support help messages.  Does support statsd, which I think is a good "also" format.

- https://github.com/segmentio/stats.  Docs are a bit rough, especially after the latest upgrade.  Does support a lot of output formats without too much work.  No less verbose to create metrics than prometheus itself.

- https://opencensus.io/quickstart/go/metrics/.  This is a google backed/derived project.  Definitely early days, but the go support appears solid.  If the community coalesces around this project, it will become _the_ library and all exposition formats become a new plugin.  Prometheus is supported out of the box.  We also gain tracing support at the same time.  It looks like their velocity has slowed down recently due to this: https://opentelemetry.io/.

[...]

PR #187, opened 2019-06-19
Proxy Client Metrics

travisgroth:
Proxy client metrics implementation for #35 

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review

### DIFF
--- a/internal/metrics/middleware.go
+++ b/internal/metrics/middleware.go
@@ -6,11 +6,11 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/pomerium/pomerium/internal/middleware/responsewriter"
+	"github.com/pomerium/pomerium/internal/tripper"
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
-
-	"github.com/pomerium/pomerium/internal/middleware/responsewriter"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


internal/tripper/chain_test.go
```diff
@@ -0,0 +1,123 @@
...
+			body, _ := ioutil.ReadAll(resp.Body)
+			mockResp := httptest.NewRecorder()
+			mockResp.Write(body)
+			mockResp.WriteString(fmt.Sprintf(",%s", id))
+			return mockResp.Result(), nil
+		})
+	}
+}
+
+func TestNew(t *testing.T) {
```
travisgroth: The tripper chain functionality is largely copied from middleware, but I wanted to redo the tests from first principles so that (a) I understood the code well (b) could focus on the side effects rather than returned data structures.  

I think this approach captured all the tests nicely, but I'm happy to port over the middleware chain tests if you'd like to keep it consistent.

internal/metrics/middleware.go
```diff
@@ -79,10 +116,37 @@ func HTTPMetricsHandler(service string) func(next http.Handler) http.Handler {
...
+}
+
+// HTTPMetricsRoundTripper creates a metrics tracking tripper for outbound HTTP Requests
+func HTTPMetricsRoundTripper(service string) func(next http.RoundTripper) http.RoundTripper {
+
+	return func(next http.RoundTripper) http.RoundTripper {
+		return tripper.RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+			startTime := time.Now()
+
+			resp, err := next.RoundTrip(r)
```
desimone: If an error gets thrown here, I assume `resp` would be nil and the next two lines may panic due to nil reference?

internal/tripper/chain_test.go
desimone: I think that makes sense. I'm generally for a bit of duplication for the sake of clarity and I think that's true here.

travisgroth: Do you want me to bring over the middleware tests or leave the tests as-is?  I'm okay either way.

internal/metrics/middleware.go
travisgroth: That's a good point.  I was thinking on the server side still...and that a response of some sort would happen - but that's not true here in the Transport layer.  Will fix.

internal/tripper/chain_test.go
travisgroth: Also an option: update the middleware tests to use this style.

desimone: Hmmm... If you think the middleware tests are relevant -- sure. If not, I don't think it's required. I'll defer to your judgement. 

The one thing that would be nice to add (sorry, I can't see coverage directly here as I'm traveling, so perhaps you already have that base covered) would be to have a mocked RoundTripper that also could test for that nil condition from before. .... maybe add a `roundTripError` type to `mockTransport`?

As for changing the existing middleware tests format, I'm a bit more cautious just because they are a complete carry over from [Alice](https://github.com/pomerium/pomerium/blob/master/3RD-PARTY#L90_ and have been pretty stable. Though, adding additional relevant tests makes sense.

travisgroth: >have a mocked RoundTripper that also could test for that nil condition from before

Yep, should be covered.  I have a then() transport/tripper that errors. 
 https://github.com/travisgroth/pomerium/blob/ca46202d21561eab90c5a59113080a5c5190b6c3/internal/metrics/middleware_test.go#L128-L132  and https://github.com/travisgroth/pomerium/blob/ca46202d21561eab90c5a59113080a5c5190b6c3/internal/metrics/middleware_test.go#L215-L221

>As for changing the existing middleware tests format, I'm a bit more cautious just because they are a complete carry over from [Alice]

Ah, I did not realize.  I think they accomplish the same thing; just a different approach.  Take a look when you're back at a keyboard for a bit.  I'm happy to leave as is; we can always revisit.

desimone: Ah! Figured that I just missed it.

desimone: > I think they accomplish the same thing; just a different approach. Take a look when you're back at a keyboard for a bit. I'm happy to leave as is; we can always revisit.

Yeah, let's revisit in another issue.

desimone:
Thanks!!

PR #195, opened 2019-06-27
Add GRPC Client Metrics

travisgroth:
GRPC client metrics for #35 

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review

### DIFF
--- a/internal/metrics/interceptors.go
+++ b/internal/metrics/interceptors.go
@@ -23,6 +23,7 @@ var (
 	grpcClientResponseSize    = stats.Int64("grpc_client_response_size_bytes", "grpc Client Response Size in bytes", "bytes")
 	grpcClientRequestDuration = stats.Int64("grpc_client_request_duration_ms", "grpc Client Request duration in ms", "ms")
 
+	// GRPCServerRequestCountView is an OpenCensus view which tracks GRPC Server requests by pomerium service, host, grpc service, grpc method, and status
 	GRPCServerRequestCountView = &view.View{
 		Name:        grpcServerRequestCount.Name(),
 		Measure:     grpcServerRequestCount,
@@ -31,6 +32,7 @@ var (
 		Aggregation: view.Count(),
 	}
 
+	// GRPCServerRequestDurationView is an OpenCensus view which tracks GRPC Server request duration by pomerium service, host, grpc service, grpc method, and statu
 	GRPCServerRequestDurationView = &view.View{
 		Name:        grpcServerRequestDuration.Name(),
 		Measure:     grpcServerRequestDuration,
@@ -45,6 +47,7 @@ var (
 		),
 	}
 
+	// GRPCServerResponseSizeView is an OpenCensus view which tracks GRPC Server request duration by pomerium service, host, grpc service, grpc method, and statu
 	GRPCServerResponseSizeView = &view.View{
 		Name:        grpcServerResponseSize.Name(),
 		Measure:     grpcServerResponseSize,
@@ -56,6 +59,7 @@ var (
 		),
 	}
 
+	// GRPCClientRequestCountView is an OpenCensus view which tracks GRPC Client requests by pomerium service, target host, grpc service, grpc method, and statu
 	GRPCClientRequestCountView = &view.View{
 		Name:        grpcClientRequestCount.Name(),
 		Measure:     grpcClientRequestCount,
@@ -63,6 +67,8 @@ var (
 		TagKeys:     []tag.Key{keyService, keyHost, keyMethod, keyStatus, keyGRPCService},
 		Aggregation: view.Count(),
 	}
+
+	// GRPCClientRequestDurationView is an OpenCensus view which tracks GRPC Client request duration by pomerium service, target host, grpc service, grpc method, and statu
 	GRPCClientRequestDurationView = &view.View{
 		Name:        grpcClientRequestDuration.Name(),
 		Measure:     grpcClientRequestDuration,
@@ -76,6 +82,8 @@ var (
 			100000,
 		),
 	}
+
+	// GRPCClientResponseSizeView  is an OpenCensus view which tracks GRPC Client response size by pomerium service, target host, grpc service, grpc method, and statu
 	GRPCClientResponseSizeView = &view.View{
 		Name:        grpcClientResponseSize.Name(),
 		Measure:     grpcClientResponseSize,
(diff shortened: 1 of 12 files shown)


internal/metrics/interceptors.go
```diff
@@ -0,0 +1,147 @@
+package metrics // import "github.com/pomerium/pomerium/internal/metrics"
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/golang/protobuf/proto"
+
+	"google.golang.org/grpc/status"
```
desimone: nit: [group imports](https://github.com/golang/go/wiki/CodeReviewComments#imports)

internal/metrics/interceptors.go
```diff
@@ -0,0 +1,147 @@
...
+		// Split the method into parts for better slicing
+		rpcInfo := strings.SplitN(method, "/", 3)
+		var rpcMethod string
+		var rpcService string
+		if len(rpcInfo) == 3 {
+			rpcService = rpcInfo[1]
+			rpcMethod = rpcInfo[2]
+		}
+
+		responseStatus, _ := status.FromError(err)
```
desimone: Potentially unhandled error here? It's hard for me to tell from the [definition](https://godoc.org/google.golang.org/grpc/status#FromError) but makes me nervous.

internal/metrics/interceptors.go
```diff
@@ -0,0 +1,147 @@
...
+			context.Background(),
+			tag.Insert(keyService, service),
+			tag.Insert(keyHost, cc.Target()),
+			tag.Insert(keyMethod, rpcMethod),
+			tag.Insert(keyGRPCService, rpcService),
+			tag.Insert(keyStatus, responseStatus.Code().String()),
+		)
+
+		responseProto := reply.(proto.Message)
+		responseSize := len([]byte(responseProto.String()))
```
desimone: Equivalent to `https://godoc.org/github.com/golang/protobuf/proto#Size` ?

internal/metrics/interceptors.go
```diff
@@ -0,0 +1,147 @@
...
+
+		if tagErr == nil {
```
desimone: If `tagErr` is not nil, I think it falls through, is that the intent?

internal/metrics/interceptors_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+			invoker := testInvoker{
+				invokeResult: tt.errorCode,
+			}
+			var reply testProto
+
+			interceptor(context.Background(), tt.method, nil, &reply, newTestCC(), invoker.UnaryInvoke)
+
+			// grpcClientResponseSize
+			data, _ := view.RetrieveData(grpcClientResponseSize.Name())
```
desimone: I know it's just tests but prefer to check err condition with`t.Fatalf`

internal/metrics/interceptors_test.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (t testInvoker) UnaryInvoke(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, opts ...grpc.CallOption) error {
+	r := reply.(*testProto)
+	r.message = "hello"
+	return t.invokeResult
+}
+
+func newTestCC() *grpc.ClientConn {
+	testCC, _ := grpc.Dial("dns:localhost:9999", grpc.WithInsecure())
```
desimone: nit: check error and throw t.Fatal .

internal/metrics/interceptors.go
```diff
@@ -0,0 +1,147 @@
...
+			Aggregation: view.Distribution(
+				1, 256, 512, 1024, 2048, 8192, 16384, 32768, 65536, 131072, 262144, 524288,
+				1048576, 2097152, 4194304, 8388608,
+			),
+		},
+	}
+)
+
+func init() {
+	view.Register(grpcViews...)
```
desimone: Would it be possible to register the views outside of init and register/invoke directly from main?

internal/metrics/tags.go
```diff
@@ -0,0 +1,13 @@
+package metrics
+
+import (
+	"go.opencensus.io/tag"
+)
+
+var (
+	keyMethod, _      = tag.NewKey("method")
```
desimone: Ignored err makes me nervous. Would it be possible to keep the declarations global but move the instantiation to an `err` checked function?

internal/metrics/interceptors.go
travisgroth: Yeah.  Intentional and safe IMO.  That's a weird implementation but the second value doesn't do much for us.  The rest of the evaluation of responseStatus handles either case correctly.

internal/metrics/interceptors.go
travisgroth: Ah, yes.  That's a better way.

internal/metrics/interceptors.go
travisgroth: Yes.  I'm not 100% clear on what an error condition looks like when creating the tag context...but we don't try to record if it doesn't succeed.  I think I should move the proto handling into that block just to save some operations in that case, though.

travisgroth: Second thoughts here.  Maybe debug log if it isn't nil?

internal/metrics/interceptors_test.go
travisgroth: Yes...I actually am pushing a generalized version of those tests, which includes this check.  Can you give that a look vs the previous (commented out) implementation?  It should work for the http metrics too.

internal/metrics/interceptors.go
desimone: Yes. Maybe a `log.Warn().Err(err).Msg("....")` type of deal?

internal/metrics/tags.go
travisgroth: Yep.  Done.  Wasn't too worried since they're static, but it's easy to do the right thing.

internal/metrics/interceptors_test.go
travisgroth: Should be all set.

internal/metrics/interceptors.go
travisgroth: 👍

internal/metrics/interceptors.go
travisgroth: I should be able to restructure it a bit.  

The original thought was that if you're using the middleware or interceptors, you want the views they feed.  However, If we wanted to turn different stat types on or off in the future, it might be easier to control the view registration than the middleware chains.

desimone:
And merged!

Issue #214, opened 2019-07-07
deployment: prepare v0.1.0 release


desimone:
Before tagging

- [x] Review and update examples 
- [x] Review and update dockerfiles
- [x] Review and update docs
- [x] Review and update changelog
- [x] Review and update upgrading guide

After tagging
- [x] Update github release docs
- [x] Publish tagged version of documentation (
- [x] Review and update helm-charts (merge after tag)

victornoel:
@desimone I will be testing current master now, if you want to wait for potential feedbacks before doing release, it's possible :)

victornoel:
@desimone ok I see I was too late, anyway, this seems to work as expected for the issues I was involved in ;)

victornoel:
@desimone @travisgroth but I don't see any image tagged v0.1.0 in docker hub though

travisgroth:
Looks like the arm images are tagged, but not amd64 for some reason.

desimone:
arm64 image should now be there.

victornoel:
@desimone I confirm, it works, I'm using it :)

desimone:
Annnd that should do it!

Issue #201, opened 2019-07-02
development: use golangci-lint for combined linting


desimone:
**Is your feature request related to a problem? Please describe.**

GolangCI-Lint is a linters aggregator. It's fast: on average 5 times faster than gometalinter. It includes all the current lints and vets plus:
```
deadcode: Finds unused code [fast: true, auto-fix: false]
errcheck: Errcheck is a program for checking for unchecked errors in go programs. These unchecked errors can be critical bugs in some cases [fast: true, auto-fix: false]
gosimple: Linter for Go source code that specializes in simplifying a code [fast: false, auto-fix: false]
govet (vet, vetshadow): Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string [fast: false, auto-fix: false]
ineffassign: Detects when assignments to existing variables are not used [fast: true, auto-fix: false]
staticcheck: Staticcheck is a go vet on steroids, applying a ton of static analysis checks [fast: false, auto-fix: false]
structcheck: Finds unused struct fields [fast: true, auto-fix: false]
typecheck: Like the front-end of a Go compiler, parses and type-checks Go code [fast: true, auto-fix: false]
unused: Checks Go code for unused constants, variables, functions and types [fast: false, auto-fix: false]
varcheck: Finds unused global variables and constants [fast: true, auto-fix: false]
and the following linters are disabled by default:

Disabled by default linters:
bodyclose: checks whether HTTP response body is closed successfully [fast: false, auto-fix: false]
depguard: Go linter that checks if package imports are in a list of acceptable packages [fast: true, auto-fix: false]
dupl: Tool for code clone detection [fast: true, auto-fix: false]
gochecknoglobals: Checks that no globals are present in Go code [fast: true, auto-fix: false]
gochecknoinits: Checks that no init functions are present in Go code [fast: true, auto-fix: false]
goconst: Finds repeated strings that could be replaced by a constant [fast: true, auto-fix: false]
gocritic: The most opinionated Go source code linter [fast: true, auto-fix: false]
gocyclo: Computes and checks the cyclomatic complexity of functions [fast: true, auto-fix: false]
gofmt: Gofmt checks whether code was gofmt-ed. By default this tool runs with -s option to check for code simplification [fast: true, auto-fix: true]
goimports: Goimports does everything that gofmt does. Additionally it checks unused imports [fast: true, auto-fix: true]
[...]

PR #215, opened 2019-07-07
Make dockerfiles consistent between archs

travisgroth:
Working release checklist in #214 

I don't think the lint command belongs in the docker file for release; we aren't using it in the container.

I think the error reported in #121 should be fixed as part of the Makefile, if anywhere.  I can add that to this PR if that sounds right.

**Checklist**:
- [x] updated CHANGELOG.md
- [x] ready for review


desimone:
> I think the error reported in #121 should be fixed as part of the Makefile, if anywhere. 

I agree. I think this would also ease the transition of #201 as well. 

> I can add that to this PR if that sounds right.  

Please :)

PR #226, opened 2019-07-16
internal/cryptutil: fixed potential panic on nil pubkey block

desimone:
It's an edge-case where if an empty public key was passed into `DecodePublicKey` a panic could be caused. 

**Checklist**:
- [x] unit tests added
- [x] ready for review

### DIFF
--- a/internal/cryptutil/marshal.go
+++ b/internal/cryptutil/marshal.go
@@ -1,5 +1,5 @@
 // Package cryptutil provides encoding and decoding routines for various cryptographic structures.
-package cryptutil
+package cryptutil // import "github.com/pomerium/pomerium/internal/cryptutil"
 
 import (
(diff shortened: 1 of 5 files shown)


internal/cryptutil/marshal.go
```diff
@@ -13,11 +13,9 @@ import (
 // DecodePublicKey decodes a PEM-encoded ECDSA public key.
 func DecodePublicKey(encodedKey []byte) (*ecdsa.PublicKey, error) {
 	block, _ := pem.Decode(encodedKey)
-	if block == nil || block.Type != "PUBLIC KEY" {
```
desimone: To be clear `block.Type != "PUBLIC KEY" ` would panic if block ==nil here.

PR #227, opened 2019-07-16
internal/metrics: status metrics

travisgroth:
The last of the metrics in scope for #35 

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review

### DIFF
--- a/cmd/pomerium/main.go
+++ b/cmd/pomerium/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
@@ -87,6 +88,7 @@ func main() {
 
 	if opt.MetricsAddr != "" {
 		go newPromListener(opt.MetricsAddr)
+		metrics.SetBuildInfo(opt.Services)
 	}
 
 	if srv, err := startRedirectServer(opt.HTTPRedirectAddr); err != nil {
@@ -162,6 +164,8 @@ func newPromListener(addr string) {
 	metrics.RegisterView(metrics.HTTPServerViews)
 	metrics.RegisterView(metrics.GRPCClientViews)
 	metrics.RegisterView(metrics.GRPCServerViews)
+	metrics.RegisterInfoMetrics()
+	metrics.RegisterView(metrics.InfoViews)
 
 	log.Info().Str("MetricsAddr", addr).Msg("cmd/pomerium: starting prometheus endpoint")
 	log.Error().Err(metrics.NewPromHTTPListener(addr)).Str("MetricsAddr", addr).Msg("cmd/pomerium: could not start metrics exporter")
@@ -206,6 +210,14 @@ func parseOptions(configFile string) (*config.Options, error) {
 	if o.LogLevel != "" {
 		log.SetLevel(o.LogLevel)
 	}
+	metrics.AddPolicyCountCallback(o.Services, func() int64 {
+		return int64(len(o.Policies))
+	})
+	checksumInt, err := strconv.ParseInt(fmt.Sprintf("0x%s", o.Checksum()), 0, 64)
+	if err != nil {
+		log.Warn().Err(err).Msg("Could not parse config checksum into integer")
+	}
+	metrics.SetConfigChecksum(o.Services, checksumInt)
 	return o, nil
 }
 
@@ -213,6 +225,7 @@ func handleConfigUpdate(opt *config.Options, services []config.OptionsUpdater) *
 	newOpt, err := parseOptions(*configFile)
 	if err != nil {
 		log.Error().Err(err).Msg("cmd/pomerium: could not reload configuration")
+		metrics.SetConfigInfo(opt.Services, false, "")
 		return opt
 	}
 	optChecksum := opt.Checksum()
@@ -232,8 +245,12 @@ func handleConfigUpdate(opt *config.Options, services []config.OptionsUpdater) *
 	for _, service := range services {
 		if err := service.UpdateOptions(*newOpt); err != nil {
 			log.Error().Err(err).Msg("cmd/pomerium: could not update options")
+			metrics.SetConfigInfo(opt.Services, false, "")
 		}
 	}
-
+	metrics.AddPolicyCountCallback(newOpt.Services, func() int64 {
+		return int64(len(newOpt.Policies))
+	})
+	metrics.SetConfigInfo(newOpt.Services, true, newOptChecksum)
 	return newOpt
 }
--- a/internal/metrics/info.go
+++ b/internal/metrics/info.go
@@ -1,0 +1,184 @@
+package metrics // import "github.com/pomerium/pomerium/internal/metrics"
+
+import (
+	"context"
+	"runtime"
+	"sync"
+	"time"
+
+	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/internal/version"
+	"go.opencensus.io/metric"
+	"go.opencensus.io/metric/metricdata"
+	"go.opencensus.io/metric/metricproducer"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+)
+
+var (
+	//buildInfo               = stats.Int64("build_info", "Build Metadata", "1")
+	configLastReload        = stats.Int64("config_last_reload_success_timestamp", "Timestamp of last successful config reload", "seconds")
+	configLastReloadSuccess = stats.Int64("config_last_reload_success", "Returns 1 if last reload was successful", "1")
+	registry                = newMetricRegistry()
+
+	// ConfigLastReloadView contains the timestamp the configuration was last
+	// reloaded, labeled by service
+	ConfigLastReloadView = &view.View{
+		Name:        configLastReload.Name(),
+		Description: configLastReload.Description(),
+		Measure:     configLastReload,
+		TagKeys:     []tag.Key{keyService},
+		Aggregation: view.LastValue(),
+	}
+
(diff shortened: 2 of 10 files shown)


internal/metrics/info.go
```diff
@@ -0,0 +1,150 @@
...
+
+// Register non-view based metrics registry globally for export
+func RegisterInfoMetrics() {
+	metricproducer.GlobalManager().AddProducer(metricRegistry)
+}
+
+// SetConfigChecksum creates the configuration checksum metric.  You must call RegisterInfoMetrics to
+// have this exported
+func SetConfigChecksum(service string, checksum int64) {
+	if configChecksum == nil {
```
travisgroth: This pattern for checking if nil and then initializing the Gauge is repeated.  Open to suggestions on making it more DRY.

cmd/pomerium/main.go
```diff
@@ -87,6 +87,7 @@ func main() {
 
 	if opt.MetricsAddr != "" {
 		go newPromListener(opt.MetricsAddr)
+		metrics.SetBuildInfo(opt.Services)
```
travisgroth: This could be moved elsewhere, potentially.

cmd/pomerium/main.go
```diff
@@ -206,13 +209,22 @@ func parseOptions(configFile string) (*config.Options, error) {
 	if o.LogLevel != "" {
 		log.SetLevel(o.LogLevel)
 	}
+	metrics.AddPolicyCountCallback(o.Services, func() int64 {
+		return int64(len(o.Policies))
+	})
+	checksumInt, err := strconv.ParseInt(fmt.Sprintf("0x%s", o.Checksum()), 0, 64)
```
desimone: Feels scary. But I suppose if this fails we get 0 not nil

internal/metrics/info.go
```diff
@@ -0,0 +1,150 @@
+package metrics // import "github.com/pomerium/pomerium/internal/metrics"
+
+import (
+	"context"
+	"runtime"
+	"time"
+
+	"github.com/pomerium/pomerium/internal/version"
```
desimone: :)

internal/metrics/info.go
```diff
@@ -0,0 +1,150 @@
...
+var (
+	//buildInfo               = stats.Int64("build_info", "Build Metadata", "1")
+	configLastReload        = stats.Int64("config_last_reload_success_timestamp", "Timestamp of last successful config reload", "seconds")
+	configLastReloadSuccess = stats.Int64("config_last_reload_success", "Returns 1 if last reload was successful", "1")
+
+	// Metrics that may/must be measured or set up before the config is fully parsed
+	// and metric views are initialized
+	buildInfo      *metric.Int64Gauge
+	policyCount    *metric.Int64DerivedGauge
+	configChecksum *metric.Int64Gauge
```
desimone: Global state always makes me a little nervous. 

What do you think about wrapping these metrics in a struct and protecting access via a mutex? It could also potentially help with abstracting out the nil checks and initial population. It feels like we want these to be [initialized global singletons](http://marcio.io/2015/07/singleton-pattern-in-go/)?  

Also, how would multiple instance effect the recording of these stats?

internal/metrics/info.go
desimone: See above comment about singleton.

internal/metrics/info.go
```diff
@@ -0,0 +1,150 @@
...
+			log.Error().Err(err).Msg("internal/metrics: failed to record config reload")
+		}
+	} else {
+		stats.Record(context.Background(), configLastReloadSuccess.M(0))
+	}
+}
+
+// Register non-view based metrics registry globally for export
+func RegisterInfoMetrics() {
+	metricproducer.GlobalManager().AddProducer(metricRegistry)
```
desimone: This is still a bit black magic to me; what are the implications of this?

cmd/pomerium/main.go
travisgroth: Yes both overflow and "not an int" conditions are safe based on my read of ParseInt() docs.

internal/metrics/info.go
```diff
@@ -0,0 +1,150 @@
...
+	m, err := buildInfo.GetEntry(
+		metricdata.NewLabelValue(service),
+		metricdata.NewLabelValue(version.FullVersion()),
+		metricdata.NewLabelValue(version.GitCommit),
+		metricdata.NewLabelValue((runtime.Version())),
+	)
+	if err != nil {
+		log.Error().Err(err).Msg("internal/metrics: failed to add build info metric")
+	}
+	m.Set(1)
```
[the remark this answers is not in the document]
travisgroth: Added a comment for clarity

internal/metrics/info.go
travisgroth: This is akin to registering a view.  Views and Metrics are handled a bit differently - the latter has the behavior we want/need for config derived info early on.  

The setup for Metric based stats is in reverse order.  Rather than Measure -> View -> Register you need a Registry and then Registry -> Metric -> Entry.  

We could do this in init() safely, which might address the metric singleton pattern better.  I will elaborate more in those comments.

I added a test to demonstrate that multiple AddProducer() calls with the same registry does not result in multiple references to the same registry internally.

internal/metrics/info.go
travisgroth: >What do you think about wrapping these metrics in a struct and protecting access via a mutex?

I considered a struct for DRYness, but discounted it.  It didn't seem like it would become any less verbose with the _current_ metric count.  We have two type signatures and three methods.

Good point on access safety.  These are currently early single threaded calls, but that isn't guaranteed later.  This implementation is not safe.   

A lighter pattern might be to do the setup in init(), which I believe is [threadsafe](https://golang.org/ref/spec#Package_initialization), though possibly controversial.  Given the rest of the "metric" type variables are initialized in package level var(), I think init() is consistent in spirit.

A third option is to provide a pair of MustNewXXXX() functions and move these into var(), but then we're back in the business of 2 methods for 3 variables.  I'm not sure I like that any better, though it gives us singleton guarantees as well.

Moving into init() has large overlap with the effort to move into a struct, so unless you specifically want to stay away from init() or there's something I'm missing in that pattern, I can try the init() implementation first.  Let me know thoughts here.

>Also, how would multiple instance effect the recording of these stats?

Just so I'm clear, do you mean calling metricRegistry.AddXXXX concurrently?  

A casual glance looks like they mutex on insertion by name.  So multiple calls with the same name should(?) result in only a single metric with the "last" description and labelkey set.  Type conflicts result in errors (I think, we can test that!).  When the metric is pulled out, it isn't immediately clear that you're getting the same reference.  I could see a race condition where you record to a variable that is later thrown away.

desimone: > I considered a struct for DRYness, but discounted it. It didn't seem like it would become any less verbose with the current metric count. We have two type signatures and three methods.
> Good point on access safety. These are currently early single threaded calls, but that isn't guaranteed later. This implementation is not safe.
> 
> A lighter pattern might be to do the setup in init(), which I believe is threadsafe, though possibly controversial. Given the rest of the "metric" type variables are initialized in package level var(), I think init() is consistent in spirit.

My thinking on a struct was not so much about DRYness but rather as a means of providing a thread-safe abstraction mutating those package scoped variables (mutex + accessor methods) and pattern for first-time initialization of metric gauges. 

So for example where we have logic like [this](https://github.com/pomerium/pomerium/blob/e0c60d64a7f283524728e3cd287463dbdf175973/internal/metrics/info.go#L53-L62) and [this](https://github.com/pomerium/pomerium/blob/e0c60d64a7f283524728e3cd287463dbdf175973/internal/metrics/info.go#L111-L120) and [this](https://github.com/pomerium/pomerium/blob/e0c60d64a7f283524728e3cd287463dbdf175973/internal/metrics/info.go#L134-L144) we could instead use a singleton and a [DoOnce](https://golang.org/pkg/sync/#Once.Do). What do you think?

desimone: > A casual glance looks like they mutex on insertion by name. So multiple calls with the same name should(?) result in only a single metric with the "last" description and labelkey set. Type conflicts result in errors (I think, we can test that!). When the metric is pulled out, it isn't immediately clear that you're getting the same reference. I could see a race condition where you record to a variable that is later thrown away.

👍

internal/metrics/info.go
desimone: > We could do this in init() safely, which might address the metric singleton pattern better. I will elaborate more in those comments.

 I think that makes sense. 

> I added a test to demonstrate that multiple AddProducer() calls with the same registry does not result in multiple references to the same registry internally.

 👍

internal/metrics/info.go
travisgroth: So just a summary - yes I am comfortable that the accessors/recorders (1) are safe but initialization (2) we should guarantee ourselves.  I'll see if the init() pattern looks clean.

Will also see about a test for failing that initialization with a conflicting type.

desimone: Sounds good to me. 👍

travisgroth: I have a [commit](https://github.com/pomerium/pomerium/pull/227/commits/e294188a9f1c7744c9642838748f0294f6539fcb) with the init() style implementation.  It is less wrapping that a struct, but the special handling started to feel a bit brittle and spread out.  The current commit represents a struct based.

I think having the struct container (using the sync.Once pattern) makes it easier to maintain and keep the code behavior in your head.  It also keeps the var() initialization pattern of other data structures.

Additional note - I've intentionally retained the external API, so that callers don't need to be aware of another struct to keep around or reference and we can refactor without much worry.

internal/metrics/info.go
travisgroth: Going to leave this as is so we can opt-in/out of various metrics later.  We're not doing anything in init() to move this next to.

PR #230, opened 2019-07-17
internal/telemetry: add tracing

### LINKED ISSUE #233: Prefer X25519 over P256 for TLS connections.

X25519 is probably a safer and faster choice than P-256 so we should use that when we can. 

- It's [safer](https://safecurves.cr.yp.to/).
- It's now [widely implemented](https://ianix.com/pub/curve25519-deployment.html).

Prefer X25519 to P256 when establishing TLS connections.

1. Keep things as they are. Both seem to be part of the upcoming [TLS1.3 spec](https://tools.ietf.org/html/draft-ietf-tls-tls13-28). 
2. Don't set this at all and use the default go preference.

https://github.com/golang/go/blob/a6a7b148f874b32a34e833893971b471cd9cdeb7/src/crypto/tls/common.go#L876-L883

The issue with 2 is we don't want to use [non-asm implemented curves](https://github.com/golang/go/tree/master/src/crypto/elliptic). 

**Additional context**

The order was previously reversed for compatibility (unsupported by go until 1.8). However, now X25519 is well supported. 

Now the world has largely moved to X25519. 

> Chrome’s implementation of TLS 1.3, for example, sends an X25519 keyshare in the first message to the server.

>   A TLS-compliant application MUST support digital signatures with
   rsa_pkcs1_sha256 (for certificates), rsa_pss_rsae_sha256 (for
   CertificateVerify and certificates), and ecdsa_secp256r1_sha256.  A
   TLS-compliant application MUST support key exchange with secp256r1
   (NIST P-256) and SHOULD support key exchange with X25519 [RFC7748].

  https://tools.ietf.org/html/draft-ietf-tls-tls13-28

desimone:
## Summary of changes

- all: add span traces throughout.
- internal/metrics: Rename to telemetry (to include traces).
- internal/telemetry: rename export to metrics. 
- internal/telemetry: add service name span to HTTPMetricsHandler. 
- internal/telemetry: removed chain dependency middleware_tests.
- internal/telemetry: wrap and encapsulate variadic view registration.
- internal/telemetry: add jaeger support for tracing.
- cmd/pomerium: move `parseOptions` to internal/config.
- cmd/pomerium: offload server handling to httputil and sub pkgs.
- internal/httputil: standardize creation/shutdown of http listeners. 
- internal/fileutil: use standardized Getwd

## Related Issues

- #35
- Closes #233 

## Screenshots
jaeger

Screen Shot 2019-07-14 at 2 24 39 PM

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review


authenticate/grpc.go
```diff
@@ -35,10 +41,8 @@ func (p *Authenticate) Validate(ctx context.Context, in *pb.ValidateRequest) (*p
 // Refresh renews a user's session checks if the session has been revoked using an access token
 // without reprompting the user.
 func (p *Authenticate) Refresh(ctx context.Context, in *pb.Session) (*pb.Session, error) {
-	// todo(bdd): add request id from incoming context
-	// md, _ := metadata.FromIncomingContext(ctx)
-	// sublogger := log.With().Str("req_id", md.Get("req_id")[0]).WithContext(ctx)
-	// sublogger.Info().Msg("tracing sucks!")
```
desimone: ! yay not anymore!

cmd/pomerium/main.go
```diff
@@ -35,15 +34,16 @@ func main() {
 		fmt.Println(version.FullVersion())
 		os.Exit(0)
 	}
-	opt, err := parseOptions(*configFile)
+	opt, err := config.ParseOptions(*configFile)
```
desimone: `ParseOptions` moved to config package.

cmd/pomerium/main.go
```diff
@@ -60,20 +60,17 @@ func main() {
 	if err != nil {
 		log.Fatal().Err(err).Msg("cmd/pomerium: proxy")
 	}
+	defer proxy.AuthenticateClient.Close()
```
desimone: On program quit, safely close the gRPC client connections.

cmd/pomerium/main.go
```diff
@@ -194,46 +184,3 @@ func wrapMiddleware(o *config.Options, mux http.Handler) http.Handler {
 	c = c.Append(middleware.Healthcheck("/ping", version.UserAgent()))
 	return c.Then(mux)
 }
-
-func parseOptions(configFile string) (*config.Options, error) {
```
desimone: Moved to `internal/config`

cmd/pomerium/main.go
```diff
@@ -194,46 +184,3 @@ func wrapMiddleware(o *config.Options, mux http.Handler) http.Handler {
...
-	if o.Debug {
-		log.SetDebugMode()
-	}
-	if o.LogLevel != "" {
-		log.SetLevel(o.LogLevel)
-	}
-	return o, nil
-}
-
-func handleConfigUpdate(opt *config.Options, services []config.OptionsUpdater) *config.Options {
```
desimone: Moved to `internal/config`

internal/httputil/tls.go
```diff
@@ -178,8 +120,8 @@ func newDefaultTLSConfig(cert *tls.Certificate) *tls.Config {
 		PreferServerCipherSuites: true,
 		// Use curves which have assembly implementations
 		CurvePreferences: []tls.CurveID{
-			tls.CurveP256,
 			tls.X25519,
+			tls.CurveP256,
```
desimone: For more details:
- https://safecurves.cr.yp.to/
- https://paragonie.com/blog/2019/03/definitive-2019-guide-cryptographic-key-sizes-and-algorithm-recommendations

Can expound on this as required.

internal/telemetry/metrics.go
```diff
@@ -0,0 +1,31 @@
...
+			Registry:  reg,
+		})
+	if err != nil {
+		return nil, fmt.Errorf("internal/telemetry: prometheus exporter: %v", err)
+	}
+	view.RegisterExporter(exporter)
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", exporter)
+	// todo: should be it's own http service... i know.
+	// mux.Handle("/metrics/debug/", http.StripPrefix("/metrics/debug", zpages.Handler))
```
desimone: pages apparently carry some overhead. Will address in FUP pull request.

internal/telemetry/middleware_test.go
```diff
@@ -25,10 +24,6 @@ func newTestMux() http.Handler {
 
 func Test_HTTPMetricsHandler(t *testing.T) {
 
-	chain := middleware.NewChain()
```
desimone: Middleware chain can be replaced with long form (otherwise I was getting a cyclical import error)

internal/telemetry/tracing.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// CollectorEndpoint is the full url to the Jaeger HTTP Thrift collector.
+	// For example, http://localhost:14268/api/traces
+	JaegerCollectorEndpoint string `mapstructure:"tracing_jaeger_collector_endpoint"`
+	// AgentEndpoint instructs exporter to send spans to jaeger-agent at this address.
+	// For example, localhost:6831.
+	JaegerAgentEndpoint string `mapstructure:"tracing_jaeger_agent_endpoint"`
+}
+
+func RegisterTracing(opts *TracingOptions) error {
```
desimone: Sets the stage for additional providers.

internal/telemetry/view.go
```diff
@@ -0,0 +1,57 @@
...
+	var views []*view.View
+	for _, v := range DefaultViews {
+		views = append(views, v...)
+	}
+	return registerView(views...)
+}
+
+// registerView registers one of the defined metrics views. It must be called
+// for metrics to see metrics in the configured exporters.
+func registerView(v ...*view.View) error {
```
desimone: nit: I replaced the slice of pointers with the variadic equivalent for consistency.

cmd/pomerium/main.go
```diff
@@ -157,19 +116,9 @@ func newProxyService(opt config.Options, mux *http.ServeMux) (*proxy.Proxy, erro
 	return service, nil
 }
 
-func newPromListener(addr string) {
```
desimone: moved into the metrics (now telemetry) package.

internal/httputil/http.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/internal/urlutil"
+)
+
+// NewHTTPServer starts a http server given a set of options and a handler.
+//
+// It is the caller's responsibility to Close() or Shutdown() the returned
+// server.
+func NewHTTPServer(opt *ServerOptions, h http.Handler) *http.Server {
```
desimone: Standardized interface (and defaults) for any HTTP server we want to ListenAndServe from given a set of handlers and options. That way we don't have a bunch of different servers with different timeouts, loggers, etc running around.

desimone:
I've tried (and failed) to come up with a way to break this into multiple PRs. 😭 At the very least, I'd like to wait for #227 .

cmd/pomerium/main.go
```diff
@@ -18,8 +16,8 @@ import (
 	"github.com/pomerium/pomerium/internal/config"
 	"github.com/pomerium/pomerium/internal/httputil"
 	"github.com/pomerium/pomerium/internal/log"
-	"github.com/pomerium/pomerium/internal/metrics"
 	"github.com/pomerium/pomerium/internal/middleware"
+	"github.com/pomerium/pomerium/internal/telemetry"
```
desimone: I'm trying to move as much as [possible outside of main](https://dave.cheney.net/practical-go/presentations/qcon-china.html#_keep_package_main_small_as_small_as_possible).

internal/middleware/reverse_proxy.go
```diff
@@ -6,11 +6,14 @@ import (
 
 	"github.com/pomerium/pomerium/internal/cryptutil"
 	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/internal/telemetry"
 )
 
 func SignRequest(signer cryptutil.JWTSigner, id, email, groups, header string) func(next http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			_, span := telemetry.StartSpan(r.Context(), "middleware.SignRequest")
```
travisgroth: Should this be keeping the returned context and calling r.WithContext() when we ServeHTTP on the next middleware?

internal/middleware/reverse_proxy.go
```diff
@@ -29,6 +32,9 @@ func SignRequest(signer cryptutil.JWTSigner, id, email, groups, header string) f
 func StripPomeriumCookie(cookieName string) func(next http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			_, span := telemetry.StartSpan(r.Context(), "middleware.StripPomeriumCookie")
```
travisgroth: Should this be keeping the returned context and calling r.WithContext() when we ServeHTTP on the next middleware?

cmd/pomerium/main.go
```diff
@@ -199,58 +144,61 @@ func wrapMiddleware(o *config.Options, mux http.Handler) http.Handler {
...
-		log.Error().Err(err).Msg("cmd/pomerium: could not reload configuration")
-		metrics.SetConfigInfo(opt.Services, false, "")
-		return opt
+func setupMetrics(opt *config.Options) {
+	if opt.MetricsAddr != "" {
+		if handler, err := telemetry.PrometheusHandler(); err != nil {
+			log.Error().Err(err).Msg("cmd/pomerium: couldn't start metrics server")
+		} else {
+			serverOpts := &httputil.ServerOptions{Addr: opt.MetricsAddr}
+			srv := httputil.NewHTTPServer(serverOpts, handler)
```
travisgroth: I like this pattern

proxy/proxy.go
```diff
@@ -236,6 +236,8 @@ type UpstreamProxy struct {
 
 // ServeHTTP handles the second (reverse-proxying) leg of pomerium's request flow
 func (u *UpstreamProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	_, span := telemetry.StartSpan(r.Context(), fmt.Sprintf("proxy.%s%s", r.Host, r.URL.Path))
```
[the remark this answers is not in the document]
desimone: That's a really good point. I think we should be a bit conservative here and if we get feedback that it's useful, then add it back.

proxy/proxy.go
```diff
@@ -236,6 +236,8 @@ type UpstreamProxy struct {
...
+	defer span.End()
 	u.handler.ServeHTTP(w, r)
```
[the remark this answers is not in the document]
desimone: Agree, but going to remove per https://github.com/pomerium/pomerium/pull/230#discussion_r306113269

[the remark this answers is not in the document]
travisgroth: Clarification of that comment - I was more wondering if we should remove the 'path' component.  The answer might be "no", though.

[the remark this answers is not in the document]
desimone: I guess we have a few options.:

1. Don't track it at all (because it will vary wildly depending on the downstream app being served, and that's out of scope for our app?)
2. Track path and hostname (great because we kinda get back-door tracing even if org doesn't have it in their apps, but bad because now we have a ton of potential traces.
3. Track just hostname not the path (easier to manage UI wise, get a sense for aggregate response times per downstream app, but we aren't very helpful if that app doesn't itself have tracing).

I feel like if we _are_ going to trace downstream routes, we might as well do it on a per-path basis.

proxy/clients/authorize_client.go
```diff
@@ -65,6 +69,9 @@ func (a *AuthorizeGRPC) Authorize(ctx context.Context, route string, s *sessions
 
 // IsAdmin takes a session and returns whether the user is an administrator
 func (a *AuthorizeGRPC) IsAdmin(ctx context.Context, s *sessions.SessionState) (bool, error) {
+	ctx, span := telemetry.StartSpan(ctx, "proxy.client.grpc.Authorize")
```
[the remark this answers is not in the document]
desimone: Yes! Good catch 😄

internal/middleware/reverse_proxy.go
desimone: Yes, very much so. great catch.

cmd/pomerium/main.go
desimone: Awesome! Me too. I think this paves the way if/when we have to serve gRPC distinctly from our HTTP(s) handlers.

internal/telemetry/tracing_test.go
```diff
@@ -0,0 +1,23 @@
...
+		wantErr bool
+	}{
+		{"jaeger", &TracingOptions{JaegerAgentEndpoint: "localhost:6831", Service: "all", Provider: "jaeger"}, false},
+		{"jaeger with debug", &TracingOptions{JaegerAgentEndpoint: "localhost:6831", Service: "all", Provider: "jaeger", Debug: true}, false},
+		{"jaeger no endpoint", &TracingOptions{JaegerAgentEndpoint: "", Service: "all", Provider: "jaeger"}, true},
+		{"unknown provider", &TracingOptions{JaegerAgentEndpoint: "localhost:0", Service: "all", Provider: "Lucius Cornelius Sulla"}, true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := RegisterTracing(tt.opts); (err != nil) != tt.wantErr {
```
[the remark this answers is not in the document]
desimone: Not in a practical way that I could figure out. Open to ideas 👍

desimone:
> One larger scope question - I was already feeling like internal/metrics package was getting a bit big. Should internal/telemetry contain internal/telemetry/{metrics,tracing} to better break up functionality? I didn't get the feeling that the metrics and tracing code were sharing much except maybe presence in some of the middleware. Those could continue live top level in internal/telemetry.

Great idea, @travisgroth .  I've done my best to do that.  I also split the metrics package into what felt like a more logical separation of concerns (`http` vs `grpc` vs `info`). Let me know what you think !

internal/telemetry/tracing_test.go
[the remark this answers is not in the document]
travisgroth: I concur.  Not seeing anything to latch onto.  Oh well.

internal/telemetry/metrics/grpc.go
```diff
@@ -12,93 +12,98 @@ import (
...
 	// request duration by pomerium service, grpc service, grpc method, and status
 	GRPCServerRequestDurationView = &view.View{
-		Name:        "grpc_server_request_duration_ms",
+		Name:        "grpc/server/request_duration_ms",
 		Measure:     ocgrpc.ServerLatency,
 		Description: "grpc Request duration in ms",
-		TagKeys:     []tag.Key{keyService, keyGRPCMethod, ocgrpc.KeyServerStatus, keyGRPCService},
-		Aggregation: grcpLatencyDistribution,
+		TagKeys:     []tag.Key{TagKeyService, TagKeyGRPCMethod, ocgrpc.KeyServerStatus, TagKeyGRPCService},
+		Aggregation: DefaultMillisecondsDistribution,
```
travisgroth: This makes sense

internal/telemetry/metrics/grpc.go
```diff
@@ -12,93 +12,98 @@ import (
...
+		GRPCServerRequestCountView,
+		GRPCServerRequestDurationView,
+		GRPCServerResponseSizeView,
+		GRPCServerRequestSizeView}
 
 	// GRPCServerRequestCountView is an OpenCensus view which counts GRPC Server
 	// requests by pomerium service, grpc service, grpc method, and status
 	GRPCServerRequestCountView = &view.View{
-		Name:        "grpc_server_requests_total",
+		Name:        "grpc/server/requests_total",
```
travisgroth: IIRC this gets fixed up on the exporter side so this should be okay

desimone: Yeah -- I took this pattern from https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ochttp/stats.go#L216

PR #237, opened 2019-07-22
Fix checksum integer overflow in metrics

travisgroth:
Fixes issue with some checksums being larger than int64

```{"level":"warn","error":"strconv.ParseInt: parsing \"0xd18a518ef7190001\": value out of range","time":"2019-07-22T08:52:08-04:00","message":"Could not parse config checksum into integer"}```

**Checklist**:
- [x] ready for review

### DIFF
--- a/cmd/pomerium/main.go
+++ b/cmd/pomerium/main.go
@@ -213,11 +213,13 @@ func parseOptions(configFile string) (*config.Options, error) {
 	metrics.AddPolicyCountCallback(o.Services, func() int64 {
 		return int64(len(o.Policies))
 	})
-	checksumInt, err := strconv.ParseInt(fmt.Sprintf("0x%s", o.Checksum()), 0, 64)
+
+	checksumDec, err := strconv.ParseUint(o.Checksum(), 16, 64)
 	if err != nil {
-		log.Warn().Err(err).Msg("Could not parse config checksum into integer")
+		log.Warn().Err(err).Msg("Could not parse config checksum into decimal")
 	}
-	metrics.SetConfigChecksum(o.Services, checksumInt)
+	metrics.SetConfigChecksum(o.Services, checksumDec)
+
 	return o, nil
 }
 
--- a/internal/metrics/info_test.go
+++ b/internal/metrics/info_test.go
@@ -49,7 +49,7 @@ func Test_SetBuildInfo(t *testing.T) {
 	}
 
 	SetBuildInfo("test_service")
-	testMetricRetrieval(registry.registry.Read(), t, wantLabels, 1, "build_info")
+	testMetricRetrieval(registry.registry.Read(), t, wantLabels, int64(1), "build_info")
 }
 
 func Test_AddPolicyCountCallback(t *testing.T) {
@@ -65,11 +65,11 @@ func Test_AddPolicyCountCallback(t *testing.T) {
 func Test_SetConfigChecksum(t *testing.T) {
 	registry = newMetricRegistry()
 
-	wantValue := int64(42)
+	wantValue := uint64(42)
 	wantLabels := []metricdata.LabelValue{{Value: "test_service", Present: true}}
 	SetConfigChecksum("test_service", wantValue)
 
-	testMetricRetrieval(registry.registry.Read(), t, wantLabels, wantValue, "config_checksum_int64")
+	testMetricRetrieval(registry.registry.Read(), t, wantLabels, float64(wantValue), "config_checksum_decimal")
 }
 
 func Test_RegisterInfoMetrics(t *testing.T) {
--- a/internal/metrics/helpers_test.go
+++ b/internal/metrics/helpers_test.go
@@ -36,7 +36,15 @@ func testDataRetrieval(v *view.View, t *testing.T, want string) {
 	}
 }
 
-func testMetricRetrieval(metrics []*metricdata.Metric, t *testing.T, labels []metricdata.LabelValue, value int64, name string) {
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: cmd/pomerium/main.go [hunk 1]
```diff
@@ -213,11 +213,15 @@ func parseOptions(configFile string) (*config.Options, error) {
 	metrics.AddPolicyCountCallback(o.Services, func() int64 {
 		return int64(len(o.Policies))
 	})
-	checksumInt, err := strconv.ParseInt(fmt.Sprintf("0x%s", o.Checksum()), 0, 64)
+
+	float := big.Float{}
```
desimone: 🚨🚨🚨Super nit-picky alert 🚨🚨🚨 

I like using the form `var float big.Float` in cases like this because I think it makes it more clear we are [not initializing the variable](https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_declaration_style).  I'll defer to your preference. 😄

internal/metrics/info_test.go
```diff
@@ -65,11 +65,11 @@ func Test_AddPolicyCountCallback(t *testing.T) {
 func Test_SetConfigChecksum(t *testing.T) {
 	registry = newMetricRegistry()
 
-	wantValue := int64(42)
+	wantValue := float64(42)
```
desimone: I can't believe I missed the answer to the ultimate question in life on the first review 😆

cmd/pomerium/main.go
```diff
@@ -213,11 +213,15 @@ func parseOptions(configFile string) (*config.Options, error) {
...
+	checksumDec, _, err := float.Parse(fmt.Sprintf("0x%s", o.Checksum()), 0)
+	//checksumInt, err := strconv.ParseFloat(fmt.Sprintf("0x%s", o.Checksum()), 64)
```
desimone:
```
cmd/pomerium/main.go:219:2: commentedOutCode: may want to remove commented-out code (gocritic)
	//checksumInt, err := strconv.ParseFloat(fmt.Sprintf("0x%s", o.Checksum()), 64)
```
Annoying, but the linter is probably right here

cmd/pomerium/main.go
travisgroth: Ah yes that's fine.  The usage of Float is weird in general; I updated to something a little more compact after a second glance.

internal/metrics/info_test.go
travisgroth: Most important value to test.

cmd/pomerium/main.go
travisgroth: Thought I had removed this. 👍

cmd/pomerium/main.go
```diff
@@ -213,11 +213,20 @@ func parseOptions(configFile string) (*config.Options, error) {
...
-	checksumInt, err := strconv.ParseInt(fmt.Sprintf("0x%s", o.Checksum()), 0, 64)
+
+	// checksumDec, _, err := big.ParseFloat(
+	// 	fmt.Sprintf("0x%s", o.Checksum()),
+	// 	0, 64, big.ToNearestEven)
+	// if err != nil {
+	// 	log.Warn().Err(err).Msg("Could not parse config checksum into decimal")
+	// }
+	// checksumFloat64, _ := checksumDec.Float64()
+	checksumDec, _ := strconv.ParseUint(fmt.Sprintf("0x%s", o.Checksum()), 0, 64)
```
desimone: errcheck :)

desimone: `strconv.ParseUint(o.Checksum(), 16, 64)` ?

EDIT: nvm.  TIL base 0 is equivalent if marked with octal prefix
```go
	case base == 0:
		// Look for octal, hex prefix.
		switch {
		case s[0] == '0' && len(s) > 1 && (s[1] == 'x' || s[1] == 'X'):
			if len(s) < 3 {
				return 0, syntaxError(fnParseUint, s0)
			}
			base = 16
			s = s[2:]
		case s[0] == '0':
			base = 8
			s = s[1:]
		default:
			base = 10
		}


```

cmd/pomerium/main.go
```diff
@@ -213,11 +213,20 @@ func parseOptions(configFile string) (*config.Options, error) {
...
 	if err != nil {
-		log.Warn().Err(err).Msg("Could not parse config checksum into integer")
+		log.Warn().Err(err).Msg("Could not parse config checksum into decimal")
```
desimone: uint :)

travisgroth: Soft counter argument - my thinking was that there's really no need to get type specific outside the code itself.  We're just trying to get it into base10.  I changed a few other refs along the same lines.

travisgroth:
>As far as I understand, we aren't incrementing/decrementing the hash value so those bytes are just a representation of the hash value of the struct right; the underlying data will be the same whether its a uint64, a float64, or a int64 right?

As discussed - the shift to a full float doesn't seem necessary after a closer look at hashstructure.  We can pass along a uint until the metrics code and then cast to float64 there.  It can't stay as unint, because the OC metrics gauges only supported signed types.  So, we need to use a float64 gauge to support the possible value range.

cmd/pomerium/main.go
travisgroth: It's actually cleaner to be explicit with the base and avoid the Sprintf().  Changed.

cmd/pomerium/main.go
desimone: Good point.

cmd/pomerium/main.go
desimone: 👍

PR #241, opened 2019-07-24
authenticate: copy redirect url by value

desimone:
Fixes regression in redirect URL being passed by reference to identity from authenticate.
 
See:
- https://github.com/pomerium/pomerium/pull/202/files#diff-54dcb0c4d47e021d5e418c259fbb6669R73
- https://github.com/travisgroth/pomerium/blob/29eee409ef9307c253124108fda0808219762293/authenticate/authenticate.go#L78-L79

**Checklist**:
- [x] unit tests added
- [x] ready for review

### DIFF
--- a/authenticate/authenticate.go
+++ b/authenticate/authenticate.go
@@ -18,7 +18,7 @@ import (
 // The checks do not modify the internal state of the Option structure. Returns
 // on first error found.
 func ValidateOptions(o config.Options) error {
-	if o.AuthenticateURL == nil {
+	if o.AuthenticateURL.String() == "" {
 		return errors.New("authenticate: 'AUTHENTICATE_SERVICE_URL' missing")
 	}
 	if o.ClientID == "" {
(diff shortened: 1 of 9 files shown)


desimone:
y no codecov?

authenticate/authenticate.go
```diff
@@ -75,7 +75,7 @@ func New(opts config.Options) (*Authenticate, error) {
 	if err != nil {
 		return nil, err
 	}
-	redirectURL := opts.AuthenticateURL
+	redirectURL := *opts.AuthenticateURL
```
travisgroth: Maybe add a warning comment here if we keep this as-is.

desimone: > The deref in Authenticate fixes the immediate problem, but I'm inclined to pass in a fully dereferenced Options struct to the services. Once inside, we can deal in pointers if needed.

> That said, this looks fine, but I'd favor the value based approach as a safer pattern. Thoughts?

I agree. Let me take another pass.

desimone:
@travisgroth updated. Also changed gRPC clients to use value URLs.

PR #255, opened 2019-08-03
deployment: use goproxy for docker

desimone:
Though `make build` uses goproxy, the caching layer of our docker images does not. 

**Checklist**:
- [ ] ready for review


desimone:
We are so close to `1.13` (I believe August is the release date?) might shelf this and refactor the images when GOPROXY becomes default.

Issue #179, opened 2019-06-10
Allow proxy for https site running w/ self-signed certificate


haozhou:
**Is your feature request related to a problem?
It's not able to reverse proxy an internal https website use self-signed certificate.
The error is 6:36PM ERR http: proxy error: x509: certificate signed by unknown authority proxy=192.168.1.2:8006

**Describe the solution you'd like**
Add config parameter that ignores the certificate validation.

alexandrem:
A `disable_ssl_cert_validation` attribute could be useful, but I'd like to be able to specify a cert/key content as well.

I'm personally looking forward to host Pomerium in Kubernetes, so that would translate to a secret name somehow. We have to think about how to expose that in a generic way however in the main policy config file by possibly injecting the full cert base64 content there.

Ideally, I think we need to rethink the way the policies and routes are managed.

desimone:
@haozhou  

> `disable_ssl_cert_validation`

Make sense to me. A contribution that added a route level config that explicitly disabled TLS validation like this would be welcome. 

Hi @alexandrem 

I am probably misunderstanding your comment but... I think much of what you are suggesting should be possible today. 

>  I'd like to be able to specify a cert/key content as well.

Again, I'm not sure I follow. That should be possible with the certificate and certificate_key [config settings](https://www.pomerium.io/reference/#certificate-authority). Or do you mean to the proxied routes? Do you mind clarifying what you had in mind?

> I'm personally looking forward to host Pomerium in Kubernetes, so that would translate to a secret name somehow

What secret name?

> We have to think about how to expose that in a generic way however in the main policy config file by possibly injecting the full cert base64 content there.

This should be possible with environmental variables today. 

> Ideally, I think we need to rethink the way the policies and routes are managed.

Open to suggestions, do you mind expanding on that? Thanks!

alexandrem:
Hi @desimone,
I was talking about the config for each individual proxied routes.

Global environment variable is only good if all backends are using the same self-signed certificate.

We either need to be able to specify base64 encoded cert content for the route config or to refer to a Kubernetes secret name that exists in the same namespace where Pomerium is deployed.

desimone:
fyi @haozhou @alexandrem  this will land in `v0.1.0`  and is one `latest` Look forward to your thoughts.

PR #259, opened 2019-08-11
proxy: enable client certificate in transport

desimone:
This would enable downstream applications to enforce mutually authenticated connections for TLS connections.  An example app can be found in the [examples repository](https://github.com/pomerium/examples/tree/master/mutual-tls).

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review


internal/config/policy.go
```diff
@@ -47,25 +49,45 @@ type Policy struct {
...
+		return fmt.Errorf("internal/config: policy bad destination url %s", err)
 	}
 
 	// Only allow public access if no other whitelists are in place
 	if p.AllowPublicUnauthenticatedAccess && (p.AllowedDomains != nil || p.AllowedGroups != nil || p.AllowedEmails != nil) {
-		return errors.New("internal/config: route marked as public but contains whitelists")
+		return fmt.Errorf("internal/config: policy route marked as public but contains whitelists")
+	}
+
+	if p.TLSClientCert != "" && p.TLSClientKey != "" {
```
travisgroth: Should probably throw an error if only one is configured.

alagendijk-minddistrict:
Could you also add an option for these to be files instead of only the base64 encoded string?

travisgroth:
>Could you also add an option for these to be files instead of only the base64 encoded string?

Agreed.  I read the cryptutil changes before sufficient coffee.

desimone:
Thanks @alagendijk-minddistrict & @travisgroth ; 
- added support to add client certs as files in addition to as base64 blobs.
- added support to add custom ca (#179) as files too.  
- updated the example repo to prefer files which are easier to read
- added a few log statements for non-standard routes.

PR #272, opened 2019-08-21
httputil: standardize http error handling

desimone:
### Summary

These changes add a http error wrapper to be used to wrap http errors. `httputil.Error` wraps the standard error interface and allows for additional context including a user friendly (and safe) error to display, error nesting, and setting HTTP status code. 

The standard error response handler has been adapted to use this wrapper to make a determination as to whether the displayed error is debug-able (and thus should have a troubleshoot your error link). 

Note: xerrors can be replaced with the release of go 1.13

### See also:
- https://github.com/golang/go/wiki/ErrorValueFAQ
- https://godoc.org/golang.org/x/xerrors
- https://go.googlesource.com/proposal/+/master/design/29934-error-values.md
- https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully
- https://crawshaw.io/blog/xerrors
- https://github.com/golang/go/blob/d9b13233378668a4fd24ac5d044e9d550cd2a8be/src/os/error.go#L39-L44

**Checklist**:
- [x] ready for review
- [x] additional context
- [x] unit tests updated


authenticate/handlers.go
```diff
@@ -54,93 +59,80 @@ func (a *Authenticate) authenticate(w http.ResponseWriter, r *http.Request, sess
...
 	}
 	return nil
 }
 
-// SignIn handles the sign_in endpoint. It attempts to authenticate the user,
-// and if the user is not authenticated, it renders a sign in page.
 func (a *Authenticate) SignIn(w http.ResponseWriter, r *http.Request) {
 	session, err := a.sessionStore.LoadSession(r)
 	if err != nil {
-		switch err {
```
desimone: n.b.: If we get an error, we don't really care if it's expected or not. We want to log it, and repropmpt. Some spill over from just updating errors, but felt related enough.

internal/middleware/middleware.go
```diff
@@ -100,6 +90,7 @@ func SameDomain(u, j *url.URL) bool {
 	if err != nil {
 		return false
 	}
+	fmt.Println(u, j)
```
travisgroth: Leftover from debugging?

proxy/handlers.go
```diff
@@ -294,28 +268,23 @@ func (p *Proxy) UserDashboard(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := p.authenticate(w, r, session); err != nil {
-		log.FromRequest(r).Error().Err(err).Msg("proxy: authenticate failed")
-		httpErr := &httputil.Error{Code: http.StatusUnauthorized, CanDebug: true}
-		httputil.ErrorResponse(w, r, httpErr)
+		p.sessionStore.ClearSession(w, r)
```
travisgroth: This looks new.  Intentional?

proxy/handlers.go
```diff
@@ -351,38 +320,29 @@ func (p *Proxy) UserDashboard(w http.ResponseWriter, r *http.Request) {
...
 
 	// reject a refresh if it's been less than the refresh cooldown to prevent abuse
 	if time.Since(iss) < p.refreshCooldown {
-		log.FromRequest(r).Error().Dur("cooldown", p.refreshCooldown).Err(err).Msg("proxy: refresh cooldown")
-		httpErr := &httputil.Error{
-			Message: fmt.Sprintf("Session must be %v old before refresh", p.refreshCooldown),
-			Code:    http.StatusBadRequest}
-		httputil.ErrorResponse(w, r, httpErr)
+		httputil.ErrorResponse(w, r,
+			httputil.Error(fmt.Sprintf("Session must be %v old before refreshing", p.refreshCooldown), http.StatusBadRequest, nil))
```
travisgroth: I feel like we should avoid %v here and cast the interval into a string or integer.

authenticate/handlers.go
```diff
@@ -54,93 +59,80 @@ func (a *Authenticate) authenticate(w http.ResponseWriter, r *http.Request, sess
...
-			return fmt.Errorf("authenticate: session valid: %v : %v", valid, err)
+			return xerrors.Errorf("session valid: %v : %w", valid, err)
 		}
 	}
 	return nil
 }
 
-// SignIn handles the sign_in endpoint. It attempts to authenticate the user,
-// and if the user is not authenticated, it renders a sign in page.
 func (a *Authenticate) SignIn(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
desimone: Thanks!

internal/middleware/middleware.go
desimone: Great catch!

proxy/handlers.go
desimone: Yes but I'll address in FUP PR with more tests and context.

authenticate/handlers.go
```diff
@@ -204,83 +197,71 @@ func (a *Authenticate) OAuthStart(w http.ResponseWriter, r *http.Request) {
...
-// getOAuthCallback completes the oauth cycle from an identity provider's callback
 func (a *Authenticate) getOAuthCallback(w http.ResponseWriter, r *http.Request) (string, error) {
-	// handle the callback response from the identity provider
 	if err := r.ParseForm(); err != nil {
-		return "", httputil.Error{Code: http.StatusInternalServerError, Message: err.Error()}
+		return "", httputil.Error("invalid signature", http.StatusBadRequest, err)
 	}
-	errorString := r.Form.Get("error")
-	if errorString != "" {
-		log.FromRequest(r).Error().Str("Error", errorString).Msg("authenticate: provider returned error")
```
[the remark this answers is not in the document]
desimone: Not sure I follow, but I think we want to return the IdP error string first here.

proxy/handlers.go
travisgroth: Makes sense to me

PR #276, opened 2019-08-23
dev: add code climate

desimone:
Thoughts?

**Checklist**:
- [x] ready for review


desimone:
Looks like codecov is getting jealous a bit :)

PR #278, opened 2019-08-23
Make auth flow diagram fonts consistent

travisgroth:
**Checklist**:
- [x] updated docs
- [x] ready for review


travisgroth:
draw.io has it's own opinion of what font that is.  Kicked it in the head, and I think we're good now.

Issue #282, opened 2019-08-27
proxy: handle CORs preflight check as middleware


desimone:
**Is your feature request related to a problem? Please describe.**

Currently, preflight checks are added ad-hoc before a proxied route's `ServeHTTP` function and not at all in the authenticate service. 

**Describe the solution you'd like**

CORS preflight checks should be handled as middleware not only for proxied routes, but for pomerium itself. 

**Explain any additional use-cases**

This ties into https://github.com/pomerium/pomerium/issues/281 where we should be providing `OPTIONS` feedback. 

**Additional context**

- https://github.com/rs/cors/blob/master/examples/alice/server.go
- https://github.com/go-chi/cors
- #86 
- #78

Issue #281, opened 2019-08-27
Gracefully handle API calls from SPA through Pomerium


victornoel:
**Is your feature request related to a problem? Please describe.**

When a Single Page Application (SPA) + its API is protected by pomerium, two types of HTTP requests happen:
- page loading by the browser (the html + spa code)
- API calls by the SPA itself from Javascript

While the first one works nicely with pomerium (i.e., when not auth, the browser redirects with 302 the user to authentication and then back to the page), the second has some undesired caveats currently.

In particular:
- When the session expires, an API call will get as a response a 302 redirect, which results in a redirect to authentication, which is of course not presented to the user (since it's happening in javascript)
- Sometimes, the browser sends an OPTIONS request to the authentication service, that then responds with a 302: this is not an expected answer for OPTIONS.

So while the SPA-related behaviour seems to work as expected, the API-related one does not: when a session ends, the SPA can't detect properly that it needs to make authentication happen again by explicitly redirecting the user to the authentication page.

As a side note, this mean that the SPA is aware it is served from authenticating proxy and that it has to do something specific when it detects the situation:
- one idea is to simply reload the page and let the browser take over (for the record, this means that the browser shouldn't rely on its cache for loading the SPA, this is a known problem in the SPA world)
- another idea is to use more complex authentication schema that gives even more control on the authentication to the SPA. It's not clear if this is out of scope of the present issue or not.

More details on the matter here:
- https://pomerium-io.slack.com/archives/CK6SVMPU0/p1566836804003000
- https://pomerium-io.slack.com/archives/CK6SVMPU0/p1566839236007300
- Recommended authentication scheme for SPAs: https://auth0.com/blog/oauth2-implicit-grant-and-spa/

**Describe the solution you'd like**

Concerning the first point, the original need is for the SPA (when making API calls) being capable of detecting that the session was terminated and that authentication should happen.

I would expect something like getting a 401 instead of a 302 when this happens. This presupposes that pomerium is able to consider differently normal browser page loading from API calls.

One option for this is to consider that if the request has the pomerium cookie, it is browser-based pag eloading, and if it has an authentication header (containing the jwt token normally available to the backend I suppose?), then it is an API call.

This solution implies though that there is the need for the frontend to become aware of the jwt token, either it being provided by the backend (extracted from the cookie) or by pomerium itself (this is where we enter the world of advanced authentication scheme for SPA I think).

[...]

Issue #262, opened 2019-08-14
proxy: handle double slash in path


alagendijk-minddistrict:
**Describe the bug**

When requesting a URL through pomerium that contains a double slash in the path pomerium redirects this request to the same URL with just a single slash
example: https://test.corp.example.com/foo//bar will be redirected to https://test.corp.example.com/foo/bar

**To Reproduce** Steps to reproduce the behavior:

1. Requested https://test.corp.example.com/foo//bar
2. Got redirect to https://test.corp.example.com/foo/bar

**Expected behavior**
Pomerium proxies a URL containing // in the path to the endpoint instead off redirecting it.

**Environment:**

- Pomerium version: v0.2.1+47f7a44
- Server Operating System/Architecture/Cloud: Ubuntu 18.04

Logs(s):

```text
10:32AM DBG http-request duration=0.215143 email= group= ip=<removed> method=GET req_id=40ccf958-5809-e524-9995-fe9592701cdb service=all size=43 status=301 url=/foo//bar user_agent="<removed>"
10:32AM DBG http-request duration=34.982312 email=<removed> group="<removed>" ip=<removed> method=GET req_id=f339ff81-11c4-eb4b-3c0a-0d6de7316960 service=all size=233 status=404 url=/foo/bar user_agent="<removed>"
```

**Additional context**

According to https://tools.ietf.org/html/rfc3986#section-3.3 a request path can not start with a double slash, but can have a double slash somewhere else in the path.
I encountered this issue with Jenkins, which does a reverse proxy check to a URL which contains its own full URL in the path.

desimone:
@alagendijk-minddistrict Just to confirm, is the double slash either immediately following the domain or at the end of the path?

I don't think [url.Parse is doing this](https://play.golang.org/p/4k3zAGOYaRB) but it could be happening within the reverse proxy by [singleJoiningSlash](https://golang.org/src/net/http/httputil/reverseproxy.go?s=2687:2730#L79) but I wouldn't expect that in the middle of the string.  Or, the path is being cleaned by Go's default ServeMux.

https://github.com/golang/go/blob/1962dc88eb89cd37d3f9f85e8e4b7ad4915db089/src/net/http/server.go#L2212-L2232

I haven't figure it out yet but I'm looking into it :)

alagendijk-minddistrict:
According to the RFC (if I understand it correctly) a double slash following the domain is not allowed. So in that case a redirect seems logical to me.
[...]

PR #300, opened 2019-09-12
all: refactor handler logic

### LINKED ISSUE #306: authenticate: oauth csrf errors
Greetings and thanks for this nice project :)

I've installed Pomerium binary v0.3.0+bc72d08 on Debian 9.11

At first it was working, but after a few minutes I get only "Internal server errors", such as:
- oauth callback : csrf failure: %!w(<nil>)

I tried to restart Pomerium and clear cookies.

What did I do wrong?

Thanks in advance

desimone:
These changes move user authentication verification & handling, csrf protection, and service mutual authentication form handlers to middleware.

**Why**:

- CSRF mitigation (both between user sessions and from the identity provider) was difficult to reason about, sprawling, and error prone. 
- Session and csrf store were both implemented on top of cookie store in a confusing (and often overlapping) way. 
- Go's built in ServeMux has started to show it's limitations. In particular, #262 and lack of [named matched hosts](https://www.gorillatoolkit.org/pkg/mux#Route.Host) which required us to implement our own route hashmap. 
- User sessions were repeatedly being loaded from session 1) during middleware checks 2) during the handler itself. Now, subsequent per-request calls requiring user state can be fetched from context which is (imho) much more idiomatic.
- Enables the follow up (related) PRs coming for:
    - #282 
    - #262 
    - #281 
- Fixes #306

**Details**
- all: prefer `FormValues` to `ParseForm` with subsequent `Form.Get`s
- all: refactor authentication stack to be checked by middleware, and accessible via request context.
- all: replace http.ServeMux with gorilla/mux’s router
- all: replace custom CSRF checks with gorilla/csrf middleware
- authenticate: extract callback path as constant.
- internal/config: implement stringer interface for policy
- internal/cryptutil: add helper func `NewBase64Key`
- internal/cryptutil: rename `GenerateKey` to `NewKey`
- internal/cryptutil: rename `GenerateRandomString` to `NewRandomStringN`
- internal/middleware: removed alice in favor of gorilla/mux
- internal/sessions: remove unused `ValidateRedirectURI` and `ValidateClientSecret`
- internal/sessions: replace custom CSRF with gorilla/csrf fork that supports custom handler protection
- internal/urlutil: add `SignedRedirectURL` to create hmac'd URLs
- internal/urlutil: add `ValidateURL` helper to parse URL options
- internal/urlutil: add `GetAbsoluteURL` which takes a request and returns its absolute URL.
- proxy: remove holdover state verification checks; we no longer are setting sessions in any proxy routes so we don’t need them.
- proxy: replace un-named http.ServeMux with named domain routes.

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review


proxy/proxy.go
```diff
@@ -34,29 +35,32 @@ const (
 	HeaderGroups = "x-pomerium-authenticated-user-groups"
 )
 
+var (
```
travisgroth: should these be const?

desimone: Yes, unfortunately Go doesn't support const structs (AFAIK). 

I split the difference and made the constant values strings and then cast them to URLs in the constructor body. Let me know what you think.

desimone:
@travisgroth incorporated changes. let me know if looks good to go, if so I'll squash and merge.

proxy/proxy.go
travisgroth: Ah.  Yeah that makes sense.

PR #316, opened 2019-09-18
authenticate: encrypt & mac oauth callback state

desimone:
The existing implementation of validating the 'state' parameter returned from the identity provider's callback verifies authenticity and guards against replay protection using HMAC. However, this potentially leaks the original intended route to the identity provider and any logs in-between. In these changes, we provide confidentiality by encrypting `redirect_uri` while still MAC'ing the timestamp, nonce (csrf) using go's [AEAD](https://golang.org/pkg/crypto/cipher/#AEAD) construction.

## details
- cryptutil: add hmac & tests
- cryptutil: rename cipher / encoders to be more clear
- cryptutil: simplify SecureEncoder interface
- cryptutil: renamed NewCipherFromBase64 to NewAEADCipherFromBase64
- cryptutil: move key & random generators to helpers

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review


internal/cryptutil/sign.go
```diff
@@ -9,6 +9,11 @@ import (
 	"gopkg.in/square/go-jose.v2/jwt"
 )
 
+const (
+	// DefaultLeeway defines the default leeway for matching NotBefore/Expiry claims.
+	DefaultLeeway = 2.0 * time.Minute
```
travisgroth: 5 might be a good number here as it builds on Kerberos time sync constraint, which is probably a baked in assumption in other adjacent systems.

authenticate/handlers.go
```diff
@@ -137,15 +138,21 @@ func (a *Authenticate) SignOut(w http.ResponseWriter, r *http.Request) {
 }
 
 // redirectToIdentityProvider starts the authenticate process by redirecting the
-// user to their respective identity provider.
-//
+// user to their respective identity provider. This function also builds the
+// 'state' parameter which is encrypted and includes authenticating data
+// for validation.
+// 'state' is : nonce|timestamp|redirect_url|encrypt(redirect_url)+mac(nonce,ts))
```
travisgroth: Is this accurate?  I'm having trouble finding all of those fields being inserted.

authenticate/handlers.go
```diff
@@ -187,19 +194,33 @@ func (a *Authenticate) getOAuthCallback(w http.ResponseWriter, r *http.Request)
...
 		return nil, httputil.Error("malformed state", http.StatusBadRequest, err)
 	}
-	// split state into its it's components (nonce:redirect_uri)
-	statePayload := strings.SplitN(string(bytes), ":", 2)
-	if len(statePayload) != 2 {
-		return nil, fmt.Errorf("state malformed, size: %d", len(statePayload))
+
+	// split state into its it's components, e.g.
+	// (nonce|timestamp|redirect_url|encrypted_data(redirect_url)+mac(nonce,ts))
+	statePayload := strings.SplitN(string(bytes), "|", 3)
```
travisgroth: There's a small dark part of me that thinks we should maybe use an off the shelf ser/deser...this should be sufficient, but it feels a little fiddly.

authenticate/handlers.go
```diff
@@ -137,15 +138,21 @@ func (a *Authenticate) SignOut(w http.ResponseWriter, r *http.Request) {
...
+// 'state' is : nonce|timestamp|redirect_url|encrypt(redirect_url)+mac(nonce,ts))
+
 // https://openid.net/specs/openid-connect-core-1_0-final.html#AuthRequest
 // https://tools.ietf.org/html/rfc6749#section-4.2.1
 func (a *Authenticate) redirectToIdentityProvider(w http.ResponseWriter, r *http.Request) {
 	redirectURL := a.RedirectURL.ResolveReference(r.URL)
 	nonce := csrf.Token(r)
-	state := fmt.Sprintf("%v:%v", nonce, redirectURL.String())
-	encodedState := base64.URLEncoding.EncodeToString([]byte(state))
+	now := time.Now().Unix()
```
desimone: @travisgroth RE: https://github.com/pomerium/pomerium/pull/316/files#r327387012

` nonce | timestamp | ${ouput of our aead call}`

authenticate/handlers.go
desimone: See : https://github.com/pomerium/pomerium/pull/316/files#r327393530

authenticate/handlers.go
desimone: I considered it as well but in this case I want/need to be able to pluck out the associated cleartext data first (the csrf and timestamp) before decrypting/verifying the mac. The [CSRF nonce is checked in middleware](https://github.com/pomerium/csrf/blob/f3318380bad3b451772163283269d0454da833b3/helpers.go#L121).   So the process is:

1. Middleware checks for the cleartext CSRF in the state query param. If invalid, reject.
1.  Check cleartext timestamp within leeway, if invalid, reject. (mac not checked yet!)
1.  Decrypt the payload (redirect_uri) and verify the authenticity of the CSRF/timestamp. If any fail, reject. 

The only weird edge-case I could see is the IdP intentionally forging the CSRF or timestamp, but that would just result in rejection ... or maybe a denial of service to the tend user who now can't log in. Of course, if you don't trust your identity provider, you've got bigger issues here. 

Hopefully that makes sense?

internal/cryptutil/sign.go
desimone: Thanks, I was going back and forth about this. Really good idea.

authenticate/handlers.go
travisgroth: Yep makes sense.  I think in practice this is fine.

authenticate/handlers.go
travisgroth: Ah, I didn't go deep enough.  👍

internal/cryptutil/sign.go
desimone: Fixed!

Issue #291, opened 2019-08-30
authorize does not check 'to' field in policy configuration


pecigonzalo:
**Describe the solution you'd like**

Currently the authorizer requires a minimal config as follows
```
policy:
  - from: https://httpbin.domain.tld
    to: http://httpbin
    allowed_domains:
      - domain.tld
  - from: https://external-httpbin.domain.tld
    to: http://httpbin
    allow_public_unauthenticated_access: true
```
If `to:` section is removed, it will complain that it is missing. Ideally, it should allow to be empty as  https://github.com/pomerium/pomerium/blob/master/authorize/identity.go#L64-L77 does not use the `to: ` field (which makes sense as policies are usually set for external domains).

This would allow for this sort of config, which simplifies and detaches `authz` and `authn` responsibilities completely, as they dont have to know about each other.
- **authz.yml**
```
policy:
  - from: https://httpbin.domain.tld
    allowed_domains:
      - domain.tld
```
- **proxy.yml**
```
authenticate_service_url: https://authn.domain.tld
authorize_service_url: https://authz.domain.tld
policy:
  - from: https://httpbin.domain.tld
    to: http://httpbin
  - from: https://external-httpbin.domain.tld
    to: http://httpbin
    allow_public_unauthenticated_access: true
```
- **authn.yml**
```
authenticate_service_url: https://authn.domain.tld
idp_provider: google
idp_client_id: REPLACE_ME
idp_client_secret: REPLACE_ME
# idp_service_account: REPLACE_ME
```

### Workaround
As a workaround, we can just set `authz` configuration to dummy values
```
policy:
  - from: https://httpbin.domain.tld
    to: https://domain.tld
    allowed_domains:
      - domain.tld
```

travisgroth:
Hi @pecigonzalo,

Thanks for opening this.  

Operationally, I think it makes more sense to keep policy as a format across components, even if they don't all need every field.  It's easier to validate you're consistent and you can share/review configurations directly.  Less chance of error.

That said, I think you've identified something we should actually improve.  `to` is the backend service, and I think the backend target of a policy entry is as important - if not more important - than the `from` that a user hit on the frontend.  

By only checking `from`, we open the door for some mistakes, unintended behavior, and potential inconsistency in access to a given service.  In a situation where different teams are responsible for proxy mappings and service access policies, the latter group would not have any way of knowing what service is actually behind a given proxy route without knowing the `to` of a route.  Arguably, they might _only_ care about `to`.

[...]

Issue #64, opened 2019-03-15
Allow running behind external TLS termination / reverse proxy


eriksw:
I'm interested in using Pomerium, but would like to run it behind an external load balancer so that I don't have to deal with Pomerium needing its own certificate, but I don't see a way (skimming the documentation) to configure Pomerium to operate behind a reverse proxy.

My specific scenario is that I'm running in AWS, and I'm already paying the hourly costs for an Application Load Balancer for other reasons, so I'd like to use the ALB to terminate TLS for Pomerium as well because there's no incremental cost aside from actual bandwidth use. ALBs (and also NLBs) support free zero-touch-renewal certificates issued by Amazon's CA, which notably can be wildcard certificates—a particularly useful thing for a use case like Pomerium.

There are two scenarios that would be nice to support:

1.  Pomerium behind an HTTPS->HTTP reverse proxy. (ELB,ALB)

    In this mode of operation, Pomerium would trust `X-Forwarded-For`, `X-Forwarded-Proto`, and `X-Forwarded-Port` headers for logging and other purposes. Out of an abundance of caution, in this mode Pomerium should probably refuse or force a redirect to https if `X-Forwarded-Proto` is `http`.

2.  Pomerium behind protocol-oblivious TLS->TCP. (NLB)

    In this mode of operation, Pomerium would not receive any special headers. It must assume (for url construction, etc...) that from the browser's perspective the browser has made an HTTPS connection.

desimone:
I know some people are currently using pomerium with AWS and ELB/ALB/NLB. 

However, in the scenarios you provided, pomerium currently only supports secured communication (typically with self-signed certs/ca) between load-balancer and pomerium and its services. Those certs not only secure communication between ingress and pomerium, but also between pomerium's services over sGRPC. 

 So you'd have something like (forgive the horrible diagram)....

<img width="1090" alt="Screen Shot 2019-03-15 at 8 23 31 PM" src="https://user-images.githubusercontent.com/1544881/54470144-3b5e9e80-4760-11e9-93a0-1ce851ab4b9d.png">

Would using self-signed certs/ca behind ALB/NLB be workable for your usecase? 


edit: I've been meaning to revisit AWS specific documentation and examples with ALB/NLB auto-certificate integration. Hopefully I'll have time to put something together this weekend.

desimone:
[...]

PR #324, opened 2019-09-26
proxy: support external access control request handling (forward auth)

desimone:
## Flow

[diagram](https://mermaidjs.github.io/mermaid-live-editor/#/view/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgQ2xpZW50IGFzIEJyb3dzZXJcbiAgICBwYXJ0aWNpcGFudCBJZFAgYXMgSWRlbnRpdHkgUHJvdmlkZXJcbiAgICBDbGllbnQtPj5JbmdyZXM6IEdFVCAvYXBwXG4gICAgSW5ncmVzLT4-UG9tZXJpdW06IC92ZXJpZnkvYXBwXG4gICAgUG9tZXJpdW0tPj5Qb21lcml1bTogQXV0aGVudGljYXRlZD9cbiAgICBQb21lcml1bS0-PiBQb21lcml1bTogTm8hXG4gICAgUG9tZXJpdW0tPj5Qb21lcml1bSBBdXRoTjogIFxuICAgIFBvbWVyaXVtIEF1dGhOLT4-Q2xpZW50OiBIVFRQIDMwMSBzaWduIGluIGNhbGxiYWNrIHVybFxuICAgIENsaWVudC0-PklkUDogIFxuICAgIElkUC0-PkNsaWVudDogIFxuICAgIENsaWVudC0-PlBvbWVyaXVtIEF1dGhOOiBIVFRQIDMwMTogT2F1dGgyIGNhbGxiYWNrIGVuZHBvaW50XG4gICAgUG9tZXJpdW0gQXV0aE4tPj4gQ2xpZW50OiBTYXZlIHNlc3Npb25cbiAgICBDbGllbnQtPj4gSW5ncmVzOiBIVFRQIDMwMSBhcHBcbiAgICBJbmdyZXMtPj4gUG9tZXJpdW06IC92ZXJpZnkvYXBwXG4gICAgUG9tZXJpdW0tPj5Qb21lcml1bTogQXV0aGVudGljYXRlZD9cbiAgICBQb21lcml1bS0-PiBQb21lcml1bTogWWVzIVxuICAgIFBvbWVyaXVtLT4-IFBvbWVyaXVtIEF1dGhaOiBBdXRob3JpemVkP1xuICAgIFBvbWVyaXVtIEF1dGhaLT4-IFBvbWVyaXVtOlllcz9cbiAgICBQb21lcml1bS0-PiBJbmdyZXM6IEhUVFAgMjAwXG4gICAgSW5ncmVzLT4-IGFwcDogIFxuICAgIGFwcC0-PiBJbmdyZXM6ICBcbiAgICBJbmdyZXMtPj4gQ2xpZW50OiAgT0shXG4iLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9fQ)

## In action

Not that exciting, but I guess that's the point. Subsequent calls to any other routes will be transparent as well. 

https://www.youtube.com/watch?v=WXE79juX7zk&feature=youtu.be

## Example config

Using [Traefik forwardauth](https://docs.traefik.io/v2.0/middlewares/forwardauth/)

```yaml
version: "3"

services:
  reverse-proxy:
    # The official v2.0 Traefik docker image
    image: traefik:v2.0
    # Enables the web UI and tells Traefik to listen to docker
    command: --api.insecure=true --providers.docker
    ports:
      # The HTTP port
      - "80:80"
      # The Web UI (enabled by --api.insecure=true)
      - "8080:8080"
    volumes:
      # So that Traefik can listen to the Docker events
      - /var/run/docker.sock:/var/run/docker.sock
  whoami:
    # A container that exposes an API to show its IP address
    image: containous/whoami
    labels:
      - "traefik.http.routers.whoami.rule=Host(`whoami.corp.beyondperimeter.com`)"
      # Create a middleware named `foo-add-prefix`
      - "traefik.http.middlewares.test-auth.forwardauth.address=https://httpbin.corp.beyondperimeter.com/.pomerium/verify/http://whoami.corp.beyondperimeter.com"
      - "traefik.http.routers.whoami.middlewares=test-auth@docker"
  echo:
    # image: kennethreitz/httpbin
    image: mendhak/http-https-echo
    labels:
      - "traefik.http.routers.httpbin.rule=Host(`whoami.corp.beyondperimeter.com`)"

```

And using [nginx-ingress](https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/annotations.md#external-authentication) . 
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: kuard
  annotations:
    kubernetes.io/ingress.class: "nginx"
    certmanager.k8s.io/issuer: "letsencrypt-prod"
    nginx.ingress.kubernetes.io/auth-url: https://httpbin.corp.beyondperimeter.com:8443/.pomerium/verify/kuard.k8s.corp.beyondperimeter.com?no_redirect=true
    nginx.ingress.kubernetes.io/auth-signin: https://httpbin.corp.beyondperimeter.com:8443/.pomerium/verify/kuard.k8s.corp.beyondperimeter.com

spec:
  tls:
    - hosts:
        - kuard.k8s.corp.beyondperimeter.com
      secretName: quickstart-example-tls
  rules:
    - host: kuard.k8s.corp.beyondperimeter.com
      http:
        paths:
          - path: /
            backend:
              serviceName: kuard
              servicePort: 80

```
**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review

### DIFF
--- a/proxy/handlers.go
+++ b/proxy/handlers.go
@@ -19,8 +19,11 @@ import (
 // registerHelperHandlers returns the proxy service's ServeMux
 func (p *Proxy) registerHelperHandlers(r *mux.Router) *mux.Router {
 	h := r.PathPrefix(dashboardURL).Subrouter()
+	// 1. Retrieve the user session and add it to the request context
 	h.Use(sessions.RetrieveSession(p.sessionStore))
+	// 2. AuthN - Verify the user is authenticated. Set email, group, & id headers
 	h.Use(p.AuthenticateSession)
+	// 3. Enforce CSRF protections for any non-idempotent http method
 	h.Use(csrf.Protect(
 		p.cookieSecret,
 		csrf.Path("/"),
@@ -32,6 +35,7 @@ func (p *Proxy) registerHelperHandlers(r *mux.Router) *mux.Router {
 	h.HandleFunc("/impersonate", p.Impersonate).Methods(http.MethodPost)
 	h.HandleFunc("/sign_out", p.SignOut).Methods(http.MethodGet, http.MethodPost)
 	h.HandleFunc("/refresh", p.ForceRefresh).Methods(http.MethodPost)
+	h.HandleFunc("/verify/{hostname}", p.Verify).Methods(http.MethodGet)
 	return r
 }
 
@@ -146,4 +150,51 @@ func (p *Proxy) Impersonate(w http.ResponseWriter, r *http.Request) {
 	}
 
 	http.Redirect(w, r, dashboardURL, http.StatusFound)
+}
+
+// Verify checks a user's credentials for an arbitrary host. If the user is
+// properly authenticated and is authorized to access the supplied host,
+// a 200 http status code is returned. If the user is not authenticated, they
+// will be redirected to the authenticate service to sign in with their identity
+// provider. If the user is unauthorized, they will be given a 403 http status.
+func (p *Proxy) Verify(w http.ResponseWriter, r *http.Request) {
+	// retrieve the target destination hostname from the url path
+	hostname, ok := mux.Vars(r)["hostname"]
+	if !ok {
+		httputil.ErrorResponse(w, r, httputil.Error("no hostname set", http.StatusBadRequest, nil))
+		return
+	}
+	// attempt to retrieve the user session from the request context, validity
+	// of the identity session is asserted by the middleware chain
+	s, err := sessions.FromContext(r.Context())
+	if err != nil {
+		httputil.ErrorResponse(w, r, err)
+		return
+	}
+	// query the authorization service to see if the session's user has
+	// the appropriate authorization to access the given hostname
+	authorized, err := p.AuthorizeClient.Authorize(r.Context(), hostname, s)
+	if err != nil {
+		httputil.ErrorResponse(w, r, err)
+		return
+	} else if !authorized {
+		errMsg := fmt.Sprintf("%s is not authorized for this route", s.RequestEmail())
+		httputil.ErrorResponse(w, r, httputil.Error(errMsg, http.StatusForbidden, nil))
+		return
+	}
+	// check the queryparams to see if this check immediately followed
+	// authentication. If so, redirect back to the originally requested hostname.
+	if isCallback := r.URL.Query().Get(disableCallback); isCallback == "true" {
+		http.Redirect(w, r, "http://"+hostname, http.StatusFound)
+		return
+	}
+
+	// User is authenticated and authorized for the given host.
+	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+	w.Header().Set(HeaderUserID, s.User)
+	w.Header().Set(HeaderEmail, s.RequestEmail())
+	w.Header().Set(HeaderGroups, s.RequestGroups())
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprintf(w, "%s is authorized for %s.", s.Email, hostname)
 }
--- a/proxy/handlers_test.go
+++ b/proxy/handlers_test.go
@@ -11,6 +11,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/gorilla/mux"
+
 	"github.com/pomerium/pomerium/internal/config"
 	"github.com/pomerium/pomerium/internal/cryptutil"
 	"github.com/pomerium/pomerium/internal/sessions"
@@ -62,7 +64,7 @@ func TestProxy_authenticate(t *testing.T) {
 	req := httptest.NewRequest(http.MethodGet, "/oauth-start", nil)
 
 	rr := httptest.NewRecorder()
-	proxy.authenticate(rr, req)
+	proxy.reqNeedsAuthentication(rr, req)
 	// expect oauth redirect
 	if status := rr.Code; status != http.StatusFound {
 		t.Errorf("handler returned wrong status code: got %v want %v", status, http.StatusFound)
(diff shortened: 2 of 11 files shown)


lukasmrtvy:
What about this example? Pomerium as auth service is hidden behind Traefik. ( Thats why I wanted to have Pomerium with optional TLS )
```
version: "3"

services:
  reverse-proxy:
    image: traefik:v1.7-alpine
    command: --entryPoints="Name:http Address::80 Redirect.EntryPoint:https" --entryPoints='Name:https Address::443 TLS' 
    ports:
      - "80:80"
      - "443:8080"
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
  pomerium:
    image: pomerium/pomerium
    labels:
      - "traefik.frontend.rule=Host:auth.example.com"
      - "traefik.port=8080" 
      - "traefik.protocol=http"
  whoami:
    image: containous/whoami
    labels:
      - "traefik.rule=Host:whoami.example.com"
      - "traefik.port=80" 
      - "traefik.protocol=http"
      - "traefik.frontend.auth.forward=https://auth.example.com"
```

desimone:
@lukasmrtvy -- Yes, that should work just fine but would require the TLS change ( #64 ) which I'm also taking another look at. In that case, Traefik would be responsible for handling all routing and certificates.

proxy/handlers_test.go
```diff
@@ -75,115 +77,6 @@ func TestProxy_authenticate(t *testing.T) {
 	}
 }
 
-// func TestProxy_PomeriumHandler(t *testing.T) {
```
desimone: Testing moved to the middleware packages.

cmd/pomerium/main.go
```diff
@@ -9,7 +9,7 @@ import (
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/gorilla/mux"
-	"github.com/spf13/viper"
+	"github.com/pomerium/viper"
```
travisgroth: Not that viper is much of a moving target, but can we avoid forking it?

proxy/handlers.go
```diff
@@ -147,3 +148,47 @@ func (p *Proxy) Impersonate(w http.ResponseWriter, r *http.Request) {
...
+func (p *Proxy) Verify(w http.ResponseWriter, r *http.Request) {
+	// retrieve the target destination hostname from the url path
+	hostname, ok := mux.Vars(r)["hostname"]
+	if !ok {
+		httputil.ErrorResponse(w, r, httputil.Error("no hostname set", http.StatusBadRequest, nil))
+		return
+	}
+	// attempt to retrieve the user session from the request context, validity
+	// of the identity session is asserted by the middleware chain
+	s, err := sessions.FromContext(r.Context())
```
travisgroth: Implementation question - How are we getting access to the cookie on subsequent auth checks?  Does the third party proxy pass the cookies as part of the subrequest?

This depends on the proxy having an endpoint on the same domain that the requests hit the third party proxy with, correct?

proxy/handlers.go
```diff
@@ -147,3 +148,47 @@ func (p *Proxy) Impersonate(w http.ResponseWriter, r *http.Request) {
 
 	http.Redirect(w, r, dashboardURL, http.StatusFound)
 }
+
+// Verify checks a user's credentials for an arbitrary host. If the user is
+// properly authenticated and is authorized to access the supplied host,
+// a 200 http status code is returned. If the user is not authenticated, they
+// will be redirected to the authenticate service to sign in with their identity
+// provider. If the user is unauthorized, they will be given a 403 http status.
+func (p *Proxy) Verify(w http.ResponseWriter, r *http.Request) {
```
travisgroth: It seems odd to run the proxy just to bounce auth checks off of.  It feels like something Authorize should do...but that's obviously a whole extra set of problems.

In a typical config where all pomerium services share the policy, the proxy service would function if you simply pointed your `from` addresses at it, correct?  Perhaps we should disable proxying in this mode?

cmd/pomerium/main.go
desimone: Probably not, unless we are okay with the potential race conditions. I can try to make these changes and request changes upstream but I'm not sure of the appetite.

Some context.
- https://github.com/spf13/viper/issues/19
- https://github.com/spf13/viper/issues/174
- https://github.com/googleforgames/open-match/issues/304
- https://github.com/spf13/viper/issues/629
- https://github.com/spf13/viper/issues/378
- https://github.com/spf13/viper/issues/353

proxy/handlers.go
desimone: Great questions!

> How are we getting access to the cookie on subsequent auth checks?

Every request is checked for access and authorization. Every request will check and validate the session cookie. AFAI understand, this means every request to an endpoint results in another request to pomerium for validation. 

> > To perform authentication, NGINX makes an HTTP subrequest to an external server where the subrequest is verified.
https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication/

> Does the third party proxy pass the cookies as part of the subrequest?

Yes, it will if pomerium sits on the same subdomain chain as the downstream application. Which paradoxically is why ...

> This depends on the proxy having an endpoint on the same domain that the requests hit the third party proxy with, correct?

For security reasons, I'd actually recommend _against_  having pomerium on the same domain because there's no explicit way to strip the session from downstream requests like we can do if we are the proxying entity.

cmd/pomerium/main.go
desimone: Probably not, unless we are okay with the potential race conditions. I can try to make these changes and request changes upstream but I'm not sure of the appetite.

Some context.
- https://github.com/spf13/viper/issues/19
- https://github.com/spf13/viper/issues/174
- https://github.com/googleforgames/open-match/issues/304
- https://github.com/spf13/viper/issues/629
- https://github.com/spf13/viper/issues/378
- https://github.com/spf13/viper/issues/353

proxy/handlers.go
desimone: This is where #291 would potential come in. We could skip setting up the reverse proxy if no `to` destination was defined in the policy (but still allow all the `.pomerium/` routes for troubleshooting and verification.

Plus, this would allow for a mixed model of some routes using pomerium directly for proxying, while other not, if that was a desired configuration setup. 

What do you think about that?

cmd/pomerium/main.go
desimone: I've undone these changes both here and in #328 based on our conversation regarding moving viper into the options struct to be instance based.

proxy/handlers.go
travisgroth: Hmm.  In this mode, would any policy be defined at all on the proxy?  It strikes me that it just serves as an integration point to Authorize and Authenticate.  The proxy wouldn't care about the `from` at all... except maybe an unprotected endpoint?

From a deployment perspective where you might be sharing the config file, it would probably be nice to be able to set this mode with a single flag.  That also makes the decision explicit.  Thoughts?

desimone: I've added a flag that sits on the main option  (vs policy) branch. I went back and forth as to whether this should be under policy or config, but it feels easier to implement in `config.

travisgroth: I think main config makes sense.

tlvenn:
This is awesome ! Any chance to have an example with ambassador as well ? Thanks in advance.

desimone:
@tlvenn not at the moment but we'd love the contribution. However, [looking at the docs](https://www.getambassador.io/reference/services/auth-service/), the approach looks extremely similar to nginx and traefik so it should work out the box.

lyndon160:
Hi @desimone, thanks for providing this. Is there a complete docker-compose example of how to use pomerium was a forward auth middleware for ingresses such as Traefik?

The example you gave above appears to be missing the pomerium service and config.

There seems to be some example https://github.com/pomerium/pomerium/tree/master/examples/traefik but it does not appear to work as expected.

Ideally, Traefik would be responsible for managing TLS comms and the pomerium would be just over HTTP. Also, the IAP would only be accessible internally as middleware, not directly via the browser. I've seen merged pull requests for both of these but nothing in the docs to reflect it, making it difficult to use.

Effectively I'm looking to use pomerium in the same way that https://github.com/thomseddon/traefik-forward-auth supports.

pom

cuonglm:
> There seems to be some example https://github.com/pomerium/pomerium/tree/master/examples/traefik but it does not appear to work as expected.

Can you elaborate what's exactly "not appear to work as expected"?

lyndon160:
Hi @cuonglm, thanks for looking into this. It could be that I have my configuration wrong, please see:

My configuration is almost identical to the one in the example except that I have different IDP creds, have changed the policy domain to my domain, and I've changed the pomerium.forwardauth.address to point to the local pomerium (http://pomerium), which is typically how forward auth and IAPs are used with Traefik.

The example config has this in: `- "traefik.http.middlewares.pomerium.forwardauth.address=https://a6acdabcde358bd08f3537f4de7df7eb.m.pipedream.net"`

This service appears to just return a 200, presumably allowing all messages through when used as a forward auth, rather than using pomerium? Why is this address used in the example?

I'm not sure what this should be replaced with. I would've thought this should point to the local pomerium instance, not some web endpoint. Configuring it to point to the local instance causes a 403 through the browser, and through postman with a valid token redirects me to my IDP login page.

I have an extract of the failure logs here: https://pastebin.com/r1LTyCt8

On top of this, it is not clear to me why the pomerium instance is exposed via traefik:
      `- "traefik.http.routers.pomerium.rule=Host(`authenticate.localhost.pomerium.io`)"`

travisgroth:
>The example config has this in: - "traefik.http.middlewares.pomerium.forwardauth.address=https://a6acdabcde358bd08f3537f4de7df7eb.m.pipedream.net"

Hey @lyndon160 this is an error introduced in a recent repo reorganization.  https://github.com/pomerium/pomerium/pull/1128 should fix it.  For the v0.9.4 version of Pomerium it should be:

```yaml
- "traefik.http.middlewares.pomerium.forwardauth.address=http://pomerium/?uri=https://httpbin.localhost.pomerium.io"
```

This will work without the `uri` parameter in the `v0.9.5` release we're doing shortly.

lyndon160:
Hey @travisgroth thanks for the help, for the browser flow, this seems to be working as expected now. However, when I use postman with the same client and password grant type, pomerium errors saying:

`pomerium_1  | 5:20PM ERR proxy: could not locate session from context error="internal/sessions: session is not found"`
` INF authorize check allow=false check-request-id=ef54f1b1-057c-4219-acff-05c50905cc3c host=httpbin.localhost.pomerium.io method=GET path= query= request-id=417bdf7e-2964-47fb-8c3c-5f5d2088500d service=authorize status=401`
`pomerium_1  | 5:20PM INF authenticate: session load error error="Bad Request: internal/sessions: session is not found"`
`pomerium_1  | 5:20PM ERR proxy: could not locate session from context error="internal/sessions: session is not found"`

The result is that postman gets a login page in return, which is not the expected behaviour.

lyndon160:
@travisgroth thanks for the help. My issue was answered in https://github.com/pomerium/pomerium/issues/1130.

Unfortunately Pomerium does not appear to suit my use case.

PR #328, opened 2019-09-29
all: support insecure transport

desimone:
all: support insecure mode

- pomerium/authenticate: add cookie secure setting
- internal/config: transport security validation moved to options
- internal/config: certificate struct hydrated 
- internal/grpcutil: add grpc server mirroring http one
- internal/grpcutil: move grpc middleware
- cmd/pomerium: use run wrapper around main to pass back errors
- cmd/pomerium: add waitgroup (block on) all servers http/grpc

See also:
- https://github.com/pomerium/pomerium/issues/64
- https://github.com/pomerium/pomerium/pull/324

**Checklist**:
- [x] updated docs
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review


internal/config/options.go
```diff
@@ -210,19 +225,33 @@ func OptionsFromViper(configFile string) (*Options, error) {
...
-	// shared key must be set for all modes other than "all"
-	if o.SharedKey == "" {
-		if o.Services == "all" {
+	if IsAll(o.Services) {
+		// mutual auth between services on the same host can be generated at runtime
+		if o.SharedKey == "" {
 			o.SharedKey = cryptutil.NewBase64Key()
-		} else {
-			return errors.New("shared-key cannot be empty")
+		}
```
desimone: Is it reasonable to be setting/hyrdating these options here in Validate or should it be done in the downstream packages?

internal/config/options.go
```diff
@@ -254,6 +283,23 @@ func (o *Options) Validate() error {
...
+			"POMERIUM WITHOUT ENCRYPTION! In this configuration, it is up to you " +
+			"to secure your instance and handle TLS termination. If exposing HTTP " +
+			"directly, you must also set `cookie_secure` to false")
+	} else if o.Cert != "" && o.Key != "" {
+		o.TLSCertificate, err = cryptutil.CertifcateFromBase64(o.Cert, o.Key)
+	} else if o.CertFile != "" && o.KeyFile != "" {
+		o.TLSCertificate, err = cryptutil.CertificateFromFile(o.CertFile, o.KeyFile)
+	}
+	// else {
+	// 	err = errors.New("no certificates supplied nor was insecure mode set")
```
desimone: Should we fail hard if neither certificates nor use insecure is set? Part of me thinks yes. I was considering falling back on insecure but that feels wrong.

travisgroth: I think that's the right call.  Explicit security relaxation is pretty common (docker and curl come to mind).

internal/config/options.go
travisgroth: I _think_ doing it centrally makes sense since more than one package may use an option.  We wouldn't want different hydration/validation/default semantics by accident.

internal/config/options.go
```diff
@@ -388,7 +437,7 @@ func (o *Options) Checksum() string {
 func ParseOptions(configFile string) (*Options, error) {
 	o, err := OptionsFromViper(configFile)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("internal/config: options from viper %v", err)
```
travisgroth: %w for new errorf calls?

internal/httputil/server.go
```diff
@@ -1,50 +1,42 @@
...
+	"time"
 
-	"github.com/pomerium/pomerium/internal/cryptutil"
 	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/internal/urlutil"
 )
 
-// NewTLSServer creates a new TLS server given a set of options, handlers, and
+// NewServer creates a new HTTP(s) server given a set of options, handlers, and
 // optionally a set of gRPC endpoints as well.
```
travisgroth: Is this still accurate with this call signature?

internal/httputil/server.go
```diff
@@ -98,15 +90,35 @@ func newDefaultTLSConfig(cert *tls.Certificate) *tls.Config {
...
+func RedirectHandler() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ct := r.Header.Get("Content-Type")
-		if r.ProtoMajor == 2 && strings.Contains(ct, "application/grpc") {
-			rpcServer.ServeHTTP(w, r)
-		} else {
-			other.ServeHTTP(w, r)
-		}
+		w.Header().Set("Connection", "close")
+		url := fmt.Sprintf("https://%s%s", urlutil.StripPort(r.Host), r.URL.String())
```
travisgroth: Will this handle alternate ports correctly?

internal/config/options.go
```diff
@@ -164,13 +180,12 @@ var defaultOptions = Options{
 		"Strict-Transport-Security": "max-age=31536000; includeSubDomains; preload",
 	},
 	Addr:                    ":443",
-	CertFile:                filepath.Join(fileutil.Getwd(), "cert.pem"),
```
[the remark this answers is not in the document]
desimone: Not strongly. I'd rather keep backward compatibility. 

However, keeping iIt would change how options fail if `insecure_server` and no certs are provided in that it will always fail on trying to read a (potentially) non existent cert file.

e.g. if you run pomerium with no certs specified, no `cert.pem` in the root dir, and no insecure flag, will error with a fopen error on those non existent cert files. 

Which I'm okay with, but could be a bit confusing.

internal/httputil/server.go
desimone: Good catch.

internal/config/options.go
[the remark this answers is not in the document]
travisgroth: Hm.  Yeah neither is ideal here.  Without considering prior defaults this is probably the right behavior - we'll just have to make sure it is addressed in usage elsewhere.  Add as a breaking change?

internal/config/options.go
desimone: 🕺

internal/config/options.go
```diff
@@ -192,10 +207,15 @@ func NewDefaultOptions() *Options {
...
 // call to validate.
-func NewMinimalOptions(authenticateURL, authorizeURL string) (*Options, error) {
+func NewMinimalOptions(authenticateURL, authorizeURL string, cert *tls.Certificate) (*Options, error) {
 	o := NewDefaultOptions()
 	o.AuthenticateURLString = authenticateURL
 	o.AuthorizeURLString = authorizeURL
+	if cert != nil {
+		o.TLSCertificate = cert
+	} else {
+		o.InsecureServer = true
```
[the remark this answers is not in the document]
desimone: I'll update the docs. Right now this is only being used as a shorthand by unit tests which feels a bit mucky

internal/httputil/server.go
desimone: No, it will only redirect to the default https.

More concerning, in adding some unit tests to this function... I'm not sure this has ever worked as intended.

travisgroth: :sadtrombone:

internal/config/options.go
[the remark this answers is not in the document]
travisgroth: I thought we were going to error rather than default to insecure.  I think we should make insecure be an explicit operator decision.

internal/config/options.go
travisgroth: 💯

internal/config/options.go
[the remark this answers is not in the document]
desimone: I've changed the public interface to configuration options a bit to make this a bit more focused and without setting any unintended settings. Let me know what you think.

internal/config/options.go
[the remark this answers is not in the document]
desimone: Agree. I've tweaked the public exported methods a bit to make sure this is explicit.

internal/config/options.go
```diff
@@ -188,148 +187,68 @@ var defaultOptions = Options{
...
 
-// NewOptions creates a new Options struct with only viper initialized
-func NewOptions() *Options {
-	o := Options{}
-	o.viper = viper.New()
-	return &o
+// NewDefaultOptions returns a new set of default options. It's the caller's
+// responsibility to do a follow up Validate call.
+func NewDefaultOptions() Options {
+	return defaultOptions
```
travisgroth: This should be a reference, shouldn't it?

internal/config/options.go
```diff
@@ -188,148 +187,68 @@ var defaultOptions = Options{
 	GRPCAddr:                ":443",
 	GRPCClientTimeout:       10 * time.Second, // Try to withstand transient service failures for a single request
 	GRPCClientDNSRoundRobin: true,
+	viper:                   viper.New(),
```
travisgroth: Since viper.New() returns pointer, don't we wind up sharing a single viper instance everywhere if we start with a copy of this struct?  I avoided this pattern because of that assumption...am I forgetting how that works?

internal/config/options.go
desimone: Updated.

internal/config/options.go
desimone: Yes, you totally had the right to begin with. I tried to remove the NewXXX() but ended up going back to it. My bad.

internal/config/options.go
```diff
@@ -25,8 +24,7 @@ import (
 // DisableHeaderKey is the key used to check whether to disable setting header
 const DisableHeaderKey = "disable"
 
-// Options are the global environmental flags used to set up pomerium's services.  Use NewXXXOptions() methods
-// for a safely initialized data structure.
```
travisgroth: Should carry this warning over.

internal/config/options.go
```diff
@@ -164,47 +179,50 @@ var defaultOptions = Options{
...
 	WriteTimeout:            0, // support streaming by default
 	IdleTimeout:             5 * time.Minute,
 	RefreshCooldown:         5 * time.Minute,
+	GRPCAddr:                ":443",
 	GRPCClientTimeout:       10 * time.Second, // Try to withstand transient service failures for a single request
 	GRPCClientDNSRoundRobin: true,
 }
 
-// NewOptions creates a new Options struct with only viper initialized
-func NewOptions() *Options {
```
travisgroth: I don't think the empty Options is used too many places but I feel like we should offer it as a baseline so you can always get a safe Options struct.

internal/config/options.go
travisgroth: I spent a few minutes trying to remember I didn't use exactly the same pattern  😆

internal/config/options.go
desimone: By empty options you mean this method right?

travisgroth: Sorry - yes just a plain NewOptions().  I know there were one or two in the tests.

internal/config/options.go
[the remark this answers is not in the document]
travisgroth: Ah I see missed how this was being used.  Still best not to leave a landmine though.  👍

internal/config/options.go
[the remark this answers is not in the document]
travisgroth: 👍

internal/config/options.go
desimone: 👍

internal/config/options.go
desimone: Agree. Done.

internal/httputil/server.go
desimone: Cool if we punt on this for now? It works as (originally) intended but not for non `:443` redirects.

travisgroth: Yeah no sense adding to the scope here.

internal/config/options.go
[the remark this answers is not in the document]
desimone: You have the right of it.`config.NewXXX` should not silently default to insecure mode.

PR #332, opened 2019-10-01
internal/config: Switch to using struct scoped viper instance

### LINKED ISSUE #213: internal/config: refactor viper instance scope
After seeing all the viper.Reset() calls in the tests, I'm thinking we should make an instance of viper instead of using the default singleton.  This would keep every viper instance scoped to an Options struct - which would reduce the chances of errors and allow Parallel() calls in many cases that we avoid now.  

I think stashing the instance of viper in the struct or passing it to the parseXXXX() methods would be all that is needed.

_Originally posted by @travisgroth in https://github.com/pomerium/pomerium/pull/202_

travisgroth:
Closes #213 

**Checklist**:
- [x] unit tests added
- [x] updated CHANGELOG.md
- [x] ready for review

### DIFF
--- a/internal/config/options.go
+++ b/internal/config/options.go
@@ -141,6 +141,9 @@ type Options struct {
 	// GRPC Service Settings
 	GRPCClientTimeout       time.Duration `mapstructure:"grpc_client_timeout"`
 	GRPCClientDNSRoundRobin bool          `mapstructure:"grpc_client_dns_roundrobin"`
+
+	// Scoped viper instance
+	viper *viper.Viper
 }
 
 var defaultOptions = Options{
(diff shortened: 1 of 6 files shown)


internal/config/options.go
```diff
@@ -188,17 +203,17 @@ func NewOptions(authenticateURL, authorizeURL string) (*Options, error) {
...
-		viper.SetConfigFile(configFile)
-		if err := viper.ReadInConfig(); err != nil {
+		o.viper.SetConfigFile(configFile)
+		if err := o.viper.ReadInConfig(); err != nil {
 			return nil, fmt.Errorf("internal/config: failed to read config: %s", err)
 		}
 	}
 
-	if err := viper.Unmarshal(&o); err != nil {
+	if err := o.viper.Unmarshal(&o); err != nil {
```
travisgroth: This is a bit weird but works in practice.  An alternative would be a wrapper struct but that doesn't feel worth it.  We could also make a new instance and then place it into the struct after Unmarshall().

internal/config/options.go
```diff
@@ -171,11 +174,23 @@ var defaultOptions = Options{
 	GRPCClientDNSRoundRobin: true,
 }
 
+func NewEmptyOptions() *Options {
```
travisgroth: I actually think this should be what `NewOptions` does and the functionality of `NewOptions` could be done with something like `NewMinimalOptions`.  Thoughts?

internal/config/options.go
```diff
@@ -171,41 +174,55 @@ var defaultOptions = Options{
...
+	o := Options{}
+	o.viper = viper.New()
+	return &o
+}
+
+// NewDefaultOptions returns an Options struct with defaults set and viper initialized
+func NewDefaultOptions() *Options {
+	o := defaultOptions
+	o.viper = viper.New()
+	return &o
```
travisgroth: Return pointer for consistency?

internal/config/options.go
desimone: Agree!

internal/config/options.go
desimone: Agree

internal/config/options.go
desimone: That's pretty cool this works. 
That said, for clarity and readability, (and unintended state from the original?) I think it might be worth a bit of verbosity to create a new uninitialized struct like you are saying. But I don't feel strongly.

travisgroth: bindEnvs() gets in the way a bit but here's an attempt https://github.com/pomerium/pomerium/pull/332/commits/60e64f971c5e9ea10cadc677dd9936303fca6208

I don't hate it :)

travisgroth:
Lint fixed - surprised that was there.  

Offhand I _think_ this should prevent any race conditions from our tests.  In practice I'm wondering if we've got some low probability ones stemming from UpdateOptions() in the services.  I believe Options itself is safe, though.

internal/config/options.go
desimone: It's fine. 👍 


🤣

internal/config/options.go
```diff
@@ -171,41 +175,63 @@ var defaultOptions = Options{
...
 	// start a copy of the default options
-	o := defaultOptions
+	o := NewDefaultOptions()
+	// New viper instance to save into Options later
+	v := viper.New()
 	// Load up config
-	o.bindEnvs()
+	err := bindEnvs(o, v)
+	if err != nil {
+		return nil, fmt.Errorf("failed to bind options to env vars: %w", err)
```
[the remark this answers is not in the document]
travisgroth: Had to be done!

PR #351, opened 2019-10-05
internal/sessions: separate query params, auth-header, and cookie stores

desimone:
## Summary
These changes standardize how session loading is done for session cookie, auth bearer token, and query params. Each implements a common interface that returns a user's session state. 

### Details
- Bearer token previously combined with session cookie.
- rearranged cookie-store to put exported methods above unexported
- added header store that implements session loader interface
- added query param store that implements session loader interface

## Related issues

- #281 (possibly solving a large portion of it)

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/sessions/cookie_store.go
+++ b/internal/sessions/cookie_store.go
@@ -10,47 +10,39 @@ import (
 	"github.com/pomerium/pomerium/internal/cryptutil"
 )
 
-// ChunkedCanaryByte is the byte value used as a canary prefix to distinguish if
-// the cookie is multi-part or not. This constant *should not* be valid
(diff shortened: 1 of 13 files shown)


internal/sessions/cookie_store.go
```diff
@@ -10,47 +10,39 @@ import (
 	"github.com/pomerium/pomerium/internal/cryptutil"
 )
 
-// ChunkedCanaryByte is the byte value used as a canary prefix to distinguish if
```
desimone: Changes in `cookie_store` are almost all rearranging of methods / reindentation. :rage2:

PR #368, opened 2019-10-14
proxy: strip added params from verify callback

### LINKED ISSUE #366: v0.4.0: extra query parameters pomerium-auth-callback breaks protected apps
## What happened?

With v0.4.0, just after authentication is triggered, pomerium adds a query parameter `?pomerium-auth-callback=true` to the url that is transmitted to the protected apps url: `https://my-app.corp.pomerium.io/?pomerium-auth-callback=true`.

The problem with that is twofold:
- the users are confused by the resulting url
- the application loose control of the location uri (in particular SPAs are used to be in full control of it), in my particular case, when the user reaches `/`, he is then redirected to a default page, but with the query parameter here, the SPA does not do the redirection.

That once I am connected through pomerium, the url is exactly as typed originally by the user.

## How'd it happen?

1. Entered a protected app url in the browser
2. Got redirected to authenticate service then to google idp for auth
3. Got redirected to the original url + the extra query parameters

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint): 0.4.0

desimone:
## Summary
This PR changes the way forward-authentication works in several respects: 

1) Forward-authentication handlers are not registered as part of the dashboard router registration process (this had the weird side effect of making it so that you could do forward-auth validation against _any_ proxied route at `${route}/.pomerium/verify`
1) Verification will now occur on at the _root_ of the forward authentication host handler. I think this is easier to read / work with. 
1) Verification and signing endpoints are now two distinct endpoints (more along the lines of what Nginx asks for). One handler that simply verifies (returns `200` or `401`) the request, and another which again verifies the request, but will return a `302` redirect to complete the authentication process. In practice, `sign-in` is a superset of verify.
    - The sign in and verify endpoint is `/`
    - The verify only endpoint is `/verify`
1. Additional query params (which are used to tell the proxy service when the authenticate service has completed authentication and can be redirected back to it's original domain) are now stripped from the redirect (Fixing #366).

This is a breaking change. But I would like to consider cherrypicking it as part of our patch release.  

\ cc @victornoel if you could give this a try that would be a big help. I've pushed a test image on docker hub with the tag `PR366v2`

```bash
docker run pomerium/pomerium:PR366v2 --version                                                                                          
v0.4.0-dirty+0e85b2b
```

## Related issues
Fixes #366 

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated CHANGELOG.md
- [x] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Fixed an issue where requests handled by forward-auth would not be redirected back to the underlying route after successful authentication and authorization. [GH-363]
+- Fixed an issue where requests handled by forward-auth would add an extraneous query-param following sign-in causing issues in some configurations. [GH-366]
 
 ## v0.4.0
 
--- a/docs/docs/reference/reference.md
+++ b/docs/docs/reference/reference.md
@@ -303,7 +303,7 @@ Forward authentication creates an endpoint that can be used with third-party pro
 
 #### NGINX Ingress
 
-Some reverse-proxies, such as nginx split access control flow into two parts: verification and sign-in redirection. Notice the additional the additional `?no_redirect=true` query param in `auth-rul` which tells Pomerium to return a `401` instead of redirecting and starting the sign-in process.
+Some reverse-proxies, such as nginx split access control flow into two parts: verification and sign-in redirection. Notice the additional path `/verify` used for `auth-url` indicating to Pomerium that it should return a `401` instead of redirecting and starting the sign-in process.
 
 ```yaml
 apiVersion: extensions/v1beta1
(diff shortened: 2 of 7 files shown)


lukasmrtvy:
@desimone 

KO:
`traefik.frontend.auth.forward.address=https://fwdauth.example.com/verify?uri=https://whoami.example.com` does not work for me ( unauthorized message is shown, without redirection to idp -> pomerium styled page )

OK:
`traefik.frontend.auth.forward.address=https://fwdauth.example.com?uri=https://whoami.example.com` works for me with proper redirect

desimone:
Cherry-picked `7d7e997e7972f79208f4e4bb9e32551e6d48839b` on

- https://github.com/pomerium/pomerium/tree/latest
- https://github.com/pomerium/pomerium/tree/0-4-0

Issue #128, opened 2019-05-20
internal/sessions : make state consistent across routes and services


desimone:
**Is your feature request related to a problem? Please describe.**

Currently, user state is mirrored between services not shared. Each proxy'd route has its own session cookie as does the authenticate service. 

For simplicity's sake, consider the workflow for an unauthenticated user:

1. User navigates to some proxied endpoint (e.g. `httpbin.corp.example.com`)
1. The proxy service handles the request and checks if the request is authenticated by checking if either the cookie is no longer valid (via a grpc call to authenticate) or if the session cookie does not exist at all. 
1. The proxy service hands off (redirects) the user to the authenticate service (with CSRF token) (authenticate.corp.example.com). 
1. The authenticate service redirects the user to her identity provider's authentication endpoint.
1. The user logs in with her identity provider (google, okta, etc)
1. Assuming a valid login, the identity provider redirects the user back to the authenticate service (`OAuthCallback`). The identity provider returns a user session which is parsed and stored in an encrypted session cookie **only on** `authenticate.corp.example.com`. 
1. The authenticate service initiates a callback to the proxy service (`authenticate.ProxyCallback`) which marshals the encrypted session state (along with a CSRF) and hands off that request and payload back to the proxy service (redirect).
1. The proxy service receives the callback (`proxy.OAuthCallback`), decrypts the payload and stores it into **another** session cookie whose scope is the proxied domain (`httpbin.corp.example.com`). For the rest of that user's session on this proxied route, that cookie will serve as her session no matter what happens on any other pomerium routed endpoint, or on the authenticate service domain itself. Of course, the session is periodically checked for validity. 

(normal user flow continues)

The result of the above is that each route maintains its own state. If a user logs out from `httpbin.corp.example.com` that does *not* automatically log her out of `git.corp.example.com` because her session state is tracked at the route level. 

Though not a security issue, having out of sync session state can create user confusion for a period of however long a session's TTL is set at.

**Describe the solution you'd like**

(Session) state should be shared amongst all pomerium http endpoints. At first glance this should be entirely possible using subdomain-wide cookies which also would remove the need to pass state over redirect calls.

**Additional context**

- https://tools.ietf.org/html/rfc2109
- https://github.com/buzzfeed/sso/issues/141
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Syntax

PR #373, opened 2019-10-31
all: support route scoped sessions

desimone:
## Summary

User state was previously managed as a global (cookie shared across a subdomain) encrypted session. These changes make session route-scoped in the form of a transparent, signed JWT. The authenticate service still maintains sensitive access tokens as encrypted state, but that session state is scoped to the authenticate route's session cookie. The proxy service no longer mutates state and instead validates the route-sessions issues by the authenticate service. 

 To make this a bit more concrete, the new flow approximately looks like the [following](https://mermaidjs.github.io/mermaid-live-editor/#/view/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgVSBhcyBVc2VyXG4gICAgcGFydGljaXBhbnQgUCBhcyBQcm94eVxuICAgIHBhcnRpY2lwYW50IEFOIGFzIEF1dGhlbnRpY2F0ZVxuICAgIHBhcnRpY2lwYW50IEFaIGFzIEF1dGhvcml6ZVxuICAgIHBhcnRpY2lwYW50IEkgYXMgSWRlbnRpdHkgUHJvdmlkZXJcbiAgICBwYXJ0aWNpcGFudCBTIGFzIFNlcnZpY2VcblxuICAgIFUtPj5QOiBkZXNpcmVkX3VybFxuICAgIFAtPj5QOiBSb3V0ZSBzZXNzaW9uPyBOby5cbiAgICBQLT4-VTogMzAyXG4gICAgVS0-PkFOOiAvc2lnbl9pblxuICAgIEFOLT4-QU46IEdsb2JhbCBzZXNzaW9uPyBOby5cbiAgICBBTi0-PlU6IDMwMlxuICAgIFUtPj5JOiBzaWduLWluIHVybCAoc3RhdGV8YXV0aGVudGljYXRlX2NhbGxiYWNrX3VybClcbiAgICBJLT4-VTogMzAyXG4gICAgVS0-PkFOOiBhdXRoZW50aWNhdGVfY2FsbGJhY2tfdXJsXG4gICAgQU4tPj5JOiBleGNoYW5nZSBhY2Nlc3MgdG9rZW5cbiAgICBJLT4-QU46IGV4Y2hhbmdlIGFjY2VzcyB0b2tlblxuICAgIEFOLT4-VTogc2F2ZSBnbG9iYWwgc2Vzc2lvblxuICAgIEFOLT4-VTogMzAyXG4gICAgVS0-PlA6IC9jYWxsYmFjayAoZGVzaXJlZF91cmx8and0KVxuICAgIFAtPj5VOiBzYXZlIHJvdXRlIHNlc3Npb24gKGp3dClcbiAgICBQLT4-VTogMzAyXG4gICAgVS0-PlA6IGRlc2lyZWRfdXJsXG4gICAgUC0-PlA6IFJvdXRlIHNlc3Npb24_IFllcy5cbiAgICBQLT4-UDogU2Vzc2lvbiB2YWxpZD8gWWVzXG4gICAgUC0-PkFaOiBBdXRoWj9cbiAgICBBWi0-PlA6IFllc1xuICAgIFAtPj5TOiBBdXRob3JpemVkIHJlcXVlc3RcbiAgICBTLT4-UDogQXV0aG9yaXplZCByZXF1ZXN0XG4gICAgUC0-PlU6IEF1dGhvcml6ZWQgcmVxdWVzdFxuIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifX0) or [edit](https://mermaidjs.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgVSBhcyBVc2VyXG4gICAgcGFydGljaXBhbnQgUCBhcyBQcm94eVxuICAgIHBhcnRpY2lwYW50IEFOIGFzIEF1dGhlbnRpY2F0ZVxuICAgIHBhcnRpY2lwYW50IEFaIGFzIEF1dGhvcml6ZVxuICAgIHBhcnRpY2lwYW50IEkgYXMgSWRlbnRpdHkgUHJvdmlkZXJcbiAgICBwYXJ0aWNpcGFudCBTIGFzIFNlcnZpY2VcblxuICAgIFUtPj5QOiBkZXNpcmVkX3VybFxuICAgIFAtPj5QOiBSb3V0ZSBzZXNzaW9uPyBOby5cbiAgICBQLT4-VTogMzAyXG4gICAgVS0-PkFOOiAvc2lnbl9pblxuICAgIEFOLT4-QU46IEdsb2JhbCBzZXNzaW9uPyBOby5cbiAgICBBTi0-PlU6IDMwMlxuICAgIFUtPj5JOiBzaWduLWluIHVybCAoc3RhdGV8YXV0aGVudGljYXRlX2NhbGxiYWNrX3VybClcbiAgICBJLT4-VTogMzAyXG4gICAgVS0-PkFOOiBhdXRoZW50aWNhdGVfY2FsbGJhY2tfdXJsXG4gICAgQU4tPj5JOiBleGNoYW5nZSBhY2Nlc3MgdG9rZW5cbiAgICBJLT4-QU46IGV4Y2hhbmdlIGFjY2VzcyB0b2tlblxuICAgIEFOLT4-VTogc2F2ZSBnbG9iYWwgc2Vzc2lvblxuICAgIEFOLT4-VTogMzAyXG4gICAgVS0-PlA6IC9jYWxsYmFjayAoZGVzaXJlZF91cmx8and0KVxuICAgIFAtPj5VOiBzYXZlIHJvdXRlIHNlc3Npb24gKGp3dClcbiAgICBQLT4-VTogMzAyXG4gICAgVS0-PlA6IGRlc2lyZWRfdXJsXG4gICAgUC0-PlA6IFJvdXRlIHNlc3Npb24_IFllcy5cbiAgICBQLT4-UDogU2Vzc2lvbiB2YWxpZD8gWWVzXG4gICAgUC0-PkFaOiBBdXRoWj9cbiAgICBBWi0-PlA6IFllc1xuICAgIFAtPj5TOiBBdXRob3JpemVkIHJlcXVlc3RcbiAgICBTLT4-UDogQXV0aG9yaXplZCByZXF1ZXN0XG4gICAgUC0-PlU6IEF1dGhvcml6ZWQgcmVxdWVzdFxuIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifX0).

mermaid-diagram-20191103150734

### New

- Session state is now route-scoped. Each managed route uses a transparent, signed JSON Web Token (JWT) to assert identity.
- Managed routes no longer need to be under the same subdomain! Access can be delegated to any route, on any domain.
- Programmatic access now also uses JWT tokens. Access tokens are now generated via a standard oauth2 token flow, and credentials can be refreshed for as long as is permitted by the underlying identity provider.
- User dashboard now pulls in additional user context fields (where supported) like the profile picture, first and last name, and so on.

Screen Shot 2019-11-03 at 8 16 38 PM

### Security

- Some identity providers (Okta and Azure) previously used mutable signifiers to set and assert group membership. Group membership for all providers now use globally unique and immutable identifiers when available.

### Changed

- Azure AD identity provider now uses globally unique and immutable `ID` for [group membership](https://docs.microsoft.com/en-us/graph/api/group-get?view=graph-rest-1.0&tabs=http).
- Okta no longer uses tokens to retrieve group membership. Group membership is now fetched using Okta's HTTP API. [Group membership](https://developer.okta.com/docs/reference/api/groups/) is now determined by the globally unique and immutable `ID` field.
- Okta now requires an additional set of credentials to be used to query for group membership set as a [service account](https://www.pomerium.io/docs/reference/reference.html#identity-provider-service-account).
- URLs are no longer validated to be on the same domain-tree as the authenticate service. Managed routes can live on any domain.

### Removed

- Force refresh has been removed from the dashboard.
- Previous programmatic authentication endpoints (`/api/v1/token`) has been removed and is no longer supported.
### Detailed

- authenticate: cookies are now constrained to the authenticate service domain only.
- authenticate: operates in a dual mode configuration and supports two types of session encoders: one to encrypt sensitive global state, and another to create transparent session cookies for managed routes.
- authenticate: session state can be loaded from query params, cookie, or authorization headers.
- authenticate: supports issuing programmatic tokens: a transparent, signed JWT access token, and a encrypted refresh token.
- encoding: create encoder abstraction modeled after standard library and that fulfills the same `Marshal` / `Unmarshal` interfaces as the `json` package.
- encoding/ecjson: created an encrypted and compressed JSON encoder. (`JWE` is unsuitable because cipher-suite choice is not nonce-reuse resistent)
- encoding/jws: support signed JSON-based data structures (JWT/JWS) as specified by rfc7515
- identity: significantly standardized identity provider interface; most now use generic standard authenticate / refresh methods.
- identity/google: replace custom rest calls with `/x/oauth2/google`
- internal/cryptutil: abstracted serialization into their own `encoding` package.
- internal/cryptutil: consolidated x509 related functions into `certificates.go`
- internal/cryptutil: merged public key/ shared secret JWT signing into `encoding` package.
- internal/httputil: carry context.Context into client request for cancellation.
- internal/httputil: fix HeaderForwardHost docs
- internal/middleware: remove validate `SameDomain` as it is no longer relevant; routes can be on any domain tree.
- internal/middleware: remove validate `ValidateClientSecret` as it is no longer relevant; authenticity for cross service http calls is covered by HMAC.
- internal/middleware: remove validate `ValidateRedirectURI` as it is no longer relevant; routes can be on any domain tree.
- internal/sessions: remove `ParentSubdomain` as it is no longer relevant.
- internal/sessions: `Header`, `Cookie`, and `QueryParam` stores use Marshal/Unmarshal interfaces to implement session store operations.
- internal/sessions: state is a compound type of oauth2 access (kept confidential, and opaque to the user) and oidc identity tokens (used as the basis for route sessions, and are transparent).
- internal/templates: updated templates to add additional identity information: issuer, issued at, expiry and profile picture if available.
- internal/templates: removed refresh ability.
- proxy: remove force refresh endpoint
- proxy: add callback endpoint which is called by authentication and is responsible for setting (but not mutating) the user's identity session cookie.
- proxy: add endpoint (`.pomerium/api/v1/login`) that takes a `redirect_uri` (e.g. `localhost`) and returns a signed login url that can be used for programmatic access. The `redirect_uri` will be called following authentication, and call to proxy's callback handler. If no `redirect_uri` is provided, the login api handler will be used and display the result of the login.
- proxy: add callback handler (`.pomerium/callback`) that receives a signed `redirect_uri` from the authenticate service. The user's intended route session is encoded in `redirect_uri`'s query params as an encrypted payload. Once decrypted, the user's route session (`pomerium_jwt`) is saved. If the callback handler is called following a programmatic request, an additional `pomerium_refresh_token` is set and contains a refresh token that can be used to refresh the provided route session (`pomerium_jwt`). All query params are cleared before being redirected to the destination url.
- proxy: programmatic sessions will return a `401` error instead of redirecting to the identity provider.
- scripts: update `programmatic_access.py` and remove bash example.

## Open questions / problems

- How does user impersonation work with per-routes?
- What's a reasonable default per-route validity period? Should it be configurable?  (Current thinking, 10 minutes and not configurable unless we get feedback)
- Should programmatic access be enabled by default? (Curren thinking, no. It should have to be explicitly enabled on a per-route basis as part of policy)

## Related issues

-  #131 
- #128
- #152
- #158

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [x] updated CHANGELOG.md
- [x] updated UPGRADING.md
- [x] ready for review


internal/cryptutil/hmac.go
```diff
@@ -7,6 +7,11 @@ import (
 	"time"
 )
 
+const (
+	// DefaultLeeway defines the default leeway for matching NotBefore/Expiry claims.
+	DefaultLeeway = 1.0 * time.Minute
```
travisgroth: Maybe lean on the kerberos default of 5m?

internal/httputil/proxy.go
```diff
@@ -9,9 +9,9 @@ import (
 	"github.com/pomerium/pomerium/internal/log"
 )
 
-// HeaderForwardHost is the header key the identifies the originating
-// IP addresses of a client connecting to a web server through an HTTP proxy
-// or a load balancer.
+// HeaderForwardHost is the header key that identifies the original host requested
+// by the client in the Host HTTP request header.
```
travisgroth: 😆

internal/cryptutil/hmac.go
desimone: I've gone back and forth about this. Just for some context, I wanted to make this value consistent with how we handle JWTs.  The JWT [spec](https://tools.ietf.org/html/rfc7519#section-4.1.4)  says the intent of leeway for jwt based time claims is to account for clock skew.  I noticed that our [JWT](https://godoc.org/github.com/square/go-jose/jwt#pkg-constants) library uses one minute leeway and went with that. 

> Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew.

 I don't know much about why Kerberos's default leeway but I think it's within a reasonable time-frame. AD seems to use five minutes [as well](https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/maximum-tolerance-for-computer-clock-synchronization). 

My intuition tells me that five minutes is quite a long time when we are looking at probably ~10-15 minute route tokens to begin with, effectively adding anywhere from 1/3-1/2 duration on either end of a token's validity. 

I think five minutes makes sense, but think we could be more aggressive as I think it's less likely (...one would hope?) that server machines would have as "drifty" time skew as desktop machines. I'd love to know what folks like [aws](https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html) use for their allowed leeway, basically.

internal/httputil/proxy.go
desimone: 😬

internal/cryptutil/hmac.go
travisgroth: 5 is definitely a hold over from when you couldn't count on fantastic clock sync.  Since this is server to server (right?) 1m could be reasonable.

desimone: > sync. Since this is server to server (right?)

Yes

> 1m could be reasonable

Let's see how it goes 👍

victornoel:
@desimone that seems great!

So to better understand, what happens in terms of cookie: there is one cookie per protected app's domain instead of one per protected super-domain?

Does this mean that I can have two pomerium instances each protecting different urls and they won't interfere with each other?

travisgroth:
LGTM overall.  👍

travisgroth:
To round out some discussion:

>What's a reasonable default per-route validity period? Should it be configurable? (Current thinking, 10 minutes and not configurable unless we get feedback)

That sounds reasonable

>Should programmatic access be enabled by default? (Curren thinking, no. It should have to be explicitly enabled on a per-route basis as part of policy)

Agreed.

desimone:
> So to better understand, what happens in terms of cookie: there is one cookie per protected app's domain instead of one per protected super-domain?

Each managed route gets its own cookie session. Global state is still maintained on authenticate, but the two never cross paths. Route cookies will be "refreshed" every ten minutes, global session state is refreshed as needed or set by the identity provider (typically an hours). What's nice here is we now have transparent per-route session cookies in the form of JWTs, while still maintaining encrypted state on the authenticate service. And we don't have to worry about cookie domain scope anymore. I hope that is ;) 

> Does this mean that I can have two pomerium instances each protecting different urls and they won't interfere with each other? 

Possibly yes, though I haven't tried that yet, and it depends what you mean by two pomerium instances. When this merges to master, I'd be love to know if this works in your configuration as expected. Hoping to test a bunch then release `0.5.0`

victornoel:
@desimone 

> Possibly yes, though I haven't tried that yet, and it depends what you mean by two pomerium instances. When this merges to master, I'd be love to know if this works in your configuration as expected. Hoping to test a bunch then release 0.5.0

I will try to do so but I'm not 100% sure I will be able. One question: will the cookies from 0.5.0 interferes from those from previous versions?

desimone:
> @desimone
> 
> > Possibly yes, though I haven't tried that yet, and it depends what you mean by two pomerium instances. When this merges to master, I'd be love to know if this works in your configuration as expected. Hoping to test a bunch then release 0.5.0
> 
> I will try to do so but I'm not 100% sure I will be able. One question: will the cookies from 0.5.0 interferes from those from previous versions?

It should. I did some investigation on why you've been seeing some cookie errors and found a possible redirect loop if the authenticate session was not getting cleared on error.  Is [this](https://youtu.be/2yol-rd1L0w) what you were seeing as well?

PR #376, opened 2019-11-07
internal/sessions: fix cookie clear session

desimone:
## Summary

CookieStore's ClearSession now properly clears the user session cookie by setting [MaxAge to -1](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie).

- internal/sessions: move encoder interface to encoding package, and rename to MarshalUnmarshaler.
- internal/encoding: move mock to own package
- authenticate: use INFO log level for authZ error.

## Related issues
- https://github.com/pomerium/pomerium/pull/373#issuecomment-551328123

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [x] updated CHANGELOG.md
- [x] ready for review

PR #375, opened 2019-11-07
deployment: prepare v0.5.0

desimone:
## Summary

Re-organized docs a bit. Community is now a sub-setion of `docs`, and broke out configuration options and examples as their own subsection.

https://5dc73df5bde5f7000877cc24--pomerium-docs.netlify.com

## Related issues

**Checklist**:

**Before tagging**

- [x] Bump version
- [x] Update docs to have a new version in config.js and in netlify
- [x] Review and update examples 
- [x] Review and update dockerfiles 
- [x] Review and update docs
- [x] Review and update changelog
- [x] Review and update upgrading guide

**After tagging**
- [ ] Update github release docs
- [ ] Set `latest` branch to track the new release
- [ ] Publish tagged version of documentation
- [ ] Review and update helm-charts (merge after tag) 
- [ ] Change docker examples back to `master` on master.

PR #382, opened 2019-11-10
internal/sessions: fix upgrade path for new sessions

desimone:

## Summary

Previously, only the exact session cookie (including matches on `Path` and `Domain` values would be cleared when doing a session clear. Because version `v0.5.0` is moving from a global session cookie whose domain is defaulted to be the parent subdomain, on an error on the new `v0.5.0` branch, the session cookie would not be cleared because `MaxAge` and `Expiry` would be set on a _new_ session which a differently scoped `Domain`.

<s>These changes change the behavior of `ClearSession` generate new, clearing session cookies for both the route-domain (e.g. no domain set, `Host-Only`) and the previous default cookie session domain scope, which was the immediate parent domain. 

This should solve the [infinite redirect](https://github.com/pomerium/pomerium/pull/375#issuecomment-552180935) issue seen when upgrading which is caused by the managed route only clearing the _locally_ scoped session cookie, not the parent domain scoped one leftover from the previous version (which, confusingly, is still a valid authenticate session if authenticate's cookie secret has not changed). 

**N.B.** This won't solve issues for those who had previously set custom domain scopes. 
**N.B.** Though this will clear old sessions, if you downgrade again, you'll see a similar issue with competing session tokens with different scopes. </s>

These changes support (from the user perspective) a graceful upgrade path. Existing user state is ignored. 

## Related issues
- #376 
- #375 

**Checklist**:
- [x] add issues
- [x] updated docs
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/sessions/cookie_store.go
+++ b/internal/sessions/cookie_store.go
@@ -96,26 +96,40 @@ func (cs *CookieStore) makeCookie(value string) *http.Cookie {
 }
 
 // ClearSession clears the session cookie from a request
-func (cs *CookieStore) ClearSession(w http.ResponseWriter, _ *http.Request) {
+func (cs *CookieStore) ClearSession(w http.ResponseWriter, r *http.Request) {
 	c := cs.makeCookie("")
 	c.MaxAge = -1
 	c.Expires = timeNow().Add(-time.Hour)
 	http.SetCookie(w, c)
 }
 
+func getCookies(r *http.Request, name string) []*http.Cookie {
+	allCookies := r.Cookies()
+	matchedCookies := make([]*http.Cookie, 0, len(allCookies))
+	for _, c := range allCookies {
+		if strings.EqualFold(c.Name, name) {
+			matchedCookies = append(matchedCookies, c)
+		}
+	}
+	return matchedCookies
+}
+
 // LoadSession returns a State from the cookie in the request.
-func (cs *CookieStore) LoadSession(req *http.Request) (*State, error) {
-	data := loadChunkedCookie(req, cs.Name)
-	if data == "" {
+func (cs *CookieStore) LoadSession(r *http.Request) (*State, error) {
+	cookies := getCookies(r, cs.Name)
+	if len(cookies) == 0 {
 		return nil, ErrNoSessionFound
 	}
-	var session State
-	err := cs.decoder.Unmarshal([]byte(data), &session)
-	if err != nil {
-		return nil, ErrMalformed
-	}
+	for _, cookie := range cookies {
+		data := loadChunkedCookie(r, cookie)
 
-	return &session, err
+		session := &State{}
+		err := cs.decoder.Unmarshal([]byte(data), session)
+		if err == nil {
+			return session, nil
+		}
+	}
+	return nil, ErrMalformed
 }
 
 // SaveSession saves a session state to a request's cookie store.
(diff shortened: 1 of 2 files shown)


desimone:
@victornoel 

Hopefully this solves the issue.... I've confirmed the fix on my side, but want to make sure it's working for you as well.

I've made a temporary docker build you can pull @ `pomerium/pomerium:PR382`.

victornoel:
@desimone unfortunetaly, it's not working.

Maybe it is related, but I had set `cookie_domain` to the higher level domain (`example.com`) while my authenticate service is served on `auth.something.example.com` and my services are served under things like `app.else.example.com`.

I noticed some strange things when I connect to the new pomerium:
- my browser has the old `_pomerium` cookie on `example.com`
- `auth.something.example.com` answers with a new **empty** cookie with a past date (I suppose to clear it) on `something.example.com` (and not `example.com` as I would have expected)
- on the callback, my browser still sends the old `_pomerium` cookie from `example.com`
- auth then redirect againt to login starting back from the first point…

I'm not a cookie expert, but maybe instead of trying to delete old cookies, can't we just ignore them? If they are not exactly on the domain of the auth service or the domain of the proxied service, then we just ignore them and create new ones?

victornoel:
@desimone by the way, the proxy service does the same strange thing when clearing the app service cookie: for an app on `app.else.example.com`, it tries to clear `else.example.com`.

desimone:
> @desimone unfortunetaly, it's not working.

😭 

> Maybe it is related, but I had set `cookie_domain` to the higher level domain (`example.com`) while my authenticate service is served on `auth.something.example.com` and my services are served under things like `app.else.example.com`.

That's it. The challenge is you cannot delete cookies by name, or site, in one swoop. As you inferred bellow, to clear a session cookie you basically have to set a _new_ session cookie with not only the same name but also `path` and `domain` scope. The challenge for us in this new version is that while we _now_ know that all the new sessions use the exact route subdomain (and `/` for path), the previous version's global session cookie could be set anywhere from the leaf of the domain tree height (`auth.x.y.z.example.com` to `example.com`).  Unfortunately, clients/browsers do not send us any associated session data  like what path, or domain it was scoped to, but only the key value pairs of name (`_pomerium`) and value (the session). So global session was valid for both authenticate and individual proxied routes in previous version, that's no longer the case which is why we originally got stuck in a redirect loop. Authenticate thinks it has a valid session, and the proxied route is sure it does not. And "clearing" the cookie is doing nothing but clearing the _new_ session cookie which is scoped to the sentinel domain (and does nothing to that prexisting cookie).

> I noticed some strange things when I connect to the new pomerium:
> 
> * my browser has the old `_pomerium` cookie on `example.com`
👍 good to know. I think if we go the "clear" strategy, we'll have to basically walk up the domain tree and set a clear session token at each level. 

> * `auth.something.example.com` answers with a new **empty** cookie with a past date (I suppose to clear it) on `something.example.com` (and not `example.com` as I would have expected)
> * on the callback, my browser still sends the old `_pomerium` cookie from `example.com`
> * auth then redirect againt to login starting back from the first point…

I hard coded the clear cookie to the default parent session domain scope, which does nothing for your environment  unfortunately. 

> I'm not a cookie expert, but maybe instead of trying to delete old cookies, can't we just ignore them? If they are not exactly on the domain of the auth service or the domain of the proxied service, then we just ignore them and create new ones?

It's something we've talked about and considered. The challenge is there's no way to tell the source of the cookie, so we'd have to iterate through each and ignore the invalid ones. The problem is, when we end up in situation where we have a (new) malformed  or otherwise bad session cookie, we'll have to clear that as well. This is further complicated route-sessions using the same session name as authenticate service. 

That said, as I typed this out I came up with an idea. Stay tuned 👍

desimone:
Welp! I think I've got it working, for real this time by ignoring any malformed state, unless _all_ found pomerium cookies are malformed, then we clear the "new" session cookies, and restart the process.
 
Screen Shot 2019-11-11 at 6 41 12 PM
Screen Shot 2019-11-11 at 6 41 52 PM

victornoel:
> Sorry for the hassle, @victornoel, do you mind trying updating pomerium/pomerium:PR382 and trying again?

Yes of course, no problem, I will try to test it today.

Thanks for taking the time to tackle this kind of features, the way cookie management work is really complex apparently ^^

victornoel:
@travisgroth @desimone it seems to work as expected! :+1: 

Some notes:
- The old cookie on `example.com` just stays there but a new one is created on the app domain as expected!
- I can still connect to apps protected with the previous version (0.3.0) of pomerium on another cluster and the old cookie is used
- If I connect again to the app protected with the new version of pomerium, it is seamless and works as expected

Great work @desimone again, thank you for the effectiveness :)

PR #387, opened 2019-11-13
authenticate: handle XHR redirect flow

desimone:
## Summary

In the event of an authentication error, the proxy service now checks incoming requests to see if they contain the `X-Requested-With` header (which is typically, but not always, a XHR request) and will  return a direct error instead of redirecting to the identity provider. 

Also,  a request header `x-pomerium-intercepted-response: true` was added to indicate to any client application when a given response was created by pomerium itself, and not the downstream application. 

See the below links for why it's rather difficult to detect whether a request is a an ajax/xhr request. 

These changes also ensure that the authenticate service (in particular, the sign_in endpoint) properly handles preflight checks which are typically made when a `302` redirect is sent to reauthenticate a user. For example...
Screen Shot 2019-11-14 at 10 20 12 AM

### Limitations

If the managed app does not include credentials with their request (e.g. [XMLHttpRequest.withCredentials](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials), pomerium won't be able to refresh a session resulting in a hard failure (either `302` or `401 depending if the `xhr` request uses the `X-Requested-With` header). 

## Related issues
- https://pomerium-io.slack.com/archives/CK6SVMPU0/p1573568647179000
- https://github.com/pomerium/pomerium/issues/281

## See also
- [What's the point of the X-Requested-With header?](https://stackoverflow.com/a/22533680)
- [x-request-with is important for security](https://markitzeroday.com/x-requested-with/cors/2017/06/29/csrf-mitigation-for-ajax-requests.html)
- [but angular doesn't add x-request-with by default](https://github.com/angular/angular.js/issues/11008)
- [this is complicated, but XMLHttpRequest is now lame and fetch is the new hot](https://developers.google.com/web/updates/2015/03/introduction-to-fetch)
- [and fetch doesn't add x-request-with by default](https://stackoverflow.com/questions/45591594/fetch-does-not-send-headers)
- [so basically, detecting an AJAX request is not always reliable](https://stackoverflow.com/a/28029666)

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/authenticate/handlers.go
+++ b/authenticate/handlers.go
@@ -53,9 +53,11 @@ func (a *Authenticate) Handler() http.Handler {
 	// Proxy service endpoints
 	v := r.PathPrefix("/.pomerium").Subrouter()
 	c := cors.New(cors.Options{
-		AllowOriginRequestFunc: middleware.ValidateRedirectURI,
-		AllowCredentials:       true,
-		AllowedHeaders:         []string{"*"},
+		AllowOriginRequestFunc: func(r *http.Request, _ string) bool {
+			return middleware.ValidateRedirectURI(r, a.sharedKey)
+		},
+		AllowCredentials: true,
+		AllowedHeaders:   []string{"*"},
 	})
 	v.Use(c.Handler)
 	v.Use(middleware.ValidateSignature(a.sharedKey))
@@ -84,7 +86,7 @@ func (a *Authenticate) VerifySession(next http.Handler) http.Handler {
 				return
 			}
 			// redirect to restart middleware-chain following refresh
-			http.Redirect(w, r, urlutil.GetAbsoluteURL(r).String(), http.StatusFound)
+			httputil.Redirect(w, r, urlutil.GetAbsoluteURL(r).String(), http.StatusFound)
 			return
 		} else if err != nil {
 			log.FromRequest(r).Info().Err(err).Msg("authenticate: verify session")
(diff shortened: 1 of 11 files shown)


desimone:
\cc @victornoel

victornoel:
@desimone seems a nice solution, let's discuss it later as planned

authenticate/handlers.go
```diff
@@ -51,6 +52,19 @@ func (a *Authenticate) Handler() http.Handler {
 
 	// Proxy service endpoints
 	v := r.PathPrefix("/.pomerium").Subrouter()
+	c := cors.New(cors.Options{
+		AllowOriginRequestFunc: func(r *http.Request, _ string) bool {
+			return middleware.ValidSignature(
```
desimone: Typically this would be determined based on the origin's location, but in our case we care not where the request comes from, but that the request has been signed by our proxy service. 

Trusted origin is particularly important when AllowCredentials is also set to true. 
- https://ejj.io/misconfigured-cors/
- https://portswigger.net/research/exploiting-cors-misconfigurations-for-bitcoins-and-bounties

go.mod
```diff
@@ -18,7 +18,8 @@ require (
 	github.com/pomerium/go-oidc v2.0.0+incompatible
 	github.com/pquerna/cachecontrol v0.0.0-20180517163645-1555304b9b35 // indirect
 	github.com/prometheus/client_golang v0.9.3
-	github.com/rs/zerolog v1.14.3
+	github.com/rs/cors v1.7.0
+	github.com/rs/zerolog v1.16.0
```
desimone: bump zero log too :)

PR #389, opened 2019-11-15
proxy: fix forward auth, request signing

desimone:
### Forward auth fix
These changes fix forward-auth support to work with the new per-route session changes made in `v0.5.0`. Previously, it was possible to simply set a shared-domain scope session cookie on the authenticate service's domain (`authenticate.x.y.z`) which, following authentication flow, would be passed to the forward authentication endpoint for validation. In the new model, the session must be set on the specific domain (sometimes referred to as [SameSite](https://www.owasp.org/index.php/SameSite) whose access is being delegated by the forward proxy and pomerium. These changes add the necessary callback flow to enable the new session management flow. 

### Entire URL is now signed, not just redirect URL

Previously, only _part_ of the URLs passed between services was signed (`redirect_uri`), which is largely a history of the project. With these changes the entire URL is signed/hmac'd plus includes explicit expiry and issue timestamps in a similar style as our JWT tokens. This is important because in many configurations, such as forward auth, the only way to pass information between services or following callbacks is using  query params (headers/cookies are not forwarded).  This ensures the authenticity of those query params and those payloads.

### Release candidate image
You can find the latest build on dockerhub. 

```
pomerium/pomerium:PR389vRC1
sha256:b7ea311f275cda7a7a4496f152a4257bb37703a3d0adb05385c68543edaa117f
```

#### Further reading
- https://pomerium-io.slack.com/archives/CK92MUAES/p1573832306087000
- https://github.com/containous/traefik/issues/2820
- https://github.com/thomseddon/traefik-forward-auth/issues/11
- https://community.containo.us/t/forwardauth-problem/2046

### Detailed
- internal/httputil: add global header constants
- scripts: update programmatic access script to use pomerium prepended redirect_uri to avoid conflicts
- internal/urlutil: abstract query params into their own package as constants
- internal/urlutil: add `SignedURL` struct enabling signing all parts of the URI
- http: remove a extraneous redundant `nosniff` header definitions in tests
- authenticate: update CSP headers for CSS changes. 

## Related issues
- Fixes #363 

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [ ] updated CHANGELOG.md
- [x] ready for review

PR #392, opened 2019-11-20
internal/frontend : serve, don't inline, static assets

desimone:
## Summary

The existing implementation loaded `template/html` code directly from within the source-code. This was convenient at the time since we only had a few templates and static files. However, this is both difficult to manage and develop against (editing html / css/ js code in an embedded go string 😱) and required explicit and strict content-security-policy settings because of all the in-lining. 

This PR keeps static resources (via statik) bundled and compressed as part of the binary, while also serving the files normally, without the need to inline. 

### Further reading
- https://developers.google.com/web/fundamentals/security/csp
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy

## Related issues

Fixes https://github.com/pomerium/pomerium/pull/389#discussion_r347746574

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [ ] updated CHANGELOG.md
- [x] ready for review


internal/httputil/constants.go
```diff
@@ -7,3 +7,11 @@ const (
 	// Especially useful when working with single page apps (SPA).
 	HeaderPomeriumResponse = "x-pomerium-intercepted-response"
 )
+
+// HeadersContentSecurityPolicy are the content security headers added to the service's handlers
+// by default includes profile photo exceptions for supported identity providers.
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src
+var HeadersContentSecurityPolicy = map[string]string{
+	"Content-Security-Policy": "default-src 'none'; style-src 'self'; img-src 'self' https://*.googleusercontent.com;",
```
desimone: > https://*.googleusercontent.com

To allow user photos to be remotely loaded on the profile page we either need to:

1) Add each identity provider, as done here.
2) Make `img-src` less strict. The concern is that a potentially hacked JWT/idp provider could serve the user with a malformed `profile_picture` claim which would allow the owner of that remote domain to perform a `xss` / injection attack. 

I'm somewhat comfortable going for 2 because if an attacker is able to forge our signed cookies, or hack the trusted identity provider, we have much bigger issues than a `xss`.

Makefile
```diff
@@ -49,6 +49,11 @@ tag: ## Create a new git tag to prepare to build a release
 	git tag -sa $(VERSION) -m "$(VERSION)"
 	@echo "Run git push origin $(VERSION) to push your new tag to GitHub."
 
+.PHONY: frontend
+frontend: ## Runs go generate on the static assets package.
+	@echo "==> $@"
+	@CGO_ENABLED=0 GO111MODULE=on go generate github.com/pomerium/pomerium/internal/frontend
```
desimone: 🎉 codegen!

desimone:
> `- Hits         2463    2099    -364     `

Welp, I see what was juicing our test coverage results now. 🙃

travisgroth:
> > `- Hits         2463    2099    -364     `
> 
> Welp, I see what was juicing our test coverage results now. 🙃

😭

internal/httputil/errors.go
```diff
@@ -92,7 +93,7 @@ func ErrorResponse(w http.ResponseWriter, r *http.Request, e error) {
 			RequestID: requestID,
 			CanDebug:  canDebug,
 		}
-		templates.New().ExecuteTemplate(w, "error.html", t)
+		template.Must(frontend.NewTemplates()).ExecuteTemplate(w, "error.html", t)
```
desimone: httputil/errors: http template should not instantiate on each call

internal/httputil/constants.go
desimone: Done.

PR #413, opened 2019-11-28
httputil : wrap handlers for additional context

### LINKED ISSUE #383: Improve user experience when authenticate callback session expires
## What happened?

A thing I experienced, is that
- after trying to log into a protected app
- when I am redirected to the IdP (google in this case)
- if I wait a bit too long to click on the account in the IdP login page
- when the IdP redirects to the auth callback, I get a `session has expired` error page from pomerium.

I would expect that waiting 30 or 60 seconds should not provoke this kind of error. Maybe either increasing the expiration time or instead of showing the error to the user, he could be redirected again to the IdP?

I'm pretty sure sometimes users tries to log in, then they can get interrupted in their work, and when come back to their browser after a few minutes, they should be able to log in.
Right now, after seeing the error, refreshing the page only shows more new errors while the situation is not so exceptional.

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint): `pomerium/pomerium:PR382` from #382.
- Server Operating System/Architecture/Cloud: AKS/Google Identity

```logs
{"level":"error","fwd_ip":["a.b.c.d"],"ip":"10.242.1.28","user_agent":"Mozilla/5.0 (X11; Linux x86_64; rv:70.0) Gecko/20100101 Firefox/70.0","referer":"https://accounts.google.com/signin/oauth/oauthchooseaccount?client_id=xxxxxx&as=xxxxxx&destination=https%3A%2F%2Fauth.corp.example.com&approval_state=xxxxx&oauthgdpr=1&xsrfsig=xxxx&flowName=GeneralOAuthFlow","req_id":"5a0fc2bd-ef62-cd37-39ad-4a6df5f8b9d6","error":"oauth callback : 400 Bad Request: internal/cryptutil: timestamp expired: internal/cryptutil: timestamp expired","http-message":"internal/cryptutil: timestamp expired","http-code":400,"time":"2019-11-11T09:35:05Z","message":"http-error"}
```

### LINKED ISSUE #395: httputil/errors: http template should not instantiate on each call
httputil/errors: http template should not instantiate on each call

_Originally posted by @desimone in https://github.com/pomerium/pomerium/pull/392/review_comment/create_

desimone:
## Summary

1) We extend standard library's [http.HandleFunc](https://golang.org/pkg/net/http/#HandlerFunc) to support errors. This improves (imho) request flow to be more explicit and safer. In particular, returning an error and not another request handler func ensures that we don't accidentally forget to return from the parent func after a error request handler is called.

For example
```
...
		httputil.ErrorResponse(w, r, err)
		return // request will actually fall through if this return is omitted!
	}
fmt.Print("some super duper secret info")
}
```
vs

```
...
		return err
	}
fmt.Print("some super duper secret info")
}
```

In practice, this mostly meant replacing `http.HandlerFunc` with  our own `httputil.HandlerFunc` and changing our handler function definitions to return an error type. 

2. We create a custom HTTP error type that allows us uniformly handle pomerium http error, and include additional request context where available like original request url (#383). 

## Related issues

- Fixes #395 
- Fixes #383 

### See also

- https://blog.questionable.services/article/http-handler-error-handling-revisited/
- https://blog.questionable.services/article/custom-handlers-avoiding-globals/

### Screenshots

Screen Shot 2019-11-30 at 2 28 02 AM
Screen Shot 2019-11-30 at 2 27 33 AM
Screen Shot 2019-11-29 at 5 48 25 PM

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [x] ready for review


internal/httputil/router_test.go
```diff
@@ -1,42 +1,39 @@
...
-	"net/http/httptest"
 	"reflect"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/gorilla/mux"
 )
 
-func TestCSRFFailureHandler(t *testing.T) {
+// func TestCSRFFailureHandler(t *testing.T) {
```
travisgroth: Should we get rid of this block?

PR #422, opened 2019-12-12
docs(background): improved sentence flow and other stuff

Rio:
## Summary
Adjusted some things I noticed while reading the background.
Note that I'm dutch and not a native English speaker.


Rio:
Didn't consider that. But you if you'd see my dutch you'd realize the opposite is also true. XD

Issue #430, opened 2019-12-22
Websocket error with linkerd dashboard


derrickburns:
## What happened?

I cannot get the linkerd dashboard live stream of calls to work. I get "websocket error: undefined`.

## What did you expect to happen?

I expect to see a stream of calls to my service.  I see the live stream if I port-forward. I do not see the live stream if I proxy with pomerium.

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint):
v0.5.1
- Server Operating System/Architecture/Cloud:
 K8s Rev: v1.14.9-eks-ba3d77

## What's your config.yaml?

```config.yaml
apiVersion: v1
data:
  config.yaml: "policy: \n  - allow_websockets: true\n    allowed_groups:\n    - eng@tidepool.org\n
    \   allowed_users: []\n    from: https://glooe-monitoring.production.tidepool.org\n
    \   to: http://glooe-grafana.gloo-system.svc.cluster.local\n  - allow_websockets:
    true\n    allowed_groups:\n    - eng@tidepool.org\n    allowed_users: []\n    from:
    https://linkerd-dashboard.production.tidepool.org\n    to: http://linkerd-dashboard.linkerd.svc.cluster.local:8080\n"
kind: ConfigMap
metadata:
  annotations:
    flux.weave.works/antecedent: pomerium:helmrelease/pomerium
  creationTimestamp: "2019-12-12T18:50:52Z"
  labels:
    app.kubernetes.io/instance: pomerium
    app.kubernetes.io/managed-by: Tiller
    app.kubernetes.io/name: pomerium
    helm.sh/chart: pomerium-4.1.2
  name: pomerium
  namespace: pomerium
  resourceVersion: "26735217"
  selfLink: /api/v1/namespaces/pomerium/configmaps/pomerium
  uid: 525ccabf-1d10-11ea-abeb-02c077500bb6
```

desimone:
@derrickburns What do does your browser console logs look like? Do you mind posting a (sanitized) HAR file?

derrickburns:
<img width="1572" alt="Screen Shot 2019-12-24 at 10 37 18 AM" src="https://user-images.githubusercontent.com/583930/71422775-6d7b2600-2639-11ea-9e0d-e47f5f27f8a5.png">

derrickburns:
[redacted.txt](https://github.com/pomerium/pomerium/files/3998910/redacted.txt)

derrickburns:
[...]

Issue #409, opened 2019-11-27
`Access-Control-Allow-Origin` error on authenticate service


victornoel:
## What happened?

I'm experimenting a strange and problematic situation, I suppose it is since v0.5.0 because I've never seen that before.

At first it seemed similar to #390 but this one is about the pomerium service not answering CORS correctly.

Basically what happens is something like this:
- An SPA is making XHR calls without problems (with the `X-Requested-With` header)
- at one point, one of the request is considered as needing reauth by the proxy
- so the proxy returns a redirect response toward the authenticate service
- the browser tries to validate this can be done with OPTIONS
- we get the following error in the browser:
```
Access to XMLHttpRequest at 'https://auth.example.com/.pomerium/sign_in?redirect_uri=https%3A%2F%2Fapp.example.com%2Fapi%stuff%2F10&sig=nSjPGT0tgnrsizrhZnWZZ0WvYSI_Zyy0UaMXkY-vdtg%3D&ts=1574843825' (redirected from 'https://app.example.com/api/stuff/10') from origin 'https://app.example.com' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

## What did you expect to happen?

It seems to me that it is authenticate that does not answer CORS requests while I think it should.

## Environment

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint): v0.5.0
- Server Operating System/Architecture/Cloud: AKS

## What did you see in the logs?

The logs are not very clear about what happens

authenticate
```logs
{
    "level": "info",
    "fwd_ip": [
        "86.234.73.194"
    ],
    "ip": "10.242.1.28",
    "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:70.0) Gecko/20100101 Firefox/70.0",
    "referer": "https://app.example.com/optimize/stuff/10",
    "req_id": "08388fd6-6fcb-99e4-075a-12f4c86c4189",
    "error": "internal/sessions: session is not found",
    "time": "2019-11-27T08:32:05Z",
    "message": "authenticate: verify session"
}
{
    "level": "debug",
    "fwd_ip": [
        "86.234.73.194"
    ],
    "ip": "10.242.1.28",
    "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:70.0) Gecko/20100101 Firefox/70.0",
[...]

Issue #390, opened 2019-11-18
`Access-Control-Allow-Origin` and AJAX calls issue


mohsen0:
## What happened?

So I am trying to secure an AWS s3 website by putting pomerium in front of it.
I have setup CORs and configured the two s3 endpoints correctly, I mean it works without proxy alright.

```yaml
  - from: https://reports.example.com
    to: http://my-reports.s3-website-eu-west-1.amazonaws.com
    allowed_groups:
      - developers@example.com
  - from: https://reports-s3.example.com
    to: https://my-reports.s3-eu-west-1.amazonaws.com
    allowed_groups:
      - developers@example.com
```

The index page is a simple HTML page, which there is some javascript code that read from the XML feed in `https://my-reports.s3-eu-west-1.amazonaws.com` by invoking an ajax call 

## What did you expect to happen?

So the setup works without the proxy, but when the proxy is in place, the `Access-Control-Allow-*` headers are not set.

```
Access-Control-Allow-Methods: GET
Access-Control-Allow-Origin: *
Access-Control-Max-Age: 0
```

and the underlying AJAX call redirected using 302 and it seems to be canceled by the browser when it tries to be authenticated by the `authenticate.example.com`.

## How'd it happen?

1. create a bucket
2. setup a website setup
3. upload https://github.com/rufuspollock/s3-bucket-listing
4. configure the pomerium as front to that bucket
5. try to access to the page.

## What's your environment like?

- Pomerium version : 0.4.0
- Server Operating System/Architecture/Cloud: kubernetes 1.12

## What's your config.yaml?

```config.yaml
  - from: https://reports.example.com
    to: http://my-reports.s3-website-eu-west-1.amazonaws.com
    allowed_groups:
      - developers@example.com
  - from: https://reports-s3.example.com
    to: https://my-reports.s3-eu-west-1.amazonaws.com
    allowed_groups:
      - developers@example.com
```

## What did you see in the logs?

browser console logs:

```logs
Access to XMLHttpRequest at 'https://reports-s3.example.com/?delimiter=/&prefix=reports/' from origin 'https://reports.example.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

[reports.example.com.har.zip](https://github.com/pomerium/pomerium/files/3859780/reports.example.com.har.zip)

## Additional context

https://github.com/pomerium/pomerium/pull/387

desimone:
@mohsen0 Thank you for the detailed issue!

[...]

PR #438, opened 2019-12-27
authenticate/proxy: add backend refresh

desimone:
## Summary

These changes use an embedded distributed cache called [groupcache](https://github.com/golang/groupcache) to store encrypted global session state and enable backend refresh without using redirect based refresh flow.

### Background 

Previously, when a user's session expired, the next request would be intercepted by pomerium. A 302 redirect would be sent back to the client (usually a browser, but hold onto that thought). The user's client would follow that redirect to the authenticate service,  which is where the user's parent session was being persisted in the form of an _encrypted_ JWT that contains values required to perform a refresh with the upstream identity provider (namely a refresh token). After the the global session was refreshed on the authenticate service, a new route scoped session would be passed back to the managed route.  Requests from other routes would similarly be issued route-scoped session derived from that global session bound to authenticate. 

This works well when the client is a browser. Session data can be set in a route scoped JWT ( #373 ) without sensitive data, and then when a refresh needed to occur we could use the _authenticate scoped_ parent session to refresh that user session with the upstream identity provider. Sensitive data (refresh token, access token) is encrypted and domain bound to the authenticate service, the refresh flow is entirely transparent to the end user, and best of all everything is completely stateless. Pomerium doesn't require a backend store. 

Unfortunately, this refresh request flow breaks down if the  application is a single page app (spa) or makes `XHR` request using `fetch` or `XMLHttpRequest`. Even though pomerium is configured to [accept credentials](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials), third party applications will not by default pass [along cross site credentials](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) (which is understandable because of security implications). As a result, cross site script requests will fail and provide errors about getting either redirect to their identity provider with an error about CORS policy.  The underlying issue is that the cross site request is _not_ passing along the global user credential like a browser based refresh request flow would, and the `xhr` is being redirected to the upstream identity provider thinking that the user has no session. 

#### Previously

[edit](https://mermaidjs.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgVSBhcyBVc2VyXG4gICAgcGFydGljaXBhbnQgUCBhcyBQcm94eVxuICAgIHBhcnRpY2lwYW50IEFOIGFzIEF1dGhlbnRpY2F0ZVxuICAgIHBhcnRpY2lwYW50IEFaIGFzIEF1dGhvcml6ZVxuICAgIHBhcnRpY2lwYW50IEkgYXMgSWRlbnRpdHkgUHJvdmlkZXJcblxuICAgIFUtPj5QOiBzZXJ2aWNlLngueS56XG4gICAgUC0-PlA6IEhhcyBzZXNzaW9uPyBZZXMuXG4gICAgUC0-PlA6IFNlc3Npb24gZXhwaXJlZD8gWWVzLlxuICAgIFAtPj5VOiAzMDJcbiAgICBVLT4-QU46IC9zaWduX2luXG4gICAgQU4tPj5BTjogR2xvYmFsIHNlc3Npb24gd2l0aCByZWZyZXNoIHRva2VuPyBZZXMuXG4gICAgQU4tPj5BTjogR2xvYmFsIHNlc3Npb24gaXMgZXhwaXJlZD8gWWVzLlxuICAgIEFOLT4-STogUmVmcmVzaCBzZXNzaW9uXG4gICAgSS0-PkFOOiBOZXcgb2F1dGgyIHRva2VuXG4gICAgQU4tPj5VOiAzMDIgdy8gc2NvcGVkIHNlc3Npb24gaW4gcXVlcnkgcGFyYW1zXG4gICAgQU4tPj5VOiBTZXQgdXBkYXRlZCBnbG9iYWwgc2Vzc2lvbiBjb29raWVcbiAgICBVLT4-UDogMzAyIHcvIHNjb3BlZCBzZXNzaW9uIGluIHF1ZXJ5IHBhcmFtc1xuICAgIFAtPj5QOiBJcyB2YWxpZCByb3V0ZSBzZXNzaW9uP1xuICAgIFAtPj5VOiBTZXQgdXBkYXRlZCBzY29wZWQgc2Vzc2lvbiBjb29raWVcblxuIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifX0) 
mermaid-diagram-20191227195349

#### Now

mermaid-diagram-20191227194811

[edit](https://mermaidjs.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgVSBhcyBVc2VyXG4gICAgcGFydGljaXBhbnQgUCBhcyBQcm94eVxuICAgIHBhcnRpY2lwYW50IEFOIGFzIEF1dGhlbnRpY2F0ZVxuICAgIHBhcnRpY2lwYW50IEFaIGFzIEF1dGhvcml6ZVxuICAgIHBhcnRpY2lwYW50IEkgYXMgSWRlbnRpdHkgUHJvdmlkZXJcblxuICAgIFUtPj5QOiBzZXJ2aWNlLngueS56XG4gICAgUC0-PlA6IEhhcyBzZXNzaW9uPyBZZXMuXG4gICAgUC0-PlA6IFNlc3Npb24gZXhwaXJlZD8gWWVzLlxuICAgIFAtPj5BTjogL3JlZnJlc2gvaWQ6MTIzNFxuICAgIEFOLT4-QU46IHNlc3Npb24gaWQ6MTIzNCBpbiBjYWNoZT8gWWVzLlxuICAgIEFOLT4-QU46IHNlc3Npb24gdmFsaWQ_IFllcy5cbiAgICBBTi0-PkFOOiBzZXNzaW9uIGV4cGlyZWQ_IFllcy5cbiAgICBBTi0-Pkk6IFJlZnJlc2ggc2Vzc2lvblxuICAgIEktPj5BTjogTmV3IG9hdXRoMiB0b2tlblxuICAgIEFOLT4-UDogTmV3IHJvdXRlIHNjb3BlZCBzZXNzaW9uXG4gICAgUC0-PlA6IElzIHZhbGlkIHJvdXRlIHNlc3Npb24_XG4gICAgUC0-PlU6IFNldCB1cGRhdGVkIHNjb3BlZCBzZXNzaW9uIGNvb2tpZVxuXG4iLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9fQ)

### Details
- authenticate: remove unused `sessionStores` slice
- config: replace FNV hash with xxhash for struct hashing
- sessions: add cache store which implements a backend session store using groupcache and falls back to another store that implements session store
- sessions: moved errors to own file
- sessions: moved store interface to own file
- sessions: replace FNV hash with xxhash for struct hashing

## Related issues

- https://github.com/pomerium/pomerium/issues/409
- https://github.com/pomerium/pomerium/issues/430
- https://github.com/pomerium/pomerium/issues/390
- https://github.com/golang/groupcache/pull/131

## TODO in future PR
- Add authenticate instances to discover each other using something like the gossip protocol

**Checklist**:
- [x] add related issues
- updated docs n/a
- [x] updated unit tests
-  updated UPGRADING.md n/a (till next pr?)
- [x] ready for review


go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	contrib.go.opencensus.io/exporter/prometheus v0.1.0
 	github.com/cespare/xxhash/v2 v2.1.1 // indirect
 	github.com/fsnotify/fsnotify v1.4.7
-	github.com/golang/groupcache v0.0.0-20191027212112-611e8accdfc9 // indirect
+	github.com/golang/groupcache v0.0.0-20191227052852-215e87163ea7
```
desimone: 😉

internal/sessions/cache/cache_store.go
```diff
@@ -0,0 +1,131 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/golang/groupcache"
+	"github.com/pomerium/pomerium/internal/encoding"
+	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/internal/sessions"
+)
+
+var _ sessions.SessionStore = &Store{}
```
desimone: https://commaok.xyz/post/compile-time-assertions/

config/options.go
```diff
@@ -475,7 +476,7 @@ type OptionsUpdater interface {
 
 // Checksum returns the checksum of the current options struct
 func (o *Options) Checksum() string {
-	hash, err := hashstructure.Hash(o, nil)
+	hash, err := hashstructure.Hash(o, &hashstructure.HashOptions{Hasher: xxhash.New()})
```
desimone: 👀 hashstructure's default hash fn is FNV which is suboptimal for a bunch of reasons. We were pulling in xxhash anyway so I figured we should use it.  Far from a promising bottleneck in both speed and quality, but it felt like a pretty "free" change. Let me know what you think 

- https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed
- https://cyan4973.github.io/xxHash/
- https://news.ycombinator.com/item?id=10674723

internal/httputil/client.go
```diff
@@ -17,14 +16,9 @@ import (
...
-var httpClient = &http.Client{
-	Timeout: time.Second * 5,
-	Transport: &http.Transport{
-		Dial: (&net.Dialer{
-			Timeout: 2 * time.Second,
-		}).Dial,
-		TLSHandshakeTimeout: 2 * time.Second,
-	},
+// DefaultClient avoids leaks by setting an upper limit for timeouts.
+var DefaultClient = &http.Client{
```
desimone: All we really care about is the [total timeout](https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/).

config/options.go
travisgroth: No concern as long as we can still cast the result to `float64` for the info metric.  The test in `/internal/telemetry/metrics` should probably be updated with sample data from this hash if it is significantly different output.

desimone: They should be identical, both implement the `Hash64` interface and return a `Sum64() uint64`; I ... vaguely remember but forgot... we cast to a float and not a uint64 for some reason to have to do with prometheus right?

travisgroth: Oh, actually we go uint64->float so should work as-is.

authenticate/authenticate.go
```diff
@@ -135,13 +148,13 @@ func New(opts config.Options) (*Authenticate, error) {
...
+		sessionStore:     cacheStore,
 		encryptedEncoder: encryptedEncoder,
-		sessionLoaders:   []sessions.SessionLoader{qpStore, headerStore, cookieStore},
-		sessionStores:    []sessions.SessionStore{cookieStore, qpStore},
+		sessionLoaders:   []sessions.SessionLoader{cacheStore, qpStore, headerStore, cookieStore},
 		// IdP
 		provider: provider,
 
 		templates: template.Must(frontend.NewTemplates()),
+		addr:      opts.Addr,
```
travisgroth: Will this work as intended with defaults for the call to groupcache.NewHTTPPoolOpts() in handlers?

config/options.go
desimone: 👍

authenticate/authenticate.go
desimone: In this PR `Addr`, and `groupcache.NewHTTPPoolOpts` are basically nops AFAIK as the pool is never used since we only have local cache.  We would need to support one of the following to be able to have more than 1 instance of the authenticate service. 

1) static peer loading via a configuration setting
2) dynamic peer loading via something like gossip
3) both?

I had planned to look at gossip protocol (todo), but it feels a bit wrong to break N-authenticate service configs on master. Thoughts?

desimone: I've removed references to the pool since we aren't using it.

proxy/middleware.go
```diff
@@ -30,23 +34,78 @@ func (p *Proxy) AuthenticateSession(next http.Handler) http.Handler {
...
+	// 1 - build a signed backend call to the authenticate service
+	refreshURI := *p.authenticateRefreshURL
+	q := refreshURI.Query()
+	q.Set("ati", s.AccessTokenID)           // hash value points to parent token
+	q.Set("aud", urlutil.StripPort(r.Host)) // request's audience, this route
+	refreshURI.RawQuery = q.Encode()
+	signedRefreshURL := urlutil.NewSignedURL(p.SharedKey, &refreshURI).String()
 
+	// 2 - Make a restful http call to authenticate service
+	req, err := http.NewRequestWithContext(r.Context(), http.MethodGet, signedRefreshURL, nil)
```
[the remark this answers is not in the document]
desimone: I've added trace spans to the refresh call (`authenticate.VerifySession/refresh`).

Metrics is a little trickier, but agree that we should have a tag for requests that make external calls. Not just for refresh but also authentication.  I'd like to handle in a follow up PR (#441 ) Thoughts?

internal/sessions/cookie/cookie_store.go
```diff
@@ -180,20 +189,26 @@ func (cs *CookieStore) setCookie(w http.ResponseWriter, cookie *http.Cookie) {
...
 func loadChunkedCookie(r *http.Request, c *http.Cookie) string {
+	if len(c.Value) == 0 {
+		return ""
+	}
+	if []byte(c.Value)[0] != ChunkedCanaryByte {
+		return c.Value
+	}
+
 	data := c.Value
 	// if the first byte is our canary byte, we need to handle the multipart bit
```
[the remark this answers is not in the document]
desimone: 👍 good catch

desimone:
@travisgroth I think that covers most of the requested changes. Let me know what you think .

internal/telemetry/metrics/const.go
```diff
@@ -34,10 +34,10 @@ var (
 // DefaultViews are a set of default views to view HTTP and GRPC metrics.
 var (
 	DefaultViews = [][]*view.View{
-		GRPCServerViews,
```
desimone: We were double registering `GRPCServerViews` 👀

internal/telemetry/metrics/const.go
```diff
@@ -34,10 +34,10 @@ var (
...
-		HTTPServerViews,
 		GRPCClientViews,
 		GRPCServerViews,
+		HTTPClientViews,
```
desimone: re-enable HTTPClientViews

desimone:
@victornoel @commixon @mohsen0 @derrickburns

If you are able to test the image `pomerium/pomerium:PR438v1` a try that would be very helpful. Thank you!

proxy/middleware.go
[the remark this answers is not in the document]
travisgroth: Agree.  A little more internal data would be good.  OpenCensus doesn't make one-off metrics very lightweight so handling this as a new issue seems fair.  Plenty of work to do for this functionality to be fully baked, so no need to bloat the first PR.

authenticate/authenticate.go
travisgroth: 👍

internal/telemetry/metrics/const.go
travisgroth: Nice catch

derrickburns:
I get the same error message and similar console output:
Screen Shot 2019-12-30 at 10 03 49 AM

wangyi-fudan:
just mention wyhash, the fastest high quality conventional hash function:
https://github.com/wangyi-fudan/wyhash

PR #446, opened 2020-01-06
authenticate : add distributed session cache

desimone:
Man this was... a doozy.

## Summary

These changes adds distributed caching (via groupcache), and automatic pool enrollment (via memberlist) to the cache system introduced in #438. This allows us to do backend refresh (required by many single page app / SPA environments), using multiple distributed authenticate services, without having to operate a third party datastore. 

If you have a single authenticate service, this should have no effect on you.  However, if you run N-instances of the authenticate service, session information will be split amongst those N nodes. The cache size will grow  proportionally to the number of authenticate nodes in the cache pool (10MB per node). To give you an idea, a _huge_ encrypted session would be ~4KB (and usually <1KB), so each node should be able to handle ~2500 (not inc. cache overhead) sessions before evicting the least frequently used keys. 

### Details 

- cache only touches / sees encrypted data
- cache pool calls including `Get`s are hmac'd like other cross-service calls

### How do we bootstrap the local cluster / membership? 

The cache pool will (in most cases) operate behind the ingress using LAN resolvable hostnames to IPS (which is what memberlist uses).

  - we must provide memberlist with seed node. We can't give it authenticate services main url because that by definition must be publicly resolvable/accessible and would return something like 41.12.12.12 which memberlist UDP/TCP packets would not cross.
  - TLS: The groupcache pool handler hangs off the main http server's handler. That means if you are running HTTP in insecure mode, the group cache client will also have to use that port and scheme. 
  - TLS SNI: if pomerium is directly serving it's own certificate, the internal call to authenticate service will result in a SNI mismatch error (it will dial something like `https://192.1.1.12/_groupcache`  and get back a cert for `authenticate.x.y.z` and the http go client will reject. 

Right now, I've "solved" this issue by reintroducing `authenticate_internal_service_url`.
Amusingly one option we could take to the above is to just make _all_ services part of the group cache (whether they be proxy, authZ or AuthN). The benefit there is we could just bootstrap using  `authorize_service_url`. It would also let us the cache as a store for other services, like AuthZ.

#### How do we test this?

I've so far only tested this in docker-compose. So far memberlist works beautifully. I can kill/scale nodes and everything just works. More sophisticated network environments, especially where there is some level of network segmentation will probably break in unanticipated ways. I'm thinking in particular of authenticate instances spread across multiple pods.  How do we best test this?

### Try it!

```bash
$ docker run pomerium/pomerium:PR446 --version
v0.5.0+d649d7e
```
And scale your authenticate instance/replicas. 
## Related issues

- https://github.com/pomerium/pomerium/pull/438
- #409 
- https://github.com/pomerium/pomerium/issues/409
- https://github.com/pomerium/pomerium/issues/430
- https://github.com/pomerium/pomerium/issues/390
- https://github.com/golang/groupcache/pull/131
- https://github.com/golang/groupcache/pull/132
- Fixes https://github.com/pomerium/pomerium/issues/281

## See also

-  https://github.com/pomerium/autocache
- https://github.com/hashicorp/memberlist
- https://github.com/golang/groupcache 
- https://medium.com/@orijtech/groupcache-instrumented-by-opencensus-6a625c3724c
- https://gist.github.com/yosisa/9915894

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


travisgroth:
The caveats around membership and bootstrapping make me wonder if this should just be an additional, but optional, internal-only service or service personality.  Splitting the cache into a new tier should give us a clean URL and no split DNS or SNI problems.  This would add a little complexity to turn `authenticate` into an RPC client, but it might be worth it in the long run.  Thoughts?

desimone:
@travisgroth 

> The caveats around membership and bootstrapping make me wonder if this should just be an additional, but optional, internal-only service or service personality. Splitting the cache into a new tier should give us a clean URL and no split DNS or SNI problems. This would add a little complexity to turn `authenticate` into an RPC client, but it might be worth it in the long run. Thoughts?

It's a great question and I've been pretty split about this. Here's how I've been thinking about the tradeoffs the last few days:

Pro-service: 
- autocache feels like an independent store abstraction. We don't care how `Get()` works, just that it does, and it ideally should be interchangeable with another store system that provides that interface. 
- current use-case is not very latency sensitive, and won't make use of the local hotcache. 
- it would be nice to be able to scale store components independent of AuthN instances.

Pro-embedded store:
- It's one less component to have to manage independently.we are asking operators to manage yet another service. while k8s makes this less of an issue, I think it does make other deployment configurations trickier, and adds to the just overall cognitive load of planning out a deployment.  Indeed we lose one of groupcache's biggest advantages, namely:
> [groupcache] does not require running a separate set of servers, thus massively reducing deployment/configuration pain. groupcache is a client library as well as a server. It connects to its own peers, forming a distributed cache.
- we close ourselves off from using the hotcache in the future. I can _definitely_ see us using groupcache with AuthZ (where every request is checked, and is latency and load sensitive) for storing external data locally for making policy decisions.
- We get TLS without having to explicitly configure _another_ set of internal certificates. In theory, we can extend our custom http.RoundTripper to have the http client accept the SNI of authenticate_service_url's [server name](https://golang.org/pkg/crypto/tls/#Config) as a valid SNI. So we would get TLS on that endpoint without needing additional configuration. But I need to test this assumption / add that to this PR to be sure. 

Toss up?: 
-  Either way we are going to have to have a configuration setting for an internal service address unless we bootstrap with `authz's` address and make that part of the pool. 
- We can avoid TLS issues by moving groupcache's pool off the main handler in either scenario. If we want TLS, we are either going to have to provide it a new set of certs, or override the accepted SNI from the cache get dialer to accept authN's external SNI. 

Overall, I'm leaning towards keeping cache being embedded. Let me know if I'm missing something, or I've got something wrong.

travisgroth:
>I think it does make other deployment configurations trickier

Yes.  I was thinking that this would be an optional service unless you were running SPAs.  However, it seems like this will be everyone in the near future.  😬 

>we close ourselves off from using the hotcache in the future. I can definitely see us using groupcache with AuthZ

That's fair and probably the best use case for it being embedded long term.

Can you write up the configuration that embedded mode would require?  I guess this depends on the SNI testing you mentioned.  Seeing the config a user would need to maintain in these scenarios would make this a bit more concrete for me.

PR #493, opened 2020-02-07
authenticate: make callback path configurable

### LINKED ISSUE #489: authenticate: make oauth callback address configurable
Investigating pomerium vs buzzfeed/sso and this is just causing a minor pain.

We'd like to re-use the client already setup in Google but the callback url there is /google/callback (it uses the provider "reference" IIRC).  And pomerium wants to use oath2/callback.  It would be nice if this was configurable for easing migrations.

desimone:
## Summary
These changes make the authentication service's callback path (oidc/oauth2 `redirect_uri`) configurable. 

## Related issues
Fixes #489

## See also

- https://tools.ietf.org/html/rfc6749#section-3.1.2
- https://developers.google.com/identity/protocols/OpenIDConnect#setredirecturi

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] ready for review


desimone:
\cc @tl-adrian-bridgett

PR #509, opened 2020-02-20
config: Expose and set default GRPC Server Keepalive Parameters

travisgroth:
## Summary
As of v0.26, go-grpc no longer polls DNS regularly for endpoint updates.  Discovery is only performed when a client detects a state change.  This is triggered during error conditions, but routine scaling operations in an otherwise functioning environment may leave endpoints underutilized.  To ensure we occasionally re-discover endpoints even when no connection errors have been encountered, the recommended pattern is to set server side options which gracefully terminate client connections at regular intervals.

This PR sets a reasonable default (5m) for this occasional termination and allows users to configure it further for their environments if needed.

Reference documentation - https://godoc.org/google.golang.org/grpc/keepalive#ServerParameters

## Related issues
https://github.com/grpc/grpc-go/pull/3165
https://github.com/grpc/grpc-go/pull/3228
https://github.com/grpc/grpc-go/issues/3353#issuecomment-582040107

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/cmd/pomerium/main.go
+++ b/cmd/pomerium/main.go
@@ -9,8 +9,6 @@ import (
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/gorilla/mux"
-	"google.golang.org/grpc"
-
 	"github.com/pomerium/pomerium/authenticate"
 	"github.com/pomerium/pomerium/authorize"
 	"github.com/pomerium/pomerium/cache"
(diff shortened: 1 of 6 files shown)


cmd/pomerium/main.go
```diff
@@ -27,6 +25,8 @@ import (
 	"github.com/pomerium/pomerium/internal/urlutil"
 	"github.com/pomerium/pomerium/internal/version"
 	"github.com/pomerium/pomerium/proxy"
+	"google.golang.org/grpc"
```
desimone: nit: import order

PR #518, opened 2020-03-04
providers: re-enable gitlab support

### LINKED ISSUE #136: internal/identity: re-enable gitlab

The docs are saying Gitlab was removed as provider because there is no access to private emails. This seems to be fixed now: https://gitlab.com/gitlab-org/gitlab-ce/issues/44435#note_156557985

I'm new here but I'd guess most old code from the former integration should work.

Lumexralph:
## Summary

Theses changes re-enable Gitlab's support as identity provider.

## Related issues

Fixes #136 

**Checklist**:
- [X] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [X] ready for review


internal/identity/gitlab.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil && err != httputil.ErrTokenRevoked {
+		return err
+	}
+
+	return nil
+}
+
+// GetSignInURL returns the sign in URL with typical oauth parameters
+func (p *GitLabProvider) GetSignInURL(state string) string {
+	return p.oauth.AuthCodeURL(state, oauth2.AccessTypeOffline, oauth2.SetAuthURLParam("prompt", "select_account"))
```
desimone: Does this require the additional params for  `prompt select_account`? I wasn't able to get auth flow working without removing this.

internal/identity/gitlab.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// UserGroups returns a slice of groups for the user.
+//
+// By default, this request returns 20 results at a time because the API results are paginated.
+// https://docs.gitlab.com/ee/api/groups.html#list-groups
+func (p *GitLabProvider) UserGroups(ctx context.Context, s *sessions.State) ([]string, error) {
```
desimone: I was unable to fetch groups.  Perhaps because I am not in any groups? However, I am getting a 200 response from the groups endpoint. 

Screen Shot 2020-03-04 at 8 38 00 PM


When I add myself to a group, the json unforunately still does not parse

```
[{"id":7332779,"web_url":"https://gitlab.com/groups/RayStrong","name":"Ray Strong","path":"RayStrong","description":"","visibility":"private","share_with_group_lock":false,"require_two_factor_authentication":false,"two_factor_grace_period":48,"project_creation_level":"developer","auto_devops_enabled":null,"subgroup_creation_level":"maintainer","emails_disabled":null,"mentions_disabled":null,"lfs_enabled":true,"avatar_url":null,"request_access_enabled":true,"full_name":"Ray Strong","full_path":"RayStrong","parent_id":null,"ldap_cn":null,"ldap_access":null}]
```

```
8:51PM INF httputil: ErrorResponse error="Internal Server Error: oauth callback : error redeeming authenticate code: internal/identity: could not retrieve groups json: cannot unmarshal array into Go value of type struct { Groups []struct { ID string \"json:\\\"id\\\"\"; Name string \"json:\\\"name\\\"\"; Path string \"json:\\\"path\\\"\"; Description string \"json:\\\"description\\\"\"; Visibility string \"json:\\\"visibility\\\"\" } \"json:\\\"value\\\"\" }" ip=192.168.1.1 req_id=cc1aca1a-b339-d7d0-45fe-1d46d67594ba user_agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.122 Safari/537.36"
```

internal/identity/gitlab.go
Lumexralph: @desimone I'll love to have a call with you to sort all these out.

It will be a good idea to have the update of the doc in a separate PR, what do you think?

Lumexralph: I have added the docs for gitlab support

internal/identity/gitlab.go
Lumexralph: I have worked on this and we should not have this issue again.

docs/docs/identity-providers/gitlab.md
```diff
@@ -0,0 +1,48 @@
...
+Scopes       | **Must** select **read_user** and **openid**
+
+Your `Client ID` and `Client Secret` will be displayed:
+
+![Gitlab OAuth Client ID and Secret](./img/gitlab/gitlab-credentials.png)
+
+Set `Client ID` and `Client Secret` in Pomerium's settings. Your [environmental variables] should look something like this.
+
+```bash
+REDIRECT_URL="https://sso-auth.corp.beyondperimeter.com/oauth2/callback"
```
desimone: to me this looks more like
```
authenticate_service_url: https://authenticate.localhost.pomerium.io
idp_provider: "gitlab"
idp_client_id: "REDACTED"
idp_client_secret: "REDACTED"
```

docs/docs/identity-providers/gitlab.md
```diff
@@ -0,0 +1,48 @@
...
+Go to the user settings which can be found in the user profile to [create an application](https://gitlab.com/profile/applications) where you will get your app credentials
+
+![create an application](./img/gitlab/gitlab-create-applications.png)
+
+On the **Applications** page, add a new application by setting the following parameters:
+
+Field        | Description
+------------ | --------------------------------------------
+Name         | The name of your web app
+Redirect URI | `https://${redirect-url}/oauth2/callback`
```
desimone: Redirected URI is an old configuration option.  The new value is `authenticate_service_url` .

internal/identity/gitlab.go
```diff
@@ -0,0 +1,127 @@
...
+		Name     string   `json:"name"`
+		Nickname string   `json:"nickname"`
+		Email    string   `json:"email"`
+		Groups   []string `json:"groups"`
+	}
+
+	params := url.Values{}
+	params.Add("access_token", oauth2Token.AccessToken)
+
+	err = httputil.Client(ctx, http.MethodGet, p.ProviderURL+oidcURL, "nil", nil, params, &userInfo)
```
desimone: This overlaps a bit with https://github.com/pomerium/pomerium/pull/529 because both require calling the user information endpoint . Would love your review on that PR, but I've confirmed this method works as well but is a little more verbose. If you are curious, check out that PR to see how the `UserInfo` endpoint can be queried from `oidc` package. 

If you are really having trouble sleeping, you can check out the RFC for the UserInfo endpoint here: https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

internal/identity/gitlab.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var userInfo struct {
+		ID       string   `json:"sub"`
+		Name     string   `json:"name"`
+		Nickname string   `json:"nickname"`
+		Email    string   `json:"email"`
+		Groups   []string `json:"groups"`
```
desimone: The problem with pulling groups during authenticate is that when the user's session is refreshed, there is the potential that the user would have moved groups. The method you had before where you assigned [UserGroupFn](https://github.com/pomerium/pomerium/pull/518/files/8846895e4176b7df717e99fcb4e53fd96e709f41#diff-063037931977d164f6403738f38404cfR64) which would be called on authenticate and on refresh would solve that problem.

internal/identity/gitlab.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, fmt.Errorf("identity/gitlab: getting user info error %w", err)
+	}
+
+	s, err := sessions.NewStateFromTokens(idToken, oauth2Token, p.RedirectURL.Hostname())
+
+	if err != nil {
+		return nil, err
+	}
+
+	s.Email = userInfo.Email
```
desimone: 👍

internal/identity/gitlab.go
```diff
@@ -0,0 +1,127 @@
...
+	Subject       string `json:"sub"`
+	Profile       string `json:"profile"`
+	Email         string `json:"email"`
+	EmailVerified bool   `json:"email_verified"`
+
+	RevokeURL string `json:"revocation_endpoint"`
+}
+
+// NewGitLabProvider returns a new GitLabProvider.
+// https://www.pomerium.io/docs/identity-providers.html#gitlab-active-directory
```
desimone: dead url

internal/identity/gitlab.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/pomerium/pomerium/internal/httputil"
+	"github.com/pomerium/pomerium/internal/sessions"
+	"github.com/pomerium/pomerium/internal/version"
+	"golang.org/x/oauth2"
+)
+
+const (
+	defaultGitLabProviderURL = "https://gitlab.com"
+	revokeURL                = "https://gitlab.com/oauth/revoke"
+	oidcURL                  = "/oauth/userinfo"
```
desimone: This can be parsed from the  `/.well-known/openid-configuration ` oidc endpoint see : https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig  and this comment https://github.com/desimone

internal/identity/gitlab.go
Lumexralph: @desimone I should have updated that part when I updated the docs. Nice catch!

docs/docs/identity-providers/gitlab.md
Lumexralph: @desimone Cool, I implement this feedback, thanks.

internal/identity/gitlab.go
Lumexralph: I will revert to the initial groups implementation, thanks.

internal/identity/gitlab.go
Lumexralph: @desimone  This will require making a new request to get all the endpoints we want, that means we won't be hard-coding the endpoints at the expense of a request, right?

internal/identity/gitlab.go
Lumexralph: @desimone Using the oidc package is really less verbose compared to this current implementation, thanks.

internal/identity/gitlab.go
desimone: That is right.

internal/identity/gitlab.go
desimone: 👍

docs/.vuepress/config.js
```diff
@@ -85,6 +85,7 @@ module.exports = {
             "identity-providers/",
             "identity-providers/azure",
             "identity-providers/cognito",
+            "identity-providers/gitlab",
```
desimone: 👀 👍

docs/docs/identity-providers/gitlab.md
```diff
@@ -0,0 +1,41 @@
...
+lang: en-US
+sidebarDepth: 0
+meta:
+  - name: keywords
+    content: gitlab oidc openid-connect identity-provider
+---
+
+# GitLab
+
+Log in to your Gitlab or create an account if you don't have one [here](https://gitlab.com/users/sign_in)
```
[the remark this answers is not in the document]
Lumexralph: @desimone What I thought here was that the user can log in or can create an account if they don't have one in the same address.  What do you think?

desimone:
\cc @Miroling does this work with your configuration as well?

desimone:
@Lumexralph fyi, #529 has been merged. 👍

Lumexralph:
> \cc @Miroling does this work with your configuration as well?

@Miroling can you please confirm if the current implementation works with your configuration?

desimone:
merged, thanks @Lumexralph !

PR #519, opened 2020-03-05
docs: make from source quickstart

desimone:
docs: make from source quickstart

- move building from source docs from contributing to a quick start
- remove dnsmasq instructions (we do not support an offline IdP at this point so moot)
- bump vuepress

**Checklist**:
- [x] updated docs
- [x] ready for review


desimone:
\cc @Lumexralph

travisgroth:
> @travisgroth looks like this is failing on tip. It's good to know but maybe we should change rules to allow TIP failures (and... switch to actions 🤣 )

https://github.com/pomerium/pomerium/blob/master/.travis.yml#L16-L17

😠

PR #529, opened 2020-03-08
identity: support oidc UserInfo Response

### LINKED ISSUE #526: Query UserInfo endpoint to gain more data about the user
Some OIDC providers don't include info like emails, groups etc. in the ID Token. They require the client to query a UserInfo endpoint to get more info about the user.  This causes the headers for email/groups etc. to be empty

Perhaps a flag to enable UserInfo endpoint query when doing OIDC authentication?

Code that was previously there to do this:
https://github.com/pomerium/pomerium/pull/28/files#diff-71c5c7f008749ada28449d5da94170fbR121-R138

desimone:
identity: support oidc UserInfo Response
- switch back to main oidc repo

## Summary 
Attempt to gather more info especially useful if initial request did not an contain email, or subject. 

https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

## Related issues

- fixes #526 

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/internal/identity/google.go
+++ b/internal/identity/google.go
@@ -8,7 +8,7 @@ import (
 	"net/http"
 	"net/url"
 
-	oidc "github.com/pomerium/go-oidc"
+	oidc "github.com/coreos/go-oidc"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	admin "google.golang.org/api/admin/directory/v1"
--- a/internal/identity/providers.go
+++ b/internal/identity/providers.go
@@ -10,7 +10,7 @@ import (
 
 	"github.com/pomerium/pomerium/internal/sessions"
 
-	oidc "github.com/pomerium/go-oidc"
+	oidc "github.com/coreos/go-oidc"
 	"golang.org/x/oauth2"
 )
 
@@ -81,6 +81,8 @@ type Provider struct {
 
 	UserGroupFn func(context.Context, *sessions.State) ([]string, error)
 
+	UserInfoEndpoint bool
+
 	// ServiceAccount can be set for those providers that require additional
 	// credentials or tokens to do follow up API calls (e.g. Google)
 	ServiceAccount string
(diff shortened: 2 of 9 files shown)


desimone:
\cc @bakshi41c

internal/identity/google.go
```diff
@@ -8,7 +8,7 @@ import (
 	"net/http"
 	"net/url"
 
-	oidc "github.com/pomerium/go-oidc"
+	oidc "github.com/coreos/go-oidc"
```
Lumexralph: @desimone I realized there is a change in the go-oidc package, would you be kind to let me know what prompted this change?

internal/identity/providers.go
```diff
@@ -117,6 +119,24 @@ func (p *Provider) Authenticate(ctx context.Context, code string) (*sessions.Sta
...
+
+	// check if provider has info endpoint, try to hit that and gather more info
+	// especially useful if initial request did not an contain email, or subject
+	// https://openid.net/specs/openid-connect-core-1_0.html#UserInfo
+	var claims struct {
+		UserInfoURL string `json:"userinfo_endpoint"`
+	}
+
+	if err := p.provider.Claims(&claims); err == nil && claims.UserInfoURL != "" {
+		userInfo, err := p.provider.UserInfo(ctx, oauth2.StaticTokenSource(oauth2Token))
```
Lumexralph: Nice! I will use this implementation in my current PR but would wait for this to be merged so that I won't need to update the changes you already made.

internal/identity/providers.go
```diff
@@ -117,6 +119,24 @@ func (p *Provider) Authenticate(ctx context.Context, code string) (*sessions.Sta
...
+		if err != nil {
+			return nil, fmt.Errorf("internal/identity: could not retrieve user info %w", err)
+		}
+		if err := userInfo.Claims(&s); err != nil {
+			return nil, err
+		}
```
Lumexralph: Nice! I will use this implementation in my current PR but would wait for this to be merged so that I won't need to update the changes you already made.

internal/identity/google.go
desimone: There were initially some changes between the upstream repo and our fork but those changes have since converged so it made sense to revert back to upstream.

internal/identity/providers.go
desimone: 👍

PR #552, opened 2020-03-21
cmd:  add cli to generate  service accounts

### LINKED ISSUE #492: Programmatic access with Google service_account

We want to implement automated end-to-end tests for our pomerium-secured web applications, which involves a Google account. However, we do not really want to use an actual user account but rather a service account.

By following the instructions on [Programmatic access](https://www.pomerium.io/docs/reference/programmatic-access.html), I could not find it possible to use a service account to do so.

Make it possible for service account to be authenticated via Pomerium without human intervention.

desimone:
## Summary

This adds a command line interface to generates pomerium service accounts using the same shared key as your proxy service. 

NOTA BENE: These personas will be valid until the expiry (defaults to one hour) or the shared key is changed. 

## Related issues

Fixes #492 

## Pictures 

Screen Shot 2020-03-21 at 11 44 53 AM

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/cmd/pomerium-cli/cli.go
+++ b/cmd/pomerium-cli/cli.go
@@ -1,0 +1,157 @@
+package main
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/fatih/color"
+	"gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+type stringSlice []string
+
+func (i *stringSlice) String() string {
+	return fmt.Sprint(*i)
+}
+
+func (i *stringSlice) Set(value string) error {
+	if len(*i) > 0 {
+		return errors.New("already set")
+	}
+	for _, dt := range strings.Split(value, ",") {
+		*i = append(*i, dt)
+	}
+	return nil
+}
+
+type serviceAccount struct {
+	// Standard claims (as specified in RFC 7519).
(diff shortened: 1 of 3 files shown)


desimone:
\cc @travisgroth if there's an easy way to make this work with Cobra that isn't 5x the LOC I haven't figured it out. If you are cool with it, I'd like to merge this (assuming passes review) as is and refactor with cli / cobra later.

cmd/pomerium-cli/cli.go
```diff
@@ -0,0 +1,157 @@
...
+		return errors.New("iss is required")
+	}
+
+	// check that we've got our shared key to sign our jwt
+	args := flags.Args()
+	if len(args) != 1 {
+		return errors.New("shared key required")
+	}
+
+	sharedKey := args[0]
```
travisgroth: I feel a little odd asking for this key as a command line arg without another option.  Stdin and/or a file path containing the secret would be good alternatives less prone to leaking in shell history, etc.

desimone:
updated !
carbon

cmd/pomerium-cli/cli.go
desimone: Updated! Still allows insecure-y argument but has option for stdin input.

carbon

desimone: \cc @freeyoung how's this look?

freeyoung: lgtm 👍

PR #553, opened 2020-03-22
authorization: log audience claim failure

desimone:
## Summary

When a authorization request fails because the `aud` claim is invalid, these changes add additional context to that request failure. 

**Checklist**:
- [x] ready for review


internal/sessions/state.go
```diff
@@ -116,18 +116,11 @@ func (s State) RouteSession() *State {
 
 // Verify returns an error if the users's session state is not valid.
 func (s *State) Verify(audience string) error {
-	if s.NotBefore != nil && timeNow().Before(s.NotBefore.Time()) {
```
desimone: These will still be checked on every request as part of AuthZ. 

Functionally, authN really only uses  Verify now to to check if the session is expired, and therefore whether or not to try and refresh the session.

PR #405, opened 2019-11-26
deployment: github actions test and release

desimone:
## Summary

Same tests for pull requests but 20x faster but without the docker builds. Travis-ci still is run when merged from fork to mainline.  Win, win?

## Related issues

- impatience
- Me testing by pushing small commits

**Checklist**:
- [x] add related issues
- [x] ready for review

Issue #584, opened 2020-04-04
deployment: add debug docker image


desimone:
**Is your feature request related to a problem? Please describe.**

We build our main images from a distroless base because:

> Restricting what's in your runtime container to precisely what's necessary for your app is a best practice employed by Google and other tech giants that have used containers in production for many years. It improves the signal to noise of scanners (e.g. CVE) and reduces the burden of establishing provenance to just what you need.

**Describe the solution you'd like**

Sometimes (especially related to DNS, or certificate issues) we _do_ want to be able to shell into a box and debug what's going on. In which case, I suggest we release a `-debug` container for each each major release, master, and latest. 

See: https://github.com/GoogleContainerTools/distroless#debug-images

**Describe alternatives you've considered**

Use busybox on the same pod. But that's not the same thing. 

**Explain any additional use-cases**

na

**Additional context**

na

\cc @travisgroth

desimone:
@travisgroth  I believe this was completed in https://github.com/pomerium/pomerium/pull/1513 ?

https://hub.docker.com/layers/pomerium/pomerium/debug/images/sha256-040a8abff61d19226a7849baff3cfbb2ee151044bf8d4882dcfc89a4547b8d48?context=explore

travisgroth:
As discussed we should add this for releases too before calling it closed.

PR #586, opened 2020-04-05
deployment: publish with github actions and goreleaser

travisgroth:
## Summary
- Switch release process to github actions
- Use goreleaser for archives and release docker images

## Related issues
- https://github.com/pomerium/pomerium/pull/405
- 🥃 

**Checklist**:
- [x] add related issues
- [x] ready for review
- [x] set up dockerhub token


.github/goreleaser.yaml
```diff
@@ -0,0 +1,181 @@
...
+    ldflags:
+      - -s -w
+      - -X github.com/pomerium/pomerium/internal/version.Version={{.Version}}
+      - -X github.com/pomerium/pomerium/internal/version.GitCommit={{.ShortCommit}}
+      - -X github.com/pomerium/pomerium/internal/version.BuildMeta={{.Timestamp}}
+      - -X github.com/pomerium/pomerium/internal/version.ProjectName=pomerium
+      - -X github.com/pomerium/pomerium/internal/version.ProjectURL=https://wwww.pomerium.io
+
+archives:
+  - name_template: "{{ .ProjectName }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
```
desimone, line 71: 👍

.travis.yml
```diff
@@ -28,23 +28,23 @@ after_success:
...
-    - dist/pomerium-darwin-amd64.md5
-    - dist/pomerium-darwin-amd64.sha256
-    - dist/pomerium-linux-amd64
-    - dist/pomerium-linux-amd64.md5
-    - dist/pomerium-linux-amd64.sha256
-    - dist/pomerium-windows-amd64
-    - dist/pomerium-windows-amd64.md5
-    - dist/pomerium-windows-amd64.sha256
-  skip_cleanup: true
+# deploy:
```
desimone, line 31: i know this is draft but just a note reminder to remove prior to merge

Dockerfile
```diff
@@ -1,18 +1,11 @@
 FROM golang:latest as build
 WORKDIR /go/src/github.com/pomerium/pomerium
-# docker build --build-arg ARCH=arm --build-arg ARM=7 .
```
desimone, line 3: 👍

.github/goreleaser.yaml
```diff
@@ -0,0 +1,181 @@
...
+    build_flag_templates:
+      - "--pull"
+      - "--label=org.opencontainers.image.created={{.Date}}"
+      - "--label=org.opencontainers.image.name={{.ProjectName}}"
+      - "--label=org.opencontainers.image.revision={{.FullCommit}}"
+      - "--label=org.opencontainers.image.version={{.Version}}"
+      - "--label=org.opencontainers.image.source={{.GitURL}}"
+      - "--label=repository=http://github.com/pomerium/pomerium"
+      - "--label=homepage=http://www.pomerium.io"
+# changelog:
```
desimone, line 170: Can probably remove unless we want to replace  [github releaser](https://github.com/pomerium/pomerium/blob/master/.github/release-drafter.yml) with go-releaser's built in change log

.github/goreleaser.yaml
```diff
@@ -0,0 +1,181 @@
...
+
+    ldflags:
+      - -s -w
+      - -X github.com/pomerium/pomerium/internal/version.Version={{.Version}}
+      - -X github.com/pomerium/pomerium/internal/version.GitCommit={{.ShortCommit}}
+      - -X github.com/pomerium/pomerium/internal/version.BuildMeta={{.Timestamp}}
+      - -X github.com/pomerium/pomerium/internal/version.ProjectName=pomerium
+      - -X github.com/pomerium/pomerium/internal/version.ProjectURL=https://wwww.pomerium.io
+
+  - id: pomerium-cli
```
desimone, line 39: 🙏 thanks!

.github/Dockerfile-release
```diff
@@ -0,0 +1,9 @@
+FROM busybox:latest as build
```
desimone, line 1: When is this used vs the Dockerfile in the root?
https://github.com/pomerium/pomerium/pull/586/files#diff-3254677a7917c6c01f55212f86c57fbf

travisgroth: Dockerfile at the root is for just a normal container build that someone might use for testing.  The release ones are just for goreleaser since they require a build to be run externally.

I think the normal build just needs to cover x86_64 for testing, so I removed the other archs.  We could maybe add back in ARM64 if you think that's a good idea.

.travis.yml
travisgroth: 💯

.github/goreleaser.yaml
travisgroth: ✔️

travisgroth:
Process wise this shouldn't change how we do releases much.  

- Update current draft release/notes as needed
- Publish draft with name matching `name_template: "v{{.Version}}"`
- goreleaser will build and upload artifacts to the published release

.github/Dockerfile-release
desimone: > Dockerfile at the root is for just a normal container build that someone might use for testing. The release ones are just for goreleaser since they require a build to be run externally.

Got it. Makes complete sense. 
> We could maybe add back in ARM64 if you think that's a good idea.
Nah, I think we are good with a generic x64 for now. Maybe add a debug build per #584 in a follow up PR.

travisgroth: That top level one could just include user tools.  I'll follow up on that.

PR #595, opened 2020-04-08
identity/provider: implement generic revoke method

### LINKED ISSUE #501: identity: support revocation_endpoint [rfc8414]

Similar to issue #494, I have an instance of pomerium integrated to keycloak using OIDC. When I call the .pomerium/sign_out link I get a 400 error, "Bad Request: internal/identity: revoke not implemented by oidc. 

I'd like to be able to provide the user a method to sign out their user, but currently this means they will see a web error message. It also will not sign them out of their keycloak login.
Keycloak provides an end_session_endpoint in the /.well-known/openid-configuration, which will sign out the user, so it would be good if pomerium could callback into that endpoint to logout the user. 

I'm not sure how the token revocation works with the other OIDC providers but being able to customise a callback would be useful. How does pomerium determine if the revoke option is available?

Lumexralph:
## Summary

We will attempt to get the identity provider's possible revocation endpoint or end_seesion_endpont from their `/.well-known/openid-configuration`, 
For example:
Google: https://accounts.google.com/.well-known/openid-configuration
Azure: https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration

If any other provider does not support these endpoints or revocation method:
* we clear the session 
* log the error if the configuration is in debug mode
* continue the flow of the system rather than return an error.
## Related issues

Closes #501

**Checklist**:
- [X] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [X] ready for review


internal/identity/errors.go
```diff
@@ -0,0 +1,16 @@
...
+var ErrRevokeNotImplemented = errors.New("revoke not implemented")
+
+// ProviderError records an error and the message and provider involved.
+type ProviderError struct {
+	Provider string
+	Source   string
+	Err      error
+}
+
+func (e ProviderError) Error() string { return e.Source + " " + e.Provider + ": " + e.Err.Error() }
```
desimone, line 16: What if e.Err is nil?

internal/identity/providers.go
```diff
@@ -188,6 +191,47 @@ func (p *Provider) IdentityFromToken(ctx context.Context, t *oauth2.Token) (*oid
...
 func (p *Provider) Revoke(ctx context.Context, token *oauth2.Token) error {
-	return fmt.Errorf("internal/identity: revoke not implemented by %s", p.ProviderName)
+	var response struct {
+		RevocationURL string `json:"revocation_endpoint"`
+		EndSessionURL string `json:"end_session_endpoint"`
+	}
+
+	// try to see if the provider has a way to revoke token
+	wellKnownURL := p.ProviderURL + "/.well-known/openid-configuration"
+	err := httputil.Client(ctx, http.MethodGet, wellKnownURL, version.UserAgent(), nil, nil, &response)
```
desimone, line 210: Instead of doing an additional call to the well known endpoint, we should already have this information when we do 
```
	// build api client to make group membership api calls
	if err := p.provider.Claims(&gp); err != nil {
		return nil, err
	}
``` 
```go
// Claims unmarshals raw fields returned by the server during discovery.

//    var claims struct {
//        ScopesSupported []string `json:"scopes_supported"`
//        ClaimsSupported []string `json:"claims_supported"`
//    }

//    if err := provider.Claims(&claims); err != nil {
//        // handle unmarshaling error
//    }

// For a list of fields defined by the OpenID Connect spec see:
// https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata
```
on provider instantiation. Though the claim name (as you have here) will vary by provider, I don't think it makes sense to re-query the configuration endpoint for every revoke request.

We should be able to pass your response struct to the oidc `Provider.Claims` method to return the desired endpoint.

internal/identity/okta.go
```diff
@@ -74,8 +74,6 @@ func NewOktaProvider(p *Provider) (*OktaProvider, error) {
 // https://developer.okta.com/docs/api/resources/oidc#revoke
 func (p *OktaProvider) Revoke(ctx context.Context, token *oauth2.Token) error {
 	params := url.Values{}
-	params.Add("client_id", p.ClientID)
```
desimone, line 77: Yeah, I did a bit more digging, and the RFC for revocation (though not part of OIDC standard) should not contain CLIENTID /Secret

https://tools.ietf.org/html/rfc7009

internal/identity/errors.go
Lumexralph: If the error is nil, then we'll have a nil pointer error, a segmentation violation runtime error, I honestly didn't think anyone would want to use the `ProviderError` without actually wrapping an error in there.

To avoid this, I can do it this way 
`func (e ProviderError) Error() string { return fmt.Sprintf("%s %s: %v", e.Source, e.Provider, e.Err) }`.

What do you think?

internal/identity/providers.go
Lumexralph: That's right, thanks for the feedback, I will work on it.

internal/identity/okta.go
Lumexralph: I read the resource, nothing concrete about the implementation but I hope most identity providers will handle it in the generic way we're implementing it.

config/policy.go
```diff
@@ -59,7 +59,7 @@ type Policy struct {
 	TLSCustomCAFile string         `mapstructure:"tls_custom_ca_file" yaml:"tls_custom_ca_file,omitempty"`
 	RootCAs         *x509.CertPool `yaml:",omitempty"`
 
-	// Contains the x.509 client certificate to to present to the downstream
+	// Contains the x.509 client certificate to present to the downstream
```
desimone, line 62: 👀

internal/identity/providers.go
```diff
@@ -150,7 +155,7 @@ func (p *Provider) Authenticate(ctx context.Context, code string) (*sessions.Sta
 	return s, nil
 }
 
-// Refresh renews a user's session using an oidc refresh token withoutreprompting the user.
+// Refresh renews a user's session using an oidc refresh token without reprompting the user.
```
desimone, line 158: good catch

internal/identity/providers.go
```diff
@@ -107,6 +110,15 @@ func (p *Provider) GetSignInURL(state string) string {
 	return p.oauth.AuthCodeURL(state, oauth2.AccessTypeOffline)
 }
 
+// check if provider has info endpoint, try to hit that and gather more info
+// especially useful if initial request did not an contain email, or subject
+// https://openid.net/specs/openid-connect-core-1_0.html#UserInfo
+var claims struct {
```
desimone, line 116: Can we avoid using global state to store the returned claims? Though we _should_ only be setting this once on initialization, I want to avoid any hidden state issues.

Lumexralph: Yeah, I also had this in mind but I wanted to know what you think about the current implementation.

I'll work on it, thanks.

authenticate/handlers.go
```diff
@@ -233,6 +233,7 @@ func (a *Authenticate) SignOut(w http.ResponseWriter, r *http.Request) error {
 	} else if err != nil {
 		return httputil.NewError(http.StatusBadRequest, err)
 	}
+
```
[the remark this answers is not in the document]
Lumexralph: Must have been when I was trying something out.

Lumexralph:
> @Lumexralph
> 
> What do you think about the following approach that could help avoid having multiple fields (`revocationURL`/`endSessionURL`) representing the same thing?
> 
> You could create a custom _type_ that wraps `string` (`revocationString`?) and have that type implement a custom json unmarshaller that check for either of the revocation json tags?
> 
> That way we could avoid having two fields that contain the same intended value, and wouldn't need the switch in Revoke. You will probably also have to implement `MarshalJSON()` too.

It looks like something nice to try out..

authenticate/handlers.go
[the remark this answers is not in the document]
desimone: 😉

internal/identity/providers.go
```diff
@@ -192,6 +195,58 @@ func (p *Provider) IdentityFromToken(ctx context.Context, t *oauth2.Token) (*oid
...
+	if err := json.Unmarshal(b, &t); err != nil {
+		return err
+	}
+	if t.EndSessionEndpoint != "" {
+		t.RevocationURL = t.EndSessionEndpoint
+	}
+	if t.RevocationEndpoint != "" {
+		t.RevocationURL = t.RevocationEndpoint
+	}
+	*p = *(*Provider)(t.Alias)
```
calebdoxsey, line 250: This is a really nice solution!

Is this particular line necessary? I think it updates the original fields via the pointer.

Also could we add a unit test to make sure `end_session_endpoint` and `revocation_endpoint` get handled properly? Seems like this might be an edge case that could easily be dropped accidentally in the future.

internal/identity/providers.go
```diff
@@ -192,6 +195,58 @@ func (p *Provider) IdentityFromToken(ctx context.Context, t *oauth2.Token) (*oid
...
+// on the supplied interface, and in our case, it will be the Provider.
+// An extra step is made to create this implementation because the endpoint
+// to revoke a user's access token can be revocation_endpoint or end_session_endpoint.
+// We want to be able to have only one field and not repeat any of the two possible
+// endpoints on the Provider field.
+func (p *Provider) UnmarshalJSON(b []byte) error {
+	type Alias Provider
+	t := &struct {
+		*Alias
+		EndSessionEndpoint string `json:"end_session_endpoint,omitempty"`
```
desimone, line 236: I took another look at this. It looks like while "revocation_endpoint" is the standard rfc7009 endpoint,
"end_session_endpoint" is something entirely different and unique to azure (... in very ... azure fashion).

This endpoint is not another alias for remotely revoking a azure session (... has it always been this way? I feel like it used to be a revocation endpoint?) but instead is simply a url which can be passed the enduser and redirect them to in order to remove their session and log out. As far as I can find, there is no way to actually revoke a user's session remotely for azure.

internal/identity/okta.go
desimone: Actually, this is required. https://developer.okta.com/docs/reference/api/oidc/#client-secret

internal/identity/providers.go
desimone: See https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#send-a-sign-out-request

Lumexralph: @desimone We just have to implement the revocation from https://openid.net/specs/openid-connect-frontchannel-1_0.html, we talked about on slack.

internal/identity/github.go
```diff
@@ -223,3 +217,33 @@ func (p *GitHubProvider) userInfo(ctx context.Context, at string, s *sessions.St
...
 	return nil
 }
+
+// Revoke method will remove all the github grants the user
+// gave pomerium application during authorization.
+//
+// https://developer.github.com/v3/apps/oauth_applications/#delete-an-app-authorization
+func (p *GitHubProvider) Revoke(ctx context.Context, token *oauth2.Token) error {
+	// build the basic authentication request
+	basicAuth := url.UserPassword(p.ClientID, p.ClientSecret)
```
desimone, line 227: Interesting, I've always used https://golang.org/pkg/net/http/#Request.SetBasicAuth

authenticate/handlers.go
```diff
@@ -226,18 +226,34 @@ func (a *Authenticate) SignOut(w http.ResponseWriter, r *http.Request) error {
...
+	redirectURL, err := urlutil.ParseAndValidateURL(r.FormValue(urlutil.QueryRedirectURI))
+	if err != nil {
+		return httputil.NewError(http.StatusBadRequest, err)
+	}
+
 	a.sessionStore.ClearSession(w, r)
 	err = a.provider.Revoke(r.Context(), s.AccessToken)
 	if errors.Is(err, identity.ErrRevokeNotImplemented) {
 		log.FromRequest(r).Warn().Err(err).Msg("authenticate: revoke not implemented")
+	} else if errors.Is(err, identity.ErrNoRevokeWithEndSessionURL) {
```
desimone, line 238: The control flow is rather awkward here.

internal/identity/providers.go
```diff
@@ -144,13 +157,13 @@ func (p *Provider) Authenticate(ctx context.Context, code string) (*sessions.Sta
 	if p.UserGroupFn != nil {
 		s.Groups, err = p.UserGroupFn(ctx, s)
 		if err != nil {
-			return nil, fmt.Errorf("internal/identity: could not retrieve groups %w", err)
+			fmt.Println(err)
```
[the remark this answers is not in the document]
Lumexralph: Oops! nice catch

authenticate/handlers.go
Lumexralph: Thanks for the fix in the PR, it was helpful.

PR #598, opened 2020-04-10
deployment: Improve docker mutable tag publishing

travisgroth:
## Summary
Update docker image publishing to:

- publish a `master` image on pushes to master
- use alpine for `master` image to allow for easier debugging
- publish `latest` images only on when the release tag is the latest version, to skip cherry picked releases to older versions

**Checklist**:
- [x] ready for review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ COPY . .
 RUN make
 RUN touch /config.yaml
 
-FROM gcr.io/distroless/static
+FROM gcr.io/distroless/base:debug
 WORKDIR /pomerium
 COPY --from=build /go/src/github.com/pomerium/pomerium/bin/* /bin/
 COPY --from=build /config.yaml /pomerium/config.yaml
--- a/.github/goreleaser.yaml
+++ b/.github/goreleaser.yaml
@@ -93,7 +93,6 @@ snapshot:
 dockers:
   - image_templates:
       - "pomerium/pomerium:{{ .Tag }}"
-      - "pomerium/pomerium:latest"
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -10,7 +10,8 @@ COPY . .
 RUN make
 RUN touch /config.yaml
 
-FROM gcr.io/distroless/static
+FROM alpine:latest
```
desimone, line 13: Thoughts on using the debug image of distroless instead to be consistent? (or should we consider switching everything back to alpine?) https://github.com/GoogleContainerTools/distroless#debug-images

travisgroth: Oh, yes.  That's a better idea.

travisgroth: Actually...there doesn't appear to be a debug image for static 🤨 

Alpine is very small surface area but distroless is certainly better from a security perspective.  I'm okay with the split here.  I'm going to save this comment so that I can hate myself in three months when we change them all to alpine 🙃

desimone: Totally okay with the split.  But if you wanted to avoid it we could run `base-debug which adds a few extra clibs we don't need.

> Most other applications (and Go apps that require libc/cgo) should start with gcr.io/distroless/base, which contains all of the packages in gcr.io/distroless/static, and glibc, libssl, openssl

see: https://github.com/GoogleContainerTools/distroless/tree/master/base#documentation-for-gcriodistrolessbase-and-gcriodistrolessstatic

travisgroth: Ah that makes sense.  Didn't dig enough.  Will change to that.

PR #600, opened 2020-04-11
forward-auth: validate using forwarded uri header

### LINKED ISSUE #451: forward auth: extract host and path from headers
Your current design for forward auth requires that the client create a URL specialized for the authorization request that includes the host and path.  

Instead, why not simply grab those values from the HTTP headers `:authority` and `:path`? 

Then your solution would work with Envoy based proxies that simply forward auth requests to the auth service.

One backward compatible way to do this is to use the headers if there is no `uri` query parameter.

------

Separately, I am working on performing the URL rewrite with the Gloo API Gateway.  Gloo provides
a VirtualService abstraction that can route based on host and headers, among other things. It also provides a way to use the Envoy extauth feature.  

The model is simple.  One enables external authorization on a VirtualService. One identifies a Kubernetes service to send requests to.  Gloo forwards the original request to the named Kubernetes service.

With the current Pomerium implementation of forward auth, this does not quite work.  I need to rewrite the URL to add the host and path to the query.

With Gloo, one can rewrite a URL and append headers.  So, what I do is add a header to the authorization request.  I deploy a second Gloo virtual service for each service to be authenticated that matches requests with the additional header.  When the virtual service matches, it rewrites the URL as needed by Pomerium, i.e. it changes the DNS name to `externalauth.domain.example` and appends the proper query parameter `?uri=...`.

branchmispredictor:
## Summary
Services using Pomerium forward-auth need to specify the `uri` that Pomerium should check authorization for. Previous invocations require this `uri` to be hardcoded in that service's config, for example in [examples/traefik/docker-compose.yaml](https://github.com/pomerium/examples/blob/de57bba733c2f669a3e5e5f6dbccb0575654d514/traefik/docker-compose.yaml#L28).

This PR adds an option to derive the `uri` from http headers that traefik automatically adds to the authorization request ([source](https://github.com/containous/traefik/blob/aa21351d0d70546d35634ae33858a93d156766f3/pkg/middlewares/auth/forward.go#L184-L206)). 

This new feature is gated behind a config option and defaults to off. ~It also requires a specific query param in the forward-auth request.~

## Updates to [examples/traefik](https://github.com/pomerium/examples/tree/master/traefik)

`docker-compose.yaml`:
```diff
      - "traefik.http.middlewares.pomerium.forwardauth.authResponseHeaders=X-Pomerium-Authenticated-User-Email,x-pomerium-authenticated-user-id,x-pomerium-authenticated-user-groups,x-pomerium-jwt-assertion"
-      - "traefik.http.middlewares.pomerium.forwardauth.address=https://fwdauth.x.y.z/?uri=https://httpbin.x.y.z"
-      - "traefik.http.middlewares.pomerium.forwardauth.trustForwardHeader=true"
+      - "traefik.http.middlewares.pomerium.forwardauth.address=https://fwdauth.x.y.z/"
```

`config.yaml`:
```diff
forward_auth_url: https://fwdauth.x.y.z
+ forward_auth_uri_from_headers: true
```

## Related issues 
Potentially fixes #451 

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md (needed?)
- [x] ready for review


docs/configuration/readme.md
```diff
@@ -420,11 +420,25 @@ Each unit work is called a Span in a trace. Spans include metadata about the wor
...
+- Config File Key: `forward_auth_uri_from_headers`
+- Type: `bool`
+- Resulting Verification URL: `https://forwardauth.corp.example.com/?uriFromHeaders`
+- Example HTTP Headers For URL:
+  - `X-Forwarded-Proto`: `https`
+  - `X-Forwarded-Host`: `service.corp.example.com:8080`
+- Optional
+- Default: `false`
+
+Some third-party proxies include the url-to-verify in HTTP headers `X-Forwarded-Proto` and `X-Forwarded-Host`. This optional tells Pomerium to use the values from those headers to determine the uri to verify. Note that the verification URL must include the parameter `?uriFromHeaders`.
```
travisgroth, line 440: `optional` -> `option`

branchmispredictor:
> Style question - `uriFromHeaders` looks more like a property of the proxy than something we should pass to both `registerFwdAuthHandlers` and `Verify` methods on every invocation. Thoughts?
> 
> General question - It looks like this is specific to "not nginx". Do we even need the `uriFromHeaders` query param for those proxies? It seems like we could get away with just hitting https://forwardauth.x.y.z/ and read the headers if we don't get a `uri` parameter.

I'd prefer if the query param wasn't there either, but didn't know if we would have enough mutual exclusivity on the handlers, or if it would make errors more difficult to debug down the road.

branchmispredictor:
Thanks for the update @desimone, the changes look good to me!

proxy/forward_auth.go
```diff
@@ -47,7 +68,9 @@ func (p *Proxy) nginxCallback(w http.ResponseWriter, r *http.Request) error {
 	return nil
 }
 
-func (p *Proxy) traefikCallback(w http.ResponseWriter, r *http.Request) error {
+// forwardedURIHeaderCallback handles the post-authentication callback from
+// forwarding proxies that support the `X-Forwarded-Uri` header per rfc7239.
```
calebdoxsey, line 72: rfc7239 seems to specify a standard `Forwarded` header which has all the `X-Forwarded-*` headers merged together:

```
Forwarded: for=192.0.2.60;proto=http;by=203.0.113.43
```

I wasn't able to find a spec detailing the `X-` style headers. Maybe we should note they're non-standard?

proxy/forward_auth.go
```diff
@@ -75,10 +98,23 @@ func (p *Proxy) Verify(verifyOnly bool) http.Handler {
 		if status := r.FormValue("auth_status"); status == fmt.Sprint(http.StatusForbidden) {
 			return httputil.NewError(http.StatusForbidden, errors.New(http.StatusText(http.StatusForbidden)))
 		}
-		uri, err := urlutil.ParseAndValidateURL(r.FormValue("uri"))
+
+		// the route to validate will either be pull from the uri queryparam or
```
calebdoxsey, line 102: nit: `pull` should be `pulled`

proxy/forward_auth.go
```diff
@@ -11,32 +11,53 @@ import (
...
 	r.StrictSlash(true)
 	r.Use(sessions.RetrieveSession(p.sessionStore))
 
+	// NGNIX's forward-auth capabilities are split across two settings:
+	// `auth-url` and `auth-signin` which correspond to `verify` and `auth-url`
+	//
+	// NOTE: Route order matters here. As such, it's rather confusing to follow
+	// 		 so each step of the process is labeled in order of the flow
+
+	// nginx 3: save the returned session post authenticate flow
```
calebdoxsey, line 29: What's the numbering after `nginx ` mean?

proxy/forward_auth.go
desimone: You are exactly right, I had this backwards. Looks like x-* style headers are actually deprecated but still wildly used. 

https://tools.ietf.org/html/rfc6648

https://stackoverflow.com/a/3561399

proxy/forward_auth.go
desimone: The relative order of the steps in the flow to (hopefully?) clarify the request flow. I've adjusted the above comment. Hope that helps.

proxy/forward_auth.go
desimone: good catch

desimone:
\cc @branchmispredictor @travisgroth @calebdoxsey tested this against nginx and should be ready for final review. I also added a few overrides for URI to hopefully support #615 path based routing.

desimone:
Thank you again @branchmispredictor for taking the lead on this.

branchmispredictor:
No problem, thanks for the edits and getting this over the finish line @desimone.

PR #611, opened 2020-04-14
*: use sharedKey's base64 decoded bytes

desimone:
- pomerium-cli: remove formatting from CLI so we can use stdout directly.
- pomerium-cli: remove additional json print

## Summary

Previously, it looks like all of pomerium's services were using the _actual_ bytes of the base64 encoded key, NOT the underlying encoded bytes.

 I would make the case this should be cherry picked and released as part of another minor. I think prior to v0.6.0 this worked using the encoded bytes. 

These changes also add documentation for the newly included cli, and simplify the output so that it could potentially be embedded in a script.

**Checklist**:
- [x] updated docs
- [x] updated UPGRADING.md
- [x] ready for review


authenticate/authenticate.go
```diff
@@ -102,14 +102,15 @@ func New(opts config.Options) (*Authenticate, error) {
...
 	sharedCipher, _ := cryptutil.NewAEADCipherFromBase64(opts.SharedKey)
-	signedEncoder, err := jws.NewHS256Signer([]byte(opts.SharedKey), opts.AuthenticateURL.Host)
+	decodedSharedSecret, _ := base64.StdEncoding.DecodeString(opts.SharedKey)
+	signedEncoder, err := jws.NewHS256Signer(decodedSharedSecret, opts.AuthenticateURL.Host)
 	if err != nil {
 		return nil, err
 	}
 
-	// private state encoder setup
+	// cookie secret is used encrypt, and sign authenticate service payloads
```
Lumexralph, line 113: you missed a `to`  `cookie secret is used to encrypt,...`

docs/docs/reference/impersonation.md
```diff
@@ -5,7 +5,7 @@ description: >-
   user to impersonate another user or group.
 ---
 
-# User impersonation
+# User impersonation & Service Accounts
```
Lumexralph, line 9: tiny suggestion: `User Impersonation & Service Accounts` to be consistent.

docs/docs/reference/impersonation.md
```diff
@@ -32,12 +38,133 @@ In certain circumstances, it's useful for an administrative user to impersonate
...
 Here's what it looks like.
 
 <video width="100%" height="600" controls=""><source src="./img/pomerium-user-impersonation.mp4" type="video/mp4">
 Your browser does not support the video tag.
 </video>
 
+### Using the command line interface
+
+Pomerium also includes a command line interface (cli) for generating arbitrary route-scoped service account sessions. Generated service accounts can be used to impersonate users, perform service-to-service communication, and facilitate end-to-end testing for applications managed by Pomerium. The cli is especially useful in situations where an administrator needs more control over the sessions she generates, or if skipping the authentication portion of pomerium's flow is required.
+
```
Lumexralph, line 50: Nice! Enjoyed reading this.

docs/docs/reference/impersonation.md
```diff
@@ -32,12 +38,133 @@ In certain circumstances, it's useful for an administrative user to impersonate
...
+### How
+
+```bash
+pomerium-sa generates a pomerium service account from a shared key.
+
```
Lumexralph, line 55: thought: do you mean `pomerium-cli` here or that's intended?

docs/docs/reference/impersonation.md
```diff
@@ -32,12 +38,133 @@ In certain circumstances, it's useful for an administrative user to impersonate
...
+  -iss string
+        Issuing Server (e.g authenticate.int.pomerium.io)
+  -sub string
+        Subject (typically User's GUID)
+  -user string
+        User (typically User's GUID)
+```
+
+For example, say for example I wanted to generate a service account a session with the following attributes:
+
```
Lumexralph, line 86: tiny suggestion: what do you think about improving the readability here... `For example, I want to generate a service account session with the following attributes:` or anything you think can improve the readability.

authenticate/authenticate.go
desimone: yes thank you good catch

docs/docs/reference/impersonation.md
desimone: thx!

docs/docs/reference/impersonation.md
desimone: Yes, good catch. Thank you.

docs/docs/reference/impersonation.md
desimone: I've removed the extraneous example text. hopefully it's more clear now.

docs/docs/reference/impersonation.md
desimone: 👍

proxy/proxy.go
```diff
@@ -100,11 +100,10 @@ func New(opts config.Options) (*Proxy, error) {
...
 	}
 
+	// shared secret is used to encrypt and sign data shared between services
 	sharedCipher, _ := cryptutil.NewAEADCipherFromBase64(opts.SharedKey)
-	decodedCookieSecret, _ := base64.StdEncoding.DecodeString(opts.CookieSecret)
-
-	// used to load and verify JWT tokens signed by the authenticate service
-	encoder, err := jws.NewHS256Signer([]byte(opts.SharedKey), opts.AuthenticateURL.Host)
+	decodedSharedSecret, _ := base64.StdEncoding.DecodeString(opts.SharedKey)
+	encoder, err := jws.NewHS256Signer(decodedSharedSecret, opts.AuthenticateURL.Host)
```
desimone, line 106: nota bene: where i goofed

desimone:
> Upgrading to this will invalidate any existing user sessions, right?

Yes, but this should _only_ affect route sessions so it should just result in the route session being cleared, and then redirected back to authN, then back to the route with a new signed cookie. I'll test, but I think it should be pretty not-noticeable to the end user. (Similar flow to when a user has a route session for site `a.y.z` and not `b.y.z`.

desimone:
accidental closure

Issue #607, opened 2020-04-13
Path based policy failed to apply, give invalid value for DNS-1123 subdomain error


psychomelet:
Policy:
```
- from: https://test.yogorobot.com/f
  to: http://test.svc.cluster.local/f
  cors_allow_preflight: true
  allow_public_unauthenticated_access: true
```

Error

> Error: UPGRADE FAILED: cannot patch "yogo-pomerium" with kind Ingress: Ingress.extensions "yogo-pomerium" is invalid: [spec.rules[4].host: Invalid value: "test.yogorobot.com/f": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.tls[7].hosts: Invalid value: "test.yogorobot.com/f": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')]

branchmispredictor:
As far as I can tell, Pomerium does not support path-based policies. See #378 and #591. In your usecase, can you just use the following (since both from and too end with `/f`)?

```yaml
- from: https://test.yogorobot.com/
  to: http://test.svc.cluster.local/
  cors_allow_preflight: true
  allow_public_unauthenticated_access: true
```

branchmispredictor:
@psychomelet, PR #615 to implement path-based route matching was just merged. Does this fix your usecase (provided you update your config to use the path-based routing)? You can use the `pomerium/pomerium:master` docker image to test it.

psychomelet:
it works great! thanks

desimone:
Great -- thanks @psychomelet @branchmispredictor !

Fixed with #615

Issue #591, opened 2020-04-06
Enhanced routing capabilites


sylr:
**Is your feature request related to a problem? Please describe.**

Pomerium policies are currently `hostname` based only. It would be really helpful if we could define policies based on several other properties such as HTTP method, path ... etc.

**Describe the solution you'd like**

[Traefik](https://github.com/containous/traefik) implements a syntax based on [vulcand/predicate](https://github.com/vulcand/predicate) which allows to describe [rules](https://docs.traefik.io/routing/routers/#rule) that will be matched against incoming requests. If pomerium could have something similar it would allow us to use pomerium to authorize specific endpoints to a subsets of groups or users.

A pomerium policy using traefik rule language would look like this:

```yaml
policy:
- rule: 
    expression: "Host(`myapp.domain.com`) && Method(`POST`) && PathPrefix(`/api`)"
    # if not defined priority defaults to the length of the expression string
    priority: 1000
  to: http://myapp.namespace.svc.cluster.local
  allowed_domains:
  - company.com
  allowed_groups:
  - admins
  - moderators
- rule: 
    expression: "Host(`myapp.domain.com`)"
    priority: 1
  to: http://myapp.namespace.svc.cluster.local
  allowed_domains:
  - company.com
  allowed_groups:
  - employees
```
 
**Describe alternatives you've considered**

I currently have a nginx between pomerium and my application which does something similar but it is not going to work starting with v0.7.0 as the `x-pomerium-authenticated-user-groups` header has been removed.

```
location /api/ {
            if ($request_method != POST) {
                break;
            }
            set $trusted 0;
            if ($http_x_pomerium_authenticated_user_groups ~ ".*(admins|moderator).*") {
                set $trusted 1;
            }
            if ($trusted = 0) {
                return 403;
            }
            ...
}
```

**Additional context**

https://github.com/containous/traefik/blob/v2.2.0/pkg/rules/rules.go
https://github.com/vulcand/route

desimone:
I believe this is _partially_ resolved as we now support [path, regex, and prefix](https://pomerium.com/reference/#path) based routing capabilities as of #615.

We do not support different routes for method.

@sylr let me know if you disagree!

Issue #378, opened 2019-11-08
authorize: Allow path-based authentication


cjwalton:
**Is your feature request related to a problem? Please describe.**

I would like to be able to whitelist certain paths for a resource, or in the reverse, only protect other paths. 

Examples might be:

```
- from: https://www.myblog.example 
   to: https://www.myblog.example
   allow_public_unauthenticated_access: true
- from: https://www.myblog.example/admin
   to: https://www.myblog.example/admin
   allowed_domains: example.com
```

**Describe the solution you'd like**

See above. There may be a better way of expressing this, but that's off the top of my head. 

**Describe alternatives you've considered**
Using pomerium in forward-auth mode, as documented here: https://www.pomerium.io/docs/reference/reference.html#forward-auth

**Explain any additional use-cases**

I simply have several applications that in general need to be protected by an authenticating proxy, but also require certain URLs to be whitelisted for public access. Examples include Jira and Confluence, Kolide Fleet

victornoel:
In the context of kubernetes and an ingress controller that is in front of pomerium, I have some idea to support this use case without touching pomerium.

I haven't yet tried it, but what I am planning to do is to have ingresses resources responsible of proxying directly to the app for the desired subpath, and to pomerium for the rest of them.

It will depend on the ingress implementation though, for example nginx can match regex so it would work: https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/

I think contour also supports them for the `Ingress` object only.

xeor:
Did it work as expected @victornoel ?

victornoel:
> Did it work as expected @victornoel ?

@xeor yes! I matched the `/` path to the pomerium proxy service and the un-protected paths (`/live` in my case, for health probes) directly to the real service.

The only gotcha is that ingresses can only refer to services of the same namespace, so in my case I had to introduce an extra service in the same namespace as the ingress with an external name (https://kubernetes.io/docs/concepts/services-networking/service/#externalname) pointing to my real service in a secondary namespace:
```
pomerium-ns/ingress -> /      -> pomerium-ns/pomerium-proxy     -> app-ns/app-service
                    -> /live  -> pomerium-ns/fake-service       -> app-ns/app-service
```

Maybe there are other ways to handle that last point though...

desimone:
Looking at adding this to next release.

abuzhynsky:
@desimone 
Thanks for the awesome project.
[...]

PR #615, opened 2020-04-16
implement path-based route matching

calebdoxsey:
## Summary
This set of changes adds support for matching routes by path in addition to host. The following 3 fields have been added:

* `prefix`: which matches a URL path based on its prefix (`path: /admin/` matches `/admin/example`)
* `path`: which matches a URL path exactly (`path: /admin` matches `/admin`, but not `/admin/example`)
* `regex`: which matches a URL path via a regular expression. The regular expression syntax is the same as Go's [regexp package](https://pkg.go.dev/regexp?tab=doc).

More restrictive routes should be placed before less restrictive ones, since they're tried in order and the first one to match will go through. ie:

```yaml
policy:
  - from: http://from.example.com
    to: http://to.example.com
    path: /admin
    allowed_domains: ["example.com"]

  - from: http://from.example.com
    to: http://to.example.com
```

Both the proxy routing code and the authorization code have been updated to account for these new fields. 

## Related issues
- https://github.com/pomerium/pomerium/issues/378
- https://github.com/pomerium/pomerium/issues/591
- https://github.com/pomerium/pomerium/issues/607

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
@desimone here's what I have so far. I plan to add a few more tests, enable the OPA tests in CI and update the documentation, but thought it would be worthwhile to get some feedback, since we may want to change field names, etc.

authorize/evaluator/opa/policy/authz.rego
```diff
@@ -53,12 +53,67 @@ allow {
...
+}
+
+parse_url(str) = { "scheme": scheme, "host": host, "path": path } {
+	[_, scheme, host, rawpath] = regex.find_all_string_submatch_n(
+		`(?:(http[s]?)://)?([^/]+)([^?#]*)`,
+		str, 1)[0]
+	path = normalize_url_path(rawpath)
+}
+
+normalize_url_path(str) = "/" {
```
desimone, line 110: 🤯

proxy/proxy_test.go
```diff
@@ -276,3 +276,73 @@ func TestNewReverseProxy(t *testing.T) {
 		t.Errorf("got body %q; expected %q", g, e)
 	}
 }
+
+func TestRouteMatcherFuncFromPolicy(t *testing.T) {
```
desimone, line 280: 😍

authorize/evaluator/opa/policy/authz.rego
```diff
@@ -53,12 +53,67 @@ allow {
...
+
+allowed_route_prefix(input_url_obj, policy) {
+	object.get(policy, "prefix", "") == ""
+}
+allowed_route_prefix(input_url_obj, policy) {
+	object.get(policy, "prefix", "") != ""
+	startswith(input_url_obj.path, policy.prefix)
+}
+
+allowed_route_path(input_url_obj, policy) {
```
desimone, line 87: does this handle the trailing slash?

proxy/proxy.go
```diff
@@ -323,3 +325,42 @@ func (p *Proxy) roundTripperFromPolicy(policy *config.Policy) http.RoundTripper
 func (p *Proxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	p.Handler.ServeHTTP(w, r)
 }
+
+// routeMatcherFuncFromPolicy returns a mux matcher function which compares an http request with a policy.
+//
+// Routes can be filtered by the `source`, `prefix`, `path` and `regex` fields in the policy config.
+func routeMatcherFuncFromPolicy(policy config.Policy) mux.MatcherFunc {
```
desimone, line 332: really clever function, I like it.

proxy/proxy_test.go
desimone: Ahh, maybe add a test for trailing slash on a naked host?

calebdoxsey: I added a test with a trailing slash and it worked.

authorize/evaluator/opa/policy/authz.rego
calebdoxsey: I added a few more test cases to check. Seems to work as expected.

calebdoxsey:
@desimone I turned on the OPA tests in CI, updated the configuration documentation, added a more thorough test of routing via prefix, and updated the `upgrading.md` file. 

I think this PR should be ready for a full review for merging at your earliest convenience.

branchmispredictor:
> Also, it's unclear to me if there's even a standard way get the forwarded path/uri from a fronting proxy. For instance, I know traefik sets X-Forwarded-Uri (if you [have insecure mode set?](https://community.containo.us/t/forwardauth-does-not-pass-x-forwarded-uri/3277/2) but I don't recall if nginx does, or what other proxies do.

For traefik, it looks like forward auth will set `X-Forwarded-Uri` to the request uri as long as `trustForwardHeader` is not set? [source](https://github.com/containous/traefik/blob/aa21351d0d70546d35634ae33858a93d156766f3/pkg/middlewares/auth/forward.go#L208-L216).

desimone:
To circle back on this somewhat, I've found that for `nginx-ingress` the original url is passed as ` X-Original-URL`. What I'm less clear about is whether that's unique to all nginx deployments, or just [ingress-nginx](https://github.com/kubernetes/ingress-nginx/blob/d18fa90cfde32a53166193e5066056aac7267a2b/rootfs/etc/nginx/template/nginx.tmpl#L949). 

Screen Shot 2020-04-18 at 9 53 10 PM

calebdoxsey:
@desimone I rebased the PR to pull in the latest master changes. Forward auth is working with nginx by passing the URI explicitly:

```yaml
'nginx.ingress.kubernetes.io/auth-url': 'https://forward-authenticate.localhost.pomerium.io/verify?uri=$scheme://$host$request_uri',
'nginx.ingress.kubernetes.io/auth-signin': 'https://forward-authenticate.localhost.pomerium.io/?uri=$scheme://$host$request_uri',
```

The code now handles that case via https://github.com/pomerium/pomerium/pull/615/files#diff-fc1035c92a7c3cbf9faaceb72312de13R118.

I will try testing with traefik tomorrow.

proxy/forward_auth.go
```diff
@@ -143,3 +141,10 @@ func (p *Proxy) Verify(verifyOnly bool) http.Handler {
 		return nil
 	})
 }
+
+func (p *Proxy) getOriginalRequest(r *http.Request, originalURL *url.URL) *http.Request {
```
desimone, line 145: Could you point me in the direction of where you first saw this copy-request pattern. I need to get in the habit of this. Would like to better understand when and where it makes sense.

authorize/evaluator/opa/policy/authz.rego
```diff
@@ -8,57 +8,120 @@ default allow = false
 # allow by email
 allow {
 	some route
-	input.host = route_policies[route].source
+	allowed_route(input.url, route_policies[route])
 	token.payload.email = route_policies[route].allowed_users[_]
 	token.valid
 	count(deny)==0
+	trace(sprintf("allow by email (route=%v email=%v)", [route, token.payload.email]))
```
desimone, line 15: figure you planned on removing the trace statements, but just in case fyi

proxy/forward_auth.go
calebdoxsey: It's probably not necessary... but I saw it with the reverse proxy code. https://github.com/golang/go/issues/20068

beorn-:
any date for a 0.7.6 release ? :)

desimone:
@beorn- We are shooting to release 0.8.0 on Monday.

fightforlife:
I just updated my installation to the latest master branch, (higher than 0.9.1)
Till now my policy looked like this:

```
# personal API
  - from: https://api.domain.com
    to: http://192.168.1.2:9002/api
    allow_websockets: true
```

But this is no longer working, since the path in the outgoing url is no longer respected? As far as I undesternd the 3 options in this change only allow to match incoming urls, not outgoing. Is that right?

PR #624, opened 2020-04-20
authorize: fix authorization check for allowed_domains to only match current route

calebdoxsey:
## Summary
I discovered this issue when working on https://github.com/pomerium/pomerium/pull/615. 

I believe the unbound `route` variable in `allowed_user_domain` was matching any route, so if the email domain was found in any of the rules, it would pass authorization. But it should've only been looking at the `allowed_domains` list of the current route.

- [x] ready for review


calebdoxsey:
Looking at the spec: [Variable Keys](https://www.openpolicyagent.org/docs/latest/policy-language/#variable-keys)

> If the variables are unused outside the reference, we prefer to replace them with an underscore (_) character. 

So an unbound variable is the same as using `_`.

Issue #623, opened 2020-04-20
Forward authentication does not account for policy routes with public access


calebdoxsey:
## What happened?
If you set `allow_public_unauthenticated_access` on a policy route and use forward authentication with that route, the proxy will always return a `301`.

## What did you expect to happen?
For unauthenticated access the proxy should return a `200`.

## Code Details
To fix this issue we need to add additional code to this block in `forward_auth.go`:

```go
			if verifyOnly {
				return httputil.NewError(http.StatusUnauthorized, err)
			}
```

Before returning the `401` the code needs to check if the policy allows unauthenticated access.

With the way the code is currently structured we rely on the mux router to determine which policy route should be used for a URL. The mux router has a `Match` function which we can leverage. It fills a `RouteMatch` struct, which contains a `Handler` field. If we create our own `Handler` type we could add a field to the `Handler` to store the policy.

Currently we simply create a reverse proxy from the std httputil package. We can easily wrap this and put it in our own httputil package:

```go
type PolicyRouteHandler struct {
  *httputil.ReverseProxy
  Policy *config.Policy
}
```

and then we just use it like:

```go
if prh, ok := match.Handler.(*httputil.PolicyRouteHandler); ok && prh.Policy.AllowPublicUnauthenticatedAccess {
// return 200
}
```

desimone:
> Currently we simply create a reverse proxy from the std httputil package. We can easily wrap this and put it in our own httputil package

+1 from me.

PR #639, opened 2020-04-28
authorize: add support for .pomerium and unauthenticated routes

calebdoxsey:
## Summary
These changes should allow for public unauthenticated routes when using forward authentication. 

### OPA Updates
Previously we would check every route in the policy and if any of them matched and all the conditions passed, we would allow access. But due to the addition of path matching, we should only check the first matching route. For example, with these rules:

```yaml
- from: https://from.example.com
  prefix: /admin
  allowed_users: ["admin@example.com"]
- from: https://from.example.com
```

Authorize should return false for a non-admin user trying to reach `https://from.example.com/admin`.

I suppose an alternative implementation would be to require *all* matching routes to pass the allow check, and not just the first one.

This wasn't broken before (in this way) because `AllowPublicUnauthenticatedAccess` in the OPA wasn't implemented.

### Other Changes
- The proxy will now do the authorize check in the forward-auth code even if there is not a valid session. This should allow access to public routes.
- We will now pass invalid JWT sessions to the authorize check and rely on the check being done there.
- I added a log message in the authenticate service to see the decisions its making. This should be roughly equivalent to logging http access requests.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/623

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review


authorize/grpc.go
```diff
@@ -25,7 +26,19 @@ func (a *Authorize) IsAuthorized(ctx context.Context, in *authorize.IsAuthorized
 		RemoteAddr: in.GetRequestRemoteAddr(),
 		URL:        getFullURL(in.GetRequestUrl(), in.GetRequestHost()),
 	}
-	return a.pe.IsAuthorized(ctx, req)
+	reply, err := a.pe.IsAuthorized(ctx, req)
+	log.Info().
+		// request
```
desimone, line 31: I wish gRPC preserved `ctx` because it would be awesome to log the associated request id here. Maybe we should add it to gRPC directly (in a follow up PR)?

We should look into whatever magic opencensus is doing to preserve their traceid between rpc services.

calebdoxsey: It's certainly possible to add extra data to the context with `SetValue`. gRPC will pass it through. I can do that in a subsequent PR.

Issue #192, opened 2019-06-23
User unauthenticated using Azure Active Directory 


sbilly:
**Describe the bug**

failed with login with my account

**To Reproduce** Steps to reproduce the behavior:

1. run pomerium: `bin/pomerium -config config.yaml`
2.  browserling : https://external-httpbin.int.xxxx.com/ and has been redirect to : `https://login.microsoftonline.com/ ...`
3. chose account and accept access the account information
4. redirect to : `https://authenticate.int.xxxx.com/oauth2/callback?....` with error: ` identity/microsoft: could not verify id_token identity/microsoft: could not retrieve groups Unauthorized  Request [masked]`

**Expected behavior**

Authenticated with my email which is registered on azure.com.

**Environment:**

- Pomerium version (retrieve with `pomerium --version`): v0.0.5+7191ed6
- Server Operating System/Architecture/Cloud: Ubuntu 16.04

Configuration file(s):

```yaml
service: "all" # optional, default is all
pomerium_debug: true # optional, default is false
log_level: debug # optional, default is debug


authenticate_service_url: https://authenticate.int.xxxx.com/
# authorize_service_url: https://authorize.int.xxxx.com/

certificate_authority_file: "./cert.pem"

shared_secret: [masked]
cookie_secret: [masked]

idp_provider: "azure"
idp_provider_url: "https://login.microsoftonline.com/[masked]/v2.0"
idp_client_id: "[masked]"
idp_client_secret: "[masked]"

policy:
  # https://external-httpbin.int.xxxx.com/
  - from: external-httpbin.int.xxxx.com
    to: httpbin.org
    allowed_domains:
      - xxxx.com
```

Logs(s):

```text
2:58PM DBG read configuration with checksum config-checksum=f22b739e4a2fec2f
2:58PM INF cmd/pomerium version=v0.0.5+7191ed6
2:58PM DBG add domain domain=xxxx.com route=external-httpbin.int.xxxx.com
2:58PM INF proxy: new route dst=httpbin.org src=external-httpbin.int.xxxx.com
2:58PM DBG proxy/clients: using a custom certificate authority
2:58PM DBG proxy/clients: grpc connection addr=authenticate.int.xxxx.com:443 cert-override-name=
2:58PM DBG proxy/clients: using a custom certificate authority
2:58PM DBG proxy/clients: grpc connection addr=authorize.int.xxxx.com:443 cert-override-name=
[...]

PR #644, opened 2020-05-02
cryptutil: add automatic certificate management

### LINKED ISSUE #193: internal/httputil: support OCSP stapling

I'm using Pomerium as a reverse proxy to my DNS over HTTPS server. Depending on the client implementation, the client trying to check certificate status over OCSP may result in deadlock (this is mentioned in the RFC https://tools.ietf.org/html/rfc8484#section-10).

OCSP stapling is considered the solution here.

Pomerium should add OCSP stapling response by default.

Use Nginx between user and Pomerium as reverse proxy.

I believe this is also good for user privacy. OCSP must respond in plaintext HTTP and thus ISP can learn what website you are accessing. This is mentioned in Chromium's Security FAQ: https://dev.chromium.org/Home/chromium-security/security-faq#TOC-What-s-the-story-with-certificate-revocation-

**Additional context**

https://github.com/mholt/certmagic/blob/master/ocsp.go is an existing implementation of OCSP stapling which relies on golang.org/x/crypto/ocsp.

### LINKED ISSUE #643: automatic configuration of certificates using Lets Encrypt

Pomerium is very domain-name centric. This typically means that operators have to provide either their own:

 - wild card certificate, or
-  separate SNI certs for each domain

Ideally, pomerium could automatically get and maintain public certificates for all managed routes (and authenticate) by leveraging lets encrypt. 

**Additional context**

### LINKED ISSUE #651: support loading multiple certificates at once

Currently, it is possible to load certificates either as a base64 encoded blob, or as a mounted file. This works fine for most services that typically only need to serve a single hostname (like authenticate), but this is problematic if an operator wants to run proxy server without using:

- A third party load balancer or reverse proxy to manage public certificates (I'm thinking things like traefik or kubernetes with certmanger) or the various cloud options.
- Using a wild card certificate (which is what most of the quick start guides use).

It should be possible place `n` certificates in shared folder and add them to the TLS configuration used by pomerium. Go is pretty clever and then should be able to sniff the incoming TLS request for the certificates SNI and pair it with correct response cert.

Certmagic has a pretty reasonable approach to this using the file system with locks

```sh
.pomerium/certs/certificates
└── acme-v02.api.letsencrypt.org-directory
    ├── authenticate.imac.bdd.io
    │   ├── authenticate.imac.bdd.io.crt
    │   ├── authenticate.imac.bdd.io.json
    │   └── authenticate.imac.bdd.io.key
    ├── httpbin.imac.bdd.io
    │   ├── httpbin.imac.bdd.io.crt
    │   ├── httpbin.imac.bdd.io.json
    │   └── httpbin.imac.bdd.io.key
    ├── httpbin122.imac.bdd.io
    │   ├── httpbin122.imac.bdd.io.crt
    │   ├── httpbin122.imac.bdd.io.json
    │   └── httpbin122.imac.bdd.io.key
    ├── httpbin12321.imac.bdd.io
    │   ├── httpbin12321.imac.bdd.io.crt
    │   ├── httpbin12321.imac.bdd.io.json
    │   └── httpbin12321.imac.bdd.io.key
    └── httpbin69.imac.bdd.io
        ├── httpbin69.imac.bdd.io.crt
        ├── httpbin69.imac.bdd.io.json
        └── httpbin69.imac.bdd.io.key
``` 

This probably isn't a problem for most users in production as they tend to use a third party load balancer to handle (and terminate) their public certificates. But it would be useful for PoC and small deployments, and removes the burden of having to generate a wildcard certificate (which typically requires DNS access). 

**Additional context**

Automatic certificate enrollment (#643 ) using lets encrypt should handle the vast majority of scenarios, but we should have a fallback mechanism to allow users to specify their own keys.

desimone:

## Summary

Previously, we instructed users to [retrieve a wildcard certificate](https://0-6-0.docs.pomerium.io/docs/reference/certificates.html#how) to use pomerium even in our get started guide in order to support HTTPS across multiple pomerium managed routes and the authenticate service.

While wildcard certs work fine, there are a large ask on our part, especially for a PoC/ home user. Specifically 

- they require DNS API access or ability to edit TXT records on a domain. 
- wildcard domains cannot be automatically updated without DNS api access.
- even if updated, pomerium had no mechanism of auto-updating that certificate without a restart
- you cannot revoke just a single site's cert

These changes add the ability for pomerium to do automatic per-route HTTPS certificate retrieval using LetsEncrypt.  Some benefits:

- automatic retrieval, storage, and updating of public certificates on a per route basis.
- supports hot reloading (with a small delay, for the http-01/ALPN flow to complete) on policy change
- OCSP support (#192)
- does not require any DNS or DNS-api tinkering

Some things to be aware of. 

- proxy/authenticate services must share file storage (by default `./certs/`) 
- user's are implicitly agreeing to LetsEncrypt's terms of services.
- does not support multiple SANs. This can be problematic if you have _many_ routes. In the future, we may support adding DNS support, but at that point it feels like most users will have delegated this task to whatever production cert management system they are using at their load balancer. 

Overall, these changes should result in much easier onboarding for folks that do not already have a cloud load balancer that manages certificates, or a dedicated system like [cert-manager](https://github.com/jetstack/cert-manager).

### GIF || GTFO

out

## Related issues

- fixes #193 
- fixes #643 
- fixes #651 

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- N/A updated UPGRADING.md 
- [x] ready for review


go.mod
```diff
@@ -3,12 +3,13 @@ module github.com/pomerium/pomerium
 go 1.12
 
 require (
-	cloud.google.com/go v0.49.0 // indirect
```
desimone, line 6: 🤷‍♂️

spencergilbert:
:open_mouth: This is super awesome @desimone

docs/configuration/examples/config/config.minimal.env
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 # See : https://www.pomerium.io/docs/reference/certificates
-export CERTIFICATE_FILE="$HOME/.acme.sh/*.corp.beyondperimeter.com_ecc/fullchain.cer"                      # optional, defaults to `./cert.pem`
-export CERTIFICATE_KEY_FILE="$HOME/.acme.sh/*.corp.beyondperimeter.com_ecc/*.corp.beyondperimeter.com.key" # optional, defaults to `./certprivkey.pem`
+export AUTOCERT=TRUE                # optional, defaults to `./cert.pem`
```
calebdoxsey, line 4: These comments need to be updated

proxy/proxy.go
```diff
@@ -218,7 +218,7 @@ func (p *Proxy) reverseProxyHandler(r *mux.Router, policy config.Policy) *mux.Ro
 	// 1. Create the reverse proxy connection
 	proxy := stdhttputil.NewSingleHostReverseProxy(policy.Destination)
 	// 2. Create a sublogger to handle any error logs
-	sublogger := log.With().Str("route", policy.Destination.Host).Logger()
+	sublogger := log.With().Str("route", policy.String()).Logger()
```
desimone, line 221: 👀 contains both source/destination

config/options.go
```diff
@@ -237,6 +245,7 @@ var defaultOptions = Options{
 	GRPCServerMaxConnectionAgeGrace: 5 * time.Minute,
 	CacheStore:                      "autocache",
 	AuthenticateCallbackPath:        "/oauth2/callback",
+	CertFolder:                      filepath.Join(".", ".pomerium"),
```
desimone, line 248: ❓ Should we have a default value for cert folder?

cmd/pomerium/main.go
```diff
@@ -247,8 +251,20 @@ func setupTracing(opt *config.Options) error {
 
 func setupHTTPRedirectServer(opt *config.Options, wg *sync.WaitGroup) error {
 	if opt.HTTPRedirectAddr != "" {
-		serverOpts := httputil.ServerOptions{Addr: opt.HTTPRedirectAddr}
-		srv, err := httputil.NewServer(&serverOpts, httputil.RedirectHandler(), wg)
+		serverOpts := httputil.ServerOptions{
+			Addr:              opt.HTTPRedirectAddr,
+			Insecure:          true,
+			Service:           "HTTP->HTTPS Redirect",
+			ReadHeaderTimeout: 5 * time.Second,
```
desimone, line 258: Changed default timeouts to be stricter given this endpoint will be used exclusively for redirects and http acme challenges.

internal/cryptutil/tls.go
```diff
@@ -0,0 +1,105 @@
...
+// https://letsencrypt.org/docs/challenge-types/#tls-alpn-01
+func NewAutocert(tlsConfig *tls.Config, hostnames []string, path string) (*tls.Config, func(h http.Handler) http.Handler, error) {
+	certmagic.DefaultACME.Agreed = true
+	certmagic.DefaultACME.CA = effectiveCA
+	cm := certmagic.NewDefault()
+
+	tlsConfig = newTLSConfigIfEmpty(tlsConfig)
+	// add existing certs to the cache, and staple OCSP
+	for _, cert := range tlsConfig.Certificates {
+		if err := cm.CacheUnmanagedTLSCertificate(cert, nil); err != nil {
```
desimone, line 32: autocert is called last so that any existing cert can be added to autocert's cache.

desimone:
\cc @calebdoxsey @travisgroth  re-requesting a review. User's can now add multiple certificate files (#651), http-01 challenges are now supported, and now operators can mix and match certificate settings (and so long as they point to valid certs, they should all be additive).

desimone:
@calebdoxsey 

> What's the behavior if the Let's Encrypt process doesn't work? (i.e. if :443 isn't accessible from the internet)

A `http-01` challenge will be attempted. If both fail, Pomerium will continue to run (`cm.ManageAsync`) but pomerium will fail to serve the end user with a certificate and you will see something like the following in your logs.

```logs
7:40PM ERR http: TLS handshake error from 192.168.1.1:55164: no certificate available for 'httpbin.imac.bdd.io' addr=:443 insecure=false service=all
```

config/options.go
```diff
@@ -252,7 +274,7 @@ func NewDefaultOptions() *Options {
 func NewOptionsFromConfig(configFile string) (*Options, error) {
 	o, err := optionsFromViper(configFile)
 	if err != nil {
-		return nil, fmt.Errorf("config: options from viper %w", err)
+		return nil, fmt.Errorf("config: options from config file %w", err)
```
desimone, line 277: @travisgroth fyi, snuck this change in. It's very much a me problem but when I'd see this in the logs I would assume the error originated from viper, and not us.

desimone:
\cc @yegle as this seems to give us backdoor OCSP stapling support via certmagic if used with `autocert: true`. #193

yegle:
This is awesome! Yeah I think certmagic is better than my half baked OCSP stapling support.

I think it's safe to close #193 as obsolete.

config/options.go
travisgroth: 👍

config/options.go
travisgroth: `.` might not be very predictable.

If we provide a default, it should be relative to `/etc`, `$HOME` or from https://github.com/adrg/xdg like `XDG_CACHE_HOME`.  The latter is nice in that it is cross platform and idiomatic to each platform.

config/options.go
```diff
@@ -57,16 +59,29 @@ type Options struct {
...
+	// and renewal from LetsEncrypt. Must be used in conjunction with CertFolder.
+	AutoCert bool `mapstructure:"autocert" yaml:"autocert,omitempty"`
+
+	// AutoCertHandler is the HTTP challenge handler used in a http-01 acme
+	// https://letsencrypt.org/docs/challenge-types/#http-01-challenge
+	AutoCertHandler func(h http.Handler) http.Handler `hash:"ignore"`
+
+	// CertFolder specifies the location to store (if using autocert), and load
+	// your TLS certificates.
+	CertFolder string `mapstructure:"certificate_folder" yaml:"certificate_folder,omitempty"`
```
travisgroth, line 72: We should make the parameter name clear.  Eg, `autocert_storage_dir` or something along those lines.

docs/configuration/readme.md
```diff
@@ -127,23 +127,63 @@ Pomerium should _never_ be exposed to the internet without TLS encryption.
...
+- Optional
+
+Turning on autocert allows Pomerium to automatically retrieve, manage, and renew public facing TLS certificates from [Let's Encrypt][letsencrypt] for each of your managed pomerium routes as well as for the authenticate service. This setting must be used in conjunction with `Certificate Folder` as Autocert must have a place to persist, and share certificate data between services. Provides [OCSP stapling](https://en.wikipedia.org/wiki/OCSP_stapling).
+
+This setting can be useful in a situation where you do not have Pomerium behind a TLS terminating ingress or proxy that is already handling your public certificates on your behalf.
 
-Certificate is the x509 _public-key_ used to establish secure HTTP and gRPC connections.
+:::warning
+
+By using autocert, you agree to the [Let's Encrypt Subscriber Agreement](https://letsencrypt.org/documents/LE-SA-v1.2-November-15-2017.pdf). There are [_strict_ usage limits](https://letsencrypt.org/docs/rate-limits/) per domain you should be aware of.
```
travisgroth, line 143: We might want to give the option of using the staging LE CA.

docs/configuration/readme.md
```diff
@@ -127,23 +127,63 @@ Pomerium should _never_ be exposed to the internet without TLS encryption.
...
-### Certificate Key
+Autocert requires that port `443` be accessible from the internet in order to complete a [TLS-ALPN-01 challenge](https://letsencrypt.org/docs/challenge-types/#tls-alpn-01).
 
-- Environmental Variable: either `CERTIFICATE_KEY` or `CERTIFICATE_KEY_FILE`
-- Config File Key: `certificate_key` or `certificate_key_file`
+:::
+
+### Certificate Folder
+
+- Environmental Variable: either `CERT_FOLDER`
```
travisgroth, line 155: See other comment in about parameter name

internal/grpc/server.go
```diff
@@ -33,17 +34,21 @@ func NewServer(opt *ServerOptions, registrationFn func(s *grpc.Server), wg *sync
...
-	if opt.TLSCertificate != nil {
-		log.Debug().Str("addr", opt.Addr).Msg("internal/grpc: serving over TLS")
-		cert := credentials.NewServerTLSFromCert(opt.TLSCertificate)
+	if len(opt.TLSCertificate) == 1 {
+		cert := credentials.NewServerTLSFromCert(&opt.TLSCertificate[0])
 		grpcOpts = append(grpcOpts, grpc.Creds(cert))
-	} else {
-		log.Warn().Str("addr", opt.Addr).Msg("internal/grpc: serving without TLS")
+	} else if !opt.InsecureServer {
+		return nil, errors.New("internal/grpc: unexpected number of certificates")
```
travisgroth, line 41: Should we retain the insecure warning?

proxy/proxy.go
travisgroth: 👍

config/options.go
desimone: @travisgroth good idea. What do you think about https://github.com/pomerium/pomerium/pull/644/commits/b391d01326b5aace6e47d45239f09315ddfed412

travisgroth: Given this:

>By default, CertMagic stores assets on the local file system in $HOME/.local/share/certmagic (and honors $XDG_DATA_HOME if set). CertMagic will create the directory if it does not exist. If writes are denied, things will not be happy, so make sure CertMagic can write to it!

Proposal: Let's use their default (which seems rational - and I think is approximately what https://github.com/pomerium/pomerium/commit/b391d01326b5aace6e47d45239f09315ddfed412 does?) and update our Docker images to set `$XDG_DATA_HOME` to `/data/`.  This leaves a nice `docker run -it -v pomerium:/data/pomerium ... pomerium/pomerium` pattern for container execution with a volume for storing certs. 

Thoughts?

desimone: Good thinking. It also means we can do away with the explicit certs directory.

travisgroth: Clarity - use the approximation of their pattern so that we store in `$XDG_DATA_HOME/pomerium`.

config/options.go
desimone: changed to `autocert_folder`

docs/configuration/readme.md
desimone: 👍

internal/grpc/server.go
desimone: I'm cool with it just being part of the Info log on the following line. Thoughts?

docs/configuration/readme.md
desimone: Good idea.

config/options.go
```diff
@@ -499,21 +530,59 @@ func (o *Options) Validate() error {
...
 
+func (o *Options) sourceHostnames() []string {
+	if len(o.Policies) == 0 {
+		return nil
+	}
+	var h []string
+	for _, p := range o.Policies {
+		h = append(h, p.Source.Hostname())
+	}
+	if o.AuthenticateURL != nil {
```
calebdoxsey, line 580: do we need to add forward auth too?

config/options.go
```diff
@@ -564,3 +633,30 @@ func HandleConfigUpdate(configFile string, opt *Options, services []OptionsUpdat
 	}
 	return newOpt
 }
+
+// homeDir returns the best guess of the current user's home
+// directory from environment variables. If unknown, "." (the
+// current directory) is returned instead.
+func homeDir() string {
```
calebdoxsey, line 640: you can use https://github.com/golang/go/blob/go1.14.2/src/os/file.go#L469

config/options.go
desimone: yes, good catch!

config/options.go
desimone: Oh very neat!

config/options.go
desimone: actually, on second thought, I think this is best left as an exercise to the user since in any forward auth configuration I'd expect the operator would have some other system managing public certificates like cert-manager, and terminating TLS at that fronting proxy. Thoughts @calebdoxsey ?

config/options.go
```diff
@@ -57,16 +60,36 @@ type Options struct {
...
+	AutoCert bool `mapstructure:"autocert" yaml:"autocert,omitempty"`
+
+	// AutoCertHandler is the HTTP challenge handler used in a http-01 acme
+	// https://letsencrypt.org/docs/challenge-types/#http-01-challenge
+	AutoCertHandler func(h http.Handler) http.Handler `hash:"ignore"`
+
+	// AutoCertFolder specifies the location to store, and load autocert managed
+	// TLS certificates.
+	// defaults to $XDG_DATA_HOME/pomerium
+	AutoCertFolder string `mapstructure:"autocert_folder" yaml:"autocert_folder,omitempty"`
```
travisgroth, line 74: nit: `directory` or `dir` rather than `folder`.  The latter doesn't fit in this context.

config/options.go
calebdoxsey: :+1:

config/options.go
desimone: agree. thanks and good catch.

PR #674, opened 2020-05-08
envoy: implement refresh session

calebdoxsey:
## Summary
This implements session refresh for envoy. Instead of doing the refresh in the proxy, we will now do it in the authorize service. 

Once refreshed we return the new JWT as a **request** x-pomerium-set-cookie header. Using a custom lua script, we remove that header before sending it upstream, and then put it back as a **response** set-cookie header.

This also adds a test to make sure the session is being refreshed properly. To get this to work I added a `token_expiration` input in the reference oidc provider, so that we could make it a very short-lived token. We then poll for a few seconds till we get a new session, and if we don't see that happen, we fail.

## Related issues
Closes pomerium/internal#10

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [x] ready for review


integration/internal/flows/flows.go
```diff
@@ -68,8 +111,11 @@ func Authenticate(ctx context.Context, client *http.Client, url *url.URL, email
 		forms := forms.Parse(res.Body)
 		if len(forms) > 0 {
 			f := forms[0]
-			f.Inputs["email"] = email
-			f.Inputs["groups"] = strings.Join(groups, ",")
+			f.Inputs["email"] = cfg.email
+			if len(cfg.groups) > 0 {
+				f.Inputs["groups"] = strings.Join(cfg.groups, ",")
+			}
+			f.Inputs["token_expiration"] = strconv.Itoa(int(cfg.tokenExpiration.Seconds()))
```
desimone, line 118: 😳

PR #684, opened 2020-05-11
envoy: implement header and query param session loading

calebdoxsey:
## Summary
This adds additional support for loading sessions from query params and headers in authorize. I also refactored the code to break out the session saving/loading code.

## Related issues
- pomerium/internal#13
- pomerium/internal#14

**Checklist**:
- [x] add related issues
- [x] ready for review


authorize/session.go
```diff
@@ -0,0 +1,71 @@
...
+
+func loadSession(req *http.Request, options config.Options, encoder encoding.MarshalUnmarshaler) ([]byte, error) {
+	var loaders []sessions.SessionLoader
+	cookieStore, err := getCookieStore(options, encoder)
+	if err != nil {
+		return nil, err
+	}
+	loaders = append(loaders,
+		cookieStore,
+		header.NewStore(encoder, "Pomerium"),
```
desimone, line 26: Should add this to `internal/httputil/headers.go` thoughts?

authorize/session.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	cookieStore, err := cookie.NewStore(cookieOptions, encoder)
+	if err != nil {
+		return nil, err
+	}
+	return cookieStore, nil
+}
+
+func getJWTSetCookieHeaders(cookieStore sessions.SessionStore, rawjwt []byte) (map[string]string, error) {
+	recorder := httptest.NewRecorder()
```
desimone, line 58: If the recorder gets flushed, could this fail?

internal/urlutil/query_params.go
```diff
@@ -11,6 +11,7 @@ const (
 	QueryIsProgrammatic    = "pomerium_programmatic"
 	QueryForwardAuth       = "pomerium_forward_auth"
 	QueryPomeriumJWT       = "pomerium_jwt"
+	QuerySession           = "pomerium_session"
```
desimone, line 14: 👍

authorize/session.go
calebdoxsey: Looking at the docs I should've used `recorder.Result().Header()` which should return the right headers even if the response was never written or flushed. I updated it.

PR #694, opened 2020-05-12
Return an error regardless of envoy's exit status

travisgroth:
## Summary
Ensure we always return an error when envoy exits, so that we exit cleanly from CTRL+C, etc.

NB: I do not believe this will catch an out of band termination like sending SIGTERM to envoy.  Will follow up on that.

**Checklist**:
- [x] ready for review

### DIFF
--- a/internal/envoy/envoy.go
+++ b/internal/envoy/envoy.go
@@ -4,6 +4,7 @@ package envoy
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


internal/envoy/envoy.go
```diff
@@ -83,7 +84,11 @@ func (srv *Server) Run(ctx context.Context) error {
 
 	// make sure envoy is killed if we're killed
 	srv.cmd.SysProcAttr = sysProcAttr
-	return srv.cmd.Run()
+	err = srv.cmd.Run()
+	if err == nil {
+		return errors.New("envoy exited without error")
```
travisgroth, line 89: This signals the errorgroup in `cmd` that we've exited.

PR #695, opened 2020-05-13
envoy: support autocert

calebdoxsey:
## Summary
This PR ports the autocert changes to envoy. There were several major changes:

1. The redirect HTTP server was moved to the config package and renamed. This was done so that we can restart it if configuration options change and so that we can attach the autocert handler. We probably need to revisit this at some point as the naming doesn't seem quite right, but it should at least be a little easier to follow.
1. I created an `AutocertManager` which handles provisioning Let's Encrypt certificates for hostnames. It does this synchronously as soon as configuration options are changed so that the certificates are available for the envoy configuration. This does mean on first startup there will be a lag before the envoy is listening to requests, but we don't have the same sort of TLS `ClientHello` listening capabilities as we did with the certmagic handler, so this seemed like the easiest approach.
1. I added a function in `cryptutil` to get a certificate based on its name. It will first try matching via domain/wildcard, then fall back to the first cert available, and finally generate a self-signed certificate if nothing could be found. It wasn't clear what the best strategy would be in that case, as disabling TLS would confuse the browser (since its still listening on :443). I suppose we could crash or stop listening entirely.

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [x] ready for review


calebdoxsey:
Seems to be some issue with the integration tests. I will investigate.

internal/cryptutil/tls.go
```diff
@@ -101,3 +102,46 @@ func newTLSConfigIfEmpty(tlsConfig *tls.Config) *tls.Config {
 		NextProtos: []string{"h2", "http/1.1"},
 	}
 }
+
+// GetCertificateForDomain returns the tls Certificate which matches the given domain name. It should handle both
+// exact matches and wildcard matches. If no certificates match the first one will be used. If there are no certificates
+// one will be generated.
```
travisgroth, line 108: nit: comment is a little off

travisgroth:
>finally generate a self-signed certificate if nothing could be found.
👍

desimone:
reviewing now

internal/cryptutil/certificates.go
```diff
@@ -2,17 +2,23 @@ package cryptutil
...
 	"fmt"
 	"io/ioutil"
+	"math/big"
+	"net"
+	"time"
 )
 
-// CertifcateFromBase64 returns an X509 pair from a base64 encoded blob.
-func CertifcateFromBase64(cert, key string) (*tls.Certificate, error) {
+// CertificateFromBase64 returns an X509 pair from a base64 encoded blob.
```
desimone, line 20: 🙄

PR #697, opened 2020-05-13
Add an option to request certificate with Must-Staple.

yegle:
## Related issues

**Checklist**:
- [ ] add related issues <-- N/A
- [x] updated docs
- [x] updated unit tests  <-- N/A
- [ ] updated UPGRADING.md <-- this file doesn't exist anymore?
- [x] ready for review


desimone:
@yegle at first glance this looks great. Just as a heads up, we are in the process of merging some other significant changes and I would like to see those get merged to master first, so it might take a bit longer for us to review and merge. Thanks again!

desimone:
@yegle Sorry for the delay / all the changes. Do you mind rebasing on latest?

yegle:
The PR is rebased to master now.

yegle:
The code as is requires you to always have `autocert_` prefix for autocert related options. So you'll be looking at a config yaml file like this:

```
autocert_folder: "foo"
autocert: true
autocert_must_staple: true
```

Do you have any preference to treat autocert related options as a map? E.g. something like this:

```
autocert:
  enabled: true
  folder: "foo"
  must_staple: true
```
?

desimone:
> The code as is requires you to always have `autocert_` prefix for autocert related options. So you'll be looking at a config yaml file like this:
> 
> ```
> autocert_folder: "foo"
> autocert: true
> autocert_must_staple: true
> ```
> 
> Do you have any preference to treat autocert related options as a map? E.g. something like this:
> 
> ```
> autocert:
>   enabled: true
>   folder: "foo"
>   must_staple: true
> ```
> 
> ?

I think the challenge is if we use a map setting those configuration options as environmental variables gets tricky / non-trivial. See the header related settings, for example. . Thoughts?

yegle:
> I think the challenge is if we use a map setting those configuration options as environmental variables gets tricky / non-trivial. See the header related settings, for example. . Thoughts?

I think you can set an env key replacer like this:

```
    viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
```

And then you can set the map config value as usual. See https://play.golang.org/p/6p426N2rZ4Y

yegle:
Re https://github.com/pomerium/pomerium/pull/697#issuecomment-637931127: I have #850 ready to review. Let me know if you want me to merge #850 into this PR.

travisgroth:
@yegle as I mentioned in #850, I'd like to keep the options format for autocert consistent with the rest of our options (eg, flat structure like `autocert_must_staple` for the time being.  Using deeper yaml structures is something we should switch to in the future, but we'd want to do it across the board for all options as part of a breaking change.  It should be separate from this improvement.

Can you update this PR to just reflect the changes needed for stapling with the current configuration options style?  Then I think we can merge this into 0.9 as it would be a feature add.

yegle:
SG, this PR as-is uses flat config structure. It's ready to review now.

travisgroth:
Ah, I see.  I saw the additional Option struct was still there, and assumed it was the same code.  Sorry about that.  Will look it over again!

yegle:
Yes the additional struct is still there (figure it will be cleaner than adding more fields). By using inline/squash the configuration still uses a flat structure.

PR #707, opened 2020-05-15
integration-tests: switch to go for backends to support TLS scenarios

calebdoxsey:
## Summary
I discovered that the nginx ingress controller does some validation on supplied certs, but we actually want to test scenarios where domain names *don't* match, so I decided to switch the backends to go instead of node.js to make it easier to start a TLS server with custom certs.

There will be another PR to actually add the new tests.

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review


integration/manifests/lib/backends.libsonnet
```diff
@@ -21,11 +21,18 @@ local service = function(name) {
...
-    }],
+    ports: [
+      {
+        name: 'http',
+        port: 80,
+        targetPort: 'http',
+      },
+      {
+        name: 'https',
+        port: 80,
```
travisgroth, line 32: Should this still be `80`?

calebdoxsey: nope :)

PR #724, opened 2020-05-18
envoy: implement policy TLS options

calebdoxsey:
## Summary
This implements the TLS policy options in envoy. Envoy returns a slightly different HTTP status code, but otherwise it should behave as before.

I discovered that Envoy supports HTTP/2, but it has to be explicitly enabled. Therefore I've enabled it for our gRPC endpoints, but none of the policy routes. We will likely need to add a configuration option to support this.

I also changed the way the xDS configuration is put together. There's now a separate cluster for each policy route, because the various TLS options mean we can't always re-use a cluster based on its name. Just keeping them entirely separate seemed like the more future-proof option.

One additional change to xDS is I switched from embedding bytes directly inline to instead writing them to a file (in the cache directory) and referencing that filename. This should be more secure and should also make the configuration easier to read.

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review


internal/controlplane/xds.go
```diff
@@ -117,3 +127,95 @@ func inlineBytes(bs []byte) *envoy_config_core_v3.DataSource {
...
+	for _, scts := range cert.SignedCertificateTimestamps {
+		envoyCert.SignedCertificateTimestamp = append(envoyCert.SignedCertificateTimestamp,
+			inlineBytes(scts))
+	}
+	return envoyCert
+}
+
+func getRootCertificateAuthority() (string, error) {
+	// from https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/ssl#arch-overview-ssl-enabling-verification
+	knownRootLocations := []string{
```
desimone, line 205: 🙄

internal/controlplane/xds.go
```diff
@@ -117,3 +127,95 @@ func inlineBytes(bs []byte) *envoy_config_core_v3.DataSource {
...
+		"/etc/ssl/certs/ca-certificates.crt",
+		"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem",
+		"/etc/pki/tls/certs/ca-bundle.crt",
+		"/etc/ssl/ca-bundle.pem",
+		"/usr/local/etc/ssl/cert.pem",
+		"/etc/ssl/cert.pem",
+	}
+
+	for _, path := range knownRootLocations {
+		if _, err := os.Stat(path); err == nil {
```
desimone, line 215: Should we maybe log something to debug to let the user know what we are using as their trust store... if for whatever reason they have a wonky configuration and use multiple root locations?

internal/controlplane/xds.go
```diff
@@ -117,3 +127,95 @@ func inlineBytes(bs []byte) *envoy_config_core_v3.DataSource {
...
+
+func getPolicyName(policy *config.Policy) string {
+	return fmt.Sprintf("policy-%x", policy.Checksum())
+}
+
+func envoyTLSCertificateFromGoTLSCertificate(cert *tls.Certificate) *envoy_extensions_transport_sockets_tls_v3.TlsCertificate {
+	envoyCert := &envoy_extensions_transport_sockets_tls_v3.TlsCertificate{}
+	var chain bytes.Buffer
+	for _, cbs := range cert.Certificate {
+		_ = pem.Encode(&chain, &pem.Block{
```
desimone, line 176: safe to throw away error here?

internal/controlplane/xds.go
```diff
@@ -117,3 +127,95 @@ func inlineBytes(bs []byte) *envoy_config_core_v3.DataSource {
...
+			Type:  "CERTIFICATE",
+			Bytes: cbs,
+		})
+		break
+	}
+	envoyCert.CertificateChain = inlineBytesAsFilename("tls-crt.pem", chain.Bytes())
+	if cert.OCSPStaple != nil {
+		envoyCert.OcspStaple = inlineBytes(cert.OCSPStaple)
```
desimone, line 184: 😍

internal/controlplane/xds_clusters.go
```diff
@@ -1,70 +1,138 @@
...
+	}
 
-	return clusters
+	sni := policy.Destination.Hostname()
+	if policy.TLSServerName != "" {
+		sni = policy.TLSServerName
+	}
+	tlsContext := &envoy_extensions_transport_sockets_tls_v3.UpstreamTlsContext{
+		CommonTlsContext: &envoy_extensions_transport_sockets_tls_v3.CommonTlsContext{
+			AlpnProtocols: []string{"http/1.1"},
```
desimone, line 76: This is where we need an additional policy setting to explicitly enable http2 right?

internal/controlplane/xds.go
calebdoxsey: yeah it only errors on the `Write` call, and that shouldn't happen with a buffer.

internal/controlplane/xds_clusters.go
calebdoxsey: Yes and this too:

```
cluster.Http2ProtocolOptions = &envoy_config_core_v3.Http2ProtocolOptions{}
```

integration/internal/cluster/certs.go
```diff
@@ -43,7 +43,7 @@ func bootstrapCerts(ctx context.Context) (*TLSCertsBundle, error) {
 		name    string
 	}{
 		{&bundle.Trusted, filepath.Join(wd, "trusted"), true, "*.localhost.pomerium.io"},
-		{&bundle.WronglyNamed, filepath.Join(wd, "wrongly-named"), true, "*.localhost.notpomerium.io"},
+		{&bundle.WronglyNamed, filepath.Join(wd, "trusted"), true, "*.localhost.notpomerium.io"},
```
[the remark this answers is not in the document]
calebdoxsey: Yeah it uses the same CA from trusted, but a cert with the wrong name.

PR #750, opened 2020-05-21
envoy: Remove 'accept-encoding' header from proxied metric requests

travisgroth:
## Summary
The OpenCensus and Envoy prometheus implementations don't have parity for response encoding.  Remove encoding header to force a plain text response.

**Checklist**:
- [x] ready for review

### DIFF
--- a/internal/telemetry/metrics/providers.go
+++ b/internal/telemetry/metrics/providers.go
@@ -55,6 +55,9 @@ func newProxyMetricsHandler(promHandler http.Handler, envoyURL url.URL) http.Han
 	return func(w http.ResponseWriter, r *http.Request) {
 		defer promHandler.ServeHTTP(w, r)
 
+		// Ensure we don't get entangled with compression from ocprom
+		r.Header.Del("Accept-Encoding")
+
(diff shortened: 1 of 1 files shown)


internal/telemetry/metrics/providers.go
```diff
@@ -55,6 +55,9 @@ func newProxyMetricsHandler(promHandler http.Handler, envoyURL url.URL) http.Han
 	return func(w http.ResponseWriter, r *http.Request) {
 		defer promHandler.ServeHTTP(w, r)
 
+		// Ensure we don't get entangled with compression from ocprom
+		r.Header.Del("Accept-Encoding")
```
desimone, line 59: Weird, but 🤷‍♂️ 👌

travisgroth: Yeah...we do this, or we get into having to uncompress the response from the internal prometheus handler to add in our metrics.  It seems easier to just ensure it doesn't use compression.

Long term we should switch to accepting metrics via RPC and putting them into the local registry directly.  The current implementation is a bit of a hack and won't work with anything other than prometheus export format.

Issue #749, opened 2020-05-21
Support x-forwarded-client-cert header for mTLS


calebdoxsey:
**Is your feature request related to a problem? Please describe.**

Pomerium sometimes sits behind a TLS-terminating load balancer which enforces client certificates for mTLS. If Pomerium needs to enforce authentication/authorization via client certificates, but won't be terminating the TLS connection, we need some mechanism for the load balancer to communicate what client certificate was used for the connection.

**Describe the solution you'd like**

We should support the `x-forwarded-client-cert` (XFCC) header. The format of this header is described here: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-client-cert.

We will likely need the `Chain` field and from that we ought to be able to verify the client certificate against a certificate authority.

In general, we can't trust the `x-forwarded-client-cert` header by default, since a client could populate the header without first proving they actually established a TLS connection using it, so this functionality will need to be disabled by default and only enabled via a configuration option when we trust the information we receive.

**Describe alternatives you've considered**

The primary alternative for this use case would be L4 load balancing so that TLS termination is done by Envoy instead of the load balancer.

cuonglm:
Can we close this @calebdoxsey ?

calebdoxsey:
This issue is not resolved but if we feel like it is not needed I will close it.

desimone:
@calebdoxsey / @cuonglm Let's keep it open. 
If we want to associate device context behind TLS-terminating proxy or load balancer using client certificates, we'll need to support this header.

kenjenkins:
Related: https://github.com/pomerium/pomerium/issues/4259

PR #751, opened 2020-05-21
authorize: add client mTLS support

calebdoxsey:
## Summary
This adds support for client mTLS certificate verification. With these changes an administrator could require users to present a client-side TLS certificate to access routes.

Authorization is implemented by using Go's x509 certificate verification function and passing that as a property to the OPA rego policy. This is called with the configured certificate authority and the peer certificate that was used to establish the TLS connection.

I initially looked into using allow/deny, but that caused a redirect to the authentication flow, which doesn't really make sense for this situation (logging in as a different user won't fix your cert). So instead I added a new variable `http_status` that when set will bypass the typical flow and return the code/message/headers directly. We may want to look into using this model for the other checks too, as it may simplify the logic of the gRPC method, but for now it will only be used for unrecoverable HTTP errors.

The peer certificate is passed to the ExtAuthz by setting the `IncludePeerCertificate` property, and turning on certificate validation for the downstream TLS context. We also set `CertificateValidationContext_ACCEPT_UNTRUSTED` so that we can do the check in ExtAuthz.

Other Changes:
- I added scripts for `protoc` and `protoc-gen-go` that should use consistently versioned binaries instead of whatever happened to be installed on a users system (because it now stores the protobuf version in a comment, so I wanted to avoid it constantly changing when re-generated). I've not tested it in macos, but I think it should work.
- I updated the error handling in authorize so that instead of just returning a plaintext error message, it now returns a pomerium-formatted error page (if your `Accept` is for `text/html`). This both looks nicer and also allows the user to sign-out and switch users when they get a 403.

## Related issues
- #749 would be a future improvement to this functionality

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review


authorize/evaluator/opa/opa.go
```diff
@@ -199,3 +244,34 @@ func readPolicy(fn string) ([]byte, error) {
...
+	case uint8:
+		return int(v)
+	case json.Number:
+		i, _ := v.Int64()
+		return int(i)
+	case string:
+		i, _ := strconv.Atoi(v)
+		return i
+	default:
+		return anyToInt(fmt.Sprint(v))
```
desimone, line 275: my recursion spider senses are going off here (even though that string base case shuold always be caught). Thoughts on throwing up a few unit tests on this?

authorize/authorize.go
```diff
@@ -14,13 +16,16 @@ import (
...
 	"github.com/pomerium/pomerium/internal/encoding/jws"
+	"github.com/pomerium/pomerium/internal/frontend"
 	"github.com/pomerium/pomerium/internal/log"
 	"github.com/pomerium/pomerium/internal/telemetry/metrics"
 	"github.com/pomerium/pomerium/internal/telemetry/trace"
 
 	"gopkg.in/square/go-jose.v2"
 )
 
+//go:generate ../scripts/protoc -I ../internal/grpc/authorize/ --go_out=plugins=grpc:../internal/grpc/authorize/ ../internal/grpc/authorize/authorize.proto
```
desimone, line 27: we put the geneate line above the package comment elsewhere. I'm fine either way but we should be consistent

authorize/grpc.go
```diff
@@ -99,6 +102,14 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v2.CheckRe
 			},
 		})
 
+	if reply.GetHttpStatus().GetCode() > 0 && reply.GetHttpStatus().GetCode() != 200 {
```
desimone, line 105: we use http status codes elsewhere, might as well use it here. `http.StatusOK`

authorize/grpc.go
```diff
@@ -89,6 +89,9 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v2.CheckRe
 	evt = evt.Str("email", reply.GetEmail())
 	evt = evt.Strs("groups", reply.GetGroups())
 	evt = evt.Str("session", string(sess))
```
desimone, line 91: i think this session can be nil if during the refresh there was an error (see line 51)

cache/grpc_test.go
```diff
@@ -1,4 +1,4 @@
-//go:generate protoc -I ../internal/grpc/cache/ --go_out=plugins=grpc:../internal/grpc/cache/ ../internal/grpc/cache/cache.proto
+//go:generate ../scripts/protoc -I ../internal/grpc/cache/ --go_out=plugins=grpc:../internal/grpc/cache/ ../internal/grpc/cache/cache.proto
```
desimone, line 1: we can probably remove since already in grpc.go?

config/options.go
```diff
@@ -238,6 +238,11 @@ type Options struct {
 	// CacheStorePath is the path to use for a given cache store. e.g. /etc/bolt.db
 	CacheStorePath string `mapstructure:"cache_store_path" yaml:"cache_store_path,omitempty"`
 
+	// ClientCA is the base64-encoded certificate authority to validate client mTLS certificates against.
+	ClientCA string `mapstructure:"client_ca" yaml:"client_ca,omitempty"`
+	// ClientCAFile a file that contains the certificate authority to validate client mTLS certificates against.\
```
desimone, line 243: nit: trailing slash. 

nit: ClientCAFile "is a file" / "points to a file"

internal/controlplane/xds_listeners.go
```diff
@@ -326,11 +328,28 @@ func (srv *Server) buildDownstreamTLSContext(options *config.Options, domain str
 		return nil
 	}
 
+	var trustedCA *envoy_config_core_v3.DataSource
+	if options.ClientCA != "" {
+		bs, err := base64.StdEncoding.DecodeString(options.ClientCA)
+		if err != nil {
+			log.Warn().Msg("client_ca does not appear to be a base64 encoded string")
```
desimone, line 335: what happens if we give it junk?

internal/httputil/httputil.go
```diff
@@ -0,0 +1,5 @@
+package httputil
+
+// StatusInvalidClientCertificate is the status code returned when a client's certificate is invalid. This is the same
```
desimone, line 3: nit: super long line

internal/controlplane/xds_listeners.go
calebdoxsey: it would let everything through. I'll add something to the config validation.

PR #784, opened 2020-05-26
docs: update traefik example and add note about forwarded headers

### LINKED ISSUE #616: Document Traefik tuning for forward auth
If your are using Traefik to proxy requests to the pomerium proxy acting as forward auth component, then, it is required to configure traefik so that it allows "X-Fowarded-*" headers.

See https://docs.traefik.io/v2.0/routing/entrypoints/#forwarded-header

calebdoxsey:
## Summary
Our docker-compose example was missing the `--entrypoints.web.forwardedheaders.insecure=true` option which is apparently required to make Traefik work. This PR updates the documentation.

## Related issues
Fixes #616 

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] ready for review

### DIFF
--- a/docs/configuration/readme.md
+++ b/docs/configuration/readme.md
@@ -543,15 +543,22 @@ spec:
 
 #### Traefik docker-compose
 
+If the `forward_auth_url` is also handled by Traefik, you will need to configure Traefik to trust the `X-Forwarded-*`
+headers as described in [the documentation](https://docs.traefik.io/v2.2/routing/entrypoints/#forwarded-headers).
+
 ```yml
 version: "3"
 
 services:
   traefik:
-    # The official v2.0 Traefik docker image
-    image: traefik:v2.0
+    # The official v2.2 Traefik docker image
+    image: traefik:v2.2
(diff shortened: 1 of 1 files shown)


sylr:
I think the example is misleading because, according to the code, `forwardedHeaders.insecure=true` is a NOOP unless associated with a **non empty** `forwardedHeaders.trustedIPs` list, see:

https://github.com/containous/traefik/blob/7928e6d0cd4c7751a14bea5f74124f8a3cb829c0/pkg/middlewares/forwardedheaders/forwarded_header.go#L55

https://github.com/containous/traefik/blob/7928e6d0cd4c7751a14bea5f74124f8a3cb829c0/pkg/middlewares/forwardedheaders/forwarded_header.go#L79

calebdoxsey:
@sylr the trustedIPs list does not have to be set if insecure is true https://github.com/containous/traefik/blob/7928e6d0cd4c7751a14bea5f74124f8a3cb829c0/pkg/middlewares/forwardedheaders/forwarded_header.go#L165

sylr:
@calebdoxsey my bad. Thank you.

PR #802, opened 2020-05-29
telemetry: service label updates

travisgroth:
## Summary
Update `service` label in telemetry to be more consistent and useful in the face of both envoy and internal metrics

- Map `services` option into something that always contains `pomerium` for easier discovery and tracking
- Ensure usage of this service name mapping across tracing, metrics and envoy stats config
- Document envoy related changes

**Checklist**:
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/telemetry/util.go
+++ b/internal/telemetry/util.go
@@ -1,0 +1,15 @@
+package telemetry
+
+// ServiceName turns a pomerium service option into the appropriate external label for telemetry purposes
+//
+// Ex:
+// service 'all' -> 'pomerium'
+// service 'proxy' -> 'pomerium-proxy'
+func ServiceName(service string) string {
+	switch service {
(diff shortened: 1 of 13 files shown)


internal/telemetry/util.go
```diff
@@ -0,0 +1,15 @@
+package telemetry
+
+// ServiceName turns a pomerium service option into the appropriate external label for telemetry purposes
+//
+// Ex:
+// service 'all' -> 'pomerium'
+// service 'proxy' -> 'pomerium-proxy'
+func ServiceName(service string) string {
+	switch service {
+	case "all", "":
```
travisgroth, line 10: Can't use the constants in config/options.go without some reorganization.

Issue #821, opened 2020-06-02
Uncaught exception when IPv6 is disabled


BenoitKnecht:
## What happened?

I tried upgrading from a functional Pomerium 0.8.3 setup to 0.9.0, but the new release immediately crashes due to an uncaught exception (backtrace below).

## How'd it happen?

1. Ran `pomerium -config=/etc/pomerium/config.yaml`

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint): `0.9.0-1590940862+914b952`
- Server Operating System/Architecture/Cloud: CentOS 7 VM (kernel 3.10.0-1062.18.1.el7.x86_64)

## What's your config.yaml?

```config.yaml
address: :443
authenticate_service_url: https://auth.example.com

certificate: <redacted>
certificate_key: <redacted>

cookie_secret: <redacted>
shared_secret: <redacted>

idp_provider: oidc
idp_provider_url: https://gitlab.example.com
idp_client_id: <redacted>
idp_client_secret: <redacted>
idp_scopes: openid email

policy:
- from: https://service.example.com
  to: http://localhost:8080
```

## What did you see in the logs?

```
Jun 02 06:39:41 vm-01.example.com pomerium[20641]: {"level":"debug","service":"envoy","name":"main","time":"2020-06-02T06:39:41Z","message":"std::terminate called! (possible uncaught exception, see trace)"}
Jun 02 06:39:41 vm-01.example.com pomerium[20641]: {"level":"debug","service":"envoy","name":"backtrace","time":"2020-06-02T06:39:41Z","message":"[bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:91] Backtrace (use tools/stack_decode.py to get line numbers):"}
Jun 02 06:39:41 vm-01.example.com pomerium[20641]: {"level":"debug","service":"envoy","name":"backtrace","time":"2020-06-02T06:39:41Z","message":"[bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:92] Envoy version: 3504d40f752eb5c20bc2883053547717bcb92fd8/1.14.1/clean-getenvoy-902f20f-envoy/RELEASE/BoringSSL"}
[...]

PR #823, opened 2020-06-02
xds: use ipv4 address when ipv6 is disabled

calebdoxsey:
## Summary
When IPv6 is entirely disabled we aren't allowed to bind `::` and envoy fails out. This change should detect if ipv6 is disabled, and switch to `0.0.0.0` if that happens.

## Related issues
- #821 

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [x] ready for review


desimone:
Should this be cherrypicked onto `0-9-0`?

BenoitKnecht:
> Should this be cherrypicked onto `0-9-0`?

It would be great if it was. At the moment, I'm not able to upgrade from 0.8.x to 0.9.x because of this regression, and I don't think there's any workaround.

BenoitKnecht:
Hold on, I don't think the fix is actually working. I cherry-picked this commit on `0-9-0` and tried to run it on the VM with IPv6 disabled, and I got the exact same backtrace.

And now that I look at it, isn't this pretty much the same as setting `address: 0.0.0.0:443` in the configuration file? That didn't work either.

PR #849, opened 2020-06-08
Remove unnecessary viper.New()

yegle:
NewDefaultOptions() already set the viper field to a non-nil value.
Trivial change.


travisgroth:
Looks like you started from a stale branch.  Can you update to fix the conflicts?  Otherwise LGTM.

travisgroth:
Upon further investigation, I think this pattern is silly looking but necessary with the current structure.  https://github.com/pomerium/pomerium/pull/850#discussion_r437825194

yegle:
How about now?

travisgroth:
I was thinking that might work.  It does look like it should be okay.

PR #850, opened 2020-06-08
Autocert options as map

yegle:
## Summary

This is to convert the autocert related options into a map, so that instead of

```
autocert: true
autocert_dir: "/path"
autocert_use_staging: true
```

You can write the config like this:

```
autocert: true
autocert_options:
  dir: "/path"
  use_staging: true
```

Note: this is intended to be fully backward compatible.

## Related issues
The relevant discussion is in #697.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


yegle:
I'm getting a very weird race condition that the test will sometimes pass but most of the time it doesn't pass.

If you can help identify the issue that would be great...

config/options.go
```diff
@@ -319,26 +312,24 @@ func optionsFromViper(configFile string) (*Options, error) {
...
-		v.SetConfigFile(configFile)
-		if err := v.ReadInConfig(); err != nil {
+		o.viper.SetConfigFile(configFile)
+		if err := o.viper.ReadInConfig(); err != nil {
 			return nil, fmt.Errorf("failed to read config: %w", err)
 		}
 	}
 
-	if err := v.Unmarshal(&o); err != nil {
+	if err := o.viper.Unmarshal(o); err != nil {
```
travisgroth, line 329: I think this is actually the cause of the intermittent failure, and why a new viper instance is used, rather than the embedded one.  If viper unmarshals from inside the struct, it can overwrite itself before complete, and that can screw up the rest of the config handling.

yegle: Hmm, I reverted this part of change but it looks like the intermittent test failure is still here.

yegle:
I think I figured out where the race condition is from and workarounded the issue.

config/autocert.go
```diff
@@ -11,6 +11,31 @@ import (
...
+	// status_request extension. This will allow the TLS client (the browser)
+	// to fail immediately if Pomerium failed to get an OCSP staple.
+	// See also https://tools.ietf.org/html/rfc7633
+	// Only used when Enable is true.
+	MustStaple bool `mapstructure:"must_staple" yaml:"must_staple,omitempty"`
+
+	// Folder specifies the location to store, and load autocert managed
+	// TLS certificates.
+	// defaults to $XDG_DATA_HOME/pomerium
+	Folder string `mapstructure:"folder" yaml:"folder,omitempty"`
```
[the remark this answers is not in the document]
yegle: Reverted to `dir` here.

config/options.go
yegle: Finally figured out where the race condition came from. The test is no longer flaky.

yegle:
Per the discussion in #697, revert this PR.

PR #874, opened 2020-06-11
authenticate: change signin/signout endpoints

### LINKED ISSUE #857: Auth Url - Change path prefix

I'm going to preface with this is my first time writing in Go, so I didn't want to submit pull requests that are probably not great.  You can take a look at my fork if you want to investigate how I've resolved this issue/feature.  My environment is Pomerium/Traefik (forward auth)/Keycloak.

In forward auth mode with Traefik, I've got the external authenticate url setup to be sso.example.com.  When you go to sso.example.com/.pomerium/, you see the dashboard, as expected.  However, when you hit the Sign Out button, it tries to execute the authentication signout instead of the proxy signout since there are two routes for the same path.

No idea if I'm doing this right, but figured there shouldn't be multiple routers for the same path no matter how it's all hosted.

In my fork I've moved the auth router pah prefix to "./pomerium/auth" to resolve the conflicts and allow the dashboard to function correctly from the externally accessible authenticate domain.

Perhaps I need to create another externally accessible dns host instead of using the authenticate host directly?  Seems a waste to use two names for the same service though.

### LINKED ISSUE #867: Sign out link doesn't work on authenticate_service_url
## What happened?

Access /.pomerium/sign_out and get error:

Bad Request: internal/urlutil: malformed unix timestamp field

Sign out the user and ask to login again

## How'd it happen?

Access /.pomerium/sign_out

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint):
pomerium/0.9.0 (+https://wwww.pomerium.io; 914b952; go1.14.3)
In docker

## What's your config.yaml?

NA

NA

cuonglm:
## Summary
So it does not conflict with proxy endpoints, and make the "Sign Out"
button always works, regardless of where we render user's dashboard
page.

## Related issues
Fixes #857
Fixes #867

**Checklist**:
- [x] add related issues
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authenticate/handlers.go
+++ b/authenticate/handlers.go
@@ -69,8 +69,8 @@ func (a *Authenticate) Mount(r *mux.Router) {
 	v.Use(middleware.ValidateSignature(a.sharedKey))
 	v.Use(sessions.RetrieveSession(a.sessionLoaders...))
 	v.Use(a.VerifySession)
-	v.Path("/sign_in").Handler(httputil.HandlerFunc(a.SignIn))
-	v.Path("/sign_out").Handler(httputil.HandlerFunc(a.SignOut))
+	v.Path("/signin").Handler(httputil.HandlerFunc(a.SignIn))
+	v.Path("/signout").Handler(httputil.HandlerFunc(a.SignOut))
 	v.Path("/refresh").Handler(httputil.HandlerFunc(a.Refresh)).Methods(http.MethodGet)
 
 	wk := r.PathPrefix("/.well-known/pomerium").Subrouter()
--- a/authenticate/handlers_test.go
+++ b/authenticate/handlers_test.go
@@ -83,7 +83,7 @@ func TestAuthenticate_Handler(t *testing.T) {
 	}
 
 	// cors preflight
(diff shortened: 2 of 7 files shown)


cuonglm:
> Fair point on the self-exploring name. But, my thought was that "/.pomerium/* would always reflect proxy methods, and "/.pomerium/auth/*" would always reflect auth methods (kinda like a namespace). That way we're not mixing and matching methods available at any level and can't run into conflicts in the future. We can also keep everything named the same so that proxy->sign_out would essentially be a passthrough to auth->sign_out (makes finding the corresponding methods easier).
> 
> I can certainly submit a pull request - my fork has this implemented and it works well so far.

To clarify, my point is that whether we change the endpoint name, or move it to new namespace have the same effects, and both will work well. I'm discourage to move to new namespace because:

 - Self explore meaning like I explain above
 - It does not help preventing conflict occurs between proxy and authenticate (like it's being conflict currently)
 - Due to the fact that current bug happens because of *name conflict*, then a fix is *rename* one of them to prevent conflict sounds more meaningful for me.

Issue #833, opened 2020-06-07
JWT Audiences


simbaja:
## What happened?

The audience for the jwt isn't working as expected.  Let's use forwardauth mode as an example.  The jwt audience field appears to be constructed by using the authenticate url host, the forward auth host, and the redirect host.  So, in this case, let's assume:

Authenticate: sso.example.com
Forward Auth: pomerium.localhost
Redirect: httpbin.example.com

This works fine for the first request you make (i.e. to httpbin).  But, let's say you have another service that you would like to access (let's say whoami.example.com).  When you attempt to go to that service, it fails saying the host isn't in the audience.

## What did you expect to happen?

I'd hope to implement single signon so that the same session can apply to anything listed under the policy (at least to the extent they're part of the same domain).

## How'd it happen?

1. Went to httpbin.example.com
2. Logged in to the identity provider, received pomerium session
3. Redirected to httpbin.example.com -> access allowed
4. Navigated to whoami.example.com -> access denied

## What's your environment like?

- Pomerium 0.9.0
- Traefik 2.2.1
- Linux/Docker
- Keycloak (through OIDC)

## What's your config.yaml?

```
---

cookie_secret: {{ pomerium_cookie_secret }}
cookie_domain: {{ domain_name }}

idp_provider: "oidc"
idp_client_id: "{{ keycloak_auth_client_id }}"
idp_client_secret: "{{ keycloak_auth_client_secret }}"
idp_scopes: openid profile email offline_access web-auth
idp_provider_url: https://idp.{{ domain_name }}/auth/realms/master

log_level: info

insecure_server: false
certificate_file: /pomerium/cert.pem
certificate_key_file: /pomerium/privkey.pem
forward_auth_url: https://pomerium
authenticate_service_url: https://sso.{{ domain_name }}

#NOTE: need to map values for user and groups if using Keycloak
jwt_claims_headers:
  - user
  - email
  - name
  - given_name
  - family_name
  - groups

policy:
  - from: https://whoami.{{ domain_name }}
    to: http://whoami
    allowed_groups:
      - Administrator
  - from: https://httpbin.{{ domain_name }}
    to: http://httpbin
    allowed_groups:
      - Administrator
```

## What did you see in the logs?

```logs

```

travisgroth:
Hi,

Our usual test setup for forward auth is with nginx and multiple routes work fine as you'd expect. Sample policy:

```yaml
  policy:
      - from: https://httpbin.localhost.pomerium.io
[...]

PR #879, opened 2020-06-11
Changes related to forward auth invalid session handling

### LINKED ISSUE #858: Forward Auth - Invalid Session Issue
## What happened?

I'm going to preface with this is my first time writing in Go, so I didn't want to submit pull requests that are probably not great.  You can take a look at my fork if you want to investigate how I've resolved this issue/feature.  My environment is Pomerium/Traefik (forward auth)/Keycloak.

In the forward auth functionality in the proxy, the IsAuthorized check is a boolean value.  However, the actual check in envoy also considers session issues (expired, not yet valid, etc).  If the session drops (i.e. the user manually logs out of the identity provider and so the session can't be refreshed), the forward auth still gives you a 403 error instead of a 401 error.

I expected the forward auth to request re-authentication if there is a session issue.  Instead it just denies access.  In my fork I added a new bool return value that indicates whether authentication is needed, which clears the session and starts the auth process.

- Pomerium 0.9.0 (forked now)
- Traefik (forward auth)
- Keycloak IDP

## What's your config.yaml?

Not Applicable

Not Applicable

simbaja:
## Summary

If the session is invalid, don't return 403 in forward auth.  Instead, clear the session and allow it to flow through to the authenticate functionality.

## Related issues

Fixes #858 

**Checklist**:
- [X] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [X] ready for review

### DIFF
--- a/proxy/forward_auth.go
+++ b/proxy/forward_auth.go
@@ -119,12 +119,15 @@ func (p *Proxy) Verify(verifyOnly bool) http.Handler {
 		}
 
 		original := p.getOriginalRequest(r, uri)
-		isAuthorized, needsAuthentication, err := p.isAuthorized(w, original)
+		authorizeResponse, err := p.isAuthorized(w, original)
 		if err != nil {
 			return httputil.NewError(http.StatusBadRequest, err)
 		}
 
-		if isAuthorized {
+		//determine if authentication is needed (if authorize returns 401)
+		authenticated := authorizeResponse.statusCode != http.StatusUnauthorized
+
+		if authorizeResponse.authorized && authenticated {
 			w.Header().Set("Content-Type", "text/plain; charset=utf-8")
 			w.WriteHeader(http.StatusOK)
 			fmt.Fprintf(w, "Access to %s is allowed.", uri.Host)
@@ -132,14 +135,14 @@ func (p *Proxy) Verify(verifyOnly bool) http.Handler {
 		}
 
 		//if unauthorized (for example, session is invalid), clear the session here too
-		if needsAuthentication {
+		if !authenticated {
 			p.sessionStore.ClearSession(w, r)
 		}
 
 		//if we have a session, and we didn't return unauthorized, return access denied
 		_, err = sessions.FromContext(r.Context())
 		hasSession := err == nil
-		if hasSession && !needsAuthentication {
+		if hasSession && authenticated {
 			return httputil.NewError(http.StatusForbidden, errors.New("access denied"))
 		}
 
--- a/proxy/forward_auth.go
+++ b/proxy/forward_auth.go
@@ -119,21 +119,27 @@ func (p *Proxy) Verify(verifyOnly bool) http.Handler {
 		}
 
 		original := p.getOriginalRequest(r, uri)
-		authorized, err := p.isAuthorized(w, original)
+		isAuthorized, needsAuthentication, err := p.isAuthorized(w, original)
 		if err != nil {
 			return httputil.NewError(http.StatusBadRequest, err)
 		}
 
-		if authorized {
+		if isAuthorized {
 			w.Header().Set("Content-Type", "text/plain; charset=utf-8")
 			w.WriteHeader(http.StatusOK)
 			fmt.Fprintf(w, "Access to %s is allowed.", uri.Host)
 			return nil
 		}
 
+		//if unauthorized (for example, session is invalid), clear the session here too
+		if needsAuthentication {
+			p.sessionStore.ClearSession(w, r)
+		}
+
+		//if we have a session, and we didn't return unauthorized, return access denied
 		_, err = sessions.FromContext(r.Context())
 		hasSession := err == nil
-		if hasSession {
+		if hasSession && !needsAuthentication {
 			return httputil.NewError(http.StatusForbidden, errors.New("access denied"))
 		}
 
--- a/proxy/middleware.go
+++ b/proxy/middleware.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	envoy_service_auth_v2 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2"
-	envoy_type "github.com/envoyproxy/go-control-plane/envoy/type"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/gorilla/mux"
 	"github.com/rs/zerolog"
@@ -18,6 +17,11 @@ import (
 	"github.com/pomerium/pomerium/internal/telemetry/trace"
 	"github.com/pomerium/pomerium/internal/urlutil"
 )
+
+type authorizeResponse struct {
+	authorized bool
+	statusCode int
+}
 
 // AuthenticateSession is middleware to enforce a valid authentication
 // session state is retrieved from the users's request context.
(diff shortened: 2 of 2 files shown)


proxy/middleware.go
```diff
@@ -45,10 +46,10 @@ func (p *Proxy) redirectToSignin(w http.ResponseWriter, r *http.Request) error {
 	return nil
 }
 
-func (p *Proxy) isAuthorized(w http.ResponseWriter, r *http.Request) (bool, error) {
+func (p *Proxy) isAuthorized(w http.ResponseWriter, r *http.Request) (bool, bool, error) {
```
cuonglm, line 49: The `(bool. bool, error)` is the sign that we should refactor this function more. How about introduce new return type, like:

```
type authResponse struct {
    authorized bool
    statusCode int
}
```

Then you can check status code in the caller to decide to clear the session or not.

proxy/forward_auth.go
```diff
@@ -119,21 +119,27 @@ func (p *Proxy) Verify(verifyOnly bool) http.Handler {
...
+		//if unauthorized (for example, session is invalid), clear the session here too
+		if needsAuthentication {
+			p.sessionStore.ClearSession(w, r)
+		}
+
+		//if we have a session, and we didn't return unauthorized, return access denied
 		_, err = sessions.FromContext(r.Context())
 		hasSession := err == nil
-		if hasSession {
+		if hasSession && !needsAuthentication {
```
cuonglm, line 142: This sounds not right to me, must we also change the `verifiedOnly` condition below?

proxy/middleware.go
simbaja: I'm new to Go, my day-to-day language is C#... so not used to multiple return values.  The method you suggest (i.e. returning a struct/class) would be what I would do in C# because only one return type is available.  Stylistically (both in general and for this particular project), is it preferable to do this approach, or have multiple returns?

proxy/forward_auth.go
simbaja: We shouldn't need to change the verifyOnly condition.  Basically, it's a flow through.  If it reaches the point after the modified block, it requires re-authentication.  The verifyOnly flag just allows it to be broken into two steps versus completing within this function (for NGINX - see the registration).

proxy/middleware.go
simbaja: > Please add a test, so we can make sure the change won't introduce any regression.

From what I'm reading (correct me if I'm wrong), I don't see any tests for when the check result is not ok, so seems like a bigger issue than just this particular change.  Should these tests be added as part of this change, or should there be a separate pull to add more coverage overall?

cuonglm: > I'm new to Go, my day-to-day language is C#... so not used to multiple return values. The method you suggest (i.e. returning a struct/class) would be what I would do in C# because only one return type is available. Stylistically (both in general and for this particular project), is it preferable to do this approach, or have multiple returns?

The thing is that if others look at the method signature, the `(bool, bool, error)` make no sense what it returns. If it's only `isAuthorized(w, req) (bool, error)`, then you can see that the `bool` indicated whether we are authorized or not. While with `isAuthorized(w, req) (bool, bool, error)`, you don't know for sure which one indicated authorization. Even if you know the first one is, then what the second return value mean.

What I propose is change it to `authorized(w, req) (*authRespone, error)`, then you can do what ever you want with `*authResponse`, and make the code much readable.

simbaja: Modified per suggestion.  I did not have time to figure out how to write a unit test for it, perhaps over the next week I can try to figure that out, but not sure work will allow.

cuonglm:
@simbaja Any updates on this PR?

simbaja:
> @simbaja Any updates on this PR?

I've made the changes to the actual code per the request.  I honestly don't know where to begin to setup a test for it - it really is dependent on the authorize and proxy sessions being out of sync.  In order to test the changed code path, we'd have to setup that situation, and I don't see any examples of how to set up that scenario.

cuonglm:
> > @simbaja Any updates on this PR?
> 
> I've made the changes to the actual code per the request. I honestly don't know where to begin to setup a test for it - it really is dependent on the authorize and proxy sessions being out of sync. In order to test the changed code path, we'd have to setup that situation, and I don't see any examples of how to set up that scenario.

You can try https://github.com/pomerium/examples/tree/master/traefik to setup a local working forward auth mode. I did, and unfortunately, the PR does not help.

I try debugging, and see that the current code will behave the right way. The main problem is what #1014 describe.

simbaja:
> > > @simbaja Any updates on this PR?
> > 
> > 
> > I've made the changes to the actual code per the request. I honestly don't know where to begin to setup a test for it - it really is dependent on the authorize and proxy sessions being out of sync. In order to test the changed code path, we'd have to setup that situation, and I don't see any examples of how to set up that scenario.
> 
> You can try https://github.com/pomerium/examples/tree/master/traefik to setup a local working forward auth mode. I did, and unfortunately, the PR does not help.
> 
> I try debugging, and see that the current code will behave the right way. The main problem is what #1014 describe.

#1014 is related to this and something I experienced as well.  My original code (not the modified version) is currently deployed in my production instance and working.  When using Traefik forward auth, you must signout using the authenticate url (it's the only way that works).  Perhaps new issues with session synchronization have occurred since the version that is working on my production instance that is preventing this PR from working correctly, but ultimately the issue is with session tracking between authorize and proxy.  

Instantiating a Traefik instance to test is not the issue (that's straightforward enough, but see some commentary on the example in #833).  We shouldn't need to do that to test the code - it's a matter of setting up the right conditions that cause the new code path to be taken, which involves a fairly complex scenario that I don't think I can setup properly given my understanding of all they layers of this system.

cuonglm:
> > > > @simbaja Any updates on this PR?
> > > 
> > > 
> > > I've made the changes to the actual code per the request. I honestly don't know where to begin to setup a test for it - it really is dependent on the authorize and proxy sessions being out of sync. In order to test the changed code path, we'd have to setup that situation, and I don't see any examples of how to set up that scenario.
> > 
> > 
> > You can try https://github.com/pomerium/examples/tree/master/traefik to setup a local working forward auth mode. I did, and unfortunately, the PR does not help.
> > I try debugging, and see that the current code will behave the right way. The main problem is what #1014 describe.
> 
> #1014 is related to this and something I experienced as well. My original code (not the modified version) is currently deployed in my production instance and working. When using Traefik forward auth, you must signout using the authenticate url (it's the only way that works). Perhaps new issues with session synchronization have occurred since the version that is working on my production instance that is preventing this PR from working correctly, but ultimately the issue is with session tracking between authorize and proxy.
> 
> Instantiating a Traefik instance to test is not the issue (that's straightforward enough, but see some commentary on the example in #833). We shouldn't need to do that to test the code - it's a matter of setting up the right conditions that cause the new code path to be taken, which involves a fairly complex scenario that I don't think I can setup properly given my understanding of all they layers of this system.

To clarify, which version your current production instance use?

simbaja:
> > > > > @simbaja Any updates on this PR?
> > > > 
> > > > 
> > > > I've made the changes to the actual code per the request. I honestly don't know where to begin to setup a test for it - it really is dependent on the authorize and proxy sessions being out of sync. In order to test the changed code path, we'd have to setup that situation, and I don't see any examples of how to set up that scenario.
> > > 
> > > 
> > > You can try https://github.com/pomerium/examples/tree/master/traefik to setup a local working forward auth mode. I did, and unfortunately, the PR does not help.
> > > I try debugging, and see that the current code will behave the right way. The main problem is what #1014 describe.
> > 
> > 
> > #1014 is related to this and something I experienced as well. My original code (not the modified version) is currently deployed in my production instance and working. When using Traefik forward auth, you must signout using the authenticate url (it's the only way that works). Perhaps new issues with session synchronization have occurred since the version that is working on my production instance that is preventing this PR from working correctly, but ultimately the issue is with session tracking between authorize and proxy.
> > Instantiating a Traefik instance to test is not the issue (that's straightforward enough, but see some commentary on the example in #833). We shouldn't need to do that to test the code - it's a matter of setting up the right conditions that cause the new code path to be taken, which involves a fairly complex scenario that I don't think I can setup properly given my understanding of all they layers of this system.
> 
> To clarify, which version your current production instance use?

Looks like it was forked based on 6f938562ca0bf1a1257551c6b26ed2fe1aaf2677 (June 6th).  I have not merged in any changes since then.

cuonglm:
@simbaja See https://github.com/pomerium/pomerium/pull/1053, I think that the reason causes this fix does not work. Would you mind rebasing this PR to master, so we can see if it works?

If you don't have time, I can make a fresh new PR for this.

cuonglm:
> @simbaja See #1053, I think that the reason causes this fix does not work. Would you mind rebasing this PR to master, so we can see if it works?
> 
> If you don't have time, I can make a fresh new PR for this.

@simbaja Please see https://github.com/pomerium/pomerium/pull/1062, if you want, please rebasing and cherry pick my additional change in the PR here, so we can go with this PR instead of mine.

cuonglm:
Close in favor of #1062

PR #883, opened 2020-06-12
git: ignore additional test file types

travisgroth:
## Summary
Add a few more file types to ignore when doing local testing.

**Checklist**:
- [x] ready for review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 *.db
 .*.yml
 .*.yaml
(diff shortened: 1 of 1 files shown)


calebdoxsey:
Why are we ignoring `.rego` files?

travisgroth:
@calebdoxsey intent is to ignore them at the root of the repo in case of testing files, similar to yaml and json.

PR #891, opened 2020-06-13
internal/telemetry/metrics: document concurrently using

cuonglm:
## Summary
Document that metricRegistry is not safe for concurrently use. While at
it, remove t.Parallel() in tests which use metricRegistry, which causes
data race, caught by:

	go test -race ./internal/telemetry/metrics

**Checklist**:

- [x] ready for review


travisgroth:
Ah yes.  It looks like the underlying registry is not concurrently safe.  Makes sense.

PR #903, opened 2020-06-15
config: add PassIdentityHeaders option

### LINKED ISSUE #702: Disable user identity headers

I'd like to be able to disable the `X-Pomerium-Jwt-Assertion` and `X-Pomerium-Claim-*` headers on a per-policy basis.  I have some policies that forward requests to services that I don't control, and these services don't consume the per-user identity information that Pomerium provides in these headers.  For privacy reasons I'd like to avoid sending PII like name and email address to these services unnecessarily.

A per-policy configuration (maybe something like `pass_user_identity`) that defaults to true but can be set to false to prevent Pomerium from adding user identity headers.  Alternately, allow `set_request_headers` to override `X-Pomerium-Jwt-Assertion` and `X-Pomerium-Claim-*`, but that seems a little hacky.

None, there doesn't seem to be any way to configure Pomerium to avoid adding `X-Pomerium-Jwt-Assertion`.

cuonglm:
## Summary
Currently, user's identity headers are always inserted to downstream
request. For privacy reason, it would be better to not insert these
headers by default, and let user chose whether to include these headers
per=policy basis.

## Related issues
Fixes #702

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review


internal/controlplane/xds_routes.go
```diff
@@ -134,6 +135,14 @@ func buildPolicyRoutes(options *config.Options, domain string) []*envoy_config_r
 			})
 		}
 
+		requestHeadersToRemove := policy.RemoveRequestHeaders
+		if !policy.PassIdentityHeaders {
```
travisgroth, line 139: Rather than stripping the JWT/Claim related headers here, we should avoid setting them in the authorize service's `Check()` call entirely.  See:

https://github.com/cuonglm/pomerium/blob/0d9b7d28faf61f018c03f8558bc199ed28f5eabe/authorize/grpc.go#L27
https://github.com/cuonglm/pomerium/blob/0d9b7d28faf61f018c03f8558bc199ed28f5eabe/authorize/grpc.go#L113
https://github.com/cuonglm/pomerium/blob/0d9b7d28faf61f018c03f8558bc199ed28f5eabe/authorize/check_response.go#L31

The header removal pattern in envoy is only needed for the cookie so we don't leak it to the upstream service.  We always have to return it, unlike the other headers.

cuonglm: @travisgroth I thought about it, too, but can't find a way to know whether the policy is turn on/off pass identity headers in authorize's `Check()`.

travisgroth: Hmm.  I see what you're saying.  All the per-route logic is inside OPA now.  I was thinking of this as a global option.

We could attach the headers to the input into OPA with a new field on the `Request` struct.  Thoughts?

docs/docs/upgrading.md
```diff
@@ -13,6 +13,11 @@ description: >-
 
 With this release, default log level has been changed to INFO.
 
+### Identity headers
+
+With this release, pomerium will not insert identity headers (X-Pomerium-Jwt-Asserttion/X-Pomerium-Claim-*) by default. To get pre 0.9.0 behavior, you
+can set `pass_identity_headers` to true per-policy basis.
```
desimone, line 19: nit: "on a per-policy basis"

config/policy.go
```diff
@@ -89,6 +89,14 @@ type Policy struct {
 	//
 	// https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header
 	PreserveHostHeader bool `mapstructure:"preserve_host_header" yaml:"preserve_host_header,omitempty"`
+
+	// PassIdentityHeaders controls whether to add user's identity headers to the downstream request.
```
desimone, line 93: nit: add a user's

internal/controlplane/xds_routes.go
desimone: I think this is reasonable policy to leave outside of our authz policy (which can/will be specific to each user's authorization policy). I think removing headers here makes sense to me because it feels more related to proxying (and routes) than authorization policy per-se (though it does use the header returned from AuthZ).

... does that make sense? @cuonglm @travisgroth

travisgroth: I do feel strange putting the data into the authz response just to remove it in lua later.  I'm not sure there's a great answer here, though.

cuonglm: I'm quite ok with doing it here for now.

The Authz should always make a response with its desired headers, because one can interact directly with the Authz instead of via proxy. Technically, it's will be strange if authorization response has no related identity information in the response headers. In our current situation, the proxy is the only one which consume Authz response, so it can do whatever it wants. For now, we just remove it per policy basis, but we can do what ever we want here in the future, not only just removing.

docs/docs/upgrading.md
cuonglm: Fixed.

config/policy.go
cuonglm: Fixed.

PR #904, opened 2020-06-15
authorize: use databroker data for rego policy

calebdoxsey:
## Summary
This updates the authorize service to use databroker data for the rego policy.

## Related issues
- For #887 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


authorize/evaluator/evaluator.go
```diff
@@ -4,48 +4,320 @@ package evaluator
...
+
+	var results []Result
+	for _, denial := range denials {
+		denial, ok := denial.([]interface{})
+		if !ok || len(denial) < 2 {
+			continue
+		}
+
+		status, err := strconv.Atoi(fmt.Sprint(denial[0]))
+		if err != nil {
```
cuonglm, line 271: Could we return this error, or warn about it. When rego changes the result structure, this can lead to hidden bug or difficult to debug code.

authorize/evaluator/evaluator.go
```diff
@@ -4,48 +4,320 @@ package evaluator
...
+	if !ok {
+		return nil
+	}
+
+	denials, ok := result["deny"].([]interface{})
+	if !ok {
+		return nil
+	}
+
+	var results []Result
```
cuonglm, line 263: Nit, we can do:

```
results := make([]Result, 0, len(denials))
```

since when we now `len(results)` will always less than or equal to `len(denials)`. This helps appending to `results` better performance.

authorize/grpc.go
```diff
@@ -31,84 +41,87 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v2.CheckRe
...
+
+	var userID string
+
+	for {
+		a.dataBrokerDataLock.RLock()
+		s, ok := a.dataBrokerData.Get(sessionTypeURL, sessionID).(*session.Session)
+		a.dataBrokerDataLock.RUnlock()
+
+		if ok {
+			if s.GetDeletedAt() != nil {
```
cuonglm, line 89: This can be simplified to just:

```
if s.GetUserId() != "" {
    userID = s.GetUserId()
    break
}
return
```

authorize/run.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/cenkalti/backoff"
+
+	"github.com/pomerium/pomerium/internal/grpc/databroker"
+	"github.com/pomerium/pomerium/internal/log"
+)
+
+// Run runs the authorize server.
+func (a *Authorize) Run(ctx context.Context) error {
+	t, ctx := tomb.WithContext(ctx)
```
cuonglm, line 18: Is there any reason to use `tomb` here instead of `errgroup`, I see we're using `errgroup` for this kind of task in other places.

authorize/run.go
```diff
@@ -0,0 +1,152 @@
...
+	log.Info().Msg("starting type sync")
+	return tryForever(ctx, func(backoff interface{ Reset() }) error {
+		stream, err := a.dataBrokerClient.SyncTypes(ctx, new(emptypb.Empty))
+		if err != nil {
+			return err
+		}
+
+		for {
+			res, err := stream.Recv()
+			if err != nil {
```
cuonglm, line 47: I think we need to check for `err == io.EOF`, then return `nil`.

authorize/run.go
```diff
@@ -0,0 +1,152 @@
...
+			RecordVersion: recordVersion,
+			Type:          typeURL,
+		})
+		if err != nil {
+			return err
+		}
+
+		for {
+			res, err := stream.Recv()
+			if err != nil {
```
cuonglm, line 101: Same as above for checking `err == io.EOF`.

authorize/run.go
```diff
@@ -0,0 +1,152 @@
...
+		a.dataBrokerData.Update(record)
+		a.dataBrokerDataLock.Unlock()
+	}
+}
+
+func tryForever(ctx context.Context, callback func(onSuccess interface{ Reset() }) error) error {
+	backoff := backoff.NewExponentialBackOff()
+	for {
+		err := callback(backoff)
+		log.Warn().Err(err).Msg("sync error")
```
cuonglm, line 145: Should we just log when `err != nil`?

internal/cmd/pomerium/pomerium.go
```diff
@@ -137,14 +143,14 @@ func setupAuthenticate(opt *config.Options, controlPlane *controlplane.Server) e
 	return nil
 }
 
-func setupAuthorize(opt *config.Options, controlPlane *controlplane.Server, optionsUpdaters *[]config.OptionsUpdater) error {
+func setupAuthorize(opt *config.Options, controlPlane *controlplane.Server, optionsUpdaters *[]config.OptionsUpdater) (*authorize.Authorize, error) {
 	if !config.IsAuthorize(opt.Services) {
-		return nil
+		return nil, nil
```
cuonglm, line 148: IMHO, we shouldn't do this. The common pattern in Go is when error is nil, it will guarantee that the pointer returned is not nil. From the caller site, it makes no sense if we still have to check for non-nil server when error return nil.

Can we return a new Server instance instead, and have a config to make it Run returns immediately. So you don't have to guard like in https://github.com/pomerium/pomerium/pull/904/files#diff-6671c9cee71180e2fc5b0e137e44b3e6R116

authorize/evaluator/evaluator.go
```diff
@@ -4,48 +4,326 @@ package evaluator
...
+func (e *Evaluator) newInput(req *Request, isValidClientCertificate bool) *input {
+	i := new(input)
+	i.DataBrokerData = req.DataBrokerData
+	i.HTTP = req.HTTP
+	i.Session = req.Session
+	i.IsValidClientCertificate = isValidClientCertificate
+	return i
+}
+
+type (
```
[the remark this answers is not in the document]
calebdoxsey: it adds an `IsValidClientCertificate` field that can be checked in the policy

authorize/evaluator/evaluator.go
```diff
@@ -4,48 +4,320 @@ package evaluator
...
+)
+
+// Result is the result of evaluation.
+type Result struct {
+	Status    int
+	Message   string
+	SignedJWT string
+}
+
+func getAllowVar(vars rego.Vars) bool {
```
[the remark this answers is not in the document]
calebdoxsey: the types in the rego policy are different

authorize/grpc.go
calebdoxsey: a session with a `DeletedAt` set will also have a `UserId` set, so we have to check `DeletedAt` first

internal/cmd/pomerium/pomerium.go
calebdoxsey: we can't return a new server instance because that would break non all-in-one mode. I will move the check out of the function

authorize/run.go
calebdoxsey: yes there was a reason but I don't remember what it was. I will switch to errgroup.

internal/identity/manager/manager.go
```diff
@@ -370,7 +370,7 @@ func (mgr *Manager) syncSessions(ctx context.Context, ch chan<- *session.Session
 		}
 
 		for _, record := range res.GetRecords() {
-			mgr.log.Info().Interface("session", record.GetData).Msg("session update")
+			mgr.log.Info().Interface("session", record.GetData()).Msg("session update")
```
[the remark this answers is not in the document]
calebdoxsey: I removed the log.

authorize/grpc.go
```diff
@@ -256,31 +242,28 @@ func getPeerCertificate(in *envoy_service_auth_v2.CheckRequest) string {
...
-	if rawJWT != nil {
-		evt = evt.Str("session", string(rawJWT))
+	if reply != nil {
+		evt = evt.Bool("allow", reply.Status == http.StatusOK)
+		evt = evt.Int("status", reply.Status)
+		evt = evt.Str("message", reply.Message)
 	}
-	if reply.GetHttpStatus() != nil {
-		evt = evt.Interface("http_status", reply.GetHttpStatus())
+	if sessionState != nil {
```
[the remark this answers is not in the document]
calebdoxsey: It's not sensitive, but I removed the log.

authorize/grpc.go
```diff
@@ -31,84 +41,87 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v2.CheckRe
...
-			rawJWT = newRawJWT
-			sessionErr = nil
-			isNewSession = true
-		} else {
-			log.Warn().Err(err).Msg("authorize: error refreshing session")
-			// set the error to expired so that we can force a new login
-			sessionErr = sessions.ErrExpired
-		}
+	if sessionState != nil {
+		a.waitForSessionAndUser(time.Second*3, sessionState.ID)
```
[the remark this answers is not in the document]
calebdoxsey: I updated it to use `WithTimeout`

PR #905, opened 2020-06-15
policy: Add consistent route identifier

### LINKED ISSUE #870: hot reload of policy results in 503
## What happened?

Hot reloading fails (returns a `503`) if the source URL changes for a given policy route's allowed user domains.

I expected to be able to hot reload policy.

## How'd it happen?

Have a policy like so. 

```yaml
policy:
  - from: https://httpbin.imac.bdd.io
    to: https://httpbin.org
    allowed_domains:
      - pomerium.com
```
Change the policy slightly

```diff
policy:
  - from: https://httpbin.imac.bdd.io
    to: https://httpbin.org
    allowed_domains:
      - pomerium.com
+      - gmail.com
```

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint): `pomerium/0.9.0 (+https://wwww.pomerium.io; 914b952; go1.14.3)`
- Server Operating System/Architecture/Cloud: docker compose

## What's your config.yaml?

See above.

travisgroth:
## Summary
- adds a route identifier that is derived only from routing information (source/destination/path) on a policy
- update envoy to use this new identifier for cluster names

Prevents unnecessary cluster churn when policy related settings are changed (eg, `allowed_users`).  This was causing transient route failures after policy changes when using HTTP2.  

## Related issues
Closes #870 

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/policy_test.go
+++ b/config/policy_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
 )
 
 func Test_PolicyValidate(t *testing.T) {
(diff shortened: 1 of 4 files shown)


calebdoxsey:
So the checksum of the route was used to avoid accidentally forwarding requests to invalid routes once they've changed. I feel like regardless of what we're using to generate the IDs, if its causing 503s, won't  that still happen, just less often?

travisgroth:
> So the checksum of the route was used to avoid accidentally forwarding requests to invalid routes once they've changed. I feel like regardless of what we're using to generate the IDs, if its causing 503s, won't that still happen, just less often?

This seems to be an issue with an http2 connection being bound to a specific cluster name inside envoy.  When the cluster names churn due to, say, adding a user, existing connections get failures for a period of time or until they time out.  Making the cluster name bound only to the route information should minimize this scenario to actual routing changes, when we really should be changing the cluster names.  

I believe this will retain the desired behavior of preventing forwarding to invalid routes.

config/policy_test.go
```diff
@@ -81,3 +82,46 @@ func TestPolicy_String(t *testing.T) {
...
+		{
+			"different path",
+			&Policy{From: "https://pomerium.io", To: "http://localhost"},
+			&Policy{From: "https://pomerium.io", To: "http://localhost", Path: "/foo"},
+			6315033228798964203,
+			false,
+		},
+	}
+
+	for _, tt := range tests {
```
cuonglm, line 118: Nit: maybe make these sub-tests run parallel, too.

travisgroth: good call

Issue #886, opened 2020-06-12
Authenticate: change session JWT


calebdoxsey:
- Change expiration to 1 week (or configurable)
- Remove unnecessary fields
- Update issuer?
- Remove `aud`
- Put session ID in `jti` field
- Put user in `sub`

ekovac:
Should probably also add the .kid field to the header so that it can be matched up with the jwks.json entry, unless this is a different JWT use.

PR #914, opened 2020-06-17
authenticate: databroker changes

calebdoxsey:
## Summary
This PR includes changes for the authenticate service to use the new databroker service.
- refresh has been removed
- various state session state fields have been removed
- the user dashboard was updated to retrieve session data from the databroker

## Related issues
- #886 
- #911
- #912
- #913

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review


cuonglm:
Integration tests fails seem reasonable:

```
    TestAuthorization/direct/domains/allowed: authorization_test.go:50: 
        	Error Trace:	authorization_test.go:50
        	Error:      	Not equal: 
        	            	expected: 200
        	            	actual  : 403
        	Test:       	TestAuthorization/direct/domains/allowed
        	Messages:   	expected OK for dogs.test
```

calebdoxsey:
@cuonglm I removed the tests which no longer applied. It would be good if we had an actual groups test, but we need to do work on the integration tests to make them more reliable, so maybe we could add it as part of that. (it requires updating our reference oidc provider so it exposes a groups endpoint, or providing some other mechanism to define the groups) Testing a real identity provider is difficult because automating login like this is often explicitly prevented (ie with a captcha)

Longer term we plan on having infrastructure setup that we can leverage. Most of the flakiness is around creating a new kubernetes cluster every time.

PR #927, opened 2020-06-18
internal/envoy: fix handleLogs causes envoy hang forever

cuonglm:
## Summary
handleLogs uses bufio scanner to process log output from envoy. When in
debug mode, envoy produces very long log line, causing the scanner
fails, handleLogs stop processing log. But envoy continue writing to its
stdout, which is now not consumed by any process, envoy hangs there
forever.

Fixing this by switching to use bufio.Reader instead. This is also the
real fix for failed integration test, which is interpreted wrongly by
me in #910.

## Related issues
Updates #903 

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/internal/envoy/envoy.go
+++ b/internal/envoy/envoy.go
@@ -2,9 +2,8 @@
 package envoy
 
 import (
-	"bytes"
-
 	"bufio"
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -14,6 +13,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strconv"
+	"strings"
 
 	envoy_config_bootstrap_v3 "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v3"
 	envoy_config_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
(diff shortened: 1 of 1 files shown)


internal/envoy/envoy.go
```diff
@@ -309,13 +309,20 @@ func (srv *Server) addTraceConfig(traceOpts *config.TracingOptions, bootCfg *env
...
 
-	s := bufio.NewScanner(stdout)
-	for s.Scan() {
-		ln := s.Text()
+	s := bufio.NewReader(rc)
+	for {
+		ln, err := s.ReadString('\n')
+		if err != nil {
+			log.Error().Err(err).Msg("failed to read log")
+			continue
```
calebdoxsey, line 323: I think the only error we should break on would be `io.EOF`:

```go
if err == io.EOF {
    break
} else if err != nil {
   //...
}
```

calebdoxsey: also we might need exponential backoff on other types of errors? Not sure the best way to handle that

cuonglm: 👍 Fixed

cuonglm: @calebdoxsey I think we don't need, it's hard to think any error in this case other than io.EOF.

PR #965, opened 2020-06-22
authenticate: move impersonate from proxy to authenticate

calebdoxsey:
## Summary
Since we moved the dashboard to the authenticate the service, we also need to move the impersonate endpoint.

## Related issues
- https://github.com/pomerium/internal/issues/52

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review


proxy/handlers_test.go
```diff
@@ -66,57 +66,6 @@ func TestProxy_Signout(t *testing.T) {
 	}
 }
 
-func TestProxy_Impersonate(t *testing.T) {
```
cuonglm, line 69: Any reason you don't move this test to authenticate, too?

PR #968, opened 2020-06-22
Sleep longer before running integration tests

cuonglm:
## Related issues

**Checklist**:
- [x] ready for review


cuonglm:
The CI was run 3 times without a failure, hope the sleep duration is longer enough to the cluster become stable.

Issue #933, opened 2020-06-19
docs: document using service accounts for identity providers


calebdoxsey:
With the recent changes to user and session storage, we are now querying group data asynchronously via a service account. That service account needs to be created within the identity provider and supplied via the `idp_service_account` configuration option.

The format of the various providers needs to be:

## Azure
`idp_service_account` is a base64-encoded, json object:

```json
{
  "client_id": "...",
  "client_secret": "...",
  "directory_id": "..."
}
```

## GitHub
`idp_service_account` is a base64-encoded, json object:

```json
{
  "username": "...",
  "personal_access_token": "..."
}
```

## Gitlab 
`idp_service_account` is a base64-encoded, json object:

```json
{
  "private_token": "..."
}
```

## Google
`idp_service_account` is a base64-encoded, json object generated by the Google Cloud console:

```json
{
  "type": "service_account",
  "project_id": "...",
  "private_key_id": "...",
  "private_key": "...",
  "client_email": "...",
  "client_id": "...",
  "auth_uri": "...",
  "token_uri": "...",
  "auth_provider_x509_cert_url": "...",
  "client_x509_cert_url": "...",
  "impersonate_user": "..."
}
```

The `impersonate_user` field is manually added to the JSON object since it's required for the Admin API.

## Okta
`idp_service_account` is a base64-encoded, json object:

```json
{
  "api_key": "..."
}
```

## Onelogin
`idp_service_account` is a base64-encoded, json object:

```json
{
  "client_id": "...",
  "client_secret": "..."
}
```

Also I think we may not have implemented Github. I will add an issue for that.

desimone:
Nice work @calebdoxsey , I'm going through these individually to make sure they work as expected right now.

PR #971, opened 2020-06-22
directory: add service account struct and parsing method

calebdoxsey:
## Summary
It looks like the docs for Google are already correct for configuring the service account. I went ahead and updated the code to use a ServiceAccount struct to match the other providers and give better error messages if `ImpersonateUser` is missing.

## Related issues
- #933

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/directory/google/google.go
+++ b/internal/directory/google/google.go
@@ -27,7 +27,7 @@ const (
 )
 
 type config struct {
-	serviceAccount string
+	serviceAccount *ServiceAccount
 	url            string
 }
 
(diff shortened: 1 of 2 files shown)


internal/directory/google/google.go
```diff
@@ -170,3 +165,46 @@ func (p *Provider) getAPIClient(ctx context.Context) (*admin.Service, error) {
...
+		return nil, err
+	}
+
+	var serviceAccount ServiceAccount
+	err = json.Unmarshal(bs, &serviceAccount)
+	if err != nil {
+		return nil, err
+	}
+
+	if serviceAccount.ImpersonateUser == "" {
```
desimone, line 205: good idea

PR #979, opened 2020-06-23
authenticate: hide impersonation form from non-admin users

### LINKED ISSUE #881: Hide impersonation options in /.pomerium path from non-admins
Currently, all users will see the impersonation form in the /.pomerium URL. This is not only confusing, but may be a source of security concerns (real or perceived). 

Since a non-admin cannot use this feature, it should not be present for non-admin users. 

Additionally, it may be desirable to optionally disable this path completely for non-admin users.

cuonglm:
## Summary
Do not show impersonation form for non-admin users

## Related issues
Fixes #881

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


authenticate/authenticate.go
```diff
@@ -181,6 +184,10 @@ func New(opts config.Options) (*Authenticate, error) {
 		return nil, err
 	}
 
+	administrator := make(map[string]struct{}, len(opts.Administrators))
```
desimone, line 187: Will this be hot reloaded?

cuonglm: @desimone I think no, currently, only policy related rules are hot reloaded.

cuonglm: https://github.com/pomerium/pomerium/issues/982

PR #1005, opened 2020-06-25
cache: fix missing parameter

travisgroth:
## Summary
Fix build error introduced in previous PR.

## Related issues
#1004 

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/cache/memberlist.go
+++ b/cache/memberlist.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	stdlog "log"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/memberlist"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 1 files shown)


cache/memberlist.go
```diff
@@ -58,7 +59,7 @@ func (c *Cache) runMemberList(ctx context.Context) error {
 		mh.log.Error().Msg("multiple cache servers not supported")
 	}
 	<-ctx.Done()
-	mh.memberlist.Leave()
+	mh.memberlist.Leave(1 * time.Second) //nolint
```
cuonglm, line 62: This error should be handled. Maybe:

```
if err := mh.memberlist.Leave(time.Second); err != nil {
		mh.log.Error().Err(err).Msg("could not broadcast leave message")
}
```

travisgroth: You're right.  Given there's not much to do here, I didn't bother.  But we should put in a log message.

PR #1008, opened 2020-06-26
Allow specify go executable in Makefile

cuonglm:
## Related issues

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


cuonglm:
> Can you provide some explanation as to why?

This is mostly for developer, e.g I have different go executables in my local:

 - go is the gotip
 - go1.14.4 is the stable version

For now, if I run `make all`, it will always use `go` (means gotip) instead of stable go1.14.4

With this PR, I can do:

```
export GO=go1.14.4
make all
```

PR #1010, opened 2020-06-26
grpc: rename internal/grpc to pkg/grpc

calebdoxsey:
## Summary
We should move the protobuf definitions out of `internal/` so we can reference them from other projects.

**Checklist**:
- [x] updated docs
- [x] updated unit tests
- [x] ready for review


.pre-commit-config.yaml
```diff
@@ -18,3 +18,4 @@ repos:
         entry: make
         args: ["lint"]
         types: ["go"]
+        pass_filenames: false
```
calebdoxsey, line 21: `make lint` was being called for every file on commit. This makes it so its only called once.

PR #1026, opened 2020-06-29
authorize/evaluator/opa: set client tls cert usage explicitly

travisgroth:
## Summary
x509.Certificate.Verify() defaults to checking for Server usage rather than Client usage.  Set VerifyOptions to check for client usage explicitly.

**Checklist**:
- [x] ready for review

### DIFF
--- a/authorize/evaluator/functions.go
+++ b/authorize/evaluator/functions.go
@@ -6,12 +6,11 @@ import (
 	"fmt"
 	"io/ioutil"
 
-	"github.com/pomerium/pomerium/internal/log"
-
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/rakyll/statik/fs"
 
 	_ "github.com/pomerium/pomerium/authorize/evaluator/opa/policy" // load static assets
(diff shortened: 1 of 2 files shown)


authorize/evaluator/functions.go
```diff
@@ -41,13 +41,14 @@ func isValidClientCertificate(ca, cert string) (bool, error) {
...
-		Roots: roots,
+		Roots:     roots,
+		KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
 	})
 	valid := verifyErr == nil
 
 	isValidClientCertificateCache.Add(cacheKey, valid)
 
-	return valid, nil
+	return valid, verifyErr
```
desimone, line 51: Is there a gotcha by having this error set here? \cc @calebdoxsey

calebdoxsey: Yes. Error means something unexpected happened. If a user supplies an invalid cert its not an error, but it returns false.

travisgroth: 👍  changed to a log message in debug

PR #1028, opened 2020-06-30
cache: fix data race in NotifyJoin

cuonglm:
## Summary
In 35af5c0b91e1f93313a6ca67ed62cd1a72b11c5b, the check for multiple
cache servers in NotifyJoin is made to be done in a goroutine. That can
lead to a data race, because the memberlist can be changed at the time
the goroutine was run. go warns about this race when test memberlist was
run with "-race".

## Related issues

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/cache/memberlist.go
+++ b/cache/memberlist.go
@@ -69,11 +69,11 @@ func (c *Cache) runMemberList(ctx context.Context) error {
 func (mh *memberlistHandler) NotifyJoin(node *memberlist.Node) {
 	mh.log.Debug().Interface("node", node).Msg("node joined")
 
-	go func() {
-		if mh.memberlist != nil && mh.memberlist.NumMembers() > 1 {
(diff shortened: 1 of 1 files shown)


calebdoxsey:
When the event is called the memberlist is locked so the call to `NumMembers` will block forever.

cuonglm:
> When the event is called the memberlist is locked so the call to `NumMembers` will block forever.

`memberlist` is locked by what mutex? If it's the internal mutex of memberlist itself, I think we're safe. There's `TestCache_runMemberList` which I added before to make sure we're ok.

cuonglm:
> When the event is called the memberlist is locked so the call to `NumMembers` will block forever.

Thanks, fixed 👍

PR #1029, opened 2020-06-30
cryptutil: move to pkg dir, add token generator

calebdoxsey:
## Summary
This makes cryptutil available to other programs and adds code for generating random tokens.

**Checklist**:
- [x] ready for review


pkg/cryptutil/token.go
```diff
@@ -0,0 +1,68 @@
+package cryptutil
```
desimone, line 1: Could we add tests for this package?

PR #1030, opened 2020-07-01
proxy: fix wrong forward auth request

### LINKED ISSUE #873: Prefix Paths aren't working with Traefik Ingress
I can not make v0.9 work with prefix/regex. No matter what I try path/regex/prefix aren't working with traefik. From the behavior it looks like the first rule doesn't get triggered.
```yaml
  - allowed_users:
    - xxx@gmail.com
    from: https://whoami.stage.example.com
    # regex: "^/(system|ui)/.*$"
    prefix: /system
    to: http://whoami
  - allow_public_unauthenticated_access: true
    from: https://whoami.stage.example.com
    to: http://whoami
```

The following screenshot is what the proxy receives from traefik. 

image (2)

cuonglm:
## Summary
When proxy receives forward auth request, it should forward the request
as-is to authorize for verification. Currently, it composes the check
request with actual path, then send the request to authorize service.

It makes the request works accidently, because the composed check
request will satisfy the policy un-intentionally. Example, for forward
auth request:

	http://pomerium/?uri=https://httpbin.localhost.pomerium.io

the composed request will look like:

```
	&envoy_service_auth_v2.AttributeContext_HttpRequest{
		Method:   "GET",
		Headers:  map[string]string{},
		Path:     "",
		Host:     "httpbin.localhost.pomerium.io",
		Scheme:   "https",
	}
```

This check request has at least two problems.

First, it will make authorize.handleForwardAuth always returns false,
even though this is a real forward auth request. Because the "Host"
field in check request is not the forward auth host, which is "pomerium"
in this case.

Second, it will accidently matches rule like:

```
	policy:
	    to: https://httpbin
	    allowed_domains:
	      - pomerium.io
```

If the rule contains other conditions, like "prefix", or "regex":

```
	policy:
	    prefix: /headers
	    to: https://httpbin
	    allowed_domains:
	      - pomerium.io
```

Then the rule will never be triggered, because the "/headers" path can
be passed in request via "X-Forwarded-Uri" (traefik), instead of
directly from the path (nginx).

To fix this, we just pass the forward auth request as-is to authorize.

## Related issues
Fixes #873

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/grpc.go
+++ b/authorize/grpc.go
@@ -177,25 +177,33 @@ func (a *Authorize) handleForwardAuth(req *envoy_service_auth_v2.CheckRequest) b
 	}
 
 	checkURL := getCheckRequestURL(req)
-	if urlutil.StripPort(checkURL.Host) == urlutil.StripPort(opts.GetForwardAuthURL().Host) {
-		if (checkURL.Path == "/" || checkURL.Path == "/verify") && checkURL.Query().Get("uri") != "" {
-			verifyURL, err := url.Parse(checkURL.Query().Get("uri"))
-			if err != nil {
-				log.Warn().Str("uri", checkURL.Query().Get("uri")).Err(err).Msg("failed to parse uri for forward authentication")
-				return false
-			}
-			req.Attributes.Request.Http.Scheme = verifyURL.Scheme
-			req.Attributes.Request.Http.Host = verifyURL.Host
-			req.Attributes.Request.Http.Path = verifyURL.Path
-			// envoy sends the query string as part of the path
-			if verifyURL.RawQuery != "" {
-				req.Attributes.Request.Http.Path += "?" + verifyURL.RawQuery
-			}
-			return true
-		}
+	if urlutil.StripPort(checkURL.Host) != urlutil.StripPort(opts.GetForwardAuthURL().Host) {
+		return false
 	}
 
-	return false
+	uriQuery := checkURL.Query().Get("uri")
+	if (checkURL.Path != "/" && checkURL.Path != "/verify") || uriQuery == "" {
+		return false
+	}
+	verifyURL, err := url.Parse(uriQuery)
+	if err != nil {
+		log.Warn().Str("uri", checkURL.Query().Get("uri")).Err(err).Msg("failed to parse uri for forward authentication")
+		return false
+	}
+
+	req.Attributes.Request.Http.Scheme = verifyURL.Scheme
+	req.Attributes.Request.Http.Host = verifyURL.Host
+	req.Attributes.Request.Http.Path = verifyURL.Path
+	if headers := req.GetAttributes().GetRequest().GetHttp().GetHeaders(); headers != nil {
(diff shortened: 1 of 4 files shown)


cuonglm:
> We should asserting that the option [forward_auth](https://www.pomerium.io/configuration/#forward-auth) has been set before defaulting to use the forwarded uri.

Option `forward_auth` is checked in https://github.com/pomerium/pomerium/blob/cuonglm/x-forwarded-uri/authorize/grpc.go#L175

authorize/grpc.go
```diff
@@ -177,25 +177,33 @@ func (a *Authorize) handleForwardAuth(req *envoy_service_auth_v2.CheckRequest) b
...
+		log.Warn().Str("uri", checkURL.Query().Get("uri")).Err(err).Msg("failed to parse uri for forward authentication")
+		return false
 	}
 
-	return false
+	req.Attributes.Request.Http.Scheme = verifyURL.Scheme
+	req.Attributes.Request.Http.Host = verifyURL.Host
+	req.Attributes.Request.Http.Path = verifyURL.Path
+	if headers := req.GetAttributes().GetRequest().GetHttp().GetHeaders(); headers != nil {
+		if xfu := headers[http.CanonicalHeaderKey("x-forwarded-uri")]; xfu != "" {
```
desimone, line 198: nit: use constant form `internal/httputil/headers.go`

cuonglm:
> The refactor of the conditions is too complex for me to understand what's changed, but if it works LGTM.

The refactor condition is the same, just that it prevents nested condition, and return earlier if the condition failed. That the prefer style in handling error in Go.

proxy/middleware.go
```diff
@@ -55,16 +55,16 @@ func (p *Proxy) isAuthorized(w http.ResponseWriter, r *http.Request) (bool, erro
 		Method:   "GET",
 		Headers:  map[string]string{},
 		Path:     r.URL.Path,
-		Host:     r.URL.Host,
+		Host:     r.Host,
```
desimone, line 58: Will this have broader implications?

proxy/forward_auth.go
```diff
@@ -118,8 +118,7 @@ func (p *Proxy) Verify(verifyOnly bool) http.Handler {
 			return httputil.NewError(http.StatusBadRequest, err)
 		}
 
-		original := p.getOriginalRequest(r, uri)
```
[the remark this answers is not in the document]
cuonglm: That's right, we don't modify the request anymore. Before this, we modify the request from forward auth url -> actual traefik request url

proxy/middleware.go
cuonglm: No, it makes the forward auth works correctly. Before this change, with request like `http://pomerium/?uri=https://httpbin.localhost.pomerium.io`, we rewrite it into `https://httpbin.localhost.pomerium.io` and send this request as a forward auth request to authorize. In authorize, you can see there's a check that if the request "Host" matches the forward auth "Host".

Also note that this change just affect forward auth mode.

Issue #1033, opened 2020-07-02
Destination (To:) path not followed since v0.9.X


fightforlife:
## What happened?

Pomerium is no longer respecting paths in the destination (To:) of the proxied request. This was working fine with v0.8.3.
Currently I dont see another way of providing destination paths. (as far as I can see I can only specify incoming paths)

Example:
```
# Some Policy
  - from: https://some.domain.com
    to: http://192.178.198.25:8887/folder
    allow_public_unauthenticated_access: true
    allow_websockets: true
```

In version 0.8.3 this is linking `https://some.domain.com` to `http://192.178.198.25:8887/folder`.
So when requesting `https://some.domain.com/test/picture.jpg` it should link to `http://192.178.198.25:8887/folder/test/picture.jpg`

Version bigger 0.8.3:
Insted it is linking to `http://192.178.198.25:8887/test/picture.jpg` without `folder`.
Or the other way around I have to open `https://some.domain.com/folder/test/picture.jpg` to get to `https://some.policy.com/test/picture.jpg`

## What did you expect to happen?

The proxied request should be forwarded to the specified URL including the path.

## How'd it happen?

1. Updated my pomerium installation from v0.8.3 to 0.10rc1

## What's your environment like?

- v0.10.0-rc1
- Ubuntu 20.04, Pomerium inside docker

## What's your config.yaml?

```
docker compose
version: "2"
services:
  pomerium:
    container_name: pomerium
    image: pomerium/pomerium:latest
    environment:
      - AUTHENTICATE_SERVICE_URL=https://authenticate.domain.com
      - AUTOCERT=true
      - AUTOCERT_USE_STAGING=false
      - AUTOCERT_DIR=/pomerium/certs/
      - IDP_PROVIDER=google
      - IDP_CLIENT_ID=XXX.apps.googleusercontent.com
      - IDP_CLIENT_SECRET=XXX
      - COOKIE_SECRET=XXX
      - ADMINISTRATORS="some.mail@gmail.com"
      - HTTP_REDIRECT_ADDR=:80
    volumes:
      - pomerium_config:/pomerium/
    ports:
      - 80:80
      - 443:443
    restart: always
```

## What did you see in the logs?

```logs
Are these needed?
```

## Additional context

As far as I can see this happens since the v0.9.X release.

cuonglm:
@fightforlife Thanks for reporting, this should be fixed in #1034

fightforlife:
[...]

Issue #880, opened 2020-06-12
Proxy: Allow mapping apex external host url to backend path


cjwalton:
Some backend applications may require a non-apex path. I'd like to be able to craft a policy statement to map to a path. This is more elegant than trying to fix this on the backend application.

Example:

  - from: https://myapp.external.domain
    to: https://myapp.internal.domain/app

Currently, it seems Pomerium only supports statements like this:

  - from: https://myapp.external.domain
    to: https://myapp.internal.domain

Vad1mo:
This is what `pather` - `prefix` are for. did you take a look at https://www.pomerium.io/configuration/#path? https://www.pomerium.io/docs/upgrading.html#using-paths-in-from-urls

calebdoxsey:
@Vad1mo I think @cjwalton wants to support stripping/adding prefixes. ie `https://myapp.external.domain/some/path` becomes `https://myapp.internal.domain/app/some/path`. That's not currently supported. 

We can support a basic string literal pretty easily, but some of the other options would be harder. (like a regular expression match) I'm also not sure what the configuration should look like. We explicitly disabled paths in the `from` parameter because it's ambiguous how they should be interpreted, but I guess a path in the `to` parameter wouldn't have that issue. But then we're only supporting adding the prefix, not stripping it, so maybe it should be an entirely different configuration option.

cjwalton:
@calebdoxsey  is correct. Specifically I am looking for a path added to the "to" parameter. 
from: host.local
to: host.local/app

PR #1034, opened 2020-07-02
internal/controlplane: set envoy prefix rewrite if present

cuonglm:
## Summary
When policy destination url has path, we should enable envoy prefix rewrite to make it works.

## Related issues
Fixes #1033

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

PR #1040, opened 2020-07-06
docs/configuration: add doc for trailing slash limitation in "To" field

cuonglm:
## Summary
Due to the limitation of envoy, it can't handle rewriting of "From"
field without path to a destination with path.

## Related issues
Updates #880
Updates #1033

**Checklist**:
- [x] add related issues
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/configuration/readme.md
+++ b/docs/configuration/readme.md
@@ -976,9 +976,33 @@ Remove Request Headers allows you to remove given request headers. This can be u
 - `yaml`/`json` setting: `to`
 - Type: `URL` (must contain a scheme and hostname)
 - Required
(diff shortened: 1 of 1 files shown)


cuonglm:
@travisgroth I think we shouldn't backport this to 0-9-0 unless we do the same for https://github.com/pomerium/pomerium/pull/1034

travisgroth:
I see.  Given it looks to be a regression in v0.9, I think we should backport the fix unless it is difficult to do so.

cuonglm:
> I see. Given it looks to be a regression in v0.9, I think we should backport the fix unless it is difficult to do so.

it's not difficult, marked it as backport.

Issue #1054, opened 2020-07-12
`certificates.cert` / `certificates.key` should accept base64 in addition to file refs.


dmitrif:
**Is your feature request related to a problem? Please describe.**

Currently using Pomerium with "known quantities" — services that are hosted with static hostnames / ports. Working on adding possibility of spinning up n-number of services that are tagged as `internal` in Consul and having them load-balanced by Pomerium.

As we are using nomad, there is consul-template which allows templating out certs from our secret store (Vault), however they do not handle indeterminate number of templates, you can either output all certs to one file or into the config file as Base64. 

**Describe the solution you'd like**

As the framework is already there to read in Base64 values for single certs, it'd be nice for the certificates subtree to also attempt to read in Base64.

**Describe alternatives you've considered**

- Attempted using Levant in combination with Nomad to template this out but Levant does not implement consul-template and instead focuses on a subset of interactions with Consul. No integration with Vault.
- Considered created a docker image that extends Pomerium and splits out a single bundle into many certificates, which will be my fallback... but I think this feature would generally be useful.

**Explain any additional use-cases**

- Integrations with service directories like Consul with a possibility of spinning up new internal sites post-pomerium deploy.

**Additional context**

N/A

nicgrayson:
@dmitrif Sorry to ping you. Do you have an example of running pomerium in nomad? I'm trying to do the same.

dmitrif:
@nicgrayson Just saw this notification. Yes I do, I have HAProxy on a different server forwarding requests to the pomerium instance.

PR #1055, opened 2020-07-12
#1054 - Change  config key parsing to attempt Base64 decoding first.

dmitrif:
## Summary
My GO skills are non-existent but this is my stab to allow `certificates` to first try base64 decoding for the entries, and only then fallback to FS resolution.

## Related issues
Addresses #1054 

**Checklist**:
- [x] add related issues
- [x] updated docs (Bypassed as they are currently ambiguous)
- [x] updated unit tests (Bypassed due to lack of coverage)
- [x] updated UPGRADING.md (Bypassed due to not being a breaking change)
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -558,9 +558,12 @@ func (o *Options) Validate() error {
 	}
 
 	for _, c := range o.CertificateFiles {
-		cert, err := cryptutil.CertificateFromFile(c.CertFile, c.KeyFile)
+		cert, err := cryptutil.CertificateFromBase64(c.CertFile, c.KeyFile)
 		if err != nil {
-			return fmt.Errorf("config: bad cert file %w", err)
+			cert, err = cryptutil.CertificateFromFile(c.CertFile, c.KeyFile)
+		}
+		if err != nil {
+			return fmt.Errorf("config: bad cert entry, base64 or file reference invalid. %w", err)
 		}
 		o.Certificates = append(o.Certificates, *cert)
 	}
--- a/config/options_test.go
+++ b/config/options_test.go
@@ -505,6 +505,42 @@ func TestHTTPRedirectAddressStripQuotes(t *testing.T) {
 	assert.Equal(t, ":80", o.HTTPRedirectAddr)
 }
 
+func TestCertificatesArrayParsing(t *testing.T) {
+	t.Parallel()
+
+	testCertFileRef := "./testdata/example-cert.pem"
+	testKeyFileRef := "./testdata/example-key.pem"
+	testCertFile, _ := ioutil.ReadFile(testCertFileRef)
+	testKeyFile, _ := ioutil.ReadFile(testKeyFileRef)
+	testCertAsBase64 := base64.StdEncoding.EncodeToString(testCertFile)
+	testKeyAsBase64 := base64.StdEncoding.EncodeToString(testKeyFile)
+
+	tests := []struct {
+		name             string
+		certificateFiles []certificateFilePair
+		wantErr          bool
+	}{
+		{"Handles base64 string as params", []certificateFilePair{{KeyFile: testKeyAsBase64, CertFile: testCertAsBase64}}, false},
+		{"Handles file reference as params", []certificateFilePair{{KeyFile: testKeyFileRef, CertFile: testCertFileRef}}, false},
+		{"Returns an error otherwise", []certificateFilePair{{KeyFile: "abc", CertFile: "abc"}}, true},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			o := NewDefaultOptions()
+			o.CertificateFiles = tt.certificateFiles
+			err := o.Validate()
+
+			if err != nil && tt.wantErr == false {
+				t.Fatal(err)
+			}
+		})
+	}
+}
+
 func TestCompareByteSliceSlice(t *testing.T) {
 	type Bytes = [][]byte
 


cuonglm:
@dmitrif would you mind adding a test for this?

dmitrif:
Thanks! 

@cuonglm / @desimone : I am happy to add a test case for this. However struggling to get the test harness to run locally with `make: opa: No such file or directory`. Is there a contributors guide by chance to get this running? Cheers!

cuonglm:
> Thanks!
> 
> @cuonglm / @desimone : I am happy to add a test case for this. However struggling to get the test harness to run locally with `make: opa: No such file or directory`. Is there a contributors guide by chance to get this running? Cheers!

`opa` is required to run rego policy tests, in this case, you can simply do:

```
go test -race $(go list ./... | grep -v vendor | grep -v github.com/pomerium/pomerium/integration)
```

dmitrif:
@cuonglm Thank you, that worked perfectly!

@cuonglm / @desimone : Added unit test coverage and updated logging to be a bit more verbose to output both the base64 error and the file parsing error if both scenarios fail.

config/options.go
```diff
@@ -557,11 +557,24 @@ func (o *Options) Validate() error {
 		o.Certificates = append(o.Certificates, *cert)
 	}
 
-	for _, c := range o.CertificateFiles {
-		cert, err := cryptutil.CertificateFromFile(c.CertFile, c.KeyFile)
-		if err != nil {
-			return fmt.Errorf("config: bad cert file %w", err)
+	for i, c := range o.CertificateFiles {
+		fileRefErr := error(nil)
```
cuonglm, line 561: use `var fileRefErr error` instead

config/options.go
```diff
@@ -557,11 +557,24 @@ func (o *Options) Validate() error {
...
+		cert, base64Err := cryptutil.CertificateFromBase64(c.CertFile, c.KeyFile)
+
```
cuonglm, line 563: Remove this empty line.

config/options.go
```diff
@@ -557,11 +557,24 @@ func (o *Options) Validate() error {
...
 
-	for _, c := range o.CertificateFiles {
-		cert, err := cryptutil.CertificateFromFile(c.CertFile, c.KeyFile)
-		if err != nil {
-			return fmt.Errorf("config: bad cert file %w", err)
+	for i, c := range o.CertificateFiles {
+		fileRefErr := error(nil)
+		cert, base64Err := cryptutil.CertificateFromBase64(c.CertFile, c.KeyFile)
+		
+		if base64Err != nil {
```
cuonglm, line 564: I don't think this need to be complicated like this. Just use one error like your original version, and said that "either a base64 or a file reference" in error message.

config/options_test.go
```diff
@@ -505,6 +505,37 @@ func TestHTTPRedirectAddressStripQuotes(t *testing.T) {
...
+	testKeyFileRef := "./testdata/example-key.pem"
+	testCertFile, _ := ioutil.ReadFile(testCertFileRef)
+	testKeyFile, _ := ioutil.ReadFile(testKeyFileRef)
+	testCertAsBase64 := base64.StdEncoding.EncodeToString(testCertFile)
+	testKeyAsBase64 := base64.StdEncoding.EncodeToString(testKeyFile)
+
+	tests := []struct {
+		name             string
+		certificateFiles []certificateFilePair
+		wantErr 				 bool
```
cuonglm, line 519: `go fmt` this.

config/options_test.go
```diff
@@ -505,6 +505,37 @@ func TestHTTPRedirectAddressStripQuotes(t *testing.T) {
...
+	}{
+		{ "Handles base64 string as params", []certificateFilePair{{ KeyFile: testKeyAsBase64, CertFile: testCertAsBase64 }}, false },
+		{ "Handles file reference as params", []certificateFilePair{{ KeyFile: testKeyFileRef, CertFile: testCertFileRef }}, false },
+		{ "Returns an error otherwise", []certificateFilePair{{ KeyFile: "abc", CertFile: "abc" }}, true },
+	}
+
+	for _, tt := range tests {
```
cuonglm, line 526: You can run parallel here.

dmitrif:
@cuonglm Thank you! Simplified back to the previous iteration and fixed the rest of CR comments.

config/options_test.go
```diff
@@ -505,6 +505,41 @@ func TestHTTPRedirectAddressStripQuotes(t *testing.T) {
...
+		certificateFiles []certificateFilePair
+		wantErr          bool
+	}{
+		{ "Handles base64 string as params", []certificateFilePair{{ KeyFile: testKeyAsBase64, CertFile: testCertAsBase64 }}, false },
+		{ "Handles file reference as params", []certificateFilePair{{ KeyFile: testKeyFileRef, CertFile: testCertFileRef }}, false },
+		{ "Returns an error otherwise", []certificateFilePair{{ KeyFile: "abc", CertFile: "abc" }}, true },
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
cuonglm, line 529: You need `tt := tt` before `t.Run`, so it won't capture wrong `tt` variable in for loop.

dmitrif: Will add, but could you please elaborate? I assumed that Go would introduce a new scope when the test func was declared. As such there should be no scope confusion about `tt`?

Thanks!

cuonglm: I invite you to read https://blog.golang.org/subtests, read `Run a group of tests in parallel` section.

Tests run without parallel is safe to not capture the range variable.

dmitrif: Thank you! I made the change, but this is a great starting point for me understanding why :) Appreciate your patience!

cuonglm: > Thank you! I made the change, but this is a great starting point for me understanding why :) Appreciate your patience!

Would you please formatting files with `gofmt -s` 😄

PR #1063, opened 2020-07-13
kubernetes apiserver integration

calebdoxsey:
## Summary
This PR adds support for kube-apiserver authentication using a service accoun token and impersonation headers.

* A new configuration policy option (`kubernetes_service_account_token`) can be used. When this option appears, an Authorization header will be added to requests along with headers for user impersonation.
* Authorization policy evaluation now returns additional information needed to build the impersonation headers.
* The rego file was modified to return the matching policy.
* Sessions can now be loaded via an `Authorization: Bearer Pomerium-` header in addition to `Authorization: Pomerium`

**Checklist**:
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


authorize/evaluator/evaluator.go
```diff
@@ -167,15 +167,15 @@ func (e *Evaluator) ParseSignedJWT(signature string) ([]byte, error) {
 	return object.Verify(&(e.jwk.(*ecdsa.PrivateKey).PublicKey))
 }
 
-// SignedJWT returns the signature of given request.
-func (e *Evaluator) SignedJWT(req *Request) (string, error) {
+// FillResult fills the result.
```
cuonglm, line 170: This `FillResult` sounds not good enough to me. The only reason you removed it because `*Result` needs the email.

I think the proper way is moving the payload filling outside of `SignedJWT`, that said, all of these code:

```
payload := map[string]interface{}{
		"iss": e.authenticateHost,
	}
	if u, err := url.Parse(req.HTTP.URL); err == nil {
		payload["aud"] = u.Hostname()
	}
	if s, ok := req.DataBrokerData.Get("type.googleapis.com/session.Session", req.Session.ID).(*session.Session); ok {
		if tm, err := ptypes.Timestamp(s.GetIdToken().GetExpiresAt()); err == nil {
			payload["exp"] = tm.Unix()
		}
		if tm, err := ptypes.Timestamp(s.GetIdToken().GetIssuedAt()); err == nil {
			payload["iat"] = tm.Unix()
		}
		if u, ok := req.DataBrokerData.Get("type.googleapis.com/user.User", s.GetUserId()).(*user.User); ok {
			payload["sub"] = u.GetId()
			payload["user"] = u.GetId()
			payload["email"] = u.GetEmail()
		}
		if du, ok := req.DataBrokerData.Get("type.googleapis.com/directory.User", s.GetUserId()).(*directory.User); ok {
			payload["groups"] = du.GetGroups()
		}
	}
```

Something like `jwtPayload(req)`, then you can pass this payload to `SignedJWT`, and filling it to `*Result` outside of `SignedJWT`.

authorize/check_response.go
```diff
@@ -25,7 +26,19 @@ func (a *Authorize) okResponse(reply *evaluator.Result) *envoy_service_auth_v2.C
 	}
 
 	requestHeaders = append(requestHeaders,
-		mkHeader(httputil.HeaderPomeriumJWTAssertion, reply.SignedJWT))
+		mkHeader(httputil.HeaderPomeriumJWTAssertion, reply.SignedJWT, false))
+
+	if reply.MatchingPolicy != nil && reply.MatchingPolicy.KubernetesServiceAccountToken != "" {
```
cuonglm, line 31: Could we make this a function, so it's easier to test and maintain.

authorize/evaluator/evaluator.go
calebdoxsey: This will end up duplicating a lot of code but I will make the changes as you requested.

authorize/evaluator/evaluator.go
```diff
@@ -134,28 +136,28 @@ func (e *Evaluator) Evaluate(ctx context.Context, req *Request) (*Result, error)
 		return nil, fmt.Errorf("error signing JWT: %w", err)
 	}
 
+	evalResult := &Result{
+		MatchingPolicy: getMatchingPolicy(res[0].Bindings.WithoutWildcards(), e.policies),
+		SignedJWT:      signedJWT,
+	}
+	evalResult.UserEmail, evalResult.UserGroups = e.UserInfo(req)
```
cuonglm, line 143: This is not what I mean in previous review. What I mean is we should move the payload in `SignedJWT` to separated function, so there's not duplicated code. That said, currently:

```
signedJWT, err := e.SignedJWT(req)
if err != nil {
	return nil, fmt.Errorf("error signing JWT: %w", err)
}
```

becomes:

```
payload := jwtPayload(req)  // This will return a map[string]interface{}
signedJWT, err := e.SignedJWT(payload)
if err != nil {
	return nil, fmt.Errorf("error signing JWT: %w", err)
}

evalResult := &Result{
	MatchingPolicy: getMatchingPolicy(res[0].Bindings.WithoutWildcards(), e.policies),
	SignedJWT:      signedJWT,
}
if e, ok := payload["email"].(string); ok {
	evalResult.UserEmail = e
}
// Do the same for group
```

PR #1072, opened 2020-07-14
Add storage backend interface

cuonglm:
## Summary

Adding storage.Backend interface to support multiple backends in databroker.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
@cuonglm just a heads-up, but I intended the interface to be the databroker one, not the internal storage one. Notably the internal storage one does not support sync methods, so if we were running multiple databroker services (which we shouldn't do, but we have a task to support), there's no way for both of them to be informed of the changes without polling.

If however sync were implemented in the storage interface, we could use redis events or similar to implement the functionality.

That said there's a task to create this interface so I guess that's the direction we need to go.

PR #1073, opened 2020-07-14
pomerium-cli k8s exec-credential

calebdoxsey:
## Summary
Add a new `pomerium-cli k8s exec-credential` command to retrieve an authorization bearer token for use with Kubernetes. It is used like this:

```yaml
- name: via-pomerium
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - k8s
      - exec-credential
      - https://k8s.localhost.pomerium.io
      command: pomerium-cli
      env: null
```

The naming was modeled after `doctl`.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


cmd/pomerium-cli/cache.go
```diff
@@ -0,0 +1,80 @@
...
+func saveCachedCredential(serverURL string, creds *ExecCredential) {
+	fn := cachedCredentialPath(serverURL)
+
+	_ = os.MkdirAll(filepath.Dir(fn), 0755)
+
+	f, err := os.Create(fn)
+	if err != nil {
+		return
+	}
+	defer f.Close()
```
cuonglm, line 77: Since when you write to `f`, you must check the error when closing it, to make sure data was written successfully.

cmd/pomerium-cli/kubernetes.go
```diff
@@ -0,0 +1,244 @@
...
+	"golang.org/x/sync/errgroup"
+)
+
+func init() {
+	kubernetesCmd.AddCommand(kubernetesExecCredentialCmd)
+	rootCmd.AddCommand(kubernetesCmd)
+}
+
+var kubernetesCmd = &cobra.Command{
+	Use: "kubernetes",
```
cuonglm, line 28: Maybe `k8s` for short and idiom.

cmd/pomerium-cli/cache.go
calebdoxsey: We are ignoring all errors in this function. What do you want me to do if there is an error.

cuonglm: @calebdoxsey I think just emit a warning the saving credentials failed. So user won't be surprise.

calebdoxsey: We will now log errors.

FWIW as far as I know, writing files is a synchronous operation in Go. Any write errors will be surfaced in the call to `.Write`.

cuonglm: @calebdoxsey Yes, but what I mean is if `f.Close()` error, then everything you wrote previously will be lost. That's why you always have to check for error return by `f.Close()` _if_ you wrote something to file. Reading from file does not have this restriction.

PR #1075, opened 2020-07-14
authorize: Force redirect scheme to https

travisgroth:
## Summary
When building the redirect parameter, `authorize` currently attempts to use the `x-forwarded-proto` header from envoy.  Unfortunately this doesn't work correctly when in insecure mode, as it seems envoy sets `x-forwarded-proto` to the scheme it received the request on.

The result is a bad redirect and/or hmac error.  Given the nature of Pomerium, the easiest thing to do here is always set scheme to `https`.  

**Checklist**:
- [x] updated unit tests
- [x] ready for review


travisgroth:
Okay.  Yes, just changing the scheme in `redirectResponse()` does seem to function correctly.  I'm also getting what looks like a good response from forward auth without making changes in `handleForwardAuth()`.

PR #1080, opened 2020-07-15
implement google cloud serverless authentication

calebdoxsey:
## Summary
This PR adds google cloud serverless authentication support, primarily for cloud run. Changes:

* Two new config options: 
   * `google_cloud_serverless_authentication_service_account`: used to define the service account (not currently implemented because I don't know how to create the identity token using a service account)
   * `enable_google_cloud_serverless_authentication`: used to enable the authentication on a policy route
* Append the authorization header (when enabled) in the Authz OK response
* Add a GCP Identity Token Source which retrieves identity tokens using the compute metadata service and caches them
* Add an option to the xDS config to disable IPv6. During testing using IPv6 on a GCE instance to the cloudrun HTTP endpoint resulted in connection issues.

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/google_cloud_serverless.go
+++ b/authorize/google_cloud_serverless.go
@@ -2,7 +2,9 @@ package authorize
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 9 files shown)


authorize/google_cloud_serverless.go
```diff
@@ -0,0 +1,121 @@
...
+	ctx context.Context,
+	reply *evaluator.Result,
+) ([]*envoy_api_v2_core.HeaderValueOption, error) {
+	if reply.MatchingPolicy == nil || !reply.MatchingPolicy.EnableGoogleCloudServerlessAuthentication {
+		return nil, nil
+	}
+
+	svcAccount := a.currentOptions.Load().GoogleCloudServerlessAuthenticationServiceAccount
+	if svcAccount != "" {
+		panic("custom service account not implemented")
```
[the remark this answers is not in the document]
desimone: > used to define the service account (not currently implemented because I don't know how to create the identity token using a service account

🔜

authorize/google_cloud_serverless.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Get gets an identity token.
+func (src *GCPIdentityTokenSource) Get(ctx context.Context, audience string) (string, error) {
+	v, err, _ := src.Do(audience, func() (interface{}, error) {
+		src.mu.Lock()
+		tok, ok := src.tokens[audience]
+		src.mu.Unlock()
+
+		if ok && tok.Expires.Add(-gcpIdentityTokenGracePeriod).After(gcpIdentityNow()) {
```
desimone, line 57: Ah, I was a little fuzzy on grace period until I read this line of code. I had assumed grace period worked in the other direction.

authorize/google_cloud_serverless.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		req.Header.Add("Metadata-Flavor", "Google")
+
+		res, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return nil, err
+		}
+		defer res.Body.Close()
+
+		bs, err := ioutil.ReadAll(res.Body)
```
desimone, line 76: I know it's super unlikely but should we use a limit reader here just to be safe?

PR #1081, opened 2020-07-15
pkg/storage: add redis storage backend

cuonglm:
## Summary

Add redis storage backend.

## Related issues

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/pkg/storage/redis/redis.go
+++ b/pkg/storage/redis/redis.go
@@ -1,0 +1,169 @@
+package redis
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"sync/atomic"
+	"time"
+
+	"github.com/golang/protobuf/ptypes"
+	"github.com/gomodule/redigo/redis"
+	"google.golang.org/protobuf/types/known/anypb"
+
+	internal_databroker "github.com/pomerium/pomerium/internal/databroker"
+	"github.com/pomerium/pomerium/pkg/grpc/databroker"
+	"github.com/pomerium/pomerium/pkg/storage"
+)
+
+var _ storage.Backend = (*DB)(nil)
+
+// DB wraps redis conn to interact with redis server.
+type DB struct {
+	pool                   *redis.Pool
+	deletePermanentlyAfter int64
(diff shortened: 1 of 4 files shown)


pkg/storage/redis/redis.go
```diff
@@ -0,0 +1,169 @@
...
+	record := new(databroker.Record)
+	record.Data = data
+	if !exists {
+		record.CreatedAt = ptypes.TimestampNow()
+	}
+	record.ModifiedAt = ptypes.TimestampNow()
+	record.Type = db.recordType
+	record.Id = id
+	record.Version = fmt.Sprintf("%012X", atomic.AddUint64(&db.lastVersion, 1))
+	b, err := json.Marshal(record)
```
calebdoxsey, line 74: we should probably use the protobuf json: https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson?tab=doc

cuonglm: Not sure what's the different, but using protjson cause the test failed, while using standard encoding/json works.

calebdoxsey: > This package produces a different output than the standard "encoding/json" package, which does not operate correctly on protocol buffer messages.

pkg/storage/redis/redis.go
```diff
@@ -0,0 +1,169 @@
...
+
+	record := new(databroker.Record)
+	record.Data = data
+	if !exists {
+		record.CreatedAt = ptypes.TimestampNow()
+	}
+	record.ModifiedAt = ptypes.TimestampNow()
+	record.Type = db.recordType
+	record.Id = id
+	record.Version = fmt.Sprintf("%012X", atomic.AddUint64(&db.lastVersion, 1))
```
calebdoxsey, line 73: Maybe we should set `lastVersion` somehow? As written I believe if you updated a record, then restarted the process, then updated the same record again you'd end up with the same version number (`0`).

PR #1082, opened 2020-07-15
redis storage backend

cuonglm:
## Summary

Add redis storage backend

## Related issues

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/storage/redis/redis.go
```diff
@@ -0,0 +1,176 @@
...
+		record = new(databroker.Record)
+		record.CreatedAt = ptypes.TimestampNow()
+	}
+
+	record.Data = data
+	record.ModifiedAt = ptypes.TimestampNow()
+	record.Type = db.recordType
+	record.Id = id
+	record.Version = fmt.Sprintf("%012X", atomic.AddUint64(&db.lastVersion, 1))
+	b, err := json.Marshal(record)
```
calebdoxsey, line 75: Using encoding/json with protobuf is not recommended by the protobuf library: https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson?tab=doc

> This package produces a different output than the standard "encoding/json" package, which does not operate correctly on protocol buffer messages.

pkg/storage/redis/redis.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+var _ storage.Backend = (*DB)(nil)
+
+// DB wraps redis conn to interact with redis server.
+type DB struct {
+	pool                   *redis.Pool
+	deletePermanentlyAfter int64
+	recordType             string
+	lastVersion            uint64
```
calebdoxsey, line 27: Maybe we should set `lastVersion` somehow? As written I believe if you updated a record, then restarted the process, then updated the same record again you'd end up with the same version number (0).

calebdoxsey:
@cuonglm to implement sync properly we also need to store a `ServerVersion`. This way if you started the databroker against a new redis instance it would reset all the data of the consumers.

All Get and Sync calls return a server version.

I'm not entirely sure how this fits with the new storage backend interface, but the original databroker protobuf interface has this capability.

cuonglm:
> @cuonglm to implement sync properly we also need to store a `ServerVersion`. This way if you started the databroker against a new redis instance it would reset all the data of the consumers.
> 
> All Get and Sync calls return a server version.
> 
> I'm not entirely sure how this fits with the new storage backend interface, but the original databroker protobuf interface has this capability.

Can you elaborate more on `the original databroker protobuf interface has this capability.`. AFAICT, the databroker protobuf is still there.

Storing `ServerVersion` is doable, though.

calebdoxsey:
@cuonglm given the way this PR works. If you restart the databroker all consumers will erase their local stores and re-query for all the data, because the server version will have changed (it is randomly generated on start). That's not how it was intended to work.

cuonglm:
> @cuonglm given the way this PR works. If you restart the databroker all consumers will erase their local stores and re-query for all the data, because the server version will have changed (it is randomly generated on start). That's not how it was intended to work.

I understand that, but I guess it current behavior, without this PR, right? Just curious about the way you said the databroker interface has this capability.

For more clarification, the `ServerVersion` must be the same between all storage backend instance right? That said, if you have 2 databroker servers, with 2 redis instances, then the `ServerVersion` between them must be the same?

pkg/storage/redis/redis.go
cuonglm: Hmm, when I try switching to `protojson`, the test failed, and somehow hangs.

cuonglm:
> @cuonglm given the way this PR works. If you restart the databroker all consumers will erase their local stores and re-query for all the data, because the server version will have changed (it is randomly generated on start). That's not how it was intended to work.

I walk through the current implementation of databroker interface, there's no place that set `db.lastVersion` when server start, would you mind pointing out?

pkg/storage/redis/redis.go
cuonglm: Ah, maybe we should use `proto.Marshal` instead, it's not necessary to use `json` marshaler

pkg/storage/redis/redis.go
cuonglm: @calebdoxsey I'm a bit confused between this `db.lastVersion` and `databroker.Server.version`, would you mind elaborating?

Also in current implementation, I see `db.lastVersion` is increased everytime we insert or update new record to DB, is it intentional?

calebdoxsey: There are two versions: the record version and the server version. If the server version changes the consumer assumes all its current data is invalidated.

The current implementation is not persistent so the server version will change any time the process is restarted. If it were persistent the server version should not change, but the record version needs to increment properly.

I guess since we've decided not to implement actual persistence of the server version, the issue I'm pointing to won't happen.

pkg/storage/redis/redis.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// GetAll retrieves all records from redis.
+func (db *DB) GetAll(ctx context.Context) []*databroker.Record {
+	return db.getAll(ctx, func(record *databroker.Record) bool { return true })
+}
+
+// List retrieves all records since given version.
+func (db *DB) List(ctx context.Context, sinceVersion string) []*databroker.Record {
+	return db.getAll(ctx, func(record *databroker.Record) bool {
```
calebdoxsey, line 111: This is an inefficient implementation. List is called anytime a sync operation happens and this will be an `O(n)` operation. In the in-memory version this is indexed.

cuonglm:
> @cuonglm if you don't want to address the issues I've raised here, that's ok. I can fix them in a later PR.

Just want to keep PR scope as small as possible.

pkg/storage/redis/redis.go
cuonglm: Yes, any idea to improve it?

cuonglm: @calebdoxsey please see my updated implementation, which is more efficient than this one.

pkg/storage/redis/redis.go
```diff
@@ -0,0 +1,241 @@
...
+	b, err := proto.Marshal(record)
+	if err != nil {
+		return err
+	}
+	cmds := map[string][]interface{}{
+		"MULTI": nil,
+		"HSET":  {db.recordType, id, string(b)},
+		"ZADD":  {db.versionSet, db.lastVersion, id},
+	}
+	if _, err := db.tx(c, cmds); err != nil {
```
calebdoxsey, line 91: Map loops in go are non-deterministic. I think the `MULTI` needs to happen first, so you'll need to use a slice instead.

pkg/storage/redis/redis.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	r.DeletedAt = ptypes.TimestampNow()
+	b, err := proto.Marshal(r)
+	if err != nil {
+		return err
+	}
+	cmds := map[string][]interface{}{
+		"MULTI": nil,
+		"HSET":  {db.recordType, id, string(b)},
+		"SADD":  {db.deletedSet, id},
```
calebdoxsey, line 160: I think we need to do a ZADD and increment the version?

cuonglm: @calebdoxsey why 🤔 What is doing here is equivalent with what the `inmemory` backend does.

pkg/storage/redis/redis.go
cuonglm: Ah yes, my bad.

pkg/storage/redis/redis.go
calebdoxsey: I think if you delete a record, List won't pick up the change.

cuonglm: `List` will pick up if the key is still not deleted by `ClearDeleted` right? So I think it should be the right behavior.

cuonglm: @calebdoxsey isn't that we're checking that `DeletedAt` is not nil to detect this case? At least in `inmemory` implementation., I don't see it increase the version when delete one.

If that's the bug, I will fix in another PR.

calebdoxsey: The in memory version changes the version number when you call `Delete`.

```
record.Version = fmt.Sprintf("%012X", atomic.AddUint64(&db.lastVersion, 1))
```

cuonglm: @calebdoxsey ah ok, missing that part, Delete call replaceOrInsert as well

cuonglm: @calebdoxsey fixed.

pkg/storage/redis/redis.go
calebdoxsey: I don't think these changes fixed the issue.

calebdoxsey: `map[string][]interface{}` was changed to `[]map[string][]interface{}` but the loop over the map is still non-deterministic. `MULTI` can be called after `HSET`.

cuonglm: LOL, sorry, that the construction is wrong, fixed.

Issue #573, opened 2020-03-29
discuss: change policy to support unique entity id's only


desimone:
**Is your feature request related to a problem? Please describe.**

Pomerium currently supports writing declarative authorization policy using `allowed_users`,`allowed_groups`, and `AllowedDomains`.

OIDC has very few ["standardized" claims](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1). 

Some providers return all the data pomerium would want as part of the returned [ID TOKEN](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) however, that's the exception to the rule. For identity providers that do not provide fields like email, and group details, the authenticate service makes follow API calls to the identity provider to fully 'hydrate" pomerium user session.

Unfortunately, currently this means `allowed_users` or `allowed_groups` means different things to pomerium's policy depending on what identity provider you are using.  

For groups, this means:
- Google uses email address (`developers@pomerium.io`) which is unique, but not immutable
- Okta uses group UUID because no other field is unique or immutable
- OneLogin uses group UUID because no other field is unique or immutable
- Azure uses group UUID because no other field is unique or immutable

For users:
- All the providers with the exception of azure reliably return email as part of either the initial id_token, or as part of the follow up rest call the the [UserInfo](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo).

This has caused confusion and possibly unintended security issues in the past when identity providers like GitLab used `name` to map group entities to policy.  

This is complicated even further when some providers like Azure AD are not guaranteed to return a user's email at all. #520 

**Describe the solution you'd like**

Policies should be written using entities unique id only (typically `sub` for user, and `id` for groups as part of the follow up group api call). 

PRO

- This would ensure that policy remain explicit and immutable. 
- Identity provider implementations would be standardized which would improve multi-IdP support amongst other things

CON

- Breaking changes. We'd be asking user's to rewrite their policies using user id and group id not user email and group name (in the case of gsuite). 
- Worse user experience (most operators know the group names they are working with, but seldom know the group ids). 

**Describe alternatives you've considered**

Keep things as is they are. 

**Explain any additional use-cases**

**Additional context**

- The planned GUI for policy management could be used to offset the UX issues of using ID to write policy (we could do a rest call in the background to lookup an entity's ID, while still letting the operator use the human friendly entity name / email).

desimone:
[...]

PR #1083, opened 2020-07-16
internal/directory/google: return both group e-mail and id

### LINKED ISSUE #1079: cache: google directory sync not returning group e-mails
## What happened?

Group membership detection does not seem to be working on master.  

## How'd it happen?

1. Started with good `v0.9` configuration with an idp service account.
2. Upgrade to `master`
3. 403 Error

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint): `v0.9.0+a5db944`
- Server Operating System/Architecture/Cloud: local all-in-one / OSX

## What's your config.yaml?

```config.yaml
address: :80
grpc_insecure: true
insecure_server: true
debug: true
log_level: debug
authenticate_service_url: https://authn.localhost.pomerium.io
shared_secret: XXXXXXX
cookie_secret: XXXXXXX
idp_provider: "google"
idp_client_id: "XXXXXXX"
idp_client_secret: "XXXXXXX"
idp_service_account: "XXXXXXX"
metrics_address: :9090
policy:
    to: https://httpbin.org
    # allowed_domains:
    #   - pomerium.com
    #   - gmail.com
    allowed_groups:
      - XXXX
      - XXXX@pomerium.com
```

```logs
{"level":"info","service":"authorize","request-id":"67b6b65b-ceb5-473f-89c9-d07f5bb2586a","check-request-id":"308fed07-8570-42c7-853e-b495beba0812","method":"GET","path":"/","host":"httpbin.localhost.pomerium.io","query":"","allow":false,"status":403,"message":"forbidden","headers":{":authority":"httpbin.localhost.pomerium.io",":method":"GET",":path":"/","Accept":"text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9","Accept-Encoding":"gzip, deflate, br","Accept-Language":"en-US,en;q=0.9","Cookie":"_pomerium_csrf=YYYYY; _pomerium=YYYYY","Sec-Fetch-Dest":"document","Sec-Fetch-Mode":"navigate","Sec-Fetch-Site":"cross-site","Sec-Fetch-User":"?1","Upgrade-Insecure-Requests":"1","User-Agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36","X-Envoy-Internal":"true","X-Forwarded-For":"192.168.0.154","X-Forwarded-Proto":"http","X-Request-Id":"308fed07-8570-42c7-853e-b495beba0812"},"time":"2020-07-15T10:49:45-04:00","message":"authorize check"}
```

## Additional context

Here's the resulting session details.  Looks like a decoding issue or an identifier we need to follow.

image

travisgroth:
## Summary
For the Google directory provider, add both the immutable group ID and normal group e-mail to a user.  

Rationale - this allows the operator's discretion as to which type of group identifier to use, and restores the ability to use the previous (though imperfect) behavior.  

## Related issues
Closes #1079 

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/internal/directory/google/google.go
+++ b/internal/directory/google/google.go
@@ -88,13 +88,15 @@ func (p *Provider) UserGroups(ctx context.Context) ([]*directory.User, error) {
 		return nil, fmt.Errorf("google: error getting API client: %w", err)
 	}
 
+	groupIDToEmails := map[string]string{}
 	var groups []string
 	err = apiClient.Groups.List().
 		Context(ctx).
 		Customer("my_customer").
 		Pages(ctx, func(res *admin.Groups) error {
 			for _, g := range res.Groups {
-				groups = append(groups, g.Id, g.Email)
+				groups = append(groups, g.Id)
+				groupIDToEmails[g.Id] = g.Email
 			}
 			return nil
 		})
@@ -109,7 +111,7 @@ func (p *Provider) UserGroups(ctx context.Context) ([]*directory.User, error) {
 			Context(ctx).
 			Pages(ctx, func(res *admin.Members) error {
 				for _, member := range res.Members {
-					userIDToGroups[member.Id] = append(userIDToGroups[member.Id], group)
+					userIDToGroups[member.Id] = append(userIDToGroups[member.Id], group, groupIDToEmails[group])
 				}
 				return nil
 			})
--- a/internal/directory/google/google.go
+++ b/internal/directory/google/google.go
@@ -94,7 +94,7 @@ func (p *Provider) UserGroups(ctx context.Context) ([]*directory.User, error) {
 		Customer("my_customer").
 		Pages(ctx, func(res *admin.Groups) error {
 			for _, g := range res.Groups {
-				groups = append(groups, g.Id)
+				groups = append(groups, g.Id, g.Email)
 			}
 			return nil
 		})


internal/directory/google/google.go
```diff
@@ -94,7 +94,7 @@ func (p *Provider) UserGroups(ctx context.Context) ([]*directory.User, error) {
 		Customer("my_customer").
 		Pages(ctx, func(res *admin.Groups) error {
 			for _, g := range res.Groups {
-				groups = append(groups, g.Id)
+				groups = append(groups, g.Id, g.Email)
```
cuonglm, line 97: I don't think this change is enough and works. In the for loop several lines below, we do a map from userid -> group id, by calling `apiClient.Member.List(group)`, using group id.

I think you need another map from groupID -> groupName, then append the groupName to `userIDToGroups` whenever you append corresponding groupID.

travisgroth: It did work when I tested.  It seems List() accepts a few different formats - https://developers.google.com/admin-sdk/directory/v1/reference/members/list.

Should probably do something similar to your suggestion regardless, though.  We're now looking up each group's members twice.  Not ideal if you had, say 5000 groups 😄

cuonglm:
> LGTM and should support the way Google previously worked (if inconsistent with the other providers).
> 
> Should we do this for _all_ the providers?
> 
> [#573 (comment)](https://github.com/pomerium/pomerium/issues/573#issuecomment-659009016)

We should, IIRC, all other providers use group ID :|

internal/directory/google/google.go
desimone: Related question, should we adjust "max results" to something like 200? The docs are very unclear what the default return value is.

cuonglm: The default is 200:

```
Groups and group members – A default and maximum of 200 entries per page.
```

https://developers.google.com/admin-sdk/directory/v1/limits

PR #1097, opened 2020-07-17
ci: Add cloudrun build

travisgroth:
## Summary
Add image tailored for cloudrun deployment.

- Pushes image to public GCR
- Add GCP secrets friendly entrypoint
- Some requisite defaults set in ENV

## Using

This image will perform substitution of env vars via [vals](github.com/variantdev/vals) syntax.  For Cloud Run, the assumption is that GCP Secrets Manager will be used for sensitive parameters - `ref+gcpsecerets://myproject/mysecret[#/mykey]`

To source the config file from Secrets Manager, set `VALS_FILES=/pomerium/config.yaml:ref+gcpsecrets://myproject/pomerium-config`

### Example

Source main config from a secret called `pomerium-config` for shared secrets, source `IDP_SERVICE_ACCOUNT ` from a dedicated secret, and augment with a policy sourced locally at deploy time.

```shell
gcloud run deploy [SERVICE] \
--image gcr.io/pomerium-io/pomerium:[version]-cloudrun \
--set-env-vars VALS_FILES='/pomerium/config.yaml:ref+gcpsecrets://myproject/pomerium-config',IDP_SERVICE_ACCOUNT='ref+gcpsecrets://myproject/pomerium-idp-service-account',POLICY=$(cat policy.yaml | base64)
```

## Additional Info 

- At this time `--max-instances` should be set to 1 to minimize session loss.
- Custom domains are required to at least differentiate the `authenticate` endpoint from proxied endpoints.
- Docs update in follow up PR to go with GCP Serverless Authentication support

**Checklist**:
- [x] ready for review

### DIFF
--- a/.github/Dockerfile-cloudrun
+++ b/.github/Dockerfile-cloudrun
@@ -10,7 +10,7 @@ COPY pomerium* /bin/
 COPY --from=build /config.yaml /pomerium/config.yaml
 COPY --from=entrypoint /bin/vals-entrypoint /bin/vals-entrypoint
 
-ENV ADDRESS ":80"
+ENV ADDRESS ":8080"
 ENV GRPC_INSECURE true
 ENV INSECURE_SERVER true
 
--- a/.github/Dockerfile-cloudrun
+++ b/.github/Dockerfile-cloudrun
@@ -1,0 +1,18 @@
+FROM pomerium/vals-entrypoint as entrypoint
+
+FROM busybox:latest as build
+RUN touch /config.yaml
+
+FROM gcr.io/distroless/base
+ENV AUTOCERT_DIR /data/autocert
+WORKDIR /pomerium
+COPY pomerium* /bin/
+COPY --from=build /config.yaml /pomerium/config.yaml
+COPY --from=entrypoint /bin/vals-entrypoint /bin/vals-entrypoint
+
+ENV ADDRESS ":80"
+ENV GRPC_INSECURE true
+ENV INSECURE_SERVER true
+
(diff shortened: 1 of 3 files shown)


.github/Dockerfile-cloudrun
```diff
@@ -0,0 +1,18 @@
+FROM pomerium/vals-entrypoint as entrypoint
+
+FROM busybox:latest as build
+RUN touch /config.yaml
+
+FROM gcr.io/distroless/base
```
travisgroth, line 6: I went back and forth on this.  If we switched to an image with a shell environment, we could make the interface for the config file slightly cleaner, eg "CONFIG=[ref]".  I _think_ sticking with our standard image is worth slightly more complexity but open to other opinions on usability.

.github/goreleaser.yaml
```diff
@@ -105,3 +105,18 @@ dockers:
       - "--label=org.opencontainers.image.source={{.GitURL}}"
       - "--label=repository=http://github.com/pomerium/pomerium"
       - "--label=homepage=http://www.pomerium.io"
+
+  - image_templates:
+      - "gcr.io/pomerium-io/pomerium:{{ .Tag }}-cloudrun"
```
travisgroth, line 110: We are not publishing a latest tag from github actions.  (a) slightly less changes to maintain (b) cloudrun pins you to a commit hash of any tag anyway.  I think cosmetic tags are apt to create slightly more confusion in usage/upgrades.

travisgroth, line 110: We may want to add other regions here.

Issue #567, opened 2020-03-27
G Suite Service Account Group Membership Validation Fails


shreyaskarnik:
## What happened?
* Running on Kubernetes with the helm chart along with nginx ingress 
* We use G Suite Service account for group membership validation. 
* We have two users A and B
* user A is part of roughly 20 something groups
* user B is part of roughly 60 something groups (including owner of some groups)
## What did you expect to happen?
* Group membership works for user A and user A can access everything downstream fine
* For user B they get an error `ERR_CONNECTION_CLOSED` on Chrome and a blank page on Firefox
* When tried on Safari user got `kCFErrorDomainCFNetwork error 303`
* Also when user A who is an Pomerium admin tried to log in they got an 403 (creating separate issue for that)

## How'd it happen?

1. User B tried to log in

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint):
```bash
pomerium/v0.6.3 (+github.com/pomerium/pomerium; 1c7d30b; go1.14)
```
- Server Operating System/Architecture/Cloud:

## What's your config.yaml?

```config.yaml
- from: https://EXT
  to: http://INT
  allowed_groups:
    - admins@user.com
```

## What did you see in the logs?

```logs
[
    {
        "level": "debug",
        "X-Forwarded-For": [
            "XXXXX"
        ],
        "X-Forwarded-Host": [
            "INTERNAL"
        ],
        "X-Forwarded-Port": [
            "443"
        ],
        "X-Forwarded-Proto": [
            "https"
        ],
        "X-Real-Ip": [
            "XXXXX"
        ],
        "ip": "XXXXX",
        "user_agent": "REMOVED",
        "req_id": "f9ec84a4-14d9-ecc1-4632-aee4d192bd76",
        "error": "Forbidden: user@user.com is not authorized for INTERNAL",
        "time": "2020-03-26T19:18:33Z",
        "message": "proxy: AuthorizeSession"
    },
    {
        "level": "info",
        "X-Forwarded-For": [
            "INTERNAL"
        ],
        "X-Forwarded-Host": [
            "INTERNAL"
        ],
        "X-Forwarded-Port": [
            "443"
        ],
        "X-Forwarded-Proto": [
            "https"
        ],
        "X-Real-Ip": [
            "INTERNAL"
        ],
        "ip": "XXXXX",
        "user_agent": "REMOVED",
[...]

PR #1098, opened 2020-07-17
config: allow setting directory sync interval and timeout

cuonglm:
## Related issues
Updates #567

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


cuonglm:
> I think we should split the docs into two entries. Thoughts?

I am inspired by https://www.pomerium.io/configuration/#global-timeouts

I'm leaning to group them together, because these 2 settings are tied with each other, like setting timeout to 10 minutes, but interval to 1 minute make no sense.

PR #1099, opened 2020-07-17
config: add support for policies stored in the databroker

calebdoxsey:
## Summary
This adds support for storing config policy in the databroker.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/databroker/config_source.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func (src *ConfigSource) rebuild(firstTime bool) {
+	src.mu.Lock()
+	defer src.mu.Unlock()
+
+	cfg := src.underlyingConfig.Clone()
+	defer func() {
+		src.computedConfig = cfg
+		if !firstTime {
```
cuonglm, line 79: Could we use `sync.Once` here instead?

calebdoxsey: This is doing the opposite of sync.Once. It runs every time except the first time.

cuonglm: @calebdoxsey Ah right 👍

Issue #981, opened 2020-06-23
Crash when ipv6 is disabled


alagendijk-minddistrict:
## What happened?

Pomerium versions 0.9.0, 0.9.1 and 0.9.2 crash when ran on an Ubuntu 18.04 machine where ipv6 is explicitly disabled by using GRUB_CMDLINE_LINUX="ipv6.disable=1" in /etc/default/grub
a build from master branch does not have this issue

## How'd it happen?

On Ubuntu 18.04 where pomerium (0.9.0, 0.9.1 or 0.9.2) works correctly:
1. Set GRUB_CMDLINE_LINUX="ipv6.disable=1" in /etc/default/grub
2. Run sudo update-grub
3. Reboot
4. start pomerium
5. See error below

## What's your environment like?

- Pomerium 0.9.0, 0.9.1 and 0.9.1 **a build from master branch does work**
- Server Operating System/Architecture/Cloud: Ubuntu 18.04

## What's your config.yaml?

```config.yaml
authenticate_service_url: https://authenticate.corp.beyondperimeter.com
insecure_server: true
policy:
  - from: https://httpbin.corp.beyondperimeter.com
    to: http://httpbin
    allowed_domains:
      - pomerium.io
    cors_allow_preflight: true
    timeout: 30s
  - from: https://external-httpbin.corp.beyondperimeter.com
    to: https://httpbin.org
    allowed_domains:
      - gmail.com
  - from: https://hello.corp.beyondperimeter.com
    to: http://hello:8080
    allowed_groups:
      - admins@pomerium.io

signing_key: [redacted]

shared_secret: [redacted]
cookie_secret: [redacted]

idp_provider: [redacted]
idp_provider_url: [redacted]
idp_client_id: [redacted]
idp_client_secret: [redacted]
```

## What did you see in the logs?

```logs
{"level":"debug","service":"envoy","name":"main","time":"2020-06-22T17:06:22+02:00","message":"std::terminate called! (possible uncaught exception, see trace)"}
{"level":"debug","service":"envoy","name":"backtrace","time":"2020-06-22T17:06:22+02:00","message":"[bazel-out/k8-opt/bin/external/envoy/source/server/_virtual_includes/backtrace_lib/server/backtrace.h:91] Backtrace (use tools/stack_decode.py to get line numbers):"}
[...]

PR #1116, opened 2020-07-20
config: Set loopback address by ipv4 IP

travisgroth:
## Summary
Update configuration defaults to use `127.0.0.1` instead of `localhost` in an attempt to avoid IPv6 issues present in some environments.  The most common problem is a mismatch between the host OS resolving `localhost` to ipv6, but a container engine not enabling ipv6.

This change should at least eliminate some errors that result in user confusion about a broken control plane or all-in-one mode.

## Related issues
https://github.com/pomerium/pomerium/issues/981
https://github.com/pomerium/pomerium/issues/821
https://github.com/pomerium/pomerium/issues/811

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -36,7 +36,7 @@ const DisableHeaderKey = "disable"
 const DefaultAlternativeAddr = ":5443"
 
 // EnvoyAdminURL indicates where the envoy control plane is listening
-var EnvoyAdminURL = &url.URL{Host: "localhost:9901", Scheme: "http"}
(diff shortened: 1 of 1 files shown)


desimone:
@travisgroth thoughts on back-porting?

travisgroth:
@desimone i went back and forth on that.  It does seem like it should be more likely to fix than break deployments.  I'll label for backporting.

PR #1118, opened 2020-07-20
directory.Group entry for groups

calebdoxsey:
## Summary
This PR changes how we store groups. We will now store a separate `directory.Group` entry in the databroker and the `directory.User` `groups` field has been renamed to `group_ids`. On evaluation we then look up the corresponding `directory.Group` based on that ID.

I updated all the directory providers to return this information and updated the manager to sync it accordingly. Other changes:

1. The Azure group name was `displayName` not `name`.
1. Using field selection with Google resulted in a 400 error when I tested it locally, so I removed it and returned everything.

## Related
(bdd edit)
- https://github.com/pomerium/pomerium/issues/573

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/directory/google/google.go
```diff
@@ -82,62 +82,62 @@ func New(options ...Option) *Provider {
...
 			for _, g := range res.Groups {
 				// Skip group without member.
 				if g.DirectMembersCount == 0 {
 					continue
 				}
-				groups = append(groups, g.Id)
-				groupIDToEmails[g.Id] = g.Email
+				groups = append(groups, &directory.Group{
+					Id:    g.Id,
+					Name:  g.Name,
```
desimone, line 103: To be backwards compatible, I think we should use Group as Name. 

https://github.com/pomerium/pomerium/blob/50f272fa707d77941ea723dea02715ea8db9d777/internal/identity/oidc/google/google.go#L121-L123

calebdoxsey: Not sure I understand the reasoning here, but I will change it.

PR #1121, opened 2020-07-21
internal/databroker: store server version

cuonglm:
## Summary
Storing server version when creating new server. After then, we can
retrieve the version from backend when server restart.

With storage backend which supports persistent, the server version
won't change after restarting.

## Related issues

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
I guess this works but I'm not sure what storing data in the databroker that's not protobuf-encoded will do.

cuonglm:
> I guess this works but I'm not sure what storing data in the databroker that's not protobuf-encoded will do.

It's protobuf though, that's any.Any 😄

I don't think there's any problem with it, all test cases prove that.

calebdoxsey:
The `Any` type stores encoded protobuf types. To properly use an `Any` type you have to use `ptypes.MarshalAny`.

Our test suite is not exhaustive. But if you're confident using an invalid Any type won't cause any issues I guess we can merge this.

cuonglm:
> The `Any` type stores encoded protobuf types. To properly use an `Any` type you have to use `ptypes.MarshalAny`.
> 
> Our test suite is not exhaustive. But if you're confident using an invalid Any type won't cause any issues I guess we can merge this.

Hmm, I re-think about that, maybe just better to stick with the protobuf documentation. Updated with proper type for server version.

PTAL, thanks.

PR #1126, opened 2020-07-22
internal/controlplane: using envoy strip host port matching

### LINKED ISSUE #959: 404 if Host header include port number 443
## What happened?

If `Host` header includes the standard 443 port number, Pomerium will return 404.

```
$ curl -I -H 'Host: foo.example.com:443' https://foo.example.com
HTTP/2 404 
date: Mon, 22 Jun 2020 07:23:02 GMT
server: envoy

$ curl -I -H 'Host: foo.example.com' https://foo.example.com 
HTTP/2 302 
content-length: 5
content-type: text/plain
location: LOGIN_URL
date: Mon, 22 Jun 2020 07:23:52 GMT
server: envoy
```

Should handle the request correctly.

## How'd it happen?

1. Ran `x`
2. Clicked `y`
3. Saw error `z`

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint):
pomerium/v0.9.0 (+github.com/pomerium/pomerium; 84dde09; go1.14.4)
Docker

## What's your config.yaml?
N/A

```logs
12:23AM INF http-request authority=foo.example.com:443 duration=0.126206 forwarded-for= method=HEAD path=/ referer= request-id=63ed7a23-69ba-48d0-990a-b3c079608df2 response-code=404 response-code-details=route_not_found service=envoy size=0 upstream-cluster= user-agent=curl/7.64.0
12:23AM INF http-request authority=foo.example.com duration=3.894968 forwarded-for= method=HEAD path=/ referer= request-id=aa62dd7c-6506-40c2-b6f6-43247ffb3849 response-code=302 response-code-details=ext_authz_denied service=envoy size=0 upstream-cluster= user-agent=curl/7.64.0
```

## Additional context

Similar but probably unrelated bug https://github.com/pomerium/pomerium/issues/352

cuonglm:
## Summary
With envoy 1.15.0 release, strip host port matching setting allows
incoming request with Host "example:443" will match again route with
domains match set to "example".

Not that this is not standard HTTP behavior, but it's more convenient
for users.

## Related issues

Fixes #959

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -10,6 +10,7 @@
 - authenticate: allow hot reloaded admin users config @cuonglm [GH-984]
 - authorize: include "kid" in JWT headers @cuonglm [GH-1046]
 - config: both base64 and file reference can be used for "certificates" @dmitrif [GH-1055]
+- envoy: enable strip host port matching @cuonglm [GH-1126]
 
 ### Changes
 
--- a/internal/controlplane/xds_listeners.go
+++ b/internal/controlplane/xds_listeners.go
@@ -225,8 +225,9 @@ func buildMainHTTPConnectionManagerFilter(options *config.Options, domains []str
 			RandomSampling: &envoy_type_v3.Percent{Value: options.TracingSampleRate * 100},
 		},
 		// See https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for
-		UseRemoteAddress: &wrappers.BoolValue{Value: true},
-		SkipXffAppend:    false,
+		UseRemoteAddress:      &wrappers.BoolValue{Value: true},
+		SkipXffAppend:         false,
+		StripMatchingHostPort: true,
 	})
 
 	return &envoy_config_listener_v3.Filter{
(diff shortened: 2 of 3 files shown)


desimone:
\cc @yegle

calebdoxsey:
We already account for this when we create the list of domains. Making this change may break existing functionality.

cuonglm:
> We already account for this when we create the list of domains. Making this change may break existing functionality.

Would you mind elaborating? Which case it will break?

AFAIU, this only affects how envoy handle request with port in Host header, and it will cause not found behavior in previous version -> valid route.

calebdoxsey:
We define routes with domains like `authorize.localhost.pomerium.io:5433`. We already add the port. 

I don't know if stripping the port from the incoming host name will still match that virtual domain or not. I would guess no, but I haven't tested it.

cuonglm:
> We define routes with domains like `authorize.localhost.pomerium.io:5433`. We already add the port.
> 
> I don't know if stripping the port from the incoming host name will still match that virtual domain or not. I would guess no, but I haven't tested it.

Reading [this](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-strip-matching-host-port), the port only remove if it is euqal to local listener port.

PR #1129, opened 2020-07-23
internal/databroker: handle new db error

cuonglm:
## Summary
Since when we now support other storage, not only memory storage, we
need to handle the error when we can't connect to storage.

## Related issues

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -3,6 +3,7 @@ package databroker
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"sort"
 	"sync"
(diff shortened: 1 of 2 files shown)


internal/databroker/server.go
```diff
@@ -122,7 +133,11 @@ func (srv *Server) Get(ctx context.Context, req *databroker.GetRequest) (*databr
 		Str("id", req.GetId()).
 		Msg("get")
 
-	record := srv.getDB(req.GetType()).Get(ctx, req.GetId())
+	db, err := srv.getDB(req.GetType())
+	if err != nil {
+		return nil, err
+	}
+	record := db.Get(ctx, req.GetId())
```
calebdoxsey, line 140: can't all these methods return errors too?

cuonglm: @calebdoxsey Yes, but current backend interface does not return error, let do it in other PR

PR #1131, opened 2020-07-23
pkg/storage: change backend interface to return error

cuonglm:
## Summary
Since when storage backend like redis can be fault in many cases, the
interface should return error for the caller to handle.

## Related issues

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -86,7 +86,7 @@ func (srv *Server) initVersion() {
 	}
 
 	// Get version from storage first.
-	if r := dbServerVersion.Get(context.Background(), serverVersionKey); r != nil {
+	if r, _ := dbServerVersion.Get(context.Background(), serverVersionKey); r != nil {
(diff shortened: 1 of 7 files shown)


internal/databroker/server.go
```diff
@@ -86,7 +86,7 @@ func (srv *Server) initVersion() {
 	}
 
 	// Get version from storage first.
-	if r := dbServerVersion.Get(context.Background(), serverVersionKey); r != nil {
+	if r, _ := dbServerVersion.Get(context.Background(), serverVersionKey); r != nil {
```
desimone, line 89: Should we handle?

cuonglm: If we can't get the version from storage, then we start with a fresh one, so I don't think what else we can do here.

PR #1135, opened 2020-07-24
pkg/storage: introduce storage.Backend Watch method

cuonglm:
## Summary
Currently, we're doing "sync" in databroker server. If we're going to
support multiple databroker servers instance, this mechanism won't work.

This commit moves the "sync" to storage backend, by adding new Sync
method. The Sync method will return a chanel for the caller. Everytime
something happens inside the storage, we notify the caller by sending a
message to the channel.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -225,7 +225,7 @@ func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBroke
 	for range ch {
 		updated, err := db.List(ctx, recordVersion)
 		if err != nil {
-			panic(err)
+			return err
 		}
 		if len(updated) > 0 {
 			sort.Slice(updated, func(i, j int) bool {
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -198,8 +198,8 @@ func (srv *Server) Set(ctx context.Context, req *databroker.SetRequest) (*databr
 	}, nil
 }
 
-func (srv *Server) doSync(ctx context.Context, recordVersion string, db storage.Backend, stream databroker.DataBrokerService_SyncServer) error {
-	updated, err := db.List(ctx, recordVersion)
+func (srv *Server) doSync(ctx context.Context, recordVersion *string, db storage.Backend, stream databroker.DataBrokerService_SyncServer) error {
+	updated, err := db.List(ctx, *recordVersion)
 	if err != nil {
 		return err
 	}
@@ -209,7 +209,7 @@ func (srv *Server) doSync(ctx context.Context, recordVersion string, db storage.
 	sort.Slice(updated, func(i, j int) bool {
 		return updated[i].Version < updated[j].Version
 	})
-	recordVersion = updated[len(updated)-1].Version
+	*recordVersion = updated[len(updated)-1].Version
 	return stream.Send(&databroker.SyncResponse{
 		ServerVersion: srv.version,
 		Records:       updated,
@@ -238,16 +238,15 @@ func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBroke
 	}
 
 	ctx := stream.Context()
-
 	// Do first sync, so we won't missed anything.
-	if err := srv.doSync(ctx, recordVersion, db, stream); err != nil {
+	if err := srv.doSync(ctx, &recordVersion, db, stream); err != nil {
 		return err
 	}
 
 	ch := db.Sync(ctx)
 
 	for range ch {
-		if err := srv.doSync(ctx, recordVersion, db, stream); err != nil {
+		if err := srv.doSync(ctx, &recordVersion, db, stream); err != nil {
 			return err
 		}
 	}
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -243,7 +243,7 @@ func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBroke
 		return err
 	}
 
-	ch := db.Sync(ctx)
+	ch := db.Watch(ctx)
 
 	for range ch {
 		if err := srv.doSync(ctx, &recordVersion, db, stream); err != nil {
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -198,6 +198,24 @@ func (srv *Server) Set(ctx context.Context, req *databroker.SetRequest) (*databr
 	}, nil
 }
 
+func (srv *Server) doSync(ctx context.Context, recordVersion string, db storage.Backend, stream databroker.DataBrokerService_SyncServer) error {
+	updated, err := db.List(ctx, recordVersion)
+	if err != nil {
+		return err
+	}
+	if len(updated) == 0 {
+		return nil
+	}
+	sort.Slice(updated, func(i, j int) bool {
+		return updated[i].Version < updated[j].Version
+	})
+	recordVersion = updated[len(updated)-1].Version
+	return stream.Send(&databroker.SyncResponse{
+		ServerVersion: srv.version,
+		Records:       updated,
+	})
+}
+
 // Sync streams updates for the given record type.
 func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBrokerService_SyncServer) error {
 	_, span := trace.StartSpan(stream.Context(), "databroker.grpc.Sync")
@@ -220,25 +238,17 @@ func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBroke
 	}
 
 	ctx := stream.Context()
+
+	// Do first sync, so we won't missed anything.
+	if err := srv.doSync(ctx, recordVersion, db, stream); err != nil {
+		return err
+	}
+
 	ch := db.Sync(ctx)
 
 	for range ch {
-		updated, err := db.List(ctx, recordVersion)
-		if err != nil {
+		if err := srv.doSync(ctx, recordVersion, db, stream); err != nil {
 			return err
-		}
-		if len(updated) > 0 {
-			sort.Slice(updated, func(i, j int) bool {
-				return updated[i].Version < updated[j].Version
-			})
-			recordVersion = updated[len(updated)-1].Version
-			err := stream.Send(&databroker.SyncResponse{
-				ServerVersion: srv.version,
-				Records:       updated,
-			})
-			if err != nil {
-				return err
-			}
 		}
 	}
 	return nil
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -18,6 +18,7 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 
 	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/internal/signal"
 	"github.com/pomerium/pomerium/internal/telemetry/trace"
 	"github.com/pomerium/pomerium/pkg/grpc/databroker"
 	"github.com/pomerium/pomerium/pkg/storage"
@@ -36,9 +37,9 @@ type Server struct {
 	cfg     *serverConfig
 	log     zerolog.Logger
 
-	mu       sync.RWMutex
-	byType   map[string]storage.Backend
-	onchange *Signal
+	mu           sync.RWMutex
+	byType       map[string]storage.Backend
+	onTypechange *signal.Signal
 }
 
 // New creates a new server.
@@ -49,8 +50,8 @@ func New(options ...ServerOption) *Server {
 		cfg:     cfg,
 		log:     log.With().Str("service", "databroker").Logger(),
 
-		byType:   make(map[string]storage.Backend),
-		onchange: NewSignal(),
+		byType:       make(map[string]storage.Backend),
+		onTypechange: signal.New(),
 	}
 	srv.initVersion()
 
@@ -109,8 +110,6 @@ func (srv *Server) Delete(ctx context.Context, req *databroker.DeleteRequest) (*
 		Str("type", req.GetType()).
 		Str("id", req.GetId()).
 		Msg("delete")
-
-	defer srv.onchange.Broadcast()
 
 	db, err := srv.getDB(req.GetType())
 	if err != nil {
@@ -182,8 +181,6 @@ func (srv *Server) Set(ctx context.Context, req *databroker.SetRequest) (*databr
 		Str("id", req.GetId()).
 		Msg("set")
 
-	defer srv.onchange.Broadcast()
-
 	db, err := srv.getDB(req.GetType())
 	if err != nil {
 		return nil, err
@@ -222,10 +219,14 @@ func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBroke
 		return err
 	}
 
-	ch := srv.onchange.Bind()
-	defer srv.onchange.Unbind(ch)
-	for {
-		updated, _ := db.List(context.Background(), recordVersion)
+	ctx := stream.Context()
+	ch := db.Sync(ctx)
+
+	for range ch {
+		updated, err := db.List(ctx, recordVersion)
+		if err != nil {
+			panic(err)
+		}
 		if len(updated) > 0 {
 			sort.Slice(updated, func(i, j int) bool {
 				return updated[i].Version < updated[j].Version
@@ -239,13 +240,8 @@ func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBroke
 				return err
 			}
 		}
-
-		select {
-		case <-stream.Context().Done():
-			return stream.Context().Err()
-		case <-ch:
-		}
 	}
+	return nil
 }
 
 // GetTypes returns all the known record types.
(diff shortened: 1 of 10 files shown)


internal/databroker/server.go
```diff
@@ -222,10 +219,14 @@ func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBroke
...
 
-	ch := srv.onchange.Bind()
-	defer srv.onchange.Unbind(ch)
-	for {
-		updated, _ := db.List(context.Background(), recordVersion)
+	ctx := stream.Context()
+	ch := db.Sync(ctx)
+
+	for range ch {
+		updated, err := db.List(ctx, recordVersion)
```
travisgroth, line 226: Is there a race condition here from the start of the request to the db.Sync() call?  It looks like we could miss an update as we set up the db.Sync().

pkg/storage/redis/redis.go
```diff
@@ -208,6 +210,50 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
 	}
 }
 
+// Sync returns a channel to the caller, when there is a change to the version set,
+// sending message to the channel to notify the caller.
+func (db *DB) Sync(ctx context.Context) chan struct{} {
```
travisgroth, line 215: nit: maybe call this something else to avoid confusion?  Sub or Watch?

pkg/storage/redis/redis.go
```diff
@@ -208,6 +210,50 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
...
+			return
+		}
+
+		psc := redis.PubSubConn{Conn: c}
+		if err := psc.PSubscribe("__keyspace*__:" + db.versionSet); err != nil {
+			log.Error().Err(err).Msg("failed to subscribe to version set channel")
+			return
+		}
+		for {
+			select {
```
travisgroth, line 236: It might be a good idea to add in a randomized timer to trigger a List() in the caller every few minutes to mitigate missed notifications.

pkg/storage/inmemory/inmemory.go
```diff
@@ -122,12 +129,24 @@ func (db *DB) List(_ context.Context, sinceVersion string) ([]*databroker.Record
...
 		record.Data = data
 	})
 	return nil
 }
 
+// Sync returns the underlying signal.Signal binding channel to the caller.
+// Then the caller can listen to the channel for detecting changes.
+func (db *DB) Sync(ctx context.Context) chan struct{} {
+	go func() {
+		defer db.onchange.Unbind(db.syncCh)
```
calebdoxsey, line 143: why do we unbind on ctx.Done? If I used a DB across multiple connections wouldn't this result in all signals being stopped?

pkg/storage/inmemory/inmemory.go
```diff
@@ -122,12 +129,24 @@ func (db *DB) List(_ context.Context, sinceVersion string) ([]*databroker.Record
...
+		<-ctx.Done()
+	}()
+
+	return db.syncCh
```
calebdoxsey, line 147: If we re-use the same channel then only one consumer will receive the notifications.

internal/databroker/server.go
cuonglm: If something happens before client call `Sync`, then client will be lost that notifications. We can't buffer the notification.

pkg/storage/redis/redis.go
cuonglm: I'm open to any better naming suggestion.

pkg/storage/redis/redis.go
```diff
@@ -208,6 +210,50 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
...
+			case <-ctx.Done():
+				log.Error().Err(ctx.Err()).Msg("stop subscribing to version set channel")
+				return
+			default:
+			}
+			switch v := psc.Receive().(type) {
+			case redis.Message:
+				log.Debug().Str("action", string(v.Data)).Msg("Got redis message")
+				if string(v.Data) == watchAction {
+					ch <- struct{}{}
```
calebdoxsey, line 246: this can block leaving an orphaned goroutine. We should do a select with ctx.Done too:

```
select {
case <-ctx.Done(): // ...
case ch <-struct{}{}: // ...
}
```

pkg/storage/redis/redis.go
```diff
@@ -208,6 +210,50 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
...
+			return
+		}
+		for {
+			select {
+			case <-ctx.Done():
+				log.Error().Err(ctx.Err()).Msg("stop subscribing to version set channel")
+				return
+			default:
+			}
+			switch v := psc.Receive().(type) {
```
calebdoxsey, line 242: We need to detect `ctx.Done` and terminate this call somehow. If I remember right this redis library has very poor context support. Some ideas:

1. Listen for ctx.Done() in a goroutine and then close the underlying connection, which should hopefully terminate the Receive() call
2. Use a relatively short timeout and retry on timeout errors. For example we could use a 1 second timeout.

pkg/storage/inmemory/inmemory.go
cuonglm: Ah right.

internal/databroker/server.go
cuonglm: I think we should always do the first sync, before we going to listen to the channel.

pkg/storage/redis/redis.go
cuonglm: Fixed

internal/databroker/server.go
travisgroth: Yes I was thinking something like that.  Either have the db call return a signal once at the start to force the caller to catch up, or have this code do a list after the channel is set up to pick up anything in between.

pkg/storage/redis/redis.go
cuonglm: Fixed.

pkg/storage/inmemory/inmemory.go
cuonglm: Fixed

pkg/storage/inmemory/inmemory.go
```diff
@@ -122,12 +127,20 @@ func (db *DB) List(_ context.Context, sinceVersion string) ([]*databroker.Record
...
 		record.Data = data
 	})
 	return nil
 }
 
+// Sync returns the underlying signal.Signal binding channel to the caller.
+// Then the caller can listen to the channel for detecting changes.
+func (db *DB) Sync(ctx context.Context) chan struct{} {
+	ch := db.onchange.Bind()
+	return ch
```
calebdoxsey, line 141: We never unbind the channel... doesn't this result in a leak of channels? Every time a user calls Sync another one will be added and never removed.

cuonglm: Oh sorry, the Unbind when ctx.Done() was removed accidently

internal/databroker/server.go
```diff
@@ -201,6 +198,24 @@ func (srv *Server) Set(ctx context.Context, req *databroker.SetRequest) (*databr
...
+	if err != nil {
+		return err
+	}
+	if len(updated) == 0 {
+		return nil
+	}
+	sort.Slice(updated, func(i, j int) bool {
+		return updated[i].Version < updated[j].Version
+	})
+	recordVersion = updated[len(updated)-1].Version
```
calebdoxsey, line 212: this updates the local variable but not the calling function. We need to use a pointer or return the new record version

internal/databroker/server.go
```diff
@@ -222,30 +237,21 @@ func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBroke
...
-			return stream.Context().Err()
-		case <-ch:
+	// Do first sync, so we won't missed anything.
+	if err := srv.doSync(ctx, recordVersion, db, stream); err != nil {
+		return err
+	}
+
+	ch := db.Sync(ctx)
+
+	for range ch {
```
calebdoxsey, line 249: This will block. We need to use a select with `ctx.Done()`

pkg/storage/storage.go
```diff
@@ -29,4 +29,7 @@ type Backend interface {
 
 	// ClearDeleted is used clear marked delete records.
 	ClearDeleted(ctx context.Context, cutoff time.Time)
+
+	// Sync is used to notify about changes happen in storage.
+	Sync(ctx context.Context) chan struct{}
```
calebdoxsey, line 34: I guess the intention here is to use the context Done as a signal to stop syncing? Otherwise there's no way to communicate that syncing should stop. This could be error prone to use.

internal/databroker/server.go
cuonglm: Not sure we need. `db.Sync(ctx)` is called with `ctx` is `stream.Context()`, when ctx .Done(), `ch` will be closed inside `db.Sync(ctx)`, so the for loop will be terminated.

pkg/storage/storage.go
cuonglm: @calebdoxsey See https://github.com/pomerium/pomerium/pull/1135#discussion_r460088546
 
`ctx.Done()` will trigger the `close(ch)` inside `Sync`, which will signal the caller to stop syncing.

internal/databroker/server.go
cuonglm: Fixed.

pkg/storage/redis/redis.go
cuonglm: @calebdoxsey @desimone Unless you have a better name, I will go with @travisgroth for `Watch`.

pkg/storage/storage.go
cuonglm: I updated the method name, and the documentation, hope it's more clear now.

pkg/storage/redis/redis.go
```diff
@@ -208,6 +210,60 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
...
+			return
+		}
+		select {
+		case <-ctx.Done():
+			log.Warn().Err(ctx.Err()).Msg("unable to notify channel")
+			return
+		case ch <- struct{}{}:
+		}
+	case error:
+		log.Debug().Err(v).Msg("redis subscribe error")
```
calebdoxsey, line 227: I think we need to distinguish between timeout errors and other errors. Otherwise we'll just loop forever.

cuonglm: I'm not sure how can we distinguish them. Also, isn't it ok to just loop forever? The loop is controlled by the context for termination.

calebdoxsey: If, for example, you get disconnected from redis, we will end up calling this function, failing and then immediately trying again. I imagine this will peg the CPU at 100% on the process.

If the error is due to something internal to redis (not sure what could cause that) we might cause a denial of service attack due to all the failed requests.

Typically for error conditions we should use exponential backoff to prevent thundering herds like this.

I'm not sure what error redis returns on timeout, but if it's a net error you can use a method call to determine if the error was due to a timeout:

```
type Error interface {
	error
	Timeout() bool   // Is the error a timeout?
	Temporary() bool // Is the error temporary?
}
```

cuonglm: Fixed.

pkg/storage/redis/redis.go
```diff
@@ -208,6 +212,75 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
...
+				log.Error().Err(ctx.Err()).Msg("failed to notify channel")
+				return
+			}
+			if ok && err.Timeout() {
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(eb.NextBackOff()):
+				}
+				goto top
```
desimone: Can we rewrite this without the goto?  I have a very hard time mentally walking through the control flow here.

pkg/storage/redis/redis.go
```diff
@@ -208,6 +212,78 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
...
+			log.Error().Err(ctx.Err()).Msg("failed to notify channel")
+			return
+		}
+		if ok && err.Timeout() {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(eb.NextBackOff()):
+			}
+			goto top
```
desimone, line 258: Can this be rewritten without using a `goto`?

pkg/storage/redis/redis.go
```diff
@@ -208,6 +212,78 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
...
+		}
+	}
+}
+
+// Watch returns a channel to the caller, when there is a change to the version set,
+// sending message to the channel to notify the caller.
+func (db *DB) Watch(ctx context.Context) chan struct{} {
+	ch := make(chan struct{})
+	go func() {
```
desimone, line 267: Are you concerned about starting this go-routine without knowing when it will stop?

pkg/storage/redis/redis.go
```diff
@@ -208,6 +212,78 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
 	}
 }
 
+func doNotify(ctx context.Context, psc *redis.PubSubConn, ch chan struct{}) error {
+	switch v := psc.ReceiveWithTimeout(time.Second).(type) {
+	case redis.Message:
+		log.Debug().Str("action", string(v.Data)).Msg("Got redis message")
```
desimone, line 218: nit: capitalized `Got`

internal/databroker/server.go
```diff
@@ -222,30 +237,20 @@ func (srv *Server) Sync(req *databroker.SyncRequest, stream databroker.DataBroke
...
-				ServerVersion: srv.version,
-				Records:       updated,
-			})
-			if err != nil {
-				return err
-			}
-		}
+	ctx := stream.Context()
+	// Do first sync, so we won't missed anything.
+	if err := srv.doSync(ctx, &recordVersion, db, stream); err != nil {
```
travisgroth, line 242: I think this should be after the Watch() returns?

pkg/storage/redis/redis.go
cuonglm: Why "without knowing"? The `ctx` is passed to make sure that the caller can control when to signal the goroutine to stop.

internal/databroker/server.go
cuonglm: Yes, right.

pkg/storage/redis/redis.go
cuonglm: Fixed

pkg/storage/redis/redis.go
cuonglm: I'm not sure there's a cleaner way. In case of timeout, the underlying connection was closed, so we need to start from scratch, with a new fresh pub sub conn. This way looks cleaner for me, instead of using nested for loop, with a break here instead of goto.

pkg/storage/redis/redis.go
desimone: Got it. I had misread the code.

pkg/storage/redis/redis.go
desimone: My concern (probably mostly from other languages) is that I have a hard time mentally modeling behavior/state with gotos. I know there is a time and place for them, but ... yeah, I'm not sure here.

@calebdoxsey thoughts?

calebdoxsey: I don't think its a good idea to use goto like this.

cuonglm: @calebdoxsey why and any suggestion to improve?

desimone: > instead of using nested for loop, with a break here

Can we try this? 

Perhaps abstracting the flow into a few separate functions can help with the janky-ness of starting "from scratch, with a new fresh pub sub conn".

cuonglm: @desimone fixed

pkg/storage/storage.go
```diff
@@ -29,4 +29,9 @@ type Backend interface {
 
 	// ClearDeleted is used clear marked delete records.
 	ClearDeleted(ctx context.Context, cutoff time.Time)
+
+	// Watch returns a channel to the caller. The channel is used to notify
+	// about changes happen in storage. When ctx is finish, Watch will close
```
desimone, line 34: nit: changes _that_ happen
nit: is finish_ed_,

pkg/storage/redis/redis.go
```diff
@@ -208,6 +212,85 @@ func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
...
+		case ch <- struct{}{}:
+		}
+	case error:
+		log.Debug().Err(v).Msg("redis subscribe error")
+		return v
+	}
+	return nil
+}
+
+func (db *DB) doNotifyLoop(ctx context.Context, ch chan struct{}, psc *redis.PubSubConn, eb *backoff.ExponentialBackOff) error {
```
desimone, line 235: Could you add some comments to the Loop functions?

cuonglm: Fixed

PR #1145, opened 2020-07-27
fix databroker restart versioning, handle missing sessions

calebdoxsey:
## Summary
There were two issues in the code:

1. When the server version changed we properly updated the server version on the initial `GetAll` call, but we didn't update it in the `Sync` call, so we didn't detect the change. 
2. When we called `forceSync` we returned `nil` when no session was found, but this left the session state intact, and a user would see an error page instead of being redirected to the login page.

To fix these issues I made the following changes:

1. Data is now properly cleared from the databroker data map instead of relying on a lookup of the version to clear the state.
1. I removed the separate databroker data map update goroutine, and do it directly now.
1. In addition to updated the server version, we now reset the record version to the empty string. This should re-sync all the records, which is necessary if the server is restarted.
1. Rather than return nil for the error when looking up a session, we return a `not found` error. This way the session state is cleared and the user is treated as if they were logged out.

These changes could lead to a redirect loop and need to be more thoroughly tested.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/grpc.go
+++ b/authorize/grpc.go
@@ -49,19 +49,6 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v2.CheckRe
 	rawJWT, _ := loadRawSession(hreq, a.currentOptions.Load(), a.currentEncoder.Load())
 	sessionState, _ := loadSession(a.currentEncoder.Load(), rawJWT)
 
-	// only accept sessions whose databroker server versions match
-	if sessionState != nil {
-		a.dataBrokerDataLock.RLock()
-		if a.dataBrokerSessionServerVersion != sessionState.Version.String() {
-			log.Warn().
-				Str("server_version", a.dataBrokerSessionServerVersion).
-				Str("session_version", sessionState.Version.String()).
-				Msg("clearing session due to invalid version")
-			sessionState = nil
-		}
-		a.dataBrokerDataLock.RUnlock()
-	}
-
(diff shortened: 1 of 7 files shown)


authorize/grpc.go
```diff
@@ -98,7 +85,7 @@ func (a *Authorize) forceSync(ctx context.Context, ss *sessions.State) error {
 	}
 	s := a.forceSyncSession(ctx, ss.ID)
 	if s == nil {
-		return nil
+		return errors.New("session not found")
```
calebdoxsey, line 88: This should cause stale sessions (that is, sessions tied to a databroker which was restarted) to be treated as empty sessions. Users will be redirected to the login page.

There's potential here for a redirect loop. (it *should* work...)

authorize/run.go
```diff
@@ -151,41 +137,36 @@ func (a *Authorize) runDataTypeSyncer(ctx context.Context, typeURL string, updat
 			}
 
 			backoff.Reset()
-			serverVersion = res.GetServerVersion()
+			if res.GetServerVersion() != serverVersion {
+				serverVersion = res.GetServerVersion()
+				recordVersion = ""
```
calebdoxsey, line 142: we weren't resetting this before. I wonder if that was causing problems. (though with the force sync I would've expected it to pick them up anyway)

PR #1146, opened 2020-07-28
authenticate: fix wrong condition checking in VerifySession

cuonglm:
## Summary
The code checks session client not nil, but using databroker client in
the body instead.

## Related issues

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
should we remove the sessionClient field from the struct too?

cuonglm:
> should we remove the sessionClient field from the struct too?

It's still used in other place, like SignOut.

PR #1153, opened 2020-07-28
handle example.com and example.com:443

calebdoxsey:
## Summary
Previously we fixed this issue by using `StripMatchingHostPort: true`. Unfortunately this made it so any routes which actually had ports in them (ie `example.com:9000`) longer match in envoy's routing table. This is because the routes were added explicitly with the port in the name.

This PR includes two changes:

1. We remove the `StripMatchingHostPort` option.
2. We add duplicate entries for any bare domain names, or domain names with default ports. `https://www.example.com` has an entry for `www.example.com` and `www.example.com:443`.

## Related issues
- https://github.com/pomerium/pomerium/issues/959
- https://github.com/pomerium/pomerium/pull/1126

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [x] ready for review


travisgroth:
Should this be backported?

desimone:
> Should this be backported?

Yes

PR #1159, opened 2020-07-29
pkg/storage/redis: add authentication support

### LINKED ISSUE #1157: pkg/storage/redis: Support Authentication

Storage backend should support redis user/password authentication.

**Additional context**

Docs here https://redis.io/topics/acl

I think we can just support the newer user + password.

cuonglm:
## Summary
Add support for redis auth.

## Related issues
Fixes #1157

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
@cuonglm would it be possible to use a data connection string instead? Something like:

`redis://username:password@host/db`

Redigo supports a `DialURL` method following this scheme: https://www.iana.org/assignments/uri-schemes/prov/redis.

This will make a future SQL provider easier to implement.

cuonglm:
> @cuonglm would it be possible to use a data connection string instead? Something like:
> 
> `redis://username:password@host/db`
> 
> Redigo supports a `DialURL` method following this scheme: https://www.iana.org/assignments/uri-schemes/prov/redis.
> 
> This will make a future SQL provider easier to implement.

Yes, you can do it already without this CL. The `username/password` mechanism is for storage that generally don't have DSN for connection. With redis, it supports both way, so it's ok

travisgroth:
> > @cuonglm would it be possible to use a data connection string instead? Something like:
> > `redis://username:password@host/db`
> > Redigo supports a `DialURL` method following this scheme: https://www.iana.org/assignments/uri-schemes/prov/redis.
> > This will make a future SQL provider easier to implement.
> 
> Yes, you can do it already without this CL. The `username/password` mechanism is for storage that generally don't have DSN for connection. With redis, it supports both way, so it's ok.

Hmm I actually tried a few combinations of URL with the current connection string option.  I think we need to use a different Dial() to support that.  But maybe I missed something?

> databroker_storage_connection_string: redis:http://pomerium:badpassword@127.0.0.1:6379/0

> databroker_storage_connection_string: redis:http://pomerium:badpassword@127.0.0.1:6379

> databroker_storage_connection_string: redis://pomerium:badpassword@127.0.0.1:6379/0

> databroker_storage_connection_string: redis://pomerium:badpassword@127.0.0.1:6379

I believe all of these gave me `too many colons in address`

calebdoxsey:
In my opinion if a DSN is able to use usernames and passwords for redis, we should have only one way of configuring redis.

cuonglm:
> In my opinion if a DSN is able to use usernames and passwords for redis, we should have only one way of configuring redis.

just concern if we want to support other storage which does not have DSN support. @travisgroth how do you think?

cuonglm:
> In my opinion if a DSN is able to use usernames and passwords for redis, we should have only one way of configuring redis.

fixed.

docs/configuration/readme.md
```diff
@@ -831,7 +831,7 @@ The backend storage that databroker server will use, available types: `memory`,
 - Config File Key: `databroker_storage_connection_string`
 - Type: `string`
 - **Required** when storage type is `redis`
-- Example: `":6379"`
+- Example: `"redis://localhost/6379/0"`
```
calebdoxsey, line 834: this should be `redis://localhost:6379/0`

cuonglm: Fixed

PR #1161, opened 2020-07-29
telmetry: add databroker storage metrics and tracing

travisgroth:
## Summary
Add metrics and tracing to redis storage backend

**Checklist**:
- [x] updated docs
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/helpers.go
+++ b/config/helpers.go
@@ -11,6 +11,12 @@ const (
 	ServiceAuthenticate = "authenticate"
 	// ServiceCache represents running the cache service component
 	ServiceCache = "cache"
+
+	// RedisName is the name of the redis storage backend
+	StorageRedisName = "redis"
+
+	// InMemoryName is the name of the in-memory storage backend
+	StorageInMemoryName = "memory"
 )
 
 // IsValidService checks to see if a service is a valid service mode
--- a/config/options.go
+++ b/config/options.go
@@ -25,8 +25,7 @@ import (
 	"github.com/pomerium/pomerium/internal/telemetry/metrics"
 	"github.com/pomerium/pomerium/internal/urlutil"
 	"github.com/pomerium/pomerium/pkg/cryptutil"
-	"github.com/pomerium/pomerium/pkg/storage/inmemory"
-	"github.com/pomerium/pomerium/pkg/storage/redis"
+	"github.com/pomerium/pomerium/pkg/storage"
 )
 
 // DisableHeaderKey is the key used to check whether to disable setting header
@@ -491,8 +490,8 @@ func (o *Options) Validate() error {
 		o.DataBrokerURLString = o.CacheURLString
 	}
 	switch o.DataBrokerStorageType {
-	case inmemory.Name:
-	case redis.Name:
+	case storage.InMemoryName:
+	case storage.RedisName:
 		if o.DataBrokerStorageConnectionString == "" {
 			return errors.New("config: missing databroker storage backend dsn")
 		}
--- a/config/options.go
+++ b/config/options.go
@@ -25,7 +25,6 @@ import (
 	"github.com/pomerium/pomerium/internal/telemetry/metrics"
 	"github.com/pomerium/pomerium/internal/urlutil"
 	"github.com/pomerium/pomerium/pkg/cryptutil"
-	"github.com/pomerium/pomerium/pkg/storage"
 )
 
 // DisableHeaderKey is the key used to check whether to disable setting header
@@ -490,8 +489,8 @@ func (o *Options) Validate() error {
 		o.DataBrokerURLString = o.CacheURLString
 	}
 	switch o.DataBrokerStorageType {
-	case storage.InMemoryName:
-	case storage.RedisName:
+	case StorageInMemoryName:
+	case StorageRedisName:
 		if o.DataBrokerStorageConnectionString == "" {
 			return errors.New("config: missing databroker storage backend dsn")
 		}
--- a/internal/telemetry/metrics/registry.go
+++ b/internal/telemetry/metrics/registry.go
@@ -106,3 +106,33 @@ func (r *metricRegistry) setConfigChecksum(service string, checksum uint64) {
 	}
 	m.Set(float64(checksum))
 }
+
+func (r *metricRegistry) addInt64DerivedGaugeMetric(name string, desc string, service string, f func() int64) {
+
+	m, err := r.registry.AddInt64DerivedGauge(name, metric.WithDescription(desc), metric.WithLabelKeys("service"))
+	if err != nil {
+		log.Error().Err(err).Str("service", service).Msg("telemetry/metrics: failed to register metric")
+		return
+	}
+
+	err = m.UpsertEntry(f, metricdata.NewLabelValue(service))
+	if err != nil {
+		log.Error().Err(err).Str("service", service).Msg("telemetry/metrics: failed to update metric")
+		return
+	}
+}
+
+func (r *metricRegistry) addInt64DerivedCumulativeMetric(name string, desc string, service string, f func() int64) {
+
+	m, err := r.registry.AddInt64DerivedCumulative(name, metric.WithDescription(desc), metric.WithLabelKeys("service"))
+	if err != nil {
+		log.Error().Err(err).Str("service", service).Msg("telemetry/metrics: failed to register metric")
+		return
+	}
+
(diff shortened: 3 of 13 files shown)


internal/telemetry/metrics/registry.go
```diff
@@ -106,3 +106,39 @@ func (r *metricRegistry) setConfigChecksum(service string, checksum uint64) {
...
+
+func (r *metricRegistry) addInt64DerivedGaugeMetric(name string, desc string, service string, f func() int64) {
+
+	m, err := r.registry.AddInt64DerivedGauge(name, metric.WithDescription(desc), metric.WithLabelKeys("service"))
+	if err != nil {
+		log.Error().Err(err).Str("service", service).Msg("telemetry/metrics: failed to register metric")
+		return
+	}
+
+	err = m.UpsertEntry(
```
desimone, line 118: nit: why is this method call multiline?

internal/telemetry/metrics/registry.go
```diff
@@ -106,3 +106,39 @@ func (r *metricRegistry) setConfigChecksum(service string, checksum uint64) {
...
+
+func (r *metricRegistry) addInt64DerivedCumulativeMetric(name string, desc string, service string, f func() int64) {
+
+	m, err := r.registry.AddInt64DerivedCumulative(name, metric.WithDescription(desc), metric.WithLabelKeys("service"))
+	if err != nil {
+		log.Error().Err(err).Str("service", service).Msg("telemetry/metrics: failed to register metric")
+		return
+	}
+
+	err = m.UpsertEntry(
```
desimone, line 136: nit: why multiline?

config/options.go
```diff
@@ -491,8 +490,8 @@ func (o *Options) Validate() error {
 		o.DataBrokerURLString = o.CacheURLString
 	}
 	switch o.DataBrokerStorageType {
-	case inmemory.Name:
-	case redis.Name:
+	case storage.InMemoryName:
```
desimone, line 493: It feels strange to have the interface defining package to have the name of the implementing storage backends vs before (and is inconsistent with how we do things in other packages)? Can you elaborate on the change?

pkg/storage/redis/redis.go
```diff
@@ -66,13 +68,18 @@ func New(address, recordType string, deletePermanentAfter int64) (*DB, error) {
 		deletedSet:             recordType + "_deleted_set",
 		lastVersionKey:         recordType + "_last_version",
 	}
+	metrics.AddRedisMetrics(db.pool.Stats)
 	return db, nil
 }
 
 // Put sets new record for given id with input data.
-func (db *DB) Put(ctx context.Context, id string, data *anypb.Any) error {
+func (db *DB) Put(ctx context.Context, id string, data *anypb.Any) (err error) {
```
desimone, line 76: Maybe this is more of a preference thing but I find named result parameters to hurt readability (though I understand why you have one here, in order to support the early defer statement), in all but very short functions. 

[Effective Go](https://golang.org/doc/effective_go.html#named-results) is the only thing I've found to weigh in on this though.

pkg/storage/redis/redis.go
```diff
@@ -66,13 +68,18 @@ func New(address, recordType string, deletePermanentAfter int64) (*DB, error) {
...
 	c := db.pool.Get()
+	_, span := trace.StartSpan(ctx, "databroker.redis.Put")
+	defer span.End()
```
desimone, line 79: Three back-to-back defer statements feels like a code smell to me. E.g. Should we be calling `defer c.Close()` here before starting the trace span?

pkg/storage/redis/redis.go
```diff
@@ -187,8 +206,12 @@ func (db *DB) Delete(ctx context.Context, id string) error {
 }
 
 // ClearDeleted clears all the currently deleted records older than the given cutoff.
-func (db *DB) ClearDeleted(_ context.Context, cutoff time.Time) {
+func (db *DB) ClearDeleted(ctx context.Context, cutoff time.Time) {
 	c := db.pool.Get()
+	_, span := trace.StartSpan(ctx, "databroker.redis.ClearDeleted")
+	defer span.End()
+	var opErr error
```
desimone, line 213: 😉

pkg/storage/storage.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/pomerium/pomerium/pkg/grpc/databroker"
 )
 
+// RedisName is the name of the redis storage backend
+const RedisName = "redis"
```
desimone, line 14: See previous comments.

pkg/storage/redis/redis.go
```diff
@@ -352,3 +375,11 @@ func (db *DB) tx(c redis.Conn, commands []map[string][]interface{}) error {
 	_, err := c.Do("EXEC")
 	return err
 }
+
+func (db DB) recordOperation(ctx context.Context, startTime time.Time, operation string, err error) {
```
desimone, line 379: Should this just be a function if db is not being used?

pkg/storage/redis/redis.go
desimone: Yeah, maybe it's just me but I'm having a really hard time keeping tack of how the defer statements are going to unwind if there's an error somewhere.

config/options.go
travisgroth: Import loop from bringing in telemetry. :(

Open to other ideas, structurally.  The main issue is config is imported by a lot of other packages, so we shouldn't have it depend on much internals.

pkg/storage/redis/redis.go
travisgroth: Yes I _think_ we can do without it, but it makes the actual error variable easier to lose track of.  

I know there's a lot of animosity towards naked returns but this seems like a good use of named parameters.

internal/telemetry/metrics/registry.go
travisgroth: It was that way in the old code.  Probably started off longer.

pkg/storage/redis/redis.go
travisgroth: Yep.

pkg/storage/redis/redis.go
travisgroth: Could change to this:

```go
	defer func() {
		c.Close()
		recordOperation(ctx, time.Now(), "put", err)
		span.End()
	}()
```

desimone: I'm not sure that does the same thing though. 🤔 
\cc @cuonglm  / @calebdoxsey thoughts?

config/options.go
desimone: I think we can avoid the import loop by removing runtime interface assertion, that way the redis/memory packages no longer explicitly need to import storage. 

https://github.com/pomerium/pomerium/blob/8ed40b225289f91c739132f21acf3a6dfa767167/pkg/storage/redis/redis.go#L28

pkg/storage/redis/redis.go
calebdoxsey: defers are called in reverse order:
```go
defer a()
defer b()
```
runs b() then a()

config/options.go
cuonglm: We still need to keep the assertion somewhere, maybe inside `pkg/storage/storage.go`?

desimone: Or, put it in a test under a different package name?

pkg/storage/redis/redis.go
travisgroth: Correction, it would be:

```go
	defer func(s time.Time) {
		c.Close()
		recordOperation(ctx, time.Since(s), "put", err)
		span.End()
	}(time.Now())
```

@desimone any opinion on which style?

config/options.go
travisgroth: Looking again, I'm thinking the config package could just hold the names.  They're primarily a construct for the configuration file.  They're used for mapping config->backend in config and then the broker init uses the same logic.    Thoughts?

config/helpers.go
```diff
@@ -11,6 +11,12 @@ const (
 	ServiceAuthenticate = "authenticate"
 	// ServiceCache represents running the cache service component
 	ServiceCache = "cache"
+
+	// RedisName is the name of the redis storage backend
+	StorageRedisName = "redis"
```
desimone, line 16: 👍

PR #1162, opened 2020-07-29
fix redirect loop, remove user/session services, remove duplicate deleted_at fields

calebdoxsey:
## Summary
The main issue here is we weren't properly detecting deleted sessions. When we retrieve a session in authenticate, we could get a record that was marked as deleted but not actually removed from the databroker. Since it still appeared to be valid, we re-used it. But the corresponding check in authorize *did* check if it was deleted, so we got stuck in a redirect loop.

Many changes in this PR:

1. Remove the duplicate `deleted_at` fields since they caused confusion in the code which led to this problem.
1. Remove the user and session clients. They didn't really serve any purpose since the databroker interface actually does everything.
1. Increase the session timeout to 1 year instead of 1 hour
1. Update the `Get` and `GetAll` databroker methods to not return deleted records. (`Sync` still does though)

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


authenticate/handlers.go
```diff
@@ -534,11 +528,11 @@ func (a *Authenticate) Dashboard(w http.ResponseWriter, r *http.Request) error {
 }
 
 func (a *Authenticate) saveSessionToDataBroker(ctx context.Context, sessionState *sessions.State, accessToken *oauth2.Token) error {
-	if a.sessionClient == nil || a.userClient == nil {
+	if a.dataBrokerClient == nil {
 		return nil
 	}
 
-	sessionExpiry, _ := ptypes.TimestampProto(time.Now().Add(time.Hour))
+	sessionExpiry, _ := ptypes.TimestampProto(time.Now().Add(time.Hour * 24 * 365))
```
desimone, line 535: Should we set this to be equal to the total lifetime of the session instead of a year? We could use a new option for this value,  -- or the existing -- cookie expiry value (which I think defaults to a week) and has the same intent.

internal/databroker/server.go
```diff
@@ -155,10 +155,17 @@ func (srv *Server) GetAll(ctx context.Context, req *databroker.GetAllRequest) (*
 	if err != nil {
 		return nil, err
 	}
-	records, err := db.GetAll(ctx)
+	all, err := db.GetAll(ctx)
 	if err != nil {
 		return nil, err
 	}
+	records := make([]*databroker.Record, 0, len(all))
+	for _, record := range all {
```
desimone, line 163: Is there a way we can combine these two loops to prevent iterating twice?

pkg/grpc/session/session.go
```diff
@@ -5,10 +5,21 @@ import (
...
 )
 
+// Delete deletes a session from the databroker.
+func Delete(ctx context.Context, client databroker.DataBrokerServiceClient, sessionID string) error {
+	any, _ := ptypes.MarshalAny(new(Session))
+	_, err := client.Delete(ctx, &databroker.DeleteRequest{
+		Type: any.GetTypeUrl(),
+		Id:   sessionID,
+	})
+	return err
```
desimone, line 20: Should we wrap the error to provide additional context?

pkg/grpc/session/session.go
```diff
@@ -28,3 +39,17 @@ func Get(ctx context.Context, client databroker.DataBrokerServiceClient, session
...
+// Set sets a session in the databroker.
+func Set(ctx context.Context, client databroker.DataBrokerServiceClient, s *Session) (*databroker.SetResponse, error) {
+	any, _ := anypb.New(s)
+	res, err := client.Set(ctx, &databroker.SetRequest{
+		Type: any.GetTypeUrl(),
+		Id:   s.Id,
+		Data: any,
+	})
+	if err != nil {
+		return nil, err
```
desimone, line 52: Should we wrap the error to provide additional context? Otherwise, we can simply return client.Set() I think

authenticate/handlers.go
desimone: edit: we actually set cookies to expire after 14 hours.

https://github.com/pomerium/pomerium/blob/3039407597d9b510faf7559f0c8d8d78ce5e7e4a/config/options.go#L263

internal/databroker/server.go
```diff
@@ -155,10 +155,17 @@ func (srv *Server) GetAll(ctx context.Context, req *databroker.GetAllRequest) (*
 	if err != nil {
 		return nil, err
 	}
-	records, err := db.GetAll(ctx)
+	all, err := db.GetAll(ctx)
 	if err != nil {
 		return nil, err
 	}
+	records := make([]*databroker.Record, 0, len(all))
```
cuonglm, line 162: Should check `len(all) == 0` here and return early.

PR #1163, opened 2020-07-30
pkg/storage/redis: add redis TLS support

### LINKED ISSUE #1156: pkg/storage/redis: Support TLS Options

We should support some basic TLS options to redis for transport security.  This should include:

- Enable/disable client TLS 
- Provide client certificate 
- Turn server verification on/off

TLS with server verification should be the default given our desired security posture.

N/A

If there are any use-cases that would help us understand the use/need/value please share them as they can help us decide on acceptance and prioritization.

**Additional context**

N/A

cuonglm:
## Summary
Add support for redis TLS.

## Related issues
Fixes #1156

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


cache/databroker.go
```diff
@@ -15,9 +20,25 @@ type DataBrokerServer struct {
 
 // NewDataBrokerServer creates a new databroker service server.
 func NewDataBrokerServer(grpcServer *grpc.Server, opts config.Options) *DataBrokerServer {
+	var tlsConfig *tls.Config
+	if opts.DataBrokerCertificate != nil {
```
travisgroth, line 24: We should still be passing along skipTLS and the CA options without client cert/key set

Issue #1151, opened 2020-07-28
pomerium-cli/k8s: credential expiration inconsistent


travisgroth:
## What happened?

It appears we get a browser auth flow about every minute, despite JWT expiration being 1 hour long.

## What did you expect to happen?

Browser flow should only happen when the actual JWT is expired, or never, after initial setup.  I think with the current backend session management, the latter is possible/correct.

## How'd it happen?

1. Set up kubernetes auth
2. run `kubectl get pods -A && sleep 65 && kubectl get pods -A`
3. Got two browser auth flows back to back

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint): 72b6347886bd4faeb12ec0025b133643de8f01ba
- Server Operating System/Architecture/Cloud: OSX

## Additional context

Credential cache after immediately logging in:

ExecCredential: `"expirationTimestamp":"2020-07-28T15:01:38.554267-04:00"`
JWT: `exp: 1595965860` (Tuesday, July 28, 2020 3:51:00 PM GMT-04:00)

PR #1169, opened 2020-07-30
pomerium-cli: fix kubernetes token caching

calebdoxsey:
## Summary
This PR fixes two issues:

1. We weren't decoding the expiry data in the JWT properly
2. We had hardcoded the token to expire after 10 seconds (presumably for testing and I forgot to change it back)

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/1151

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/cmd/pomerium-cli/kubernetes.go
+++ b/cmd/pomerium-cli/kubernetes.go
@@ -14,9 +14,9 @@ import (
 
 	"github.com/skratchdot/open-golang/open"
 	"github.com/spf13/cobra"
-	jose "github.com/square/go-jose/v3"
 	"golang.org/x/crypto/ssh/terminal"
 	"golang.org/x/sync/errgroup"
+	jose "gopkg.in/square/go-jose.v2"
 )
 
 func init() {
--- a/cmd/pomerium-cli/kubernetes.go
+++ b/cmd/pomerium-cli/kubernetes.go
@@ -14,7 +14,7 @@ import (
 
 	"github.com/skratchdot/open-golang/open"
 	"github.com/spf13/cobra"
-	"github.com/square/go-jose/jwt"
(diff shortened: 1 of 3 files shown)


cmd/pomerium-cli/kubernetes.go
```diff
@@ -14,7 +14,7 @@ import (
 
 	"github.com/skratchdot/open-golang/open"
 	"github.com/spf13/cobra"
-	"github.com/square/go-jose/jwt"
+	jose "github.com/square/go-jose/v3"
```
desimone, line 17: Confusingly we use "gopkg.in/square/go-jose.v2/jwt" elsewhere already which is the package name recommended in their docs (though I'm not really sure of the difference). 

Thoughts on using v2 vs v3 which is marked as [unstable for the time being](https://github.com/square/go-jose#versions)?

calebdoxsey: The jwt package didn't work for me. (from what I could tell we have to use the Claims call which requires the secret key, which we don't have) They switched from gopkg because its no longer needed with go modules. I can try to use the v2 package if we'd rather use that.

desimone: > The jwt package didn't work for me.

Yeah, sorry, I mean the go-jose package in general (including the jwt sub-package). 

> They switched from gopkg because its no longer needed with go modules

I'm fine either way, but would like to be consistent. Should we change all references then so don't have multiple versions of the same package (both `go-jose` and `go-jose/jwt`? 

> I can try to use the v2 package if we'd rather use that.

I think until v3 is marked stable, I think it makes sense to do that.

PR #1171, opened 2020-07-30
fix lint errors

travisgroth:
## Summary
Fix lint checks

**Checklist**:
- [x] ready for review

### DIFF
--- a/pkg/storage/redis/redis.go
+++ b/pkg/storage/redis/redis.go
@@ -209,7 +209,9 @@ func (db *DB) ClearDeleted(ctx context.Context, cutoff time.Time) {
 	_, span := trace.StartSpan(ctx, "databroker.redis.ClearDeleted")
 	defer span.End()
 	var opErr error
-	defer recordOperation(ctx, time.Now(), "clear_deleted", opErr)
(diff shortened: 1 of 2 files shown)


pkg/storage/redis/redis.go
```diff
@@ -228,7 +228,7 @@ func (db *DB) ClearDeleted(ctx context.Context, cutoff time.Time) {
 				{"ZREM": {db.versionSet, id}},
 				{"SREM": {db.deletedSet, id}},
 			}
-			opErr = db.tx(c, cmds)
+			opErr = db.tx(c, cmds) //nolint captured in defer()
```
calebdoxsey, line 231: actually I think this is a legit issue. opErr won't be updated in the defer. It would have to be:

```
startTime := time.Now()
defer func() {
    recordOperation(ctx, startTime, "clear_deleted", opErr)
}()
```

https://play.golang.org/p/mm8kW28L_JA

travisgroth: Good call.  Thanks.

I think this style work as desired https://play.golang.org/p/1sO_dC18MXn

PR #1205, opened 2020-08-04
proxy: refactor handler setup code

travisgroth:
## Summary
Fixes missing dashboard after #1204 changes.

## Related issues
#1204 

**Checklist**:
- [x] add related issues
- [x] updated unit tests
- [x] ready for review


proxy/proxy.go
```diff
@@ -195,15 +200,7 @@ func (p *Proxy) UpdatePolicies(opts *config.Options) error {
 		h.PathPrefix("/").Handler(p.registerFwdAuthHandlers())
 	}
 
-	for _, policy := range opts.Policies {
-		if err := policy.Validate(); err != nil {
```
desimone, line 199: If a policy fails to validate, what happens?

travisgroth: That should be getting caught as part of [options parsing itself](https://github.com/pomerium/pomerium/blob/e9f208e288abe3d8685d09f999325b1f53d42795/config/options.go#L376-L383).  This proxy code no longer does anything with the policy data anymore from what I can tell.

PR #1210, opened 2020-08-05
pkg/storage/redis: fix multiple data race

cuonglm:
## Summary
There are two data race in current code:

 - Call to close pub sub conn and renew pub sub conn
 - Call to close notify channel and send data to it

Fixing them by:

 - Moving pub sub conn creation/renew in the doNotifyLoop
 - Add a lock to guard before close/send data to channel, and also add
 another check for context was done at the beginning of notify loop.

Verifying by running:

```
	for _ in {1..100}; do
	  go test -race -count=1 ./pkg/storage/redis/...
	done
```

with no failure.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/storage/redis/redis.go
+++ b/pkg/storage/redis/redis.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
@@ -38,6 +39,7 @@ type DB struct {
 	versionSet             string
 	deletedSet             string
 	tlsConfig              *tls.Config
+	notifyChMu             sync.Mutex
 }
 
 // New returns new DB instance.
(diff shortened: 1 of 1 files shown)


pkg/storage/redis/redis.go
```diff
@@ -257,7 +273,10 @@ func (db *DB) doNotifyLoop(ctx context.Context, ch chan struct{}, psc *redis.Pub
 			case <-ctx.Done():
 				log.Warn().Err(ctx.Err()).Msg("context done, stop receive from redis channel")
 				return
-			case ch <- struct{}{}:
+			default:
+				db.notifyChMu.Lock()
+				ch <- struct{}{}
```
calebdoxsey, line 278: this can block and leave the notify mutex permanently locked

cuonglm: In theory, yes. In practice, `doNotifyLoop` is the only one that send data to the channel, so it can't block.

cuonglm: If you worried about that, then I think we can add another select case to ctx.Done(), PR is coming.

cuonglm: @calebdoxsey fixed.

PR #1226, opened 2020-08-07
internal/databroker: make Sync send data in smaller batches

cuonglm:
## Summary
GRPC streaming is better at sending multiple smaller message instead of
a big one.

Benchmark result for sending 10k messages at once vs multiple batches,
each with 100 messages:

```
name     old time/op  new time/op  delta
Sync-12  14.5ms ± 3%  12.4ms ± 2%  -14.40%  (p=0.000 n=10+9)
```

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -30,6 +30,7 @@ import (
 const (
 	recordTypeServerVersion = "server_version"
 	serverVersionKey        = "version"
+	syncBatchSize           = 100
 )
 
 // Server implements the databroker service using an in memory database.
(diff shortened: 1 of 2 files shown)


internal/databroker/server.go
```diff
@@ -225,10 +226,19 @@ func (srv *Server) doSync(ctx context.Context, recordVersion *string, db storage
 		return updated[i].Version < updated[j].Version
 	})
 	*recordVersion = updated[len(updated)-1].Version
-	return stream.Send(&databroker.SyncResponse{
-		ServerVersion: srv.version,
-		Records:       updated,
-	})
+	for i := 0; i < len(updated); i += syncBatchSize {
```
desimone, line 229: I'm not clever enough to know just by reading this if there could be an off by one error in this for loop. Do you mind adding some tests ? :)

cuonglm: Added.

PR #1237, opened 2020-08-10
internal/autocert: re-use cert if renewing failed but cert not expired

### LINKED ISSUE #1232: Certificate renewal process blocks envoy from starting on startup
## What happened?
* Valid certificates eligible for renewal, port 80 firewalled to the pomerium instance. Pomerium serving traffic just fine.
* Pomerium container restarts
* No routes resolve, no traffic flows through the proxy.

* Pomerium to continue serving with the old certs.

If anything, even if the certs were expired, I'd have expected pomerium to continue serving with expired certs, rather than block access alltogether. As long as the certs aren't revoked, they should still be usable.

## How'd it happen?

1. Restarted pomerium container
2. Everything failed silently.

- Pomerium version (retrieve with `pomerium --version` or `/ping` endpoint): 0.9.6 (tried from 0.9.0 to 0.10.0, same behavior)
- Server Operating System/Architecture/Cloud: Docker on DSM (Synology)

## What's your config.yaml?

The only settings that matter are

auto_cert: true

Anything else was irrelevant.

Only INFO logs from the certificate maintenance routine. Letsencrypt was timing out trying to access the /.well-known/.. endpoint. 100% of the logs were related to my firewall issues, which are not related to this bug.

## Additional context

The biggest problem I think is that pomerium is failing completely silently. No logs other than "Letsencrypt is timing out". Which is technically what's broken, but it doesn't explain why envoy isn't running, there are valid certificates present.

Also, huge red herring, because the trick is Letsencrypt is actually timing out on the "http" endpoint. Having missed that little tidbit, I was thinking that the letsencrypt timeout was a side-effect (Since the https endpoints, which should be valid, were not resolving - envoy wasn't running at all).

cuonglm:
## Summary

If autocert failed to renew a valid cert, we should continue using that cert instead of aborting pomerium.

## Related issues
Fixes #1232

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/autocert/manager.go
+++ b/internal/autocert/manager.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"sort"
 	"sync"
+	"time"
 
 	"github.com/caddyserver/certmagic"
 
@@ -105,10 +106,14 @@ func (mgr *Manager) updateAutocert(cfg *config.Config) error {
 			cert, err = cm.CacheManagedCertificate(domain)
 		}
 		if err == nil && cert.NeedsRenewal(cm) {
+			expired := time.Now().After(cert.Leaf.NotAfter)
 			log.Info().Str("domain", domain).Msg("renewing certificate")
 			err = cm.RenewCert(context.Background(), domain, false)
-			if err != nil {
+			if err != nil && expired {
 				return fmt.Errorf("autocert: failed to renew client certificate: %w", err)
+			}
+			if !expired {
+				log.Warn().Err(err).Msg("renew client certificated failed, use existing cert")
 			}
 			cert, err = cm.CacheManagedCertificate(domain)
 		}
--- a/internal/autocert/manager.go
+++ b/internal/autocert/manager.go
@@ -3,6 +3,7 @@ package autocert
 
 import (
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


internal/autocert/manager.go
```diff
@@ -105,11 +106,15 @@ func (mgr *Manager) updateAutocert(cfg *config.Config) error {
...
 		}
 		if err == nil && cert.NeedsRenewal(cm) {
+			expired := time.Now().After(cert.Leaf.NotAfter)
 			log.Info().Str("domain", domain).Msg("renewing certificate")
 			err = cm.RenewCert(context.Background(), domain, false)
-			if err != nil {
+			if err != nil && expired {
 				return fmt.Errorf("autocert: failed to renew client certificate: %w", err)
 			}
+			if !expired {
```
jaguarcy: Will this log even if cert successfully renews since you don't update expired after cm.RenewCert?

cuonglm: Ah thanks, fair point 👍

cuonglm: Fixed.

PR #1252, opened 2020-08-11
config: warn if custom scopes set for builtin providers

### LINKED ISSUE #1144: identity: warn if custom scopes are set

It's easy to get into trouble with oauth2 scopes especially with known providers. 

(e.g. Google rejects the offline_access scope, but requires you to add it as part of the signin url). Setting the wrong scopes can result in very undefined behavior with the identity manager service, and elsewhere. 

https://github.com/pomerium/pomerium/blob/72b6347886bd4faeb12ec0025b133643de8f01ba/internal/identity/oidc/oidc.go#L28

We should provide a warning message in the logs that setting custom scopes may result in undefined behavior and provide a link to the docs.

We do warn in the documentation, but it's an easy mistake to make. 

**Additional context**

- https://www.pomerium.io/configuration/#identity-provider-scopes
- https://pomerium-io.slack.com/archives/CK92MUAES/p1595884715359500?thread_ts=1595534105.299000&cid=CK92MUAES

cuonglm:
## Summary

Warn if custom scopes set for builtin providers.

## Related issues
Fixes #1144

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -19,6 +19,12 @@ import (
 	"github.com/spf13/viper"
 	"gopkg.in/yaml.v2"
 
+	"github.com/pomerium/pomerium/internal/directory/azure"
+	"github.com/pomerium/pomerium/internal/directory/github"
+	"github.com/pomerium/pomerium/internal/directory/gitlab"
+	"github.com/pomerium/pomerium/internal/directory/google"
+	"github.com/pomerium/pomerium/internal/directory/okta"
+	"github.com/pomerium/pomerium/internal/directory/onelogin"
 	"github.com/pomerium/pomerium/internal/identity/oauth"
 	"github.com/pomerium/pomerium/internal/log"
 	"github.com/pomerium/pomerium/internal/telemetry"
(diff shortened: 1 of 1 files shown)


config/options.go
```diff
@@ -651,6 +657,16 @@ func (o *Options) Validate() error {
 		return fmt.Errorf("config: server must be run with `autocert`, " +
 			"`insecure_server` or manually provided certificates to start")
 	}
+
+	switch o.Provider {
+	case azure.Name, github.Name, gitlab.Name, google.Name, okta.Name, onelogin.Name:
+		if len(o.Scopes) > 0 {
+			docLink := "https://www.pomerium.io/reference/#identity-provider-scopes"
```
desimone, line 664: could you move whole error string to const at top of the tin?

cuonglm: Done

config/options.go
```diff
@@ -30,6 +36,11 @@ import (
 // DisableHeaderKey is the key used to check whether to disable setting header
 const DisableHeaderKey = "disable"
 
+const (
+	idpCustomScopesDocLink = "https://www.pomerium.io/reference/#identity-provider-scopes"
+	idpCustomScopesWarnMsg = "config: using custom scopes may result in undefined behavior, see: " + idpCustomScopesDocLink
```
desimone, line 41: nit++ but I think we should just have one variable here. `warnCustomScopes` and just use multi-line breaks so it's readable. Similar to `ErrXYZ`. Thoughts?

desimone: Or, just return the full warning string without setting either the message or the link to a variable if its not going to be used elsewhere.

cuonglm: My intentional is that the dock link can be changed more frequently than the error message. so put it in a constant so we can grep or edit easier. Multi-line breaks seems not appropriate because it causes the message divided into 2 lines.

In all case, the compiler can optimize that for us, whether use separated constants or big one.

Issue #1254, opened 2020-08-11
config: support reloading more settings


calebdoxsey:
Currently we support reloading some, but not all configuration settings. With the introduction of loading settings from the databroker, we should support reloading as many configuration settings as we can.

Here are the settings that currently support automatic reloading:

* `insecure_server`
* `http_redirect_addr`
* `timeout_read`
* `timeout_write`
* `timeout_idle`
* `administrators`
* `default_upstream_timeout`
* `grpc_address`
* `grpc_insecure`
* `grpc_server_max_connection_age`
* `grpc_server_max_connection_age_grace`
* `client_ca`
* `client_ca_file`
* `google_cloud_serverless_authentication_service_account`
* `autocert`
* `autocert_use_staging`
* `autocert_must_staple`
* `autocert_dir`

Here are the settings that we should try to support:

* ~`debug`~
* ~`log_level`~
* ~`proxy_log_level`~
* ~`shared_secret`~
* ~`address`~
* ~`authenticate_service_url`~
* ~`authenticate_callback_path`~
* ~`cookie_name`~
* ~`cookie_secret`~
* ~`cookie_domain`~
* ~`cookie_secure`~
* ~`cookie_http_only`~
* ~`cookie_expire`~
* ~`idp_client_id`~
* ~`idp_client_secret`~
* ~`idp_provider`~
* ~`idp_provider_url`~
* ~`scopes`~
* ~`idp_service_account`~
* ~`idp_refresh_directory_timeout`~
* ~`idp_refresh_directory_interval`~
* ~`request_params`~
* ~`authorize_service_url`~
* ~`override_certificate_name`~
* ~`certificate_authority`~
* ~`certificate_authority_file`~
* ~`signing_key`~
* ~`jwt_claims_headers`~
* ~`refresh_cooldown`~
* ~`metrics_address`~
* ~`tracing_provider`~
* ~`tracing_sample_rate`~
* ~`tracing_jaeger_collector_endpoint`~
* ~`tracing_jaeger_agent_endpoint`~
* ~`tracing_zipkin_endpoint`~
* ~`forward_auth_url`~

And here are the settings we won't support:

* `services`
* `certificates`
* `cache_service_url`
* `databroker_service_url`

calebdoxsey:
I'll try to break these up into separate PRs. I'll start with logging+telemetry which seems straightforward.

calebdoxsey:
I believe all the configuration settings have been updated. If we discover additional problems we can address them in separate issues.

PR #1255, opened 2020-08-11
config: allow reloading of telemetry settings

calebdoxsey:
## Summary
This updates the way we handle configuration to support reloading without restarting for the following options:

- `debug`
- `log_level`
- `proxy_log_level`
- `metrics_address`
- `tracing_provider`
- `tracing_sample_rate`
- `tracing_jaeger_collector_endpoint`
- `tracing_jaeger_agent_endpoint`
- `tracing_zipkin_endpoint`

I also had to move the tracing options from config to trace to avoid an import cycle.

## Related issues
- #1254 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


travisgroth:
FYI - for the tracing options to become fully dynamic, we'll need to move the envoy tracing config from the [bootstrap](https://github.com/pomerium/pomerium/blob/master/internal/envoy/envoy.go#L264-L310) file into the control plane.  This wasn't possible until envoy 1.15.

internal/envoy/envoy.go
```diff
@@ -64,12 +64,6 @@ func NewServer(opts *config.Options, grpcPort, httpPort string) (*Server, error)
 		opts:     opts,
 	}
```
desimone, line 66: Should we update the docs https://www.pomerium.io/reference/#proxy-log-level ?

internal/envoy/envoy.go
```diff
@@ -88,7 +82,7 @@ func (srv *Server) Run(ctx context.Context) error {
 
 	srv.cmd = exec.CommandContext(ctx, envoyPath,
 		"-c", configFileName,
-		"--log-level", srv.logLevel,
+		"--log-level", "trace",
```
desimone, line 85: Will this be updated by config/log.go or is that just for the main log level? Is the envoy log level now hardcoded to trace?

desimone:
\cc @travisgroth because this is his area :)

internal/envoy/envoy.go
calebdoxsey: this is so all the logs are available in the handler. Whether they are displayed or not is a result of the zerolog log level.

calebdoxsey:
@travisgroth apparently its still not supported:

> Please be aware that envoy.tracers.opencensus provider can only be configured once in Envoy lifetime. Any attempts to reconfigure it or to use different configurations for different HCM filters will be rejected. Such a constraint is inherent to OpenCensus itself. It cannot be overcome without changes on OpenCensus side.

internal/envoy/envoy.go
calebdoxsey: It didn't really change. I updated the docs to make it clear the setting only applies to access logs.

internal/envoy/envoy.go
desimone: @calebdoxsey perfect

travisgroth:
> @travisgroth apparently its still not supported:
> 
> > Please be aware that envoy.tracers.opencensus provider can only be configured once in Envoy lifetime. Any attempts to reconfigure it or to use different configurations for different HCM filters will be rejected. Such a constraint is inherent to OpenCensus itself. It cannot be overcome without changes on OpenCensus side.

🙄  that's awesome.  okay.

internal/envoy/envoy.go
travisgroth: I recall we had some performance problems when we initial switched over to envoy.  https://github.com/pomerium/pomerium/pull/779 was pulling in a default that, I believe, set us to trace and slowed down performance significantly.  I might be misremembering but we should check this.

PR #1259, opened 2020-08-12
envoy: add support for hot-reloading bootstrap configuration

calebdoxsey:
## Summary
This is a follow-up to #1255 . Envoy has some settings which can't be dynamically loaded, like the tracing options, and the log level. This PR changes the way the envoy process is run so that we can support restarting it. Envoy has built-in support for zero-downtime restarts, so we take advantage of those options.

An envoy process has a `base-id` that we first set dynamically with `--dynamic-base-id`. This `base-id` is written to a temporary file. On restart we use this `base-id` and also pass in an incremented `restart-epoch` counter. If `restart-epoch` is greater than 0, envoy knows to look in shared memory for the existing sockets/stats/etc and transfers the over. Any existing requests will be drained.

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/envoy/envoy.go
+++ b/internal/envoy/envoy.go
@@ -157,7 +157,7 @@ func (srv *Server) update(cfg *config.Config) {
 func (srv *Server) run() error {
 	args := []string{
 		"-c", configFileName,
-		"--log-level", "trace",
+		"--log-level", firstNonEmpty(srv.options.logLevel),
 		"--log-format", "[LOG_FORMAT]%l--%n--%v",
 		"--log-format-escaped",
 	}
--- a/internal/envoy/envoy.go
+++ b/internal/envoy/envoy.go
@@ -4,7 +4,6 @@ package envoy
 import (
 	"bufio"
 	"bytes"
-	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


internal/envoy/envoy.go
```diff
@@ -319,17 +397,22 @@ func (srv *Server) parseLog(line string) (name string, logLevel string, msg stri
...
 	defer rc.Close()
+
+	bo := backoff.NewExponentialBackOff()
+
 	s := bufio.NewReader(rc)
 	for {
 		ln, err := s.ReadString('\n')
 		if err != nil {
-			if err == io.EOF {
+			if errors.Is(err, io.EOF) || errors.Is(err, os.ErrClosed) {
```
calebdoxsey, line 407: Saw this come up when the process was closed and spam stdout. We should now treat `ErrClosed` like `EOF`, but I also added exponential backoff to avoid the infinite loop if another error happens.

internal/envoy/envoy.go
```diff
@@ -40,53 +43,133 @@ const (
...
+		tracingOptions: *tracingOptions,
+	}
+
+	if cmp.Equal(srv.options, options, cmp.AllowUnexported(serverOptions{})) {
+		log.Debug().Str("service", "envoy").Msg("envoy: no config changes detected")
+		return
+	}
+	srv.options = options
+
+	// release the previous process so we can hot-reload
```
travisgroth, line 135: Can/should we do this after a successful spawn of the new process?

internal/envoy/envoy.go
travisgroth: 👍

internal/envoy/envoy.go
calebdoxsey: I think this is a bit tricky, cause there's not an easy way for us to tell when the old process is ready to be killed. 

The danger would be that we don't kill the child process... but then if pomerium is restarted, I think it will pick up the envoy that was already running, because the temporary base id file still exists. So maybe this is mostly a non-issue?

travisgroth: Just to make sure I'm clear on the behavior here...envoy self terminates the old process, right?  The `Release()` call just cleans up our references to the running process and doesn't actively kill it.  I didn't think we had to tell the old envoy process when to be killed.

Assuming that's all true - if we weren't able to start a new envoy,  we'd continue running with the old envoy (granted, without a reference to it) until a new one starts successfully.  That seems okay.

However, if we were in that state and received a kill signal, wouldn't we want to terminate the original envoy?  I don't think we'd be able to with this order of operations.

You're probably right that in a restart scenario, the abandoned process would get taken over with a new envoy if it comes up.  However, I think it's reasonable and a little less surprising if envoy terminates when the pomerium process does.

Thoughts?  There might be something I'm missing

calebdoxsey: I moved the process release to after we start the new process.

PR #1260, opened 2020-08-12
config: validate databroker settings

calebdoxsey:
## Summary
I forgot that we need to validate the options after updating them to populate some of the fields. With this change we'll validate the config after we update it in the databroker, and if validation fails we'll continue to use the existing config (and log the error).

Also I noticed we weren't skipping invalid policies like our error message said we were.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/databroker/config_source.go
+++ b/internal/databroker/config_source.go
@@ -74,12 +74,6 @@ func (src *ConfigSource) rebuild(firstTime bool) {
 	defer src.mu.Unlock()
 
 	cfg := src.underlyingConfig.Clone()
-	defer func() {
-		src.computedConfig = cfg
-		if !firstTime {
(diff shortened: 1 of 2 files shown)


internal/databroker/config_source.go
```diff
@@ -105,6 +99,7 @@ func (src *ConfigSource) rebuild(firstTime bool) {
 				log.Warn().Err(err).
 					Str("policy", policy.String()).
 					Msg("databroker: invalid policy, ignoring")
+				continue
```
desimone, line 102: sorry, I totally missed this.

Issue #1256, opened 2020-08-11
Pomerium cache exceeds Okta rate limit while querying for group membership info


bjoernw:
## What happened?
Okta gives you a 500 token rate limit and given a large enough number of okta groups and members the calls will exceed 500 (I'm not sure what the period is) which causes the data sync to fail. The end result is that the `allowed_groups` feature does not work.

https://developer.okta.com/docs/reference/api/groups/#list-groups
current batch size: 100

https://developer.okta.com/docs/reference/api/groups/#list-group-members 
current batch size: 100
recommended batch size: 200

## What did you expect to happen?
I would expect all routes protected with `allowed_groups` to work for users in those groups.

## How'd it happen?

1. Deployed 0.10.0
2. Tried any route with `allowed_groups`

## What's your environment like?

- 0.10.0
- Kubernetes + Istio

## What did you see in the logs?

```logs
{"level":"warn","service":"identity_manager","error":"okta: error querying for groups: okta: error query api status_code=429: 429 Too Many Requests","time":"2020-08-10T19:15:15Z","message":"failed to refresh directory users and groups"}
```

desimone:
Related https://github.com/pomerium/pomerium/issues/27#issuecomment-632437713

PR #1264, opened 2020-08-12
internal/directory/okta: increase default batch size to 200

cuonglm:
## Summary
See: https://developer.okta.com/docs/reference/api/groups/#list-groups-with-membership-updated-after-timestamp

## Related issues
Updates #1256

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/directory/okta/okta.go
+++ b/internal/directory/okta/okta.go
@@ -61,7 +61,7 @@ func WithServiceAccount(serviceAccount *ServiceAccount) Option {
 
 func getConfig(options ...Option) *config {
 	cfg := new(config)
-	WithBatchSize(100)(cfg)
+	WithBatchSize(200)(cfg)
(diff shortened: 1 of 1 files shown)


desimone:
I'd like to have a more comprehensive fix to this problem. And I don't know that we should be setting batch sizes at all since the getting the users in a group defaults to 1000 which would further circumvent the issue.

cuonglm:
> I'd like to have a more comprehensive fix to this problem. And I don't know that we should be setting batch sizes at all since the getting the users in a group defaults to 1000 which would further circumvent the issue.

See the link included in PR description, though the doc claims default is 1000, the example and note on `limit` shows that only 200 returns for most org.

This is not a fully fix, that's why the description says `Updates`.

PR #1271, opened 2020-08-13
Rate limit for okta

### LINKED ISSUE #1256: Pomerium cache exceeds Okta rate limit while querying for group membership info

cuonglm:
## Summary

Add QPS config for identity provider, and apply it for okta.

## Related issues

Fixes #1256

**Checklist**:
- [x] add related issues
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/directory/okta/okta.go
```diff
@@ -133,6 +152,9 @@ func (p *Provider) getGroups(ctx context.Context) ([]*directory.Group, error) {
 		RawQuery: fmt.Sprintf("limit=%d", p.cfg.batchSize),
 	}).String()
 	for groupURL != "" {
+		if err := p.limiter.Wait(ctx); err != nil {
+			return nil, fmt.Errorf("reached limit: %w", err)
```
calebdoxsey, line 156: I think this error message isn't quite right. Wait will error out if the context is cancelled or the deadline exceeded, but doesn't error out if the limit is exceeded.

calebdoxsey: Also can we move this call to the `apiGet` so we only have to do it once?

cuonglm: It can return error in case we reached the burst size.

Yes, moving to `apiGet` is right 👍

PR #1278, opened 2020-08-14
Improving okta API requests

cuonglm:
## Summary

This PR improves how we query okta API, by doing 2 things:

 - In case rate limit reached, we honor the response header from okta, sleep until that time and performing query again.
 - Using filters to query for updated groups only (both profile and membership), hence reducing number of requests we need to perform to okta API server.

## Related issues
Updates #1256 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/directory/okta/okta.go
+++ b/internal/directory/okta/okta.go
@@ -27,6 +27,9 @@ const Name = "okta"
 // See https://developer.okta.com/docs/reference/rate-limits/#okta-api-endpoints-and-per-minute-limits
 const defaultQPS = 100 / 60
 
+// Okta use ISO-8601, see https://developer.okta.com/docs/reference/api-overview/#media-types
+const filterDateFormat = "2006-01-02T15:04:05.999Z"
+
 type config struct {
 	batchSize      int
 	httpClient     *http.Client
@@ -86,9 +89,11 @@ func getConfig(options ...Option) *config {
 
 // A Provider is an Okta user group directory provider.
 type Provider struct {
-	cfg     *config
-	log     zerolog.Logger
-	limiter *rate.Limiter
+	cfg         *config
+	log         zerolog.Logger
+	limiter     *rate.Limiter
+	lastUpdated *time.Time
+	groups      map[string]*directory.Group
 }
 
 // New creates a new Provider.
@@ -101,6 +106,7 @@ func New(options ...Option) *Provider {
 		cfg:     cfg,
 		log:     log.With().Str("service", "directory").Str("provider", "okta").Logger(),
 		limiter: rate.NewLimiter(rate.Limit(cfg.qps), int(cfg.qps)),
+		groups:  make(map[string]*directory.Group),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


internal/directory/okta/okta.go
```diff
@@ -145,7 +157,51 @@ func (p *Provider) UserGroups(ctx context.Context) ([]*directory.Group, []*direc
...
+	if p.lastUpdated == nil {
+		p.lastUpdated = &now
+	}
+
+	u := &url.URL{Path: "/api/v1/groups"}
+	q := u.Query()
+	q.Set("limit", strconv.Itoa(p.cfg.batchSize))
+	q.Set("filter", fmt.Sprintf(`lastUpdated+gt+"%[1]s"+or+lastMembershipUpdated+gt+"%[1]s"`, p.lastUpdated.UTC().Format(filterDateFormat)))
+	u.RawQuery = q.Encode()
+	p.lastUpdated = &now
```
calebdoxsey, line 171: This may be ok as is, but sometimes clocks are out of sync and it's generally better to rely on whatever timestamps the server returns. For example, we could loop over all the returned records and use the latest `lastUpdated` or `lastMembershipUpdated` timestamp.

internal/directory/okta/okta.go
```diff
@@ -214,6 +272,25 @@ func (p *Provider) apiGet(ctx context.Context, uri string, out interface{}) (htt
...
+				time.Sleep(time.Until(time.Unix(limitReset, 0)))
+				continue
+			}
+		}
+		if res.StatusCode/100 == 2 {
+			break
+		}
+		return nil, fmt.Errorf("okta: error query api status_code=%d: %s", res.StatusCode, res.Status)
+	}
 	res, err := p.cfg.httpClient.Do(req)
```
calebdoxsey, line 294: Are we accidentally making the same request twice here?

cuonglm: Ops, sorry, copy-paste error 😄

internal/directory/okta/okta.go
cuonglm: 👍

internal/directory/okta/okta.go
```diff
@@ -145,7 +157,61 @@ func (p *Provider) UserGroups(ctx context.Context) ([]*directory.Group, []*direc
...
+func (p *Provider) getUpdatedGroup(ctx context.Context) ([]*directory.Group, error) {
+	now := time.Now()
+	if p.lastUpdated == nil {
+		p.lastUpdated = &now
+	}
+
+	u := &url.URL{Path: "/api/v1/groups"}
+	q := u.Query()
+	q.Set("limit", strconv.Itoa(p.cfg.batchSize))
+	q.Set("filter", fmt.Sprintf(`lastUpdated+gt+"%[1]s"+or+lastMembershipUpdated+gt+"%[1]s"`, p.lastUpdated.UTC().Format(filterDateFormat)))
```
calebdoxsey, line 169: is this the only difference between `getUpdatedGroup` and `getGroups`? Maybe it would be better to just only set this parameter if `p.lastUpdated` is not nil? Then the first time we would just retrieve all the groups?

cuonglm: Updated.

PR #1280, opened 2020-08-14
proxy: move properties to atomically updated state

calebdoxsey:
## Summary
The same as #1277 but for the proxy service. I also added a new `GetGRPCClientConn` method so that we could re-use gRPC connections if none of the options change. 

In this case just rebuilding the state everytime seemed like the easiest approach, and this lets us do that for gRPC connections without having to disconnect/reconnect unless something changes.

## Related issues
- #1254 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/grpc/client.go
```diff
@@ -144,3 +146,46 @@ func grpcTimeoutInterceptor(timeout time.Duration) grpc.UnaryClientInterceptor {
...
+	cc, err := NewGRPCClientConn(opts)
+	if err != nil {
+		return nil, err
+	}
+
+	grpcClientConns.m[name] = grpcClientConnRecord{
+		conn: cc,
+		opts: opts,
+	}
+	return current.conn, nil
```
cuonglm, line 190: I guess you want to return `grpcClientConns.m[name]` instead?

pkg/grpc/client.go
```diff
@@ -144,3 +146,46 @@ func grpcTimeoutInterceptor(timeout time.Duration) grpc.UnaryClientInterceptor {
...
+
+	current, ok := grpcClientConns.m[name]
+	if ok {
+		if cmp.Equal(current.opts, opts) {
+			return current.conn, nil
+		}
+
+		err := current.conn.Close()
+		if err != nil {
+			log.Error().Err(err).Msg("grpc: failed to close existing connection")
```
cuonglm, line 177: I think we should `Warn` only, as we are going to ignore this anyway.

proxy/forward_auth.go
```diff
@@ -18,7 +18,10 @@ import (
 func (p *Proxy) registerFwdAuthHandlers() http.Handler {
 	r := httputil.NewRouter()
 	r.StrictSlash(true)
-	r.Use(sessions.RetrieveSession(p.sessionStore))
+	r.Use(func(h http.Handler) http.Handler {
+		state := p.state.Load()
```
cuonglm, line 22: Maybe combine this to one line for consistent with other places?

pkg/grpc/client.go
```diff
@@ -144,3 +146,46 @@ func grpcTimeoutInterceptor(timeout time.Duration) grpc.UnaryClientInterceptor {
...
+// GetGRPCClientConn returns a gRPC client connection for the given name. If a connection for that name has already been
+// established the existing connection will be returned. If any options change for that connection, the existing
+// connection will be closed and a new one established.
+func GetGRPCClientConn(name string, opts *Options) (*grpc.ClientConn, error) {
+	grpcClientConns.Lock()
+	defer grpcClientConns.Unlock()
+
+	current, ok := grpcClientConns.m[name]
+	if ok {
+		if cmp.Equal(current.opts, opts) {
```
desimone, line 171: Could we add test for `GetGRPCClientConn`.

proxy/state.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/pomerium/pomerium/internal/sessions"
+	"github.com/pomerium/pomerium/internal/sessions/cookie"
+	"github.com/pomerium/pomerium/internal/sessions/header"
+	"github.com/pomerium/pomerium/internal/sessions/queryparam"
+	"github.com/pomerium/pomerium/internal/urlutil"
+	"github.com/pomerium/pomerium/pkg/cryptutil"
+	"github.com/pomerium/pomerium/pkg/grpc"
+)
+
+type proxyState struct { // SharedKey used to mutually authenticate service communication
```
desimone, line 25: typo?

proxy/state.go
```diff
@@ -0,0 +1,124 @@
...
+	value atomic.Value
+}
+
+func newAtomicProxyState(state *proxyState) *atomicProxyState {
+	aps := new(atomicProxyState)
+	aps.Store(state)
+	return aps
+}
+
+func (aps *atomicProxyState) Load() *proxyState {
```
desimone, line 118: How big of a performance hit is Load? Should we add a benchmark just to get a sense?

pkg/grpc/client.go
desimone: Counterpoint, we "ignore" if there is an error here too, but I definitely think this should be bubbled up as an error.

https://github.com/pomerium/pomerium/pull/1280/files#diff-3ac192716a1df66bbd1008d327af8792R93

proxy/state.go
calebdoxsey: Load is very lightweight. It consists of two pointer loads from memory, so almost no different than a normal pointer. Store uses the `cmpxchg` processor instruction.

Because we always do a full replace of the struct, we don't have to worry about modifying memory used by multiple threads, so we can avoid a mutex.

pkg/grpc/client.go
calebdoxsey: Yep that was a mistake. Nice catch.

pkg/grpc/client.go
desimone: > equality is determined by recursively comparing the primitive kinds on both values, much like reflect.DeepEqual. Unlike reflect.DeepEqual, unexpected fields are not compared by default;

Mostly a note for myself, but options does not have any unexported fields so this is better.

PR #1292, opened 2020-08-17
authenticate: move databroker connection to state

calebdoxsey:
## Summary
This PR moves the databroker connection in authenticate to the state. This way if the connection changes it will pick up the new connection automatically.

I also added a call to `ValidateOptions` every time we rebuild the state.

## Related issues
- #1254 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authenticate/handlers.go
+++ b/authenticate/handlers.go
@@ -144,14 +144,17 @@ func (a *Authenticate) VerifySession(next http.Handler) http.Handler {
 	return httputil.HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
 		ctx, span := trace.StartSpan(r.Context(), "authenticate.VerifySession")
 		defer span.End()
+
+		state := a.state.Load()
+
 		sessionState, err := a.getSessionFromCtx(ctx)
 		if err != nil {
 			log.FromRequest(r).Info().Err(err).Msg("authenticate: session load error")
 			return a.reauthenticateOrFail(w, r, err)
 		}
 
-		if a.dataBrokerClient != nil {
-			_, err = session.Get(ctx, a.dataBrokerClient, sessionState.ID)
+		if state.dataBrokerClient != nil {
+			_, err = session.Get(ctx, state.dataBrokerClient, sessionState.ID)
 			if err != nil {
 				log.FromRequest(r).Info().Err(err).Str("id", sessionState.ID).Msg("authenticate: session not found in databroker")
 				return a.reauthenticateOrFail(w, r, err)
@@ -232,7 +235,7 @@ func (a *Authenticate) SignIn(w http.ResponseWriter, r *http.Request) error {
 	if r.FormValue(urlutil.QueryIsProgrammatic) == "true" {
 		newSession.Programmatic = true
 
-		pbSession, err := session.Get(ctx, a.dataBrokerClient, s.ID)
+		pbSession, err := session.Get(ctx, state.dataBrokerClient, s.ID)
 		if err != nil {
 			return httputil.NewError(http.StatusBadRequest, err)
 		}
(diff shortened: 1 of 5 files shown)


authenticate/state.go
```diff
@@ -52,6 +56,10 @@ func newAuthenticateState() *authenticateState {
 }
 
 func newAuthenticateStateFromConfig(cfg *config.Config) (*authenticateState, error) {
+	if err := ValidateOptions(cfg.Options); err != nil {
```
desimone, line 59: nit: if we make this `err` function scoped we can remove the uninitizialized err further down

authenticate/handlers.go
```diff
@@ -474,10 +477,12 @@ func (a *Authenticate) getSessionFromCtx(ctx context.Context) (*sessions.State,
...
 
 func (a *Authenticate) deleteSession(ctx context.Context, sessionID string) error {
-	if a.dataBrokerClient == nil {
+	state := a.state.Load()
+
+	if state.dataBrokerClient == nil {
 		return nil
 	}
-	err := session.Delete(ctx, a.dataBrokerClient, sessionID)
+	err := session.Delete(ctx, state.dataBrokerClient, sessionID)
```
desimone, line 485: nit: I think we can just return here?

authenticate/handlers.go
```diff
@@ -474,10 +477,12 @@ func (a *Authenticate) getSessionFromCtx(ctx context.Context) (*sessions.State,
 }
 
 func (a *Authenticate) deleteSession(ctx context.Context, sessionID string) error {
-	if a.dataBrokerClient == nil {
+	state := a.state.Load()
+
+	if state.dataBrokerClient == nil {
```
desimone, line 482: If databrokerclient is nil here, do we want to error?

authenticate/handlers.go
```diff
@@ -489,32 +494,34 @@ func (a *Authenticate) isAdmin(user string) bool {
...
 func (a *Authenticate) Dashboard(w http.ResponseWriter, r *http.Request) error {
+	state := a.state.Load()
+
 	s, err := a.getSessionFromCtx(r.Context())
 	if err != nil {
 		s.ID = uuid.New().String()
 	}
 
-	pbSession, err := session.Get(r.Context(), a.dataBrokerClient, s.ID)
+	pbSession, err := session.Get(r.Context(), state.dataBrokerClient, s.ID)
```
desimone, line 504: I don't think session.Get/Set/Delete are currently handling if `dataBrokerClient` is nil?

authenticate/handlers.go
```diff
@@ -144,14 +144,17 @@ func (a *Authenticate) VerifySession(next http.Handler) http.Handler {
...
+
 		sessionState, err := a.getSessionFromCtx(ctx)
 		if err != nil {
 			log.FromRequest(r).Info().Err(err).Msg("authenticate: session load error")
 			return a.reauthenticateOrFail(w, r, err)
 		}
 
-		if a.dataBrokerClient != nil {
-			_, err = session.Get(ctx, a.dataBrokerClient, sessionState.ID)
+		if state.dataBrokerClient != nil {
```
desimone, line 156: If the databroker client is nil, the request will go through? I don't think that is what we want to do here?

cuonglm:
> There are several instances where we do a `nil` check on the `dataBrokerClient` prior to making a series of calls to `session`/`user`/`directory`, should we move the nil check logic here to the underlying Get/Set/Delete functions instead of making it the callers responsibility to make sure the client is not nil?

I don't think we need to check that `dataBrokerClient` is nil. Technically, it can be nil, but we always initialize it in all of our code path, the only way it can be nil if we construct an instance of `Authenticate` directly, without calling update options. That will lead to more loudly and sooner panic.

calebdoxsey:
@cuonglm @desimone The nil check is unnecessary. It was part of the original code, but I've removed it. I believe it was added to let the tests pass, which I've now updated.

PR #1294, opened 2020-08-17
cache: support databroker option changes

calebdoxsey:
## Summary
This updates the cache (databroker) service to update the internal databroker server whenever configuration options change.

To implement this I added `Close` methods to `redis`/`inmemory` and re-create the databases if the options change such that a rebuild is necessary. In the case of `inmemory` this would cause all data to be erased, but for redis the switchover should be largely invisible.

Many of the databroker options aren't available in the dashboard currently but we will likely want to add them.

## Related issues
- #1254 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -3,7 +3,10 @@ package databroker
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
 	"reflect"
 	"sort"
 	"sync"
@@ -11,6 +14,7 @@ import (
 
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/google/go-cmp/cmp"
 	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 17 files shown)


internal/databroker/server.go
```diff
@@ -335,29 +369,56 @@ func (srv *Server) SyncTypes(req *emptypb.Empty, stream databroker.DataBrokerSer
...
-			return nil, err
+			return nil, "", err
 		}
 	}
-	return db, nil
+	return db, version, nil
 }
 
 func (srv *Server) newDB(recordType string) (db storage.Backend, err error) {
+	caCertPool := x509.NewCertPool()
```
desimone, line 405: should we pull this out into it's own function? buildTLSConfig or something?

internal/databroker/server.go
```diff
@@ -70,19 +78,19 @@ func New(options ...ServerOption) *Server {
...
+				db.ClearDeleted(context.Background(), tm)
 			}
 		}
 	}()
 	return srv
 }
 
 func (srv *Server) initVersion() {
-	dbServerVersion, err := srv.getDB(recordTypeServerVersion)
+	dbServerVersion, _, err := srv.getDB(recordTypeServerVersion, false)
```
desimone, line 93: this is the only place we do not lock when getting db right?

cuonglm: We don't need because this's only run once at server startup.

PR #1300, opened 2020-08-18
azure: support deriving credentials from client id, client secret and provider url

### LINKED ISSUE #1299: azure: derive credentials from client id/secret and provider URL
Currently we require custom service account credentials for Azure AD, but we actually have all the needed parameters in other configuration settings.

The client id and secret can be used directly. The directory (tenant) id can be derived from the provider URL. For example for this URL:

- `https://login.microsoftonline.com/0303f438-3c5c-4190-9854-08d3eb31bd9f/v2.0`

We can use `0303f438-3c5c-4190-9854-08d3eb31bd9f` as the directory id.

We can still support passing custom credentials, but we should fallback to using these settings.

calebdoxsey:
## Summary
This PR updates the Azure directory provider so that it supports deriving service account credentials from the existing configuration options. 

## Related issues
Fixes #1299 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/directory/provider.go
+++ b/internal/directory/provider.go
@@ -24,17 +24,12 @@ type Group = directory.Group
 // A User is a directory User.
 type User = directory.User
 
+// Options are the options specific to the provider.
+type Options = directory.Options
+
 // A Provider provides user group directory information.
 type Provider interface {
 	UserGroups(ctx context.Context) ([]*Group, []*User, error)
-}
-
-// Options are the options specific to the provider.
-type Options struct {
-	ServiceAccount string
-	Provider       string
-	ProviderURL    string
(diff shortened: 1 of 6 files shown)


internal/directory/provider.go
```diff
@@ -24,19 +24,14 @@ type Group = directory.Group
 // A User is a directory User.
 type User = directory.User
 
+// Options are the options specific to the provider.
+type Options = directory.Options
```
calebdoxsey, line 28: this was moved to avoid an import cycle

PR #1306, opened 2020-08-20
pkg/storage/redis: update tests to use local certs + upstream image

### LINKED ISSUE #1304: Refactor redis test code
Remove the external/custom container dependency in the tests by mounting the generated TLS certs directly.

travisgroth:
## Summary
- Update docker interaction to mount certs locally
- Change to using the upstream redis image

## Related Issues
Fixes #1304

**Checklist**:
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/storage/redis/redis_test.go
+++ b/pkg/storage/redis/redis_test.go
@@ -48,12 +48,12 @@ func tlsConfig(rawURL string, t *testing.T) *tls.Config {
 	return tlsConfig
 }
 
-func runWithRedisDockerImage(repo, tag string, env []string, withTLS bool, testFunc func(t *testing.T), t *testing.T) {
+func runWithRedisDockerImage(t *testing.T, runOpts *dockertest.RunOptions, withTLS bool, testFunc func(t *testing.T)) {
 	pool, err := dockertest.NewPool("")
 	if err != nil {
 		t.Fatalf("Could not connect to docker: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/storage/redis/redis_test.go
```diff
@@ -88,27 +88,29 @@ func TestDB(t *testing.T) {
...
 		t.Skip("Github action can not run docker on MacOS")
 	}
-	redisTLSEnv := []string{
-		"ALLOW_EMPTY_PASSWORD=yes",
-		"REDIS_TLS_ENABLED=yes",
-		"REDIS_TLS_CERT_FILE=/tls/redis.crt",
-		"REDIS_TLS_KEY_FILE=/tls/redis.key",
-		"REDIS_TLS_CA_FILE=/tls/ca.crt",
+
+	cwd, err := os.Getwd()
```
calebdoxsey, line 92: we might need to lookup the path of the source code. I'm not sure the current working directory is set consistently (ie if I run `go test ./...` wouldn't it be set to the project dir?)

travisgroth: Good question.  It does work in the same directory as the tests and at the top level (eg make test), so it seems like it does the right thing WRT the execution location.  That might be special behavior for tests?

calebdoxsey: Cool it seems to work.

PR #1353, opened 2020-08-31
internal/controlplane: add telemetry http handler

### LINKED ISSUE #535: Propagate trace headers

Distributed tracing is useful for tracking requests across a series of services. 
Pomerium already implements tracing through Jaeger for internal comms,
but these are not coordinated with other services

- Forward Auth: Extract Trace ID from headers (maybe a security concern?)
- Proxy: Inject Trace ID headers into outgoing requests

wrap Pomerium in a tracing sideca, but lose visibility into internal tracing

**Additional context**

"all" is not a very useful name in all-in-one mode, may I suggest "pomerium"

travisgroth:
## Summary
Ensure we're getting proper tracing with propagation on the control plane http endpoints.

This is arguably a bug fix, so marking it for v0.10.x.

## Related issues
closes #535 

**Checklist**:
- [x] add related issues
- [x] ready for review

### DIFF
--- a/internal/controlplane/http.go
+++ b/internal/controlplane/http.go
@@ -11,6 +11,7 @@ import (
 	"github.com/pomerium/pomerium/internal/httputil"
 	"github.com/pomerium/pomerium/internal/log"
 	"github.com/pomerium/pomerium/internal/middleware"
+	"github.com/pomerium/pomerium/internal/telemetry"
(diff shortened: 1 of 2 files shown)


travisgroth:
pre-commit failure is a missing `go mod tidy`.  Would rather do that separately.

cuonglm (approved):

PR #1358, opened 2020-09-01
certmagic: improve logging

calebdoxsey:
## Summary
Currently our logs for certmagic aren't JSON-encoded which can break log parsers. This change upgrades certmagic, which now has support for zap logging. 

Initially I tried building a translation layer from zap to zerolog, but that proved difficult, and I realized as long as we use JSON, it doesn't actually matter which library sends the log lines to stdout. I added a new `ZapLogger` method in `log` to return the zap logger and changed the timestamp format to match what we're doing with zerolog.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/autocert/manager.go
```diff
@@ -43,6 +44,7 @@ func New(src config.Source) (*Manager, error) {
 	certmagic.Default.Storage = &certmagic.FileStorage{
 		Path: src.GetConfig().Options.AutocertOptions.Folder,
 	}
+	certmagic.Default.Logger = log.ZapLogger().With(zap.String("service", "certmagic"))
```
desimone, line 47:
```suggestion
	certmagic.Default.Logger = log.ZapLogger().With(zap.String("service", "autocert"))
```

To avoid confusion?

cuonglm (approved):

PR #1361, opened 2020-09-01
proxy: disable control-plane robots.txt for public unauthenticated routes

### LINKED ISSUE #837: Allow to toggle robots.txt in policy

I am using pomerium for internal and public websites, which I selfhost. (not alot of traffic) 
Currently pomerium overwrites the robots.txt file on any policy, even with `unauthenticated_public_acces=true`. 

I think this is a correct security measure, but it would be nice if this was toggleable per policy. Something like `allow_robots=true`. This way we dont need to setup an extra reverse proxy like nginx for small personal blog. 

One could use an additional webserver like caddy or nginx to use as a reverse proxy alongside pomerium. But it would be much easier for low traffic sites to just be able to active the robots.txt file.

**Additional context**

Current robots.txt response:
`User-agent: *
Disallow: /`

Thanks in advance!

calebdoxsey:
## Summary
For public unauthenticated routes we want `robots.txt` to be forwarded to the backend. This PR implements that change.

## Related issues
Closes #837 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/grpc.go
+++ b/authorize/grpc.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
-	"regexp"
 	"strings"
 
 	"github.com/golang/protobuf/ptypes"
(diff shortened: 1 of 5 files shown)


cuonglm (approved):

fightforlife:
@calebdoxsey Is there a plan when this will be included in the latest releases?

calebdoxsey:
@fightforlife we're planning to release this with v0.11 because its a new feature. We don't have a concrete date, but likely in the next month.

fightforlife:
Okay, totally understandable! Thank you for the feedback.

PR #1362, opened 2020-09-01
proxy: support websocket timeouts

### LINKED ISSUE #1327: Configurable timeout when using websockets

Websockets based connections never expire, even when a user logs out.  Currently we have no request timeout when websockets are enabled.

Ideally, users could provide a maximum timeout for these requests so that connections must be reestablished and, thus, reauthorized occasionally.

N/A

N/A

**Additional context**

Related to #1279

calebdoxsey:
## Summary
With this change we support using the `timeout` property on a policy with `allow_websockets`, though we still don't use the global timeout.

## Related issues
Closes #1327 

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/controlplane/xds_routes.go
+++ b/internal/controlplane/xds_routes.go
@@ -223,8 +223,12 @@ func getRequestHeadersToRemove(options *config.Options, policy *config.Policy) [
 func getRouteTimeout(options *config.Options, policy *config.Policy) *durationpb.Duration {
 	var routeTimeout *durationpb.Duration
 	if policy.AllowWebsockets {
-		// disable the route timeout for websocket support
-		routeTimeout = ptypes.DurationProto(0)
+		if policy.UpstreamTimeout != 0 {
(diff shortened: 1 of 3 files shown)


internal/controlplane/xds_routes.go
```diff
@@ -223,8 +223,12 @@ func getRequestHeadersToRemove(options *config.Options, policy *config.Policy) [
 func getRouteTimeout(options *config.Options, policy *config.Policy) *durationpb.Duration {
 	var routeTimeout *durationpb.Duration
 	if policy.AllowWebsockets {
-		// disable the route timeout for websocket support
-		routeTimeout = ptypes.DurationProto(0)
+		if policy.UpstreamTimeout != 0 {
+			routeTimeout = ptypes.DurationProto(policy.UpstreamTimeout)
+		} else {
+			// disable the default route timeout for websocket support
```
desimone, line 229: I think this comment is now misplaced?

PR #1374, opened 2020-09-04
authorize: add support for service accounts

calebdoxsey:
## Summary
This PR adds support for service accounts stored in the databroker. These service accounts have an associated `id` and `user_id` and a future PR will allow admins to delete them. (they also have an optional expiration)

The `pomerium-cli` command was updated to add the service account to the databroker.

I also discovered that the databroker wasn't updating new types properly.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/databroker/server.go
```diff
@@ -390,6 +390,7 @@ func (srv *Server) getDB(recordType string, lock bool) (db storage.Backend, vers
 		if db == nil {
 			db, err = srv.newDB(recordType)
 			srv.byType[recordType] = db
+			defer srv.onTypechange.Broadcast()
```
desimone, line 393: can you explain this a bit more, I'm having trouble following the flow.

desimone (approved):
LGTM, one question.

internal/databroker/server.go
desimone: Thanks, makes sense.

PR #1408, opened 2020-09-16
azure: add support for nested groups

### LINKED ISSUE #1397: Question: Nested groups for Azure Authentication
Hello, 

I have a question regardign the implementation of Pomerium with Azure AD, we have created a pomerium deployment with a `policy` that uses `allowed_groups` feature. 

The Issue is, the group specified in the Pomerium config consists of other groups ( groups are nested), but when the token for the user is returned, the membership of the user is not correct, if the user is not **directly** assigned to that group. I have searched around and have not found any configuration that would allow for checking transitive attributes of groups for user.

 I have tried giving the Service principal permissions to read directory [(See microsoft docs)](https://docs.microsoft.com/en-us/graph/api/group-list-transitivemembers?view=graph-rest-1.0&tabs=http), but that has not solved the issue. Is there a way for pomerium to see transitive group membership of Azure AD?

Thank you for your help

calebdoxsey:
## Summary
With Azure active directory a group may contain other groups. For example, group 1 may contain group 2, and group 2 may contain user 1, therefore user 1 is a member of group 2 and group 1. These changes implement that behavior.

I pulled out the code to do that into a `groupLookup` type which implements the recursion using a stack and also should prevent cycles.

## Related issues
Fixes #1397 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/directory/azure/azure.go
+++ b/internal/directory/azure/azure.go
@@ -111,24 +111,20 @@ func (p *Provider) UserGroups(ctx context.Context) ([]*directory.Group, []*direc
 		return nil, nil, err
 	}
 
-	userIDToGroupIDs := map[string][]string{}
+	groupLookup := newGroupLookup()
 	for _, group := range groups {
-		userIDs, err := p.listGroupMembers(ctx, group.Id)
+		groupIDs, userIDs, err := p.listGroupMembers(ctx, group.Id)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


desimone (approved):
LGTM sans the broken test.

PR #1447, opened 2020-09-23
proxy/forward_auth: don't reset forward auth path if X-Forwarded-Uri is not set

whs:
## Summary
When using Pomerium forward auth with nginx, after authentication succeeded it will always redirect to domain root.

This is due to a codepath that is intended for reading Traefik original URL always hit when Traefik's header is missing.

**Checklist**:
- [ ] updated unit tests - I don't have test added - if you require one please suggest where should I add it
- [x] ready for review


desimone (approved):

desimone:
@whs do you mind fixing the contributor check? 

Thanks again for the PR.

whs:
Ok fixed, my work recently changed their primary email domain and I didn't have it on GitHub

Issue #1452, opened 2020-09-24
Domain/URL Rewrite


followsound:
**Is your feature request related to a problem? Please describe.**

Off the back of a conversation with @desimone relating to #523, it discussed that we could get around the issue of not having multiple IdP support by using domain/URL rewrites.

One could have two (or more) instances of Pomerium running at various domains, with the destination domain being rewritten by Pomerium to avoid applications expecting a single URL format.

**Describe the solution you'd like**

Consider a self-hosted GitLab, and two IdPs - GSuite and AzureAD (note here users are only present in a single IdP), with the GitLab hosted at gitlab.company.com. The AzureAD is a contracting company that needs access to our companies' code.

One Pomerium could run to serve `gitlab.company.com` with GSuite, and another to run `contract.gitlab.company.com` with AzureAD configured as the IdP.

However, in this situation GitLab (and other applications) would try to rewrite the URL to `gitlab.company.com`, forcing the AzureAD users to auth with GSuite.

URL rewriting would solve this by setting a URL rewrite to `gitlab.company.com` in the second Pomerium running `contract.gitlab.company.com`

**Describe alternatives you've considered**

It would be possible with paid products, such as Cloudflare Access, and is possible with oauth2proxy (but with lots of faffing).

**Explain any additional use-cases**

Currently, my solution is to add the contractors to our GSuite IdP. However, this is not scalable, and would not work if a company was bought out, and maintained two separate IdPs.

**Additional context**

My hope would be this solution is something that can be a stop-gap until multiple IdP support is discussed, but it would get the job done - just in a less elegant way.

calebdoxsey:
Hi @followsound . We've just merged a change to master that will allow you to customize host header rewriting via 3 possible options:

1. `host_rewrite` which will rewrite the host to a new literal value.
2. `host_rewrite_header` which will rewrite the host to match an incoming header value.
3. `host_path_regex_rewrite_pattern`, `host_path_regex_rewrite_substitution` which will rewrite the host according to a regex matching the path.

Let me know if that works for you. Thx.

followsound:
Thanks @calebdoxsey will take a look ASAP 😊

PR #1457, opened 2020-09-24
config: add support for host header rewriting

calebdoxsey:
## Summary
This PR adds new policy config options for customizing host header rewriting. I basically copied the available options in envoy: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#config-route-v3-routeaction.

I updated the readme with an explanation of the value:

### Host Rewrite

- `yaml`/`json` settings: `host_rewrite`, `host_rewrite_header`, `host_path_regex_rewrite_pattern`, `host_path_regex_rewrite_substitution`
- Type: `string`
- Optional
- Example: `host_rewrite: "example.com"`

The `host` header can be customized via 3 mutually exclusive options:

1. `host_rewrite` which will rewrite the host to a new literal value.
2. `host_rewrite_header` which will rewrite the host to match an incoming header value.
3. `host_path_regex_rewrite_pattern`, `host_path_regex_rewrite_substitution` which will rewrite the host according to a regex matching the path. For example with the following config:

    ```yaml
    host_path_regex_rewrite_pattern: "^/(.+)/.+$"
    host_path_regex_rewrite_substitution: \1
    ```

    Would rewrite the host header to `example.com` given the path `/example.com/some/path`.

These options correspond to the envoy route action host related options, which can be found [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#config-route-v3-routeaction).

## Related issues
- Implements #1452 

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/reference/readme.md
+++ b/docs/reference/readme.md
@@ -1057,6 +1057,28 @@ prefix_rewrite: /
 
 A request to `https://from.example.com/admin` would be forwarded to `https://to.example.com/`.
 
+### Host Rewrite
+
+- `yaml`/`json` settings: `host_rewrite`, `host_rewrite_header`, `host_path_regex_rewrite_pattern`, `host_path_regex_rewrite_substitution`
+- Type: `string`
+- Optional
+- Example: `host_rewrite: "example.com"`
+
+The `host` header can be customized via 3 mutually exclusive options:
+
+1. `host_rewrite` which will rewrite the host to a new literal value.
+2. `host_rewrite_header` which will rewrite the host to match an incoming header value.
+3. `host_path_regex_rewrite_pattern`, `host_path_regex_rewrite_substitution` which will rewrite the host according to a regex matching the path. For example with the following config:
+
+    ```yaml
+    host_path_regex_rewrite_pattern: "^/(.+)/.+$"
+    host_path_regex_rewrite_substitution: \1
+    ```
+
+    Would rewrite the host header to `example.com` given the path `/example.com/some/path`.
+
+These options correspond to the envoy route action host related options, which can be found [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#config-route-v3-routeaction).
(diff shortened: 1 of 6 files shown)


calebdoxsey:
@desimone @travisgroth here's an implementation of the host header rewriting. I added all the possible options I could find in envoy, but we could remove some (or rename them) if we don't think they're useful.

We already had one option (`preserve_host_header`), so these changes should also handle that.

desimone (commented):
Code LGTM. Nice work. 

I think it makes sense to include all the rewrite options that envoy provides; and can't think of a good reason not to.  @travisgroth thoughts?

travisgroth:
> I think it makes sense to include all the rewrite options that envoy provides; and can't think of a good reason not to. @travisgroth thoughts?

Agreed.  Seems pretty low overhead, and is better than handling them one at a time as use cases pop up. 👍

docs/reference/readme.md
```diff
@@ -1057,6 +1057,28 @@ prefix_rewrite: /
...
+1. `host_rewrite` which will rewrite the host to a new literal value.
+2. `host_rewrite_header` which will rewrite the host to match an incoming header value.
+3. `host_path_regex_rewrite_pattern`, `host_path_regex_rewrite_substitution` which will rewrite the host according to a regex matching the path. For example with the following config:
+
+    ```yaml
+    host_path_regex_rewrite_pattern: "^/(.+)/.+$"
+    host_path_regex_rewrite_substitution: \1
+    ```
+
+    Would rewrite the host header to `example.com` given the path `/example.com/some/path`.
```
desimone, line 1078:
```suggestion
Would rewrite the host header to `example.com` given the path `/example.com/some/path`.
```

desimone (approved):

PR #1460, opened 2020-09-25
directory: add explicit RefreshUser endpoint for faster sync

calebdoxsey:
## Summary
This implements refreshing user groups on login which may help in cases where the initial directory sync takes a while.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/directory/github/github.go
```diff
@@ -230,6 +272,77 @@ func (p *Provider) getUser(ctx context.Context, userLogin string) (*apiUserObjec
 	return &res, nil
 }
 
+func (p *Provider) listUserOrganizationTeams(ctx context.Context, userSlug string, orgSlug string) ([]int, error) {
+	// GitHub's Rest API doesn't have an easy way of querying this data, so we use the GraphQL API.
```
desimone, line 276: Wow this is wild. Nice work. I don't have a ton of experience with graphql but...uh, I kinda hope they don't go fully this direction.

cache/directory.go
```diff
@@ -0,0 +1,44 @@
...
+	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/pomerium/pomerium/pkg/grpc/databroker"
+	"github.com/pomerium/pomerium/pkg/grpc/directory"
+)
+
+// RefreshUser refreshes a user's directory information.
+func (c *Cache) RefreshUser(ctx context.Context, req *directory.RefreshUserRequest) (*emptypb.Empty, error) {
+	c.mu.Lock()
```
desimone, line 16: Sorry, I don't fully understand the mutex here. Can you explain?

desimone (approved):
Code LGTM; I haven't had a chance to manually test all of the providers yet.

cache/directory.go
calebdoxsey: When config is updated it will reset the directory provider: https://github.com/pomerium/pomerium/pull/1460/files/047a53f8a71ec724941ebaf9ecce91a529232823#diff-f7b51df04a56b62e71e037c6cacb4d6bR141

The mutex is necessary to avoid a race on reading/writing the variable.

calebdoxsey:
I will need to merge in the auth0 changes to this refactor.

cache/directory.go
desimone: Got it. Thanks

PR #1467, opened 2020-09-28
directory: additional user info

calebdoxsey:
## Summary
This adds email and name to the directory user information so that it can be searched.

I tested the various providers:

- Github rarely returns email because the user has to have selected it to be public. There didn't look to be an easy way to get it anywhere else in the API, so we'll fill it if its available but otherwise it will be blank.
- Gitlab was similar to Github.
- Azure doesn't return the email consistently, but it does return the address in the User Principal Name, so I wrote a function to use that if no email is provided

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/directory/google/google.go
```diff
@@ -110,6 +110,24 @@ func (p *Provider) UserGroups(ctx context.Context) ([]*directory.Group, []*direc
 		return nil, nil, fmt.Errorf("google: error getting groups: %w", err)
 	}
 
+	userLookup := map[string]apiUserObject{}
```
calebdoxsey, line 113: our docs already recommend the `user.readonly` permission, so this worked without having to change my svc account

internal/directory/azure/azure_test.go
```diff
@@ -92,14 +92,20 @@
...
 			Id:       "azure/user-2",
 			GroupIds: []string{"test"},
+			Name:     "User 2",
+			Email:    "user2@example.com",
 		},
 		{
 			Id:       "azure/user-3",
 			GroupIds: []string{"test"},
+			Name:     "User 3",
+			Email:    "user3@example.com",
```
desimone, line 108: Could we add a test for when the email is parsed from the UPN?

desimone (approved):
> This adds email and name to the directory user information so that it can be searched.
> 
> I tested the various providers:
> 
> Github rarely returns email because the user has to have selected it to be public. There didn't look to be an easy way to get it anywhere else in the API, so we'll fill it if its available but otherwise it will be blank.

👍

> Gitlab was similar to Github.

👍👍 

As an aside, it's crazy how much gitlab looks like the Github API. Even where it doesn't make a ton of sense. 

> Azure doesn't return the email consistently, but it does return the address in the User Principal Name, so I wrote a function to use that if no email is provided

Sounds good. One nit might be to add a test case that covers when we parse the UPN response.

PR #1415, opened 2020-09-17
proxy: forward_auth honour uri query string when checking policy

whs:
## Summary
Currently forward proxy check the policy for the FORWARD_AUTH_HOST/validate endpoint. This PR changes it to the endpoint given by uri query string, allowing per-endpoint policy like non-forward auth setup.

**Checklist**:
- [ ] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review

PR #1295, opened 2020-08-18
Move forward auth to separated service

cuonglm:
## Summary
This commit move forward auth to separated service. While at it, also
cleanup some un-used code in proxy.

Refactoring forward auth service will be done in future commits.

**Checklist**:
- [ ] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

PR #1482, opened 2020-10-01
forward-auth: use envoy's ext_authz check

desimone:
## Summary

These changes modify the way we currently handle third-party fronting proxies (traefik/nginx/etc) forward-auth requests. Previously, would forward all forward-auth requests (skipping envoy's ext_authz) to pomerium's forward-auth go handlers which would make a follow up call to the authorize service over gRPC. 

This PR instead attempts to consolidate authorization paths so that all requests, even when fronted by another proxy, still leverage envoy's internal `ext_authz` call. 

The benefit being:

- single code path for authorization (envoy --> ext_authz --> authZ) 
- proxy no longer needs to talk to the authorize service
- we can now extract forward-auth into it's own service, if we want to do that.

These changes should be backward compatible, and introduce no breaking changes. 

## Related issues

- Closes https://github.com/pomerium/internal/issues/153
- https://github.com/pomerium/pomerium/pull/1295
- https://github.com/pomerium/pomerium/pull/1313
- https://github.com/pomerium/pomerium/pull/1415

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey (approved):
LGTM

PR #1495, opened 2020-10-07
authenticate: remove unused paths, generate cipher at startup, remove qp store

desimone:
## Summary

Small improvements to authenticate that I noticed while making unrelated changes. 

- Generate shared cipher on startup, not per request.
- Use a render json helper.
- Remove unused API path declarations.
- Remove unused queryparam store. 

## Related issues

- No specific issues but the cipher could be a performance (entropy exhaustion?) under stress.

**Checklist**:

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/authenticate/handlers.go
+++ b/authenticate/handlers.go
@@ -3,7 +3,6 @@ package authenticate
 import (
 	"context"
 	"encoding/base64"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
@@ -66,8 +65,8 @@ func (a *Authenticate) Mount(r *mux.Router) {
 	v := r.PathPrefix("/.pomerium").Subrouter()
 	c := cors.New(cors.Options{
 		AllowOriginRequestFunc: func(r *http.Request, _ string) bool {
-			options := a.options.Load()
-			err := middleware.ValidateRequestURL(r, options.SharedKey)
+			state := a.state.Load()
+			err := middleware.ValidateRequestURL(r, string(state.sharedSecret))
 			if err != nil {
 				log.FromRequest(r).Info().Err(err).Msg("authenticate: origin blocked")
 			}
@@ -89,12 +88,6 @@ func (a *Authenticate) Mount(r *mux.Router) {
 	wk := r.PathPrefix("/.well-known/pomerium").Subrouter()
 	wk.Path("/jwks.json").Handler(httputil.HandlerFunc(a.jwks)).Methods(http.MethodGet)
 	wk.Path("/").Handler(httputil.HandlerFunc(a.wellKnown)).Methods(http.MethodGet)
-
-	// programmatic access api endpoint
-	api := r.PathPrefix("/api").Subrouter()
-	api.Use(func(h http.Handler) http.Handler {
-		return sessions.RetrieveSession(a.state.Load().sessionLoaders...)(h)
-	})
 }
 
 // Well-Known Uniform Resource Identifiers (URIs)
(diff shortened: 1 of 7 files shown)


authenticate/handlers.go
```diff
@@ -89,12 +89,6 @@ func (a *Authenticate) Mount(r *mux.Router) {
 	wk := r.PathPrefix("/.well-known/pomerium").Subrouter()
 	wk.Path("/jwks.json").Handler(httputil.HandlerFunc(a.jwks)).Methods(http.MethodGet)
 	wk.Path("/").Handler(httputil.HandlerFunc(a.wellKnown)).Methods(http.MethodGet)
-
-	// programmatic access api endpoint
-	api := r.PathPrefix("/api").Subrouter()
```
desimone, line 94: no longer used

authenticate/handlers.go
```diff
@@ -111,23 +105,18 @@ func (a *Authenticate) wellKnown(w http.ResponseWriter, r *http.Request) error {
...
-	return nil
+	return renderJSON(w, wellKnownURLS)
 }
 
 func (a *Authenticate) jwks(w http.ResponseWriter, r *http.Request) error {
 	state := a.state.Load()
+	return renderJSON(w, state.jwk)
+}
 
+func renderJSON(w http.ResponseWriter, v interface{}) error {
```
desimone, line 116: should probably pull into httputil.

authenticate/handlers.go
```diff
@@ -151,12 +140,12 @@ func (a *Authenticate) VerifySession(next http.Handler) http.Handler {
 			return a.reauthenticateOrFail(w, r, err)
 		}
 
-		if state.dataBrokerClient != nil {
```
desimone, line 154: I might be missing something here but it feels more clear to err on a nil databroker client, not continue?

authenticate/handlers.go
```diff
@@ -179,11 +168,6 @@ func (a *Authenticate) SignIn(w http.ResponseWriter, r *http.Request) error {
 	options := a.options.Load()
 	state := a.state.Load()
 
-	sharedCipher, err := cryptutil.NewAEADCipherFromBase64(options.SharedKey)
```
desimone, line 182: This was... probably very expensive to do on a per request basis and could ... maybe even exhaust the entropy source? IDK.

authenticate/state.go
```diff
@@ -79,12 +82,15 @@ func newAuthenticateStateFromConfig(cfg *config.Config) (*authenticateState, err
...
+	// shared cipher to encrypt data before passing data between services
+	state.sharedSecret, _ = base64.StdEncoding.DecodeString(cfg.Options.SharedKey)
+	state.sharedCipher, _ = cryptutil.NewAEADCipher(state.sharedSecret)
+
 	// private state encoder setup, used to encrypt oauth2 tokens
 	state.cookieSecret, _ = base64.StdEncoding.DecodeString(cfg.Options.CookieSecret)
 	state.cookieCipher, _ = cryptutil.NewAEADCipher(state.cookieSecret)
 	state.encryptedEncoder = ecjson.New(state.cookieCipher)
 
-	qpStore := queryparam.NewStore(state.encryptedEncoder, urlutil.QueryProgrammaticToken)
```
desimone, line 87: We don't support non-browser based auth flow anymore.

internal/httputil/handlers.go
```diff
@@ -27,6 +28,20 @@ func Redirect(w http.ResponseWriter, r *http.Request, url string, code int) {
...
+// RenderJSON replies to the request with the specified struct as JSON and HTTP code.
+// It does not otherwise end the request; the caller should ensure no further
+// writes are done to w.
+// The error message should be application/json.
+func RenderJSON(w http.ResponseWriter, code int, v interface{}) {
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+	w.WriteHeader(code)
+	if err := json.NewEncoder(w).Encode(v); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
```
calebdoxsey, line 40: we likely need to encode the JSON to a buffer in order to be able to set the header like this. Otherwise we will have already written the headers when the error occurs.

calebdoxsey (approved):

internal/httputil/handlers.go
desimone: Good catch (... and... shame on me because we definitely discussed this _exact_ issue last week !). 

I updated the code using a bytes buffer and added a test to check the header value. Let me know what you think.

calebdoxsey (approved):

PR #1503, opened 2020-10-09
databroker: require JWT for access

### LINKED ISSUE #1501: databroker: authenticate peers
Currently we don't authenticate peers for API calls to the databroker. We should add this.

We can use a JWT in gRPC metadata signed by the shared key.

calebdoxsey:
## Summary
This adds authentication to the databroker so that a JWT signed by the shared key is required to call methods.

## Related issues
- Fixes #1501 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (approved):
LGTM!

PR #1515, opened 2020-10-14
add policy to allow any authenticated user

pflipp:
## Summary

Provide a policy to allow requests to a upstream for any authenticated user.
while this option may be obvious for public providers (e.g Google) I think it is useful when you control the idP (own OIDC-Provider, self-hosted gitlab ...) and can trust all registered users.

**Checklist**:
- [ ] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone:
Hey @pflipp, thanks for the PR! Will review with the team and get back to your shortly!

pflipp:
hey @desimone, what do you think, is it a feature which could land in pomerium?
I really need this feature but if you think it doesn't fit we could close this PR and I would search for another way to solve my problem ;)

desimone:
Hey @pflipp  -- apologies for the delayed response. 

I think this is a reasonable ask and thank you for the implementation. One thing we've gone back and forth about internally is whether (and if it is possible?) to add all your users to an "everyone" group and have the desired functionality that way. I figure there's probably a reason for it, but wanted to better understand the use-case.

Thanks!

pflipp:
Hi @desimone, no problem and thank you for taking your time.
one use-case was to allow access to all registered users from a self hosted gitlab instance. here we already "solved" it with the "everyone" group. off course it blowed up the registration process as I have to remember to add every new user to this group. but it works ;)

my second use-case is that I use ory/hydra as oidc provider which does not support groups, as groups are not part of the oidc spec. since i opened this PR, https://github.com/pomerium/pomerium/pull/1542 got merged, so a solution could be to introduce an "everyone" claim.

while both approaches work it blows up the processes with use-case specific implementations which could be avoided with this policy.

chauffer:
There's definitely solid usecases for not having to maintain an 'everyone' group, and I hope this feature lands in Pomerium

desimone:
@pflipp Thanks for elaborating. I agree, I think this functionality makes sense. 

Do you mind rebasing your changes?

docs/reference/readme.md
```diff
@@ -1101,6 +1101,17 @@ These options correspond to the envoy route action host related options, which c
...
 If this setting is enabled, no whitelists (e.g. Allowed Users) should be provided in this route.
 
+### Allow Any Authenticated User
+
+- `yaml`/`json` setting: `allow_any_authenticated_user`
+- Type: `bool`
+- Optional
+- Default: `false`
+
+**Use with caution:** allow all requests to all users which have an account at the used provider
```
desimone, line 1111:
```suggestion
**Use with caution:** This setting will allow all requests for any user which is able to authenticate with our given identity provider. For instance, if you are using a corporate GSuite account, an unrelated gmail user will be able to access the underlying upstream. 

Use of this setting means Pomerium **will not enforce centralized authorization policy** for this route. The upstream is responsible for handling any authorization. 
```

desimone (changes requested):

pflipp:
@desimone nice :)

I rebased and squashed my commits.

desimone (approved):
LGTM, Thanks @pflipp

calebdoxsey (approved):

PR #1538, opened 2020-10-20
fix databroker requiring signed jwt

calebdoxsey:
## Summary
Although we were using an interceptor to enforce the shared key on the databroker it was only being attached to the internal gRPC server, not the control plane gRPC server. This change makes the call to require the signed JWT explicit in the databroker server implementation so that it works for either gRPC server.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -71,7 +71,7 @@ func New(cfg *config.Config) (*Cache, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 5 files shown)


travisgroth (approved):
LGTM though this seems valid:

https://codeclimate.com/github/pomerium/pomerium/pull/1538

Issue #1375, opened 2020-09-06
authorize: allow access based on arbitrary identity claims


desimone:
**Is your feature request related to a problem? Please describe.**

Today, we have a limited set of ways a user can define their access control policy with declarative `yaml`. Namely, oeprators can set access based on `allowed_users`, `allowed_groups`, and `allowed_users`.  

This feature would be paricularly useful for users :

- with [custom claims](https://openid.net/specs/openid-connect-core-1_0.html#AdditionalClaims) (for example, we've seen cases where users have alternative values for user id) 
- are using a OIDC bridge like Dex (where groups are pseudo-added to the [id_token](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims))

What's especially confusing  -- which, I am responsible for -- is that `allowed_users` corresponds to the user's `email` claim, not the `sub` claim which is meant to be a globally unique user identifier (RFC 7519). 

**Describe the solution you'd like**

We should allow operators to allow access to user's based on the equality of arbitrary claims. For example, a policy could look like...

```yaml
policy:
  - from: https://httpbin.localhost.pomerium.io
    to: http://httpbin.org
    allowed_domains:
      - pomerium.com
      - gmail.com
    allowed_idp_claims:
      sub:
        - 111432655977273150308
        - 111432655977273152308
      groups:
        - dev@pomerium.com
        - admin@pomerium.com
      custom_claim:
        - "anything in your id_token, really"
        - "https://openid.net/specs/openid-connect-core-1_0.html#TokenResponse"

```

To be clear, this would only before claims data that exists as part of the _original_ idp provided identity_token / jwt. This would _not_ be for writing policy against arbitrary datastore data. 

**Describe alternatives you've considered**

**Explain any additional use-cases**

**Additional context**

This would likely solve several issues including:

- https://github.com/pomerium/pomerium/issues/1112
- https://github.com/pomerium/pomerium/issues/834
- https://github.com/pomerium/pomerium/issues/1342 (For IdP claims only, not external auth sources or dynamic access policy)

desimone:
Closing. After a discussion with the team, I think it makes more sense to allow arbitrary adding and removal of claims in the attestation header directly. 

See: https://github.com/pomerium/pomerium/issues/1376#issuecomment-689054282

sly-net:
I think this is a first move towards more fine-grained authorization policies.

For my personal needs (as a Dex user), I'd like to be able to allow specific user_id to access backend applications.

```json
[...]

PR #1542, opened 2020-10-23
authorize: implement allowed_idp_claims

calebdoxsey:
## Summary
This PR implements a new type of authorization on arbitrary claims stored in the session or user objects in the databroker and derived from idp tokens. For example azure returns a claim for `family_name`, so this change would allow you to create a policy like:

```yaml
policy:
  - from: ...
    to: ...
    allowed_idp_claims:
      family_name:
        - Doxsey
```

A set of claims is defined as a map of strings to a list of values. The above policy would match a set of claims like:

```json
{
  "family_name":["Doxsey"],
  "given_name":["Caleb"] 
}
```

Nested maps are flattened. For example:

```json
{ "a": { "b": ["c"] } }
```

Becomes:

```json
{ "a.b": ["c"] }
```

In addition values are always a list, so 

```json
{ "a": "b" }
```

Becomes

```json
{ "a": ["b"] }
```

The way the policy is defined is that we test each map key and if any of the values in the policy matches any of the values in the claim, we allow access. This should match the behaviour of our other `allowed` rules.

## Related issues
- Implements #1375 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (approved):
LGTM but the test failures seem correct.

Issue #1559, opened 2020-10-29
Issue with URL Encoded strings in URI


davidcollom:
## What happened?
When visiting a URL with `%0A` in the URI we receive a 403 response from nginx-ingress, and no request make it to the backend service.

## What did you expect to happen?
A successful request to the backend service

## How'd it happen?

1. Visit https://web-test.example.com/random/%0A/path/ - Fail (403)
2. Visit https://web-test.example.com/?query=testing%0A1234 - Fail (403)
3. Visit https://web-test.example.com/random/ - Success (200)

## What's your environment like?

Running a deployed HELM chart in Kubernetes with the images pointing to master

## What's your config.yaml?

```config.yaml
config:
  rootDomain: example.com
  policy:
    - from: https://web-test.example.com
      to: https://web-test.example.com
      allowed_idp_claims:
        teams:
          - test-group

authenticate:
  idp:
    provider: oidc
    clientID: <>
    clientSecret: <>
    url: <>
    scopes:
    - openid
    - email
    - groups

forwardAuth:
  enabled: true

image:
  tag: master
  pullPolicy: Always
```

## What did you see in the logs?

```logs
pomerium-authorize: {"level":"info","service":"authorize","request-id":"<>","check-request-id":"<>","method":"GET","path":"","host":"forwardauth.example.com","query":"","allow":false,"status":403,"message":"forbidden","user":"<>","groups":[],"time":"<>","message":"authorize check"}
pomerium-proxy: {"level":"info","service":"envoy","upstream-cluster":"","method":"GET","authority":"forwardauth.example.com","path":"/verify?uri=https://web-test.example.com?query=testing%0A123","user-agent":"<>","referer":"","forwarded-for":"<>","request-id":"<>","duration":31.994513,"size":63492,"response-code":403,"response-code-details":"ext_authz_denied","time":"<>","message":"http-request"}
```

calebdoxsey:
@davidcollom it looks like we should be using `$escaped_request_uri` instead of `$request_uri`. Could you try swapping it out in your annotations:

```
nginx.ingress.kubernetes.io/auth-url: "https://forwardauth.localhost.pomerium.io/verify?uri=$scheme://$host$escaped_request_uri"
nginx.ingress.kubernetes.io/auth-signin: "https://forwardauth.localhost.pomerium.io/?uri=$scheme://$host$escaped_request_uri"
```

If that fixes it I will update our docs and examples.

davidcollom:
[...]

PR #1563, opened 2020-10-30
handle properly encoded nginx uris

calebdoxsey:
## Summary
When setting up forward-auth with the nginx ingress controller we recommend urls like:

```
nginx.ingress.kubernetes.io/auth-url: "https://forwardauth.localhost.pomerium.io/verify?uri=$scheme://$host$request_uri"
```

However this causes problems when the request uri has special characters. Nginx also expose an `escaped_request_uri`, but when this option is used various parts of our forward-auth flow stop working.

This PR includes several changes:

1. When replacing the path on the envoy CheckRequest we re-escape it because the Go path is already decoded.
2. When converting the CheckRequest into an HTTP url we now parse the relative path via `u.Parse` to handle encoded paths.
3. The control plane routes for `/?uri=` and `/verify?uri=` no longer require the session and redirect URI query parameters, because these are embedded in the single `uri` query parameter.
4. The fallback `/?uri=` route now ends up handling both `/verify?uri=` and `/?uri=` when the final session passes through, so I added code to detect this and handle it appropriately.

I left much of the existing code paths in place, so I believe this should be backwards compatible. We should also update the documentation to recommend `escaped_request_uri`.

## Related issues
- #1559 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/grpc.go
+++ b/authorize/grpc.go
@@ -46,12 +46,7 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v2.CheckRe
 	if isForwardAuth {
 		// update the incoming http request's uri to match the forwarded URI
 		fwdAuthURI := getForwardAuthURL(hreq)
-		in.Attributes.Request.Http.Scheme = fwdAuthURI.Scheme
-		in.Attributes.Request.Http.Host = fwdAuthURI.Host
-		in.Attributes.Request.Http.Path = fwdAuthURI.Path
-		if fwdAuthURI.RawQuery != "" {
-			in.Attributes.Request.Http.Path += "?" + fwdAuthURI.RawQuery
-		}
+		replaceCheckRequestURL(in, fwdAuthURI)
 	}
 
 	rawJWT, _ := loadRawSession(hreq, a.currentOptions.Load(), state.encoder)
@@ -283,13 +278,21 @@ func getCheckRequestURL(req *envoy_service_auth_v2.CheckRequest) *url.URL {
 		Host:   h.GetHost(),
 	}
 	u.Host = urlutil.GetDomainsForURL(u)[0]
-	// envoy sends the query string as part of the path
-	path := h.GetPath()
-	if idx := strings.Index(path, "?"); idx != -1 {
-		u.Path, u.RawQuery = path[:idx], path[idx+1:]
+
+	// first try using the parse method which properly handles encoding and sets Raw+Path/Query and the regular
+	// Path/Query.
+	if nu, err := u.Parse(h.GetPath()); err == nil {
+		u = nu
 	} else {
-		u.Path = path
+		// envoy sends the query string as part of the path
+		path := h.GetPath()
+		if idx := strings.Index(path, "?"); idx != -1 {
+			u.Path, u.RawQuery = path[:idx], path[idx+1:]
+		} else {
+			u.Path = path
+		}
 	}
+
 	return u
 }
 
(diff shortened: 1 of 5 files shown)


MattiasGees:
We tested this, but it removed all authentication from the endpoint and people were able to access it as it was just open.

internal/controlplane/xds_routes.go
```diff
@@ -72,8 +72,7 @@ func buildPomeriumHTTPRoutes(options *config.Options, domain string) []*envoy_co
 	if config.IsProxy(options.Services) && options.ForwardAuthURL != nil && hostMatchesDomain(options.GetForwardAuthURL(), domain) {
 		routes = append(routes,
 			// disable ext_authz and pass request to proxy handlers that enable authN flow
-			buildControlPlanePathAndQueryRoute("/verify", []string{urlutil.QueryForwardAuthURI, urlutil.QuerySessionEncrypted, urlutil.QueryRedirectURI}),
-			buildControlPlanePathAndQueryRoute("/", []string{urlutil.QueryForwardAuthURI, urlutil.QuerySessionEncrypted, urlutil.QueryRedirectURI}),
+			buildControlPlanePathAndQueryRoute("/verify", []string{urlutil.QueryForwardAuthURI}),
```
desimone, line 75: I _think_ this is fine in practice but I am worried it will break things.

authorize/grpc.go
```diff
@@ -283,13 +278,21 @@ func getCheckRequestURL(req *envoy_service_auth_v2.CheckRequest) *url.URL {
...
 	}
 	u.Host = urlutil.GetDomainsForURL(u)[0]
-	// envoy sends the query string as part of the path
-	path := h.GetPath()
-	if idx := strings.Index(path, "?"); idx != -1 {
-		u.Path, u.RawQuery = path[:idx], path[idx+1:]
+
+	// first try using the parse method which properly handles encoding and sets Raw+Path/Query and the regular
+	// Path/Query.
+	if nu, err := u.Parse(h.GetPath()); err == nil {
```
desimone, line 284: Another option...?

We could get the path from the ` X-Original-URL` heaader which -- at least for nginx -- contains the "properly" encoded original URL?

```conf
			proxy_set_header            X-Original-URL          $scheme://$http_host$request_uri;

```

desimone: What do you think? This would require users setting this in their nginx config, but this is also the default for nginx-ingress.

proxy/forward_auth.go
```diff
@@ -129,6 +133,23 @@ func (p *Proxy) startAuthN(w http.ResponseWriter, r *http.Request) error {
 		uri.Path = xfu
 	}
 
+	rawRedirectURI := uri.Query().Get(urlutil.QueryRedirectURI)
```
desimone, line 136: Oh.. this is very subtle and slightly confusing because we match on query params values in both the go, and envoy code (but I get why we can't because of the encoding).... maybe add a comment as to why this code block would be hit?

desimone: I wish there was a better way to do this... hmm...

authorize/grpc.go
```diff
@@ -331,3 +334,13 @@ func logAuthorizeCheck(
 
 	evt.Msg("authorize check")
 }
+
+func replaceCheckRequestURL(req *envoy_service_auth_v2.CheckRequest, newURL *url.URL) {
+	req.Attributes.Request.Http.Scheme = newURL.Scheme
+	req.Attributes.Request.Http.Host = newURL.Host
+	// the go path is decoded, but the check request path is encoded, so we need to re-encode it
+	req.Attributes.Request.Http.Path = newURL.EscapedPath()
```
desimone, line 342: Maybe we can instead use `X-Original-URL` per other comment if set to grab the unencoded path?

desimone (commented):
Made some comments. Can we discuss in a zoom?

desimone:
Thanks for testing @MattiasGees .... definitely not the desired effect ! Stay tuned.

PR #1582, opened 2020-11-07
Fix command in Kubernetes Quick start docs

wesleyw72:
Fixed the 'cd' command which currently tries to change to a directory which doesn't exist.


desimone (approved):
Thanks @wesleyw72 !

PR #1586, opened 2020-11-09
authenticate: oidc frontchannel-logout endpoint

### LINKED ISSUE #1581: authenticate: endpoint for Single Sign Out with oidc frontchannel-logout

Logging out on the idP from a different client does not destroy the session on Pomerium.

A endpoint like `/oauth2/sessions/frontchannel-logout` which just terminates the session (if it exists).

Many idPs let you register a endpoint to terminate the session within an iframe during the logout procedure.

OIDC describes the procedure for backchannel logout as the most reliable way for Single Sign Out (in some cases the browser might not transfer the session cookies within the iframe).
In addition it introduces much more work as it requires a way to locate the pomerium session-id from the idP session-id.

**Additional context**

https://openid.net/specs/openid-connect-frontchannel-1_0.html
https://openid.net/specs/openid-connect-backchannel-1_0.html

pflipp:
## Summary
implementation of an endpoint to provide oidc frontchannel-logout

## Related issues
Fixes #1581

**Checklist**:
- [x] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authenticate/handlers.go
+++ b/authenticate/handlers.go
@@ -101,9 +101,9 @@ func (a *Authenticate) wellKnown(w http.ResponseWriter, r *http.Request) error {
 	wellKnownURLS := struct {
 		// URL string referencing the client's JSON Web Key (JWK) Set
 		// RFC7517 document, which contains the client's public keys.
-		JSONWebKeySetURL string `json:"jwks_uri"`
-		OAuth2Callback   string `json:"authentication_callback_endpoint"`
-		FrontchannelLogoutURI   string `json:"frontchannel_logout_uri"`
+		JSONWebKeySetURL      string `json:"jwks_uri"`
+		OAuth2Callback        string `json:"authentication_callback_endpoint"`
+		FrontchannelLogoutURI string `json:"frontchannel_logout_uri"`
 	}{
 		state.redirectURL.ResolveReference(&url.URL{Path: "/.well-known/pomerium/jwks.json"}).String(),
 		state.redirectURL.ResolveReference(&url.URL{Path: "/oauth2/callback"}).String(),
@@ -246,7 +246,7 @@ func (a *Authenticate) SignOut(w http.ResponseWriter, r *http.Request) error {
 	ctx, span := trace.StartSpan(r.Context(), "authenticate.SignOut")
 	defer span.End()
 
-	rawIDToken := a.processSignOutWithContext(ctx, w, r)
+	rawIDToken := a.revokeSession(ctx, w, r)
 
 	redirectString := ""
 	if sru := a.options.Load().SignOutRedirectURL; sru != nil {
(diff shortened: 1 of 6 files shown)


jerana (approved):

internal/controlplane/xds_routes.go
```diff
@@ -66,7 +66,7 @@ func buildPomeriumHTTPRoutes(options *config.Options, domain string) []*envoy_co
 	}
 	// if we're handling authentication, add the oauth2 callback url
 	if config.IsAuthenticate(options.Services) && hostMatchesDomain(options.GetAuthenticateURL(), domain) {
-		routes = append(routes, buildControlPlanePathRoute(options.AuthenticateCallbackPath, false))
+		routes = append(routes, buildControlPlanePathRoute(options.AuthenticateCallbackPath, false), buildControlPlanePathRoute("/oauth2/session/frontchannel-logout", false))
```
desimone, line 69: Coudl you add this to the next line to be consistent with the style?

authenticate/handlers.go
```diff
@@ -603,3 +608,67 @@ func (a *Authenticate) saveSessionToDataBroker(
 
 	return nil
 }
+
+// FrontchannelLogout signs the user out
```
desimone, line 612: Could you add a bit more context in the docs here and reference the OIDC spec?

authenticate/handlers.go
```diff
@@ -603,3 +608,67 @@ func (a *Authenticate) saveSessionToDataBroker(
...
+	w.Header().Set("Cache-Control", "no-cache, no-store")
+	w.Header().Set("Pragma", "no-cache")
+
+	if err := a.processFrontchannelLogoutWithContext(ctx, w, r); err != nil {
+		log.FromRequest(r).Debug().Err(err).Msg("failed processing frontchannel logout")
+
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintf(w, "%s", "mising required parameters")
+
+		return nil
```
desimone, line 627: We actually wrap the standard HTTP request handler to return an error page / response. You should be able to return the error here directly with any wrapped message.

authenticate/handlers.go
```diff
@@ -603,3 +608,67 @@ func (a *Authenticate) saveSessionToDataBroker(
...
+	}
+
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprintf(w, "%s", "OK")
+
+	return nil
+}
+
+func (a *Authenticate) processFrontchannelLogoutWithContext(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
```
desimone, line 636: This function does a very similar (identical?) process as the logout function. Cand we abstract and resuse that code as part of a subfunction?

https://github.com/pflipp/pomerium/blob/d0cfab7020816098a73426d3234a22cdb182634a/authenticate/handlers.go#L247-L263

authenticate/handlers.go
```diff
@@ -61,6 +61,11 @@ func (a *Authenticate) Mount(r *mux.Router) {
 	r.Path("/robots.txt").HandlerFunc(a.RobotsTxt).Methods(http.MethodGet)
 	// Identity Provider (IdP) endpoints
 	r.Path("/oauth2/callback").Handler(httputil.HandlerFunc(a.OAuthCallback)).Methods(http.MethodGet)
+	s := r.PathPrefix("/oauth2/session").Subrouter()
+	s.Use(func(h http.Handler) http.Handler {
+		return sessions.RetrieveSession(a.state.Load().sessionLoaders...)(h)
+	})
+	s.Path("/frontchannel-logout").Handler(httputil.HandlerFunc(a.FrontchannelLogout)).Methods(http.MethodGet)
```
desimone, line 68: Does the URL have to be under oauth2/session? From the spec, it looks to be unspecified. If so, could we put this under `.pomerium` ?

desimone: Also, it probably makes sense to add this endpoint to the wellKnown response handler? 

```
  "frontchannel_logout_session_supported": true
```

And perhaps a https://ldapwiki.com/wiki/Frontchannel_logout_uri

desimone:
@pflipp -- Sorry this has taken me so long to review. Thanks again for the PR.  Left some comments and had some follow up questions that I couldn't figure out in reviewing the spec on my own.

authenticate/handlers.go
pflipp: since nobody is interested in the result of the call, I removed the BadRequest response. Showing the errors in the logs should be enough.

internal/controlplane/xds_routes.go
pflipp: removed this line as the route is now under `.pomerium`

authenticate/handlers.go
pflipp: I added `frontchannel_logout_uri` to the wellknown handler as I think this may help someone to find out the correct url to register. But I don't understand how `"frontchannel_logout_session_supported": true` would help anybody?
  
Maybe we should document this feature in the Docs/Guides, could you give me a hint where it fits?

desimone: > added frontchannel_logout_uri to the wellknown handler as I think this may help someone to find out the correct url to register. 

Agree. 

> But I don't understand how "frontchannel_logout_session_supported": true would help anybody?

I'm with you :) 

> Maybe we should document this feature in the Docs/Guides, could you give me a hint where it fits?

That would be great. I feel like this would fit well in `docs/docs/topics` and maybe have an article called "logging a user out" or "end a user's session" or something like that. 

Thoughts?

desimone (commented):
Thanks again for the PR. This look just about read to go. Requested a few minor changes to include context into the warning logs + I think short doc piece on how to use this functionality would be great.

authenticate/handlers.go
pflipp: > Thoughts?

I named the topic "Single Sign-out" as I think it describes it better than "logging out" which already is handled by the sign-out route

desimone (approved):
LGTM. Thanks @pflipp  -- I reworded some of the docs, functions, and ran go-fmt on the files in a separate commit. Thank you for the contribution!

PR #1624, opened 2020-11-25
replace GetAllPages with InitialSync, improve merge performance

### LINKED ISSUE #1619: Improve speed of authorize service startup for large directories

Currently we load directory data in small batches to prevent message size issues, but in very large directories the current implementation can lead to excessively long startup times.  

Ideas discussed offline include paging the redis query retrieval or switching to a pure streaming model.

Just an increased batch size, but this is a brittle solution.

n/a

**Additional context**

n/a

calebdoxsey:
## Summary
This replaces the `GetAllPages` method with a call to `Sync` and a new option to turn off waiting for changes. This should improve performance since we don't have to repeat querying all the records for each page call, while still preventing the message too large error. It also simplifies the code and in the future we can remove the `GetAll` method entirely.

I also updated the manager sync code so that it will update records 10 at a time. This helps with the initial directory sync into the databroker.

## Related issues
Fixes #1619 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


authorize/run.go
```diff
@@ -105,10 +105,10 @@ func (a *Authorize) runDataTypeSyncer(ctx context.Context, typeURL string) error
 		}
 
 		serverVersion = res.GetServerVersion()
-		recordVersion = res.GetRecordVersion()
 
 		for _, record := range res.GetRecords() {
 			a.updateRecord(record)
+			recordVersion = record.GetVersion()
```
desimone, line 111: nit: might be helpful to have declaration of `recordVersion` closer to its initialization

internal/encoding/base64.go
```diff
@@ -0,0 +1,23 @@
...
+	"encoding/json"
+	"strings"
+)
+
+// DecodeBase64OrJSON decodes a JSON string that can optionally be base64 encoded.
+func DecodeBase64OrJSON(in string, out interface{}) error {
+	in = strings.TrimSpace(in)
+
+	// the service account can be base64 encoded
+	if !strings.HasPrefix(in, "{") {
```
desimone, line 14: I believe JSON can also be prefixed with `[`.

desimone: https://stackoverflow.com/questions/5034444/can-json-start-with#:~:text=From%20what%20I%20can%20read,an%20array%20element%20in%20JSON.

desimone: Alternatively... https://golang.org/pkg/encoding/json/#Valid

internal/encoding/base64.go
```diff
@@ -0,0 +1,23 @@
+package encoding
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+)
+
+// DecodeBase64OrJSON decodes a JSON string that can optionally be base64 encoded.
+func DecodeBase64OrJSON(in string, out interface{}) error {
```
desimone, line 10: Might be worth adding a few tests to this?

pkg/grpc/databroker/databroker.go
```diff
@@ -34,27 +37,32 @@ func ApplyOffsetAndLimit(all []*Record, offset, limit int) (records []*Record, t
 	return records, len(all)
 }
 
-// GetAllPages calls GetAll for all pages of data.
-func GetAllPages(ctx context.Context, client DataBrokerServiceClient, in *GetAllRequest) (*GetAllResponse, error) {
-	var res GetAllResponse
-	var pageToken string
+// InitialSync performs a sync with no_wait set to true and then returns all the results.
+func InitialSync(ctx context.Context, client DataBrokerServiceClient, in *SyncRequest) (*SyncResponse, error) {
```
desimone, line 41: Could we replcae the existing test for `GetAllPages` with the new `InitialSync` versino.

internal/identity/manager/manager.go
```diff
@@ -61,6 +66,8 @@ type Manager struct {
 	directoryGroupsRecordVersion string
 
 	directoryNextRefresh time.Time
+
+	dataBrokerSemaphore *semaphore.Weighted
```
desimone, line 70: Interesting use of semaphores. Cool.

desimone (approved):
LGTM! Really clever use of semaphores. Added a few questions around tests, and a few minor suggestions.

authorize/run.go
calebdoxsey: Not sure I follow. Where should I put the declaration?

PR #1636, opened 2020-11-30
wait for initial sync to complete before starting control plane

### LINKED ISSUE #1620: authorize: defer listener startup until directory is loaded

When loading directory data from the databroker service at startup, authorize will return 403 errors for group based roles.  In distributed deployments, this can lead to rolling restarts causing perceived outages by end users.

Ideally, we would not start listening/serving until the initial directory load is complete.  This would provide a signal to external orchestration systems for when the service is truly ready and can help pace rollouts.  It would also prevent a proxy from connecting to an uninitialized authorize service.

There may be a more in-band way to do this with gRPC but I think most folks use a TCP port knock for gRPC health/ready checks.

n/a

**Additional context**

In practices this is only noticeable when there is a large directory, though the behavior is theoretically possible at any size with group based policies.

calebdoxsey:
## Summary
This updates the authorize service so that we can wait until the initial sync is complete before starting the listener in the control plane. Assuming there's a health check on the service, and there's more than one authorize service, this should provide a more consistent experience when restarting the service.

## Related issues
Fixes #1620 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/authorize.go
+++ b/authorize/authorize.go
@@ -27,6 +27,8 @@ type Authorize struct {
 
 	dataBrokerDataLock sync.RWMutex
 	dataBrokerData     evaluator.DataBrokerData
+
+	dataBrokerInitialSync map[string]chan struct{}
 }
 
 // New validates and creates a new Authorize service from a set of config options.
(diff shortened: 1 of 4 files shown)


desimone (approved):
LGTM

PR #1666, opened 2020-12-09
improve ca cert error message, use GetCertPool for databroker storage

### LINKED ISSUE #1606: Support Certificate Chaining for Redis Connections

Many users use managed redis instances in the cloud, which manage their own certificates using a set of intermediate certificates and rotate them often. Currently it looks like Pomerium will only recognize certificate chains of length 1, so whenever the provider rotates their intermediate certificate, Pomerium has to be reconfigured with the new certificate as `databroker_storage_ca_file`.

Since providers often use trusted roots that rarely change, either verify TLS directly from system trust store, or allow the root certificate to be pinned as `databroker_storage_ca_file`. In any case this would require Pomerium to support certificate chaining to validate redis connections.

I'm currently pinning the intermediate CA directly, but this is not ideal as there are several of the CAs and the CAs are rotated regularly.

This is mostly applicable for users of managed cloud redis instances. I encountered this issue on Azure, but most cloud providers support managed redis.

**Additional context**

calebdoxsey:
## Summary
We have a `GetCertPool` method in `cryptuil` which we weren't using in the databroker. This method has better error checking, so I switched the databroker to use it. I also changed some of the error messages to make it clear that the data needs to be PEM-encoded.

## Related issues
Fixes #1606 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/cmd/pomerium-cli/kubernetes.go
+++ b/cmd/pomerium-cli/kubernetes.go
@@ -3,8 +3,6 @@ package main
 import (
 	"context"
 	"crypto/tls"
-	"crypto/x509"
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


desimone (approved):
LGTM!

PR #1696, opened 2020-12-16
TCP client command

calebdoxsey:
## Summary
This PR adds support for a new `pomerium-cli` command to establish a TCP tunnel via HTTP Connect. Changes:

1. New `tcp` sub-command
2. Using `base36` for file names in a cache directory (was afraid hex might have filenames that were too long, but also wanted to avoid case sensitivity)
3. Add `authclient` package that makes a generic version of our kubernetes browser login flow. A subsequent PR will update the kubernetes command.
4. Add a `cliutil` package that has a cache for storing JWTs.
5. Add a `tcptunnel` package that contains the TCP tunneling code.

The TCP tunnel can either start a local TCP listener, or be called with a Reader+Writer. (the latter is used for stdin/stdout support) When a connection is established an HTTP CONNECT request is made to the server with an authorization JWT if one is available.

If a `200` is returned, then we proxy traffic over the network connection directly.
If a `30x` is returned, and there is no JWT, we run the browser login flow to get one.

The reason that we attempt a connection even if we don't have a JWT is two-fold:

1. We will be running this against a Pomerium proxy server, not the authentication server, so we need the redirect response to know the authenticate server URL.
2. If you enable public access, there's no need for a JWT.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/authclient/authclient.go
+++ b/internal/authclient/authclient.go
@@ -1,0 +1,135 @@
+// Package authclient contains an CLI authentication client for Pomerium.
+package authclient
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/skratchdot/open-golang/open"
+	"golang.org/x/sync/errgroup"
+)
+
+var openBrowser = open.Run
+
+// An AuthClient retrieves an authentication JWT via the Pomerium login API.
+type AuthClient struct {
+	cfg *config
+}
+
+// New creates a new AuthClient.
+func New(options ...Option) *AuthClient {
+	return &AuthClient{
+		cfg: getConfig(options...),
+	}
+}
+
+// GetJWT retrieves a JWT from Pomerium.
+func (client *AuthClient) GetJWT(ctx context.Context, serverURL *url.URL) (rawJWT string, err error) {
+	li, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return "", fmt.Errorf("failed to start listener: %w", err)
+	}
+	defer func() { _ = li.Close() }()
+
+	incomingJWT := make(chan string)
+	eg, ctx := errgroup.WithContext(ctx)
+	eg.Go(func() error {
+		return client.runHTTPServer(ctx, li, incomingJWT)
+	})
+	eg.Go(func() error {
+		return client.runOpenBrowser(ctx, li, serverURL)
+	})
+	eg.Go(func() error {
+		select {
+		case rawJWT = <-incomingJWT:
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+		return nil
+	})
+	err = eg.Wait()
+	if err != nil {
+		return "", err
+	}
+
+	return rawJWT, nil
+}
+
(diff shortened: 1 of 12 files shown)


internal/authclient/authclient.go
```diff
@@ -0,0 +1,135 @@
+// Package authclient contains an CLI authentication client for Pomerium.
```
desimone, line 1:
```suggestion
// Package authclient contains a CLI authentication client for Pomerium.
```

internal/cliutil/jwtcache.go
```diff
@@ -0,0 +1,143 @@
...
+	err := ioutil.WriteFile(path, []byte(rawJWT), 0600)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (cache *LocalJWTCache) hash(str string) string {
+	h := blake2s.Sum256([]byte(str))
```
desimone, line 81: Curious why not use SHA-256 via cryputil?

internal/cliutil/jwtcache.go
```diff
@@ -0,0 +1,143 @@
...
+	return base36.EncodeBytes(h[:])
```
desimone, line 82: Human-friendly-ness I assume?

internal/tcptunnel/tcptunnel.go
```diff
@@ -0,0 +1,211 @@
...
+		}()
+	}
+}
+
+// Run establishes a TCP tunnel via HTTP Connect and forwards all traffic from/to local.
+func (tun *Tunnel) Run(ctx context.Context, local io.ReadWriter) error {
+	rawJWT, err := tun.cfg.jwtCache.LoadJWT(tun.jwtCacheKey())
+	switch {
+	case err == nil:
+	case errors.Is(err, cliutil.ErrExpired):
```
desimone, line 93: Can you explain this switch statement?

desimone: I read this as "If there is no error, or is one I expect, continue with tun.Run"

desimone (commented):
LGTM, a few comments. Would probably prefer to re-use the same crypto primitives where possible.

internal/cliutil/jwtcache.go
desimone: > Using base36 for file names in a cache directory (was afraid hex might have filenames that were too long, but also wanted to avoid case sensitivity)


RTFM :)

internal/cliutil/jwtcache.go
desimone: Oh, the hashes are smaller. Got it. Yeah. Makes sense.

> BLAKE2s is optimized for 8- to 32-bit platforms and produces digests of any size between 1 and 32 bytes

calebdoxsey: I'll swap it. I think it's the same size.

Blake2s comes from ChaCha20 and is faster, but in this case it wouldn't matter.

desimone (approved):
Great stuff. LGTM 

\cc @travisgroth for +2

travisgroth (approved):
LGTM

PR #1700, opened 2020-12-17
unimpersonate button

calebdoxsey:
## Summary
This updates the dashboard so that when impersonating a user there's now an unimpersonate button.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/frontend/assets/html/dashboard.go.html
```diff
@@ -225,7 +227,11 @@ <h2>Sign-in-as</h2>
                 class="button full"
                 type="submit"
               >
-                Impersonate
+                {{if .State.ImpersonateEmail}}
+                  Unimpersonate
```
desimone, line 231: Does this.... work? Do any backend changes need to be made to clear the set status?

calebdoxsey: Yes it seems to work. It sets the `ImpersonateEmail` in the state to the empty string, which is the default unimpersonated state.

desimone (approved):

PR #1722, opened 2020-12-29
k8s cmd: use authclient package

calebdoxsey:
## Summary
As part of the TCP tunnel command we added a new `authclient` package. This PR refactors the existing `k8s` command to use that package instead of its own implementation.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (approved):
LGTM!

PR #1727, opened 2020-12-29
tcptunnel: handle invalid http response codes

calebdoxsey:
## Summary
This PR updates the TCP tunnel code to handle non-200 HTTP response codes. We will retry-authentication once for any redirect codes, but otherwise we will delete the saved JWT to make sure we can restart the login process.

I also noticed that we were attempting to do programmatic login against the authenticate service URL, but AFAIK the programmatic login is only available on the proxy URL, so I updated the URL that's used in this case.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/tcptunnel/tcptunnel.go
+++ b/internal/tcptunnel/tcptunnel.go
@@ -98,10 +98,10 @@ func (tun *Tunnel) Run(ctx context.Context, local io.ReadWriter) error {
 	default:
 		return fmt.Errorf("tcptunnel: failed to load JWT: %w", err)
 	}
-	return tun.run(ctx, local, rawJWT)
+	return tun.run(ctx, local, rawJWT, 0)
 }
 
-func (tun *Tunnel) run(ctx context.Context, local io.ReadWriter, rawJWT string) error {
+func (tun *Tunnel) run(ctx context.Context, local io.ReadWriter, rawJWT string, retryCount int) error {
 	log.Info().
 		Str("dst", tun.cfg.dstHost).
 		Str("proxy", tun.cfg.proxyHost).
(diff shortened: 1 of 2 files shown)


internal/tcptunnel/tcptunnel.go
```diff
@@ -160,15 +160,18 @@ func (tun *Tunnel) run(ctx context.Context, local io.ReadWriter, rawJWT string)
 		http.StatusFound,
 		http.StatusTemporaryRedirect,
 		http.StatusPermanentRedirect:
-		if rawJWT == "" {
+		if retryCount == 0 {
 			_ = remote.Close()
 
-			authURL, err := url.Parse(res.Header.Get("Location"))
```
calebdoxsey, line 166: If my understanding is correct, we can't use the returned redirect URL because it won't have the programmatic login endpoint.

An alternative solution would be to add support for programmatic login to the authenticate service but I'm not sure there's any advantage to that.

desimone: > If my understanding is correct, we can't use the returned redirect URL because it won't have the programmatic login endpoint.

Yes.

> An alternative solution would be to add support for programmatic login to the authenticate service but I'm not sure there's any advantage to that.

For now this makes sense. I think the way we are currently handling programmatic auth flow has some legacy logic where the audience claim is tied to the proxy's domain (which is set from the proxy handler endpoint). This made more sense when we had client-side maintained state, but makes less sense now where I think we don't really care about the upstream audience.

desimone (approved):

Issue #1670, opened 2020-12-10
proxy: Memory leak


davidcollom:
## What happened?
We've noticed an increase in memory utilisation when we update pomerium's config.yaml without restarting it.

## What did you expect to happen?
Memory profile stays relatively 

## How'd it happen?

1. Ran `pomerium` in proxy mode only
2. edit config.yaml
2.1. Add/Remove Policies
3. See memory increase
4. Restart proxy
5. See memory return to a much lower level

This was with around 62 policies defined:
![Screenshot 2020-12-10 at 6 55 43 pm](https://user-images.githubusercontent.com/1504448/101818312-92eb9300-3b1b-11eb-8ca5-a5f0fe3f1aaa.png)
@ ~18:50 we restarted the proxy pod.

We have seen this hit over 6GB of memory before being OOM'd due to pod limits being hit, on restart utilisation is around 250-270MB

## What's your environment like?

- Pomerium version (~v0.11.0-rc1~ v0.11.0):
- GKE: v1.17.12-gke.1504

## What's your config.yaml?

```config.yaml
address: :443
authenticate_service_url: https://XXXXXXXXXXXXXXX
authorize_service_url: https://pomerium-authorize.pomerium.svc.cluster.local
cache_service_url: https://pomerium-cache.pomerium.svc.cluster.local
certificate_authority_file: /pomerium/ca.pem
certificate_file: /pomerium/cert.pem
certificate_key_file: /pomerium/privkey.pem
databroker_storage_type: memory
forward_auth_url: https://XXXXXXXXXXXXXXX
grpc_address: :443
idp_client_id: askljabnskjgbasjfbsajhkfbasjkbvaksjbvjksabnvkjas
idp_provider: oidc
idp_provider_url: https://XXXXXXXXXXXXXXX
idp_scopes:
- openid email groups profile
policy:
- allowed_idp_claims:
    perms:
    - XXXXXXXXXXXXXXX
    - XXXXXXXXXXXXXXX
  from: XXXXXXXXXXXXXXX
  to: XXXXXXXXXXXXXXX
- allowed_idp_claims:
    perms:
    - XXXXXXXXXXXXXXX
    - XXXXXXXXXXXXXXX
  from: XXXXXXXXXXXXXXX
  to: XXXXXXXXXXXXXXX
[..SAME SECTIONS REPEATED 64 TIMES...]
```

travisgroth:
Hi @davidcollom,

[...]

PR #1732, opened 2021-01-04
use incremental API for envoy xDS

calebdoxsey:
## Summary
This PR changes the envoy code to use the incremental API instead of the SotW API. This helps reduce memory usage by only sending new clusters/listeners when they change.

Other changes:
1. Pulled out the incremental delta sync code to its own package.
2. Added a `marshalAny` local function to enable deterministic marshaling
3. Fixed a source of non-determinism in `toEnvoyHeaders` by sorting the map keys before looping over them
4. Added exponential try to the identity manager to handle the case when data isn't available yet (I think it would just crash before, but it was uncommon. It now happens more often because of the nature of incremental xDS updates)
5. Added a `HashProto` method for computing the hash of a protobuf message. This uses the deterministic marshaling option to ensure map order doesn't change the result of the hash. This hash is used as the version number for clusters and listeners.

## Related issues
#1670 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/identity/manager/manager.go
```diff
@@ -775,3 +779,21 @@ func isTemporaryError(err error) bool {
 	}
 	return false
 }
+
+func exponentialTry(ctx context.Context, f func() error) error {
```
desimone, line 783: Maybe indicate this goes forever?

pkg/cryptutil/hash.go
```diff
@@ -28,3 +29,14 @@ func HashPassword(password []byte) ([]byte, error) {
 func CheckPasswordHash(hash, password []byte) error {
 	return bcrypt.CompareHashAndPassword(hash, password)
 }
+
+// HashProto hashes a protobuf message. It sets `Deterministic` to true to ensure
+// the encoded message is always the same. (ie map order is lexographic)
+func HashProto(msg proto.Message) []byte {
```
desimone, line 35: Could we add tests to this func?

desimone (approved):
LGTM, a few minor requests at your discretion.

internal/identity/manager/manager.go
calebdoxsey: it does not go forever. It goes until the function succeeds or context is Done.

pkg/cryptutil/hash.go
calebdoxsey: I added a test but it will rarely succeed in the non deterministic case too.

internal/controlplane/xds.go
```diff
@@ -19,51 +20,48 @@ import (
...
+	listenerTypeURL = "type.googleapis.com/envoy.config.listener.v3.Listener"
+)
+
+func (srv *Server) buildDiscoveryResources() map[string][]*envoy_service_discovery_v3.Resource {
+	resources := map[string][]*envoy_service_discovery_v3.Resource{}
+	cfg := srv.currentConfig.Load()
+	for _, cluster := range srv.buildClusters(&cfg.Options) {
+		any, err := anypb.New(cluster)
+		if err != nil {
+			// anypb will only panic due to developer error.
```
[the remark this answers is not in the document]
calebdoxsey: My point was it was safe to ignore the error because it doesn't happen at runtime. When it errors it means the function was used improperly by the developer. (a condition where I view it as acceptable to panic)

I removed the panic.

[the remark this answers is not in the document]
desimone: I understood, I was just joking :)

PR #1744, opened 2021-01-06
tcp: prevent idle stream timeouts for TCP and Websocket routes

### LINKED ISSUE #1740: TCP connections disconnected after ~5 minutes
## What happened?
When using SSH if the connection is idle for ~ 5 minutes it will be forcibly disconnected.

Not be disconnected.

## How'd it happen?
```time ssh ssh.int.nas.bdd.io
3:28PM INF tcptunnel: opening connection dst=ssh.int.nas.bdd.io:22 proxy=ssh.int.nas.bdd.io:443 secure=true
3:28PM INF tcptunnel: connection established
bdd@nas:~$ 3:33PM INF tcptunnel: connection closed
                                                  client_loop: send disconnect: Broken pipe
ssh ssh.int.nas.bdd.io  0.20s user 0.14s system 0% cpu 5:01.92 total```


## Additional context
Likely cause is a default envoy setting that needs to be changed for TCP connections (and probably websockets too).

calebdoxsey:
## Summary
Envoy has an idle stream timeout which defaults to 5 minutes if not set. This change sets it to 0 (thus disabling idle timeouts) for TCP or websocket routes.

The upstream timeout is an independent setting and still available via config options.

Before:

before

After: (was open indefinitely, I manually closed it)

after

## Related issues
Fixes #1740 

**Checklist**:
- [x] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/controlplane/xds_routes.go
+++ b/internal/controlplane/xds_routes.go
@@ -192,6 +192,7 @@ func buildPolicyRoutes(options *config.Options, domain string) []*envoy_config_r
 		requestHeadersToAdd := toEnvoyHeaders(policy.SetRequestHeaders)
 		requestHeadersToRemove := getRequestHeadersToRemove(options, &policy)
 		routeTimeout := getRouteTimeout(options, &policy)
+		idleTimeout := getRouteIdleTimeout(&policy)
 		prefixRewrite, regexRewrite := getRewriteOptions(&policy)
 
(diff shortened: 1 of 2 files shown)


desimone (approved):
LGTM!

PR #1768, opened 2021-01-12
remove user impersonation and service account cli

calebdoxsey:
## Summary
This PR removes user impersonation.

**Checklist**:
- [ ] add related issues
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review

PR #1765, opened 2021-01-11
authorize: move impersonation into session/service account

calebdoxsey:
## Summary
This PR moves impersonation out of the session state into the databroker session / service account.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

PR #1774, opened 2021-01-13
authenticate: update user info screens

desimone:
## Summary

These changes update the look and feel of the error, and user info page (previously called the dashboard).

Screen Shot 2021-01-13 at 12 36 18 PM
.

### Details

```
authenticate: update user info screens

- rename "dashboard" to userinfo to avoid confusion
- don't leak version from error page.
- fix typo in state.go
- make statik deterministic on modtime

```
## Related issues

- #1768 
- #1765 

**Checklist**:
- [ ] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


internal/frontend/templates.go
```diff
@@ -75,6 +78,14 @@ func NewTemplates() (*template.Template, error) {
 		"formatTime": func(tm time.Time) string {
 			return tm.Format("2006-01-02 15:04:05 MST")
 		},
+		"toJSON": func(v interface{}) string {
+			output, _ := json.Marshal(v)
+			return string(output)
+		},
+		"formatPbTime": func(ts *timestamppb.Timestamp) string {
+			t, _ := ptypes.Timestamp(ts)
```
calebdoxsey, line 86: I think you can use `ts.AsTime()`

calebdoxsey (approved):

internal/frontend/templates.go
desimone: Indeed, I don't think I'm using either new func.

PR #1775, opened 2021-01-13
config: detect underlying file changes

calebdoxsey:
## Summary
This PR updates the config source so that any time an underlying file is changed it will trigger changing the envoy config. It also includes a new file manager for Envoy which will write content-addressable files instead of referencing them directly. This way if their contents change they will appear as new files to envoy.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


config/config_source.go
```diff
@@ -1,9 +1,14 @@
 package config
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
+	"io/ioutil"
 	"sync"
 
 	"github.com/fsnotify/fsnotify"
```
desimone, line 9: unrelated, but would be nice to standardize on a single notify package here.

desimone (approved):

internal/controlplane/filemgr/config.go
```diff
@@ -0,0 +1,33 @@
...
+	return func(cfg *config) {
+		cfg.cacheDir = cacheDir
+	}
+}
+
+func newConfig(options ...Option) *config {
+	cfg := new(config)
+	cacheDir, err := os.UserCacheDir()
+	if err != nil {
+		cacheDir = filepath.Join(os.TempDir())
```
travisgroth, line 26: If `os.UserCacheDir` fails, I believe it would be safer to create a more random subdirectory (eg, ioutil.TempDir) and ensure the permissions are properly restricted before using it.  Thoughts?

travisgroth (commented):
Overall LGTM pending @desimone's thoughts.  One minor proposed change, though.

config/config_source.go
calebdoxsey: We would have to fork viper, or not use its change detection to do this.

internal/controlplane/filemgr/config.go
calebdoxsey: I added a uuid and updated the permissions to be 700.

config/config_source.go
desimone: Ah okay, this is fine then.

PR #1779, opened 2021-01-15
reduce memory usage by handling http/2 coalescing via a lua script

calebdoxsey:
## Summary
This updates the XDS code to return less routing configuration by only generating routes for the matching domains in a virtual host. This will run aground of #1739, so we also use a Lua script to detect a mismatched domain name and return a 421 instead of a 404 in that case.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/controlplane/xds_filters.go
```diff
@@ -0,0 +1,139 @@
...
+		getExtAuthZSetCookieFilter(),
+		getCleanUpstreamFilter(),
+	}
+
+	if tlsDomain != "" && tlsDomain != "*" {
+		fs = append(fs, getFixMisdirectedFilter(tlsDomain))
+	}
+
+	fs = append(fs, &envoy_http_connection_manager.HttpFilter{
+		Name: "envoy.filters.http.router",
```
desimone, line 31: nit:`s/"envoy.filters.http.router"/wellknown.Router`

desimone (approved):
LGTM. One nit, and a suggestion that we are consistent (if possible) in how we leverage `lua` code.

calebdoxsey:
This PR broke routing with ports.

PR #1798, opened 2021-01-21
config: support multiple authorize and databroker URLs

calebdoxsey:
## Summary
Support multiple authorize and databroker URLs in the config.

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (approved):
LGTM. Would love @wasaga eyes as well for a +2

authenticate/authenticate.go
```diff
@@ -40,9 +40,6 @@ func ValidateOptions(o *config.Options) error {
 	if o.AuthenticateCallbackPath == "" {
 		return errors.New("authenticate: 'AUTHENTICATE_CALLBACK_PATH' is required")
 	}
-	if err := urlutil.ValidateURL(o.DataBrokerURL); err != nil {
-		return fmt.Errorf("authenticate: invalid 'DATABROKER_SERVICE_URL': %w", err)
-	}
 	return nil
 }
```
wasaga, line 44: shouldn't `ValidateOptions` be merged with `func (o *Options) Validate()` ?

config/options.go
```diff
@@ -242,9 +242,9 @@ type Options struct {
 	ForwardAuthURLString string   `mapstructure:"forward_auth_url" yaml:"forward_auth_url,omitempty"`
 	ForwardAuthURL       *url.URL `yaml:",omitempty"`
 
-	// DataBrokerURL is the routable destination of the databroker service's gRPC endpiont.
-	DataBrokerURLString string   `mapstructure:"databroker_service_url" yaml:"databroker_service_url,omitempty"`
-	DataBrokerURL       *url.URL `yaml:",omitempty"`
+	// DataBrokerURLString is the routable destination of the databroker service's gRPC endpoint.
+	DataBrokerURLString  string   `mapstructure:"databroker_service_url" yaml:"databroker_service_url,omitempty"`
+	DataBrokerURLStrings []string `mapstructure:"databroker_service_urls" yaml:"databroker_service_urls,omitempty"`
```
wasaga, line 247: maybe handle them similar to `To` via StringSlice for consistency?

config/options.go
```diff
@@ -718,22 +716,32 @@ func (o *Options) GetAuthenticateURL() *url.URL {
...
+}
+
+func (o *Options) getURLs(strs ...string) []*url.URL {
+	var urls []*url.URL
+	if o != nil {
+		for _, str := range strs {
+			u, err := urlutil.ParseAndValidateURL(str)
+			if err == nil {
+				urls = append(urls, u)
+			}
```
wasaga, line 736: error has to be propagated up

wasaga (changes requested):
+ update docs?

authenticate/authenticate.go
calebdoxsey: Pomerium can be run as separate services with separate configs.

config/options.go
calebdoxsey: mapstructure handles string slices natively. For `to` I wrote a custom type because we had to maintain compatibility with the previous format. Unfortunately this requires a very complex mapstructure decoding hook to handle properly.

I can do the same for `databroker_service_url` though I'd rather avoid it. If you disagree I can proceed with making `databroker_service_url` a `StringSlice` and handle single strings or lists of strings.

config/options.go
calebdoxsey: Error handling was done in Validate. Returning an error will complicate code downstream. Do you still want me to do this?

wasaga: in options.Validate there's only check for DataBrokerUrlString, not for a slice https://github.com/pomerium/pomerium/blob/1a33d753c24d6e2bdc677e80235712b7cd47dd83/config/options.go#L577

calebdoxsey: ok I will make the changes as requested.

calebdoxsey: I attempted to make these changes and as expected this will modify several hundred lines of code and touch most of the existing XDS code, as well as all of our tests. I am not comfortable making such a change in this PR. I will close this PR and start over.

PR #1821, opened 2021-01-25
config: fix databroker policies

calebdoxsey:
## Summary
The `Policies` slice is erased any time `Validate` is called. Because of this the policies added by the databroker were being inadvertently removed upstream.

This change adds another field called `AdditionalPolicies` to avoid this. I briefly looked into solving this by adding a new field to `config.Config` but it involved changing hundreds of files and I wasn't confident I could make those changes safely.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -112,6 +112,9 @@ type Options struct {
 	PolicyEnv  string   `yaml:",omitempty"`
 	PolicyFile string   `mapstructure:"policy_file" yaml:"policy_file,omitempty"`
 
+	// AdditionalPolicies are any additional policies added to the options.
+	AdditionalPolicies []Policy `yaml:"-"`
+
 	// AuthenticateURL represents the externally accessible http endpoints
 	// used for authentication requests and callbacks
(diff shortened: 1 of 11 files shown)


config/options.go
```diff
@@ -404,6 +407,12 @@ func (o *Options) parsePolicy() error {
 			return err
 		}
 	}
+	for i := range o.AdditionalPolicies {
+		p := &o.AdditionalPolicies[i]
+		if err := p.Validate(); err != nil {
+			return err
+		}
+	}
```
wasaga, line 415: maybe range over o.GetAllPolicies() ?

wasaga (approved):
it looks like the fastest way to address the problem

config/options.go
calebdoxsey: I tried that first, but unfortunately the `Validate` policy function actually populates needed fields in the policy and the slice is a slice of structs, not pointers, so we end up modifying a local copy of the data instead of the one in the actual slice

PR #1828, opened 2021-01-26
grpc: use custom resolver

calebdoxsey:
## Summary
To support multiple custom addresses we can't rely on gRPC's builtin resolver. This PR implements a custom resolver that supports multiple addresses like: `pomerium:///1.2.3.4:1234,2.3.4.5:2345`. 

## Related issues
#1798 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/grpc/client.go
```diff
@@ -117,11 +119,6 @@ func NewGRPCClientConn(opts *Options) (*grpc.ClientConn, error) {
 		dialOptions = append(dialOptions, grpc.WithTransportCredentials(cert))
 	}
 
-	if opts.ClientDNSRoundRobin {
```
wasaga, line 120: what is the relationship with `grpc_client_dns_roundrobin` config option now? 
probably that config option need be deprecated now as it has no effect?

pkg/grpc/resolver_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	defer func() { _ = li2.Close() }()
+
+	srv2 := grpc.NewServer()
+	grpc_testing.RegisterTestServiceServer(srv2, &resolverTestServer{
+		username: "srv2",
+	})
+	go func() { _ = srv2.Serve(li2) }()
+
+	cc, err := grpc.Dial("pomerium:///"+strings.Join([]string{
```
wasaga, line 50: maybe I'm missing something but how can I make a secure grpc call using this resolver?

pkg/grpc/client.go
calebdoxsey: I tried to use `dns:///` as before. But it doesn't work with localhost for some reason. I guess to maintain backwards compatibility we should preserve the existing behavior?

pkg/grpc/resolver_test.go
calebdoxsey: I think gRPC uses a separate option to configure TLS. Is that not working for you?

pkg/grpc/client.go
wasaga: but before we couldn't really specify multiple endpoints, right? thus I think we just need mark that option deprecated in docs and stop copying it from `config.Options` to `grpc.Options` ?

wasaga:
it seems they also recommend to set up a [health check](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) service when loading balancing is enabled but it's probably a subject for a separate PR

pkg/grpc/resolver_test.go
wasaga: right; so that's just for the hostnames/ports

PR #1830, opened 2021-01-27
upstream endpoints load balancer weights

wasaga:
## Summary

Endpoints may be specified with load balancing weight after it. When using a compatible load balancer set via `lb_policy` option, traffic would be distributed across the endpoints according to the relative weight distribution. Weight is positive `uint32`. 

Additionally, that eliminates previously computed `Destination` parameter from options. 

## Related issues

Fixes https://github.com/pomerium/internal/issues/274

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


config/custom.go
```diff
@@ -95,58 +99,227 @@ func (slc *StringSlice) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+}
+
+func DecodePolicyHookFunc() mapstructure.DecodeHookFunc {
+	return func(f, t reflect.Type, data interface{}) (interface{}, error) {
+		if t != reflect.TypeOf(Policy{}) {
 			return data, nil
 		}
 
-		ps, ok := m[policyKey].([]interface{})
+		log.Printf("POLICY HOOK %#v\n", data)
```
calebdoxsey, line 224: stray log

config/custom.go
```diff
@@ -95,58 +99,227 @@ func (slc *StringSlice) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+		mp, err := serializable(data)
+		if err != nil {
+			return nil, err
+		}
+		ms, ok := mp.(map[string]interface{})
 		if !ok {
-			return data, nil
+			return nil, errKeysMustBeStrings
```
calebdoxsey, line 232: Have you tested this? When I originally wrote this function the type was a `map[interface{}]interface{}` even though all the keys were strings.

internal/controlplane/xds_clusters.go
```diff
@@ -328,6 +332,7 @@ func buildLbEndpoints(endpoints []Endpoint) []*envoy_config_endpoint_v3.LbEndpoi
 					Address: buildAddress(u.Host, defaultPort),
 				},
 			},
+			// LoadBalancingWeight: e.loadBalancerWeight,
```
calebdoxsey, line 335: stray comment?

internal/databroker/config_source.go
```diff
@@ -83,7 +83,14 @@ func (src *ConfigSource) rebuild(firstTime bool) {
 
 	seen := map[uint64]struct{}{}
 	for _, policy := range cfg.Options.GetAllPolicies() {
-		seen[policy.RouteID()] = struct{}{}
+		id, err := policy.RouteID()
+		if err != nil {
+			log.Warn().Err(err).
+				Str("policy", policy.String()).
+				Msg("databroker: processing policy, ignoring")
```
calebdoxsey, line 90: is this the right log message?

calebdoxsey (changes requested):
Doesn't seem to work:

```
Jan 27 07:55:39 pomerium[34575]: {"level":"warn","error":"config: failed to parse policy: config: policy bad destination url parse \"http://127.0.0.2:8701,1\": invalid port \":8701,1\" after host","time":"2021-01-27T07:55:39-07:00","message":"databroker: invalid config detected, ignoring"}
Jan 27 07:55:39 pomerium[34575]: {"level":"error","error":"config: options from config file \"/home/caleb/src/github.com/calebdoxsey/pomerium-notes/cfg/pomerium-okta.yaml\": validation error config: failed to parse policy: config: policy bad destination url parse \"http://127.0.0.2:8701,1\": invalid port \":8701,1\" after host","time":"2021-01-27T07:55:39-07:00","message":"config: could not reload configuration"}
```

With:

```yaml
policy:
  - from: https://dashboard.caleb-pc-linux.doxsey.net
    to:
      - http://dashboard.localhost.pomerium.io:8701,1
      - http://127.0.0.2:8701,1
    allowed_domains:
      - pomerium.com
      - gmail.com
    pass_identity_headers: true
    allow_websockets: true
```

Did I not format the url correctly?

wasaga:
@calebdoxsey `config: policy bad destination url` error message  you observe does not exist in my branch - it is removed togehter with `Destination` field.

config/custom.go
wasaga: yes; the call to `serializable` above does `map[interface{}]interface{} => map[string]interface{}` conversion; thus it should be an unreachable error condition, but exists just in case to assert type conversion.  

```golang
		mp, err := serializable(data)
		if err != nil {
			return nil, err
		}

```

calebdoxsey:
@wasaga sorry about that. for some reason it was using master and not your branch. Maybe I accidently checked out with `-b`.

Now I'm seeing an error on this policy:

```

  - from: https://redirect-test.caleb-pc-linux.doxsey.net
    redirect:
      host_redirect: google.com
```
```
Jan 27 08:34:04 pomerium[50838]: {"level":"warn","error":"either no weights should be provided, or all endpoints must have non-zero weight specified","policy":"https://redirect-test.caleb-pc-linux.doxsey.net → ?","time":"2021-01-27T08:34:04-07:00","message":"databroker: processing policy, ignoring"}
Jan 27 08:34:04 pomerium[50838]: panic: runtime error: invalid memory address or nil pointer dereference
Jan 27 08:34:04 pomerium[50838]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1034320]
Jan 27 08:34:04 pomerium[50838]: goroutine 1 [running]:
Jan 27 08:34:04 pomerium[50838]: github.com/pomerium/pomerium/config.(*LogManager).OnConfigChange(0xc00040de60, 0x0)
Jan 27 08:34:04 pomerium[50838]:         /home/caleb/src/github.com/pomerium/pomerium/config/log.go:32 +0x60
Jan 27 08:34:04 pomerium[50838]: github.com/pomerium/pomerium/config.NewLogManager(0x1a4b720, 0xc0003e8900, 0x0)
Jan 27 08:34:04 pomerium[50838]:         /home/caleb/src/github.com/pomerium/pomerium/config/log.go:18 +0xa7
Jan 27 08:34:04 pomerium[50838]: github.com/pomerium/pomerium/internal/cmd/pomerium.Run(0x1a5eac0, 0xc000122000, 0x7ffc5322b8c5, 0x4c, 0x0, 0x0)
Jan 27 08:34:04 pomerium[50838]:         /home/caleb/src/github.com/pomerium/pomerium/internal/cmd/pomerium/pomerium.go:43 +0x11b
Jan 27 08:34:04 pomerium[50838]: main.run(0x1a5eac0, 0xc000122000, 0x0, 0x0)
Jan 27 08:34:04 pomerium[50838]:         /home/caleb/src/github.com/pomerium/pomerium/cmd/pomerium/main.go:30 +0x133
Jan 27 08:34:04 pomerium[50838]: main.main()
Jan 27 08:34:04 pomerium[50838]:         /home/caleb/src/github.com/pomerium/pomerium/cmd/pomerium/main.go:19 +0x3d
Jan 27 08:34:04 systemd[11503]: pomerium.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Jan 27 08:34:04 systemd[11503]: pomerium.service: Failed with result 'exit-code'.
```

wasaga:
I assumed `to` is [always required as per current live doc](https://pomerium.io/reference/#to) - I now see a separate use case with redirect support I was not aware of, will handle it.

calebdoxsey (approved):

wasaga:
@calebdoxsey one more pass if you don't mind? had to do a large merge with main branch.

calebdoxsey (approved):

wasaga:
@calebdoxsey could you please take a look at last commit (env variables)

calebdoxsey (approved):
I have never used the base64 env variable before for policies. If it works I guess this is fine.

config/options.go
```diff
@@ -385,16 +378,7 @@ func optionsFromViper(configFile string) (*Options, error) {
...
-	if o.PolicyEnv != "" {
-		policyBytes, err := base64.StdEncoding.DecodeString(o.PolicyEnv)
-		if err != nil {
-			return fmt.Errorf("could not decode POLICY env var: %w", err)
-		}
-		if err := yaml.Unmarshal(policyBytes, &policies); err != nil {
-			return fmt.Errorf("could not unmarshal policy yaml: %w", err)
-		}
-	} else if err := o.viperUnmarshalKey("policy", &policies); err != nil {
+	if err := o.viper.UnmarshalKey("policy", &policies, viperPolicyHooks); err != nil {
```
desimone, line 381: I believe no one is using policy as a base64 blob anymore. That funcitonality has not been around (as far as I am aware) since v0.0.4.

internal/urlutil/url.go
```diff
@@ -83,8 +83,8 @@ func GetAbsoluteURL(r *http.Request) *url.URL {
 //
 // For standard HTTP (80)/HTTPS (443) ports, it returns `example.com` and `example.com:<port>`.
 // Otherwise, return the URL.Host value.
-func GetDomainsForURL(u *url.URL) []string {
-	if IsTCP(u) {
+func GetDomainsForURL(u url.URL) []string {
```
desimone, line 86: Why not a pointer here (and elsewhere)?

internal/controlplane/xds_routes.go
```diff
@@ -528,18 +531,8 @@ func hasPublicPolicyMatchingURL(options *config.Options, requestURL *url.URL) bo
 	return false
 }
 
-func mustParseURL(str string) *url.URL {
+func mustParseURL(t *testing.T, str string) *url.URL {
```
desimone, line 534: Should this be in a `_test.go` file?

.vscode/launch.json
```diff
@@ -7,6 +7,8 @@
       "request": "launch",
       "mode": "debug",
       "program": "${workspaceRoot}/cmd/pomerium",
+      "env" : {
```
desimone, line 10: Accidental? (breaks the existing json)

config/constants.go
```diff
@@ -3,19 +3,35 @@ package config
...
 
 const (
-	policyKey    = "policy"
 	toKey        = "to"
 	envoyOptsKey = "_envoy_opts"
 )
 
 var (
-	errKeysMustBeStrings = errors.New("cannot convert nested map: all keys must be strings")
+	errKeysMustBeStrings          = errors.New("cannot convert nested map: all keys must be strings")
```
desimone, line 17: Should these errors be prepended with the package name (e.g. `pomerium/config:`)

config/constants.go
```diff
@@ -3,19 +3,35 @@ package config
...
+	errEmptyUrls                  = errors.New("url list is empty")
+	errEitherToOrRedirectRequired = errors.New("policy should have either `to` or `redirect` defined")
 )
 
 var (
 	protoPartial = protojson.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true}
 )
+
+var (
+	viperPolicyHooks = viper.DecodeHook(mapstructure.ComposeDecodeHookFunc(
```
desimone, line 31: Ah, I almost missed this code. Do you mind adding a few comments?

desimone (approved):
A few comments and suggestions. Many of the changes seem related to migrating a functions from using pointers of `url.URL` could you elaborate on that change? 

Otherwise, LGTM!

config/constants.go
wasaga: they're eventually wrapped by other messages that do mention `config`, thus I guess it'll be excessive. 

my thinking here is 
1. start collecting all errors into constants so it'll be easy to review errors user may see 
2. make every user facing error having a short URL pointing to KB article regarding problem root cause and suggested actions to fix

config/options.go
wasaga: - this is part of integration test that was failing
- `POLICY_FILE` seems to be removed, yes 
- `POLICY` env is supported as per docs https://www.pomerium.io/reference/#policy

internal/urlutil/url.go
wasaga: - I fixed a subtle pointer-related bug that occurred when multiple upstreams were present in the config, and that started changing pointers to non-pointers backwards - there were more in previous commits 
- use of pointers style is officially discouraged by the Go team https://github.com/golang/go/wiki/CodeReviewComments#pass-values
- indeed if you read thru i.e. K8S or any other protobuf centric code, there would be tons of pointers passed around because they explicitly prevent that, however that does not make that idiomatic code style. 
- pointers are potential source of faults - in principle you need check they're not nil
- i.e. I missed a check upon a condition in `options.Validate()` and that resulted in SIGSEGV in so many functions all over the place that I gave up fixing them. For a server software it's preferable not to be in a crash loop, IMHO. 
- in such functions they rarely provide any saving memory or GC wise; they're allocated on stack and cost nothing

wasaga:
use of `base64` `POLICY` is mentioned all over the documentation - i.e. kubernetes, cloud run

internal/urlutil/url.go
desimone: Thanks for the context and I take your point.

I think `url.URL` is a small enough struct to not really matter, and to your point, avoids many of the foot-guns related to passing pointers. 

But, my reading of the "pass-values" code review comments is not that the generic use of pointers is officially discouraged, but rather 

> "Don't pass pointers as function arguments just to save a few bytes."

With the caveat that 

> This advice does not apply to large structs, or even small structs that might grow.

To be fair, [url.URL](https://golang.org/pkg/net/url/#URL) is not large, but it's not tiny ( like the given example `*string`) either. 

I've gone back and forth about whether passing `url.URL` as a pointer made sense, and ended up defaulting to what I saw most often in the standard library which is to use a pointer. The example that jumps immediately to mind is [http.Request](https://golang.org/pkg/net/http/#Request). 

 I agree with all your other points.

config/constants.go
desimone: makes sense to me.

PR #1852, opened 2021-02-03
Revert "authorize: remove DataBrokerData input"

calebdoxsey:
Reverts pomerium/pomerium#1847

PR #1853, opened 2021-02-03
rego: handle null

calebdoxsey:
## Summary
I guess rego treats `null` as true. When the databroker would receive a service account, it would cause the databroker data to no longer be empty, which resulted in `null` being set for the session. 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (approved):
.... Javascript has competition?

desimone:
I'm going to merge this in favor of https://github.com/pomerium/pomerium/pull/1852

PR #1869, opened 2021-02-08
authorize: remove HTML forbidden page, always return 404s

calebdoxsey:
## Summary
This PR removes the special HTML forbidden page and always returns 404s for forbidden errors. This is to avoid leaking information about the existence of a route. 495s will still be returned for client certificate issues.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/check_response.go
+++ b/authorize/check_response.go
@@ -1,11 +1,9 @@
 package authorize
 
 import (
-	"bytes"
 	"net/http"
 	"net/url"
 	"sort"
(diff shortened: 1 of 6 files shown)


calebdoxsey:
Tabling this for now. The regression in user experience is pretty severe, and it doesn't really prevent a leak of information, since the initial 401 has to happen.

PR #1870, opened 2021-02-08
proxy: implement pass-through for authenticate backend

calebdoxsey:
## Summary
Currently its not possible to use the Pomerium proxy in front of the authenticate service. To support a single network ingress proxy, these changes implement pass-through to the authenticate backend. There are 3 changes:

1. `.pomerium` routes are not added
2. `ext_authz` is disabled
3. The cookie is not stripped

This should allow support for routes like:

```yaml
policy:
  - name: authenticate
    from: https://authenticate.localhost.pomerium.io
    to: https://pomerium-authenticate
    host_rewrite: "authenticate.localhost.pomerium.io"
    tls_skip_verify: true
```

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/controlplane/xds_routes.go
+++ b/internal/controlplane/xds_routes.go
@@ -49,57 +49,63 @@ func (srv *Server) buildGRPCRoutes() ([]*envoy_config_route_v3.Route, error) {
 
 func (srv *Server) buildPomeriumHTTPRoutes(options *config.Options, domain string) ([]*envoy_config_route_v3.Route, error) {
 	var routes []*envoy_config_route_v3.Route
-	// enable ext_authz
-	r, err := srv.buildControlPlanePathRoute("/.pomerium/jwt", true)
-	if err != nil {
-		return nil, err
-	}
-	routes = append(routes, r)
 
-	// disable ext_authz and passthrough to proxy handlers
-	r, err = srv.buildControlPlanePathRoute("/ping", false)
-	if err != nil {
-		return nil, err
-	}
-	routes = append(routes, r)
-	r, err = srv.buildControlPlanePathRoute("/healthz", false)
-	if err != nil {
-		return nil, err
-	}
-	routes = append(routes, r)
-	r, err = srv.buildControlPlanePathRoute("/.pomerium", false)
-	if err != nil {
-		return nil, err
-	}
-	routes = append(routes, r)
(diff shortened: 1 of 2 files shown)


internal/controlplane/xds_routes.go
```diff
@@ -49,57 +49,63 @@ func (srv *Server) buildGRPCRoutes() ([]*envoy_config_route_v3.Route, error) {
...
-		return nil, err
-	}
-	routes = append(routes, r)
-	r, err = srv.buildControlPlanePathRoute("/.well-known/pomerium", false)
-	if err != nil {
-		return nil, err
-	}
-	routes = append(routes, r)
-	r, err = srv.buildControlPlanePrefixRoute("/.well-known/pomerium/", false)
+	authenticateURL, err := options.GetAuthenticateURL()
```
desimone, line 53: I think we lost the associated comment here?

internal/controlplane/xds_routes.go
```diff
@@ -255,31 +266,9 @@ func (srv *Server) buildPolicyRoutes(options *config.Options, domain string) ([]
...
-								},
-							},
-						},
-					},
-				},
-			},
+			Name:                   fmt.Sprintf("policy-%d", i),
+			Match:                  match,
+			Metadata:               &envoy_config_core_v3.Metadata{},
 			RequestHeadersToAdd:    requestHeadersToAdd,
```
desimone, line 272: Could even inline the declaration here?
```go
		requestHeadersToAdd := toEnvoyHeaders(policy.SetRequestHeaders)
		requestHeadersToRemove := getRequestHeadersToRemove(options, &policy)
```

internal/controlplane/xds_routes.go
```diff
@@ -521,3 +539,11 @@ func hasPublicPolicyMatchingURL(options *config.Options, requestURL url.URL) boo
 	}
 	return false
 }
+
+func isProxyFrontingAuthenticate(options *config.Options, authenticateURL *url.URL, domain string) bool {
```
desimone, line 543: Does this function need both options and `authenticateURL` ?

Could we just grab it from options?

```go
	authenticateURL, err := options.GetAuthenticateURL()
	if err != nil {
		return nil, err
	}
```

desimone (approved):

PR #1875, opened 2021-02-10
docs: additional load balancing documentation

travisgroth:
## Summary

- Add some more context links and examples for the health checking and load balancing options
- Add the load balancing feature set to `UPGRADING`

## Related issues
https://github.com/pomerium/pomerium/pull/1804
https://github.com/pomerium/pomerium/pull/1796

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/docs/topics/load-balancing.md
+++ b/docs/docs/topics/load-balancing.md
@@ -51,7 +51,7 @@ policy:
 ```yaml
 policies:
   - from: tcp+https://tcp-service.localhost.pomerium.io
-    to: 
+    to:
       - tcp://tcp-1.local
       - tcp://tcp-2.local
     health_checks:
@@ -60,7 +60,7 @@ policies:
         unhealthy_threshold: 3
         healthy_threshold: 1
         tcp_health_check:
-          send: 
+          send:
             text: "50494E47" #PING
           receive:
             text: "504F4E47" #PONG
--- a/docs/docs/topics/load-balancing.md
+++ b/docs/docs/topics/load-balancing.md
@@ -26,11 +26,12 @@ In the presence of multiple upstreams, make sure to specify either an active or 
 
 :::
 
-### Active Health Checks
+## Active Health Checks
 
 Active health checks issue periodic requests to each upstream to determine its health.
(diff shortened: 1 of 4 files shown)


docs/docs/topics/load-balancing.md
```diff
@@ -65,11 +85,26 @@ policy:
...
+policy:
+  - from: https://myapp.localhost.pomerium.io
+    to:
+      - http://myapp-srv-1:8080
+      - http://myapp-srv-2:8080
+      - http://myapp-srv-3:8080
+      - http://myapp-srv-4:8080
+      - http://myapp-srv-5:8080
+    lb_policy: LEAST_REQUEST
+    least_request_lb_config:
```
travisgroth, line 105: @wasaga please double check me on this

desimone (commented):
LGTM; only question is do we always want to point at `latest` in the envoy documentation or pin to a version?

travisgroth:
> LGTM; only question is do we always want to point at `latest` in the envoy documentation or pin to a version?

I think a pin would be ideal, but we have a lot of links out to envoy and we're not usually very far behind.  Maybe not a major issue?

Something like [this](https://www.npmjs.com/package/vuepress-plugin-global-variables) would be helpful so we'd only have to update the version in one place, but I'm not sure if pulling that in is a good idea.  Thoughts?

desimone:
> I think a pin would be ideal, but we have a lot of links out to envoy and we're not usually very far behind. Maybe not a major issue?

Not major at all, and I don't think a concern for this API surface. I know there are a few places where we are on v2 of the API and envoy has removed support for v2 in 1.7

> Something like [this](https://www.npmjs.com/package/vuepress-plugin-global-variables) would be helpful so we'd only have to update the version in one place, but I'm not sure if pulling that in is a good idea. Thoughts?

Probably not bad idea at all. I didn't know this package existed.

travisgroth:
👍  i'll try adding that plugin.

desimone (approved):

travisgroth:
> 👍 i'll try adding that plugin.

This breaks some of our doc generation scripts and the formatting in a weird way, so I'm going to punt for the moment.

PR #1879, opened 2021-02-11
databroker: refactor databroker to sync all changes

calebdoxsey:
## Summary
The current version of the databroker does not guarantee that every change will be sent by the `Sync` method. This PR refactors the databroker to fix this.

## Interface Changes
`Record` changes:
- `version` is now a `uint64` to guarantee ordering
- `created_at` was removed
`DataBroker` changes:
- `Delete` has been removed. Use `Put` with a non-nil `DeletedAt`
- `Sync` only returns a single record at a time
- `record_version` and `server_version` are `uint64`s
- `SyncLatest` replaces `GetAll`. Streaming is used to avoid max message size issues.
- `Types` methods have been removed. `Sync` and `SyncLatest` return everything.
`Backend` changes:
- `List`, `Delete`, `ClearDeleted` and `Watch` have been removed
- `Sync` has been added which returns a `RecordStream`
- Methods were updated to take/return `Record`s

### In-Memory
- The in-memory storage provider is implemented using a btree and a map. 
- The btree contains all the changes as they occur sorted by version, which is incremented any time a change occurs. 
- The map is from `type,id` to the latest version of each record. 
- The store maintains an expiry and periodically removes changes if they've not been modified recently.

### Redis
- The redis storage provider is implemented using a hash and a sorted set. 
- The sorted set contains all the changes as they occur sorted by version, which is incremented any time a change occurs. 
- The hash is from `type,id` to the latest version of each record.
- We also store the global version and increment it when adding changes.
- We use a `TxPipeline` to make the changes. If the version changes in the interim, we retry the transaction. This should be safe with multiple databrokers.
- When a change occurs we use a pubsub channel to signal other databrokers that something has changed. This is used by `Sync` to get the latest record changes.
- Periodically we expire old changes by looping over the sorted set and checking the last modified date. We use `ZREM` with the key in case multiple databrokers try to expire the same key at the same time (it will simply be ignored if it doesn't exist)
- Members of the hash are only removed if `DeletedAt` is non-nil

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


.golangci.yml
```diff
@@ -98,6 +98,7 @@ linters:
     # - wsl
 
 issues:
+  exclude-use-default: false
```
calebdoxsey, line 101: we were skipping some issues codeclimate was picking up

internal/databroker/server.go
```diff
@@ -125,52 +91,32 @@ func (srv *Server) UpdateConfig(options ...ServerOption) {
...
 
-	db, _, err := srv.getDB(req.GetType(), true)
+	db, _, err := srv.getBackend(true)
 	if err != nil {
 		return nil, err
 	}
-	record, err := db.Get(ctx, req.GetId())
+	record, err := db.Get(ctx, req.GetType(), req.GetId())
 	if err != nil {
 		return nil, status.Error(codes.NotFound, "record not found")
```
wasaga, line 121: I think errors need be mapped - `storage.ErrNotFound => codes.NotFound`, everything else to `codes.InternalError`

wasaga (commented):
I'm still reading through it, will add more tomorrow.

internal/identity/manager/manager.go
```diff
@@ -288,9 +268,12 @@ func (mgr *Manager) mergeGroups(ctx context.Context, directoryGroups []*director
...
-					Id:   id,
+				_, err = mgr.cfg.Load().dataBrokerClient.Put(ctx, &databroker.PutRequest{
+					Record: &databroker.Record{
+						Type:      any.GetTypeUrl(),
+						Id:        id,
+						DeletedAt: timestamppb.Now(),
+					},
 				})
 				if err != nil {
 					return fmt.Errorf("failed to delete directory group: %s", id)
```
wasaga, line 279: this error doesn't seem to be visible anywhere, no? there's no `eg.Wait()`

internal/identity/manager/manager.go
```diff
@@ -288,9 +268,12 @@ func (mgr *Manager) mergeGroups(ctx context.Context, directoryGroups []*director
 				}
 				defer mgr.dataBrokerSemaphore.Release(1)
 
-				_, err = mgr.cfg.Load().dataBrokerClient.Delete(ctx, &databroker.DeleteRequest{
-					Type: any.GetTypeUrl(),
-					Id:   id,
+				_, err = mgr.cfg.Load().dataBrokerClient.Put(ctx, &databroker.PutRequest{
```
wasaga, line 271: if there are multiple instances of databroker component running, then there would be concurrent sync processes running as well?

wasaga (commented):
there seems to be a lack of observability for `identity` package, but addressing that should probably be part of a separate PR.

internal/databroker/server.go
```diff
@@ -272,197 +168,159 @@ func (srv *Server) Query(ctx context.Context, req *databroker.QueryRequest) (*da
...
 	}
-	return db, version, nil
+	return backend, version, nil
 }
 
-func (srv *Server) newDB(recordType string) (db storage.Backend, err error) {
+func (srv *Server) newBackend() (backend storage.Backend, err error) {
 	caCertPool, err := cryptutil.GetCertPool("", srv.cfg.storageCAFile)
 	if err != nil {
 		log.Warn().Err(err).Msg("failed to read databroker CA file")
```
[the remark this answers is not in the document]
calebdoxsey: I don't know. I didn't change this behavior.

pkg/grpc/databroker/syncer.go
```diff
@@ -0,0 +1,173 @@
...
+
+		if err != nil {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(syncer.backoff.NextBackOff()):
+			case <-syncer.closed:
+				return context.Canceled
+			}
+		}
```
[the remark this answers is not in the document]
calebdoxsey: it is reset in `init`.

internal/databroker/server.go
```diff
@@ -288,31 +296,37 @@ func (srv *Server) SyncLatest(req *databroker.SyncLatestRequest, stream databrok
...
+		backend, err = srv.newBackend()
+		srv.backend = backend
+		if err != nil {
+			return nil, 0, err
+		}
+	}
+	return backend, version, nil
+}
+
+func (srv *Server) getBackendLocked() (backend storage.Backend, version uint64, err error) {
```
wasaga, line 313: the semantics I was referring to is different - the `Locked` is a method that is executed when mutex is already acquired

internal/databroker/server.go
```diff
@@ -288,31 +296,37 @@ func (srv *Server) SyncLatest(req *databroker.SyncLatestRequest, stream databrok
 	})
 }
 
-func (srv *Server) getBackend(lock bool) (backend storage.Backend, version uint64, err error) {
+func (srv *Server) getBackend() (backend storage.Backend, version uint64, err error) {
```
wasaga, line 299: i.e. that function is supposed to be called `getBackendLocked`

internal/databroker/server.go
```diff
@@ -288,31 +296,37 @@ func (srv *Server) SyncLatest(req *databroker.SyncLatestRequest, stream databrok
...
+	backend = srv.backend
+	version = srv.version
+	if backend == nil {
+		var err error
+		backend, err = srv.newBackend()
```
wasaga, line 304: `newBackendLocked`

wasaga (changes requested):

pkg/grpc/databroker/syncer.go
```diff
@@ -55,36 +54,37 @@ type Syncer struct {
 	recordVersion uint64
 	serverVersion uint64
 
-	closeOnce sync.Once
-	closed    chan struct{}
+	closeCtx       context.Context
```
wasaga, line 57: I believe only cancel function need be stored not the context itself

pkg/grpc/databroker/syncer.go
```diff
@@ -55,36 +54,37 @@ type Syncer struct {
...
+	closeCtxCancel func()
 }
 
 // NewSyncer creates a new Syncer.
 func NewSyncer(handler SyncerHandler, options ...SyncerOption) *Syncer {
+	closeCtx, closeCtxCancel := context.WithCancel(context.Background())
+
```
wasaga, line 64: rm and store the context in `Run` method instead

wasaga (changes requested):

pkg/grpc/databroker/databroker.go
```diff
@@ -8,6 +8,16 @@ import (
...
 
+// TypeURLs for Any.
+const (
+	ConfigTypeURL         = "type.googleapis.com/pomerium.config.Config"
+	DirectoryGroupTypeURL = "type.googleapis.com/directory.Group"
+	DirectoryUserTypeURL  = "type.googleapis.com/directory.User"
+	SessionTypeURL        = "type.googleapis.com/session.Session"
+	ServiceAccountTypeURL = "type.googleapis.com/user.ServiceAccount"
+	UserTypeURL           = "type.googleapis.com/user.User"
+)
```
wasaga: if we're moving that into grpc package, we can make them all enums

internal/identity/manager/manager.go
calebdoxsey: There was supposed to be an `eg.Wait`. I added it.

internal/identity/manager/manager.go
calebdoxsey: yes. They would overwrite each other's changes.

pkg/grpc/databroker/databroker.go
calebdoxsey: sorry I don't understand. Go doesn't have enums. What do you want me to change?

pkg/grpc/databroker/syncer.go
calebdoxsey: This could lead to a nil reference exception if the syncer is closed without calling run. I can still make this change if you want but I'll have to add a mutex.

pkg/grpc/databroker/databroker.go
wasaga: I mean as long as all types are moved to `pkg/grpc/databroker`, we can also declare all types explicitly in `.proto` via enums and have the record types explicitly typed. 

or if we leave them just in Go code, then we can still keep them strongly typed like that 

```golang
type RecordType string

const (
	ConfigTypeURL         = RecordType("type.googleapis.com/pomerium.config.Config")
	DirectoryGroupTypeURL = RecordType("type.googleapis.com/directory.Group")
	DirectoryUserTypeURL  = RecordType("type.googleapis.com/directory.User")
	SessionTypeURL        = RecordType("type.googleapis.com/session.Session")
	ServiceAccountTypeURL = RecordType("type.googleapis.com/user.ServiceAccount")
	UserTypeURL           = RecordType("type.googleapis.com/user.User")
)
```

wasaga (approved):

pkg/grpc/databroker/databroker.go
calebdoxsey: I see. These type URLs are for the any type. I can update the code to generate them at runtime.

calebdoxsey: Oh nevermind, that will introduce a cyclic dependency. I can make a method in grpctuil I guess.

internal/identity/manager/manager.go
wasaga: i think it deserves a separate discussion https://github.com/pomerium/internal/issues/309

desimone (approved):
Nice work @calebdoxsey . 

LGTM; the code changes are more than I can keep in my head in a single review session,  but after a few passes everything looks reasonable to me, though It's possible I may have missed something, so apologies there if I did. In addition to code review, I've done manual testing and these changes perform as expected.

PR #1892, opened 2021-02-15
in-memory service registry

wasaga:
- implements simple service registry service  with in-memory cache. 
- for simplicity, always resides together with data broker service.
- other components advertise their available services by periodically reporting them
- the only current use case is embedded prometheus server running part of the console, that needs to know endpoints to scrape metrics from.


internal/registry/inmemory.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+type inMemoryServer struct {
+	ttl time.Duration
+	// onchange is used to broadcast changes to listeners
+	onchange *signal.Signal
+	// regs is {service,endpoint} -> expiration time mapping
+	regs map[inMemoryKey]*timestamppb.Timestamp
+	// mu holds lock for regs
+	mu sync.RWMutex
```
calebdoxsey, line 25: it's customary to put what the lock protects below it. ie:

```go
type inMemoryServer struct {
	ttl time.Duration
	// onchange is used to broadcast changes to listeners
	onchange *signal.Signal

	mu sync.RWMutex
	// regs is {service,endpoint} -> expiration time mapping
	regs map[inMemoryKey]*timestamppb.Timestamp
}
```

internal/registry/inmemory.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (s *inMemoryServer) periodicCheck(ctx context.Context) {
+	after := s.ttl * purgeAfterTTLFactor
+	for {
+		select {
+		case <-ctx.Done():
+			log.Info().Msg("grpc.service_registry.PeriodicCheck/Stop")
+			return
+		case <-time.After(after):
```
calebdoxsey, line 52: should probably use a `time.Ticker`

internal/registry/inmemory.go
```diff
@@ -0,0 +1,188 @@
...
+	updated, err := s.lockAndReport(req.Services)
+	if err != nil {
+		return nil, err
+	}
+
+	if updated {
+		s.onchange.Broadcast()
+	}
+
+	return &pb.RegisterResponse{CallBackAfter: ptypes.DurationProto(s.ttl / callAfterTTLFactor)}, nil
```
calebdoxsey, line 73: This is my fault for using the wrong package in many places, but we should use `durationpb` instead of `ptypes`.

internal/registry/reporter.go
```diff
@@ -0,0 +1,128 @@
...
+	after := minTTL
+	for {
+		select {
+		case <-time.After(after):
+			resp, err := client.Report(ctx, req)
+			if err != nil {
+				log.Error().Err(err).Msg("grpc.service_registry.Report")
+				after = backoff.NextBackOff()
+				continue
+			}
```
calebdoxsey, line 121: should call `backoff.Reset`

calebdoxsey (approved):
I saw a few small things but lgtm.

One potential issue with this approach is it requires pull network access for each service. A push-based approach to the databroker would mitigate that issue (since we already require services to be able to call the databroker, that's not a new requirement, but vice-versa might be). I don't know if this is an actual problem or not.

internal/registry/inmemory.go
wasaga: as discussed, that's very minor improvement vs need for manual resource management

wasaga:
- components report their whereabouts to registry, running on databroker node
- there is no new pull requirements as metrics are pulled by prometheus as previously.
- metrics continue to be optional component, if their address is not specified or it is unreachable, pomerium would continue to function.

internal/registry/inmemory.go
```diff
@@ -0,0 +1,192 @@
...
+	defer s.mu.RUnlock()
+
+	return s.getServicesLocked(kinds)
+}
+
+func (s *inMemoryServer) getServicesLocked(kinds map[pb.ServiceKind]bool) []*pb.Service {
+	out := make([]*pb.Service, 0, len(s.regs))
+	for k := range s.regs {
+		if len(kinds) == 0 {
+		} else if _, exists := kinds[k.kind]; !exists {
```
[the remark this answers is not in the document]
desimone: Also, a test for the continue case would be helpful

internal/registry/reporter.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/pomerium/pomerium/config"
+	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/pkg/grpc"
+	pb "github.com/pomerium/pomerium/pkg/grpc/registry"
+
+	"github.com/cenkalti/backoff/v4"
+)
+
+type Reporter struct {
```
desimone, line 20: missing doc strings on publicly exported struct

pkg/grpc/registry/registry.proto
```diff
@@ -0,0 +1,82 @@
+syntax = "proto3";
+
+package registry;
+option go_package = "github.com/pomerium/pomerium/pkg/grpc/registry";
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/any.proto";
+import "validate/validate.proto";
```
desimone, line 9: to answer my own question, I assume being used here?

pkg/grpc/registry/registry.proto
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// RegistrationSnapshot represents current state of the services
+message ServiceList {
+    repeated Service services = 1;
+}
+
+// Registry is invoked by services to inform
+service Registry {
+    // Register is periodically sent by each service to confirm it is still serving with the registry
```
desimone, line 73:
```suggestion
    // Report is periodically sent by each service to confirm it is still serving with the registry
```

pkg/grpc/registry/registry.proto
```diff
@@ -0,0 +1,82 @@
...
+    // data is persisted with a certain TTL
+    rpc Report(RegisterRequest) returns (RegisterResponse);
+    // List returns current snapshot of the services known to the registry
```
desimone, line 76:
```suggestion
    // List returns the current snapshot of the services known to the registry
```

pkg/grpc/registry/registry.proto
```diff
@@ -0,0 +1,82 @@
...
+message ListRequest {
+    repeated ServiceKind kinds = 1 [(validate.rules).repeated.min_items = 1];
+}
+
+message ServiceRegistration {
+    Service service = 1;
+    google.protobuf.Timestamp expires_at = 2;
+}
+
+// RegistrationSnapshot represents current state of the services
```
desimone, line 66:
```suggestion
// RegistrationSnapshot represents the current state of the services
```

pkg/grpc/registry/registry.proto
```diff
@@ -0,0 +1,82 @@
...
+    CONSOLE = 6;
+    // prometheus metrics compatible http endpoint
+    PROMETHEUS_METRICS = 7;
+}
+
+message Service {
+    // kind is service kind
+    ServiceKind kind = 1 [(validate.rules).enum = {not_in: [0]}];
+    // network endpoints this service may be reachable at
+    string endpoint = 3 [(validate.rules).repeated.min_items = 1];
```
desimone, line 34: What would "all-in-one" mode report as? All of them repeated?

internal/registry/reporter.go
```diff
@@ -0,0 +1,129 @@
...
+	backoff.MaxElapsedTime = 0
+
+	req := &pb.RegisterRequest{Services: services}
+	after := minTTL
+	for {
+		select {
+		case <-time.After(after):
+			resp, err := client.Report(ctx, req)
+			if err != nil {
+				log.Error().Err(err).Msg("grpc.service_registry.Report")
```
desimone, line 118: Maybe useful to log with ctx here? 
```suggestion
				log.Ctx(ctx).Error().Err(err).Msg("grpc.service_registry.Report")
```

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.1.1
 	github.com/coreos/go-oidc/v3 v3.0.0
 	github.com/envoyproxy/go-control-plane v0.9.9-0.20201210154907-fd9021fe5dad
+	github.com/envoyproxy/protoc-gen-validate v0.1.0
```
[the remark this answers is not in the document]
desimone: https://github.com/pomerium/pomerium/pull/1892#discussion_r577225810

internal/cmd/pomerium/constants.go
```diff
@@ -0,0 +1,7 @@
+package pomerium
+
+import "time"
+
+const (
+	registryTTL = time.Minute
```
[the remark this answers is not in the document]
wasaga: it always starts like that, and there are really two outcomes: 

- it grows and then you have `constants.go` with tons of constants that's easy to review
- they get scattered across the code

although that's not really documented anywhere, I think a lot of projects follow that semantics.

pkg/grpc/registry/registry.proto
wasaga: yes. however currently we only advertise metrics endpoint.

internal/cmd/pomerium/constants.go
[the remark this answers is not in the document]
desimone: > although that's not really documented anywhere, I think a lot of projects follow that semantics

I got curious so in another comment I looked at what go std library does. 

https://github.com/pomerium/pomerium/pull/1892#discussion_r577222457

internal/registry/inmemory.go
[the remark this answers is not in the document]
wasaga: added comments and test

PR #1906, opened 2021-02-17
config: support map of jwt claim headers

### LINKED ISSUE #1855: authorize: allow arbitrary claims in assertion / identity headers

Today, it possible to get a [user's identity](https://www.pomerium.com/docs/topics/getting-users-identity.html#verification) either from a signed attestation header (`X-Pomerium-Jwt-Assertion`) or via a subset of optional unisgned [identity headers](https://www.pomerium.com/reference/#jwt-claim-headers) (`X-Pomerium-Claim-Email` / `X-Pomerium-Claim-Group`).

However, is not possible to add identity details in a flexible way to support systems that may expect an unsigned identity header with a different header key (commonly, `x-forwarded-user`) . 

Also, upstream applications may desire more flexibility in what identity details (both to enrich, and prevent leaking unwanted details) are included in the signed assertion jwt header. 

For unsigned identity headers:
- It should be possible to define the desired header key (e.g. `x-forwarded-user`) for a given claim value. 
- It should be possible to assign an arbitrary datastore value (possibly using rego syntax? here I'm less sure about the implementation) to a given header key. 

For the signed assertion header:

- It should be possible to define an arbitrary set of identity claims to the assertion JWT. 

To prevent leaking data, both should be configurable on a per-policy (vs route?) basis. 

I suppose we could support arbitrary header rewriting as a workaround, but that feels janky, and does not handle the assertion use case. 

```
   RewriteEngine On
   RewriteCond %{LA-U:REMOTE_USER} (.+)
   RewriteRule . - [E=RU:%1,NS]
   RequestHeader set X-Forwarded-User "%{RU}e" env=RU
```

If there are any use-cases that would help us understand the use/need/value please share them as they can help us decide on acceptance and prioritization.

**Additional context**

- [Jenkins](https://plugins.jenkins.io/reverse-proxy-auth-plugin/) wanting `X-Forwarded-User` `X-Forwarded-Groups`
- [zammad](https://docs.zammad.org/en/latest/appendix/single-sign-on.html#conceptual-overview) wanting `X-Forwarded-User`

calebdoxsey:
## Summary
Update the JWT claim headers to support a map in addition to a slice:

```yaml
jwt_claims_headers:
  X-Email: email
```

This should make it easier to support downstream services which expect IDP claims in pre-defined headers which can't be customized.

## Related issues
- #1855 

## Checklist
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


config/custom_test.go
```diff
@@ -11,6 +11,27 @@ import (
...
+		var hdrs JWTClaimHeaders
+		err := json.Unmarshal([]byte(`{"x":"y"}`), &hdrs)
+		assert.NoError(t, err)
+		assert.Equal(t, JWTClaimHeaders{"x": "y"}, hdrs)
+	})
+	t.Run("array", func(t *testing.T) {
+		var hdrs JWTClaimHeaders
+		err := json.Unmarshal([]byte(`["x", "y"]`), &hdrs)
+		assert.NoError(t, err)
+		assert.Equal(t, JWTClaimHeaders{"x-pomerium-claim-x": "x", "x-pomerium-claim-y": "y"}, hdrs)
```
calebdoxsey, line 25: this our existing format, so should be backwards compatible

config/custom.go
```diff
@@ -220,53 +285,66 @@ func (urls WeightedURLs) Flatten() ([]string, []uint32, error) {
...
 	}
 }
 
+// DecodePolicyHookFunc creates a mapstructure DecodeHookFunc.
 func DecodePolicyHookFunc() mapstructure.DecodeHookFunc {
 	return func(f, t reflect.Type, data interface{}) (interface{}, error) {
-		if t != reflect.TypeOf(Policy{}) {
-			return data, nil
-		}
+		switch t {
```
wasaga, line 321: my understanding of how mapstructure hooks work is that you create a separate hook for each type, instead of making a large single function which tries to handle multiple ones. 

also then you could test them individually.

config/custom_test.go
```diff
@@ -11,6 +11,27 @@ import (
...
+	})
+	t.Run("string", func(t *testing.T) {
+		var hdrs JWTClaimHeaders
+		err := json.Unmarshal([]byte(`"x, y"`), &hdrs)
+		assert.NoError(t, err)
+		assert.Equal(t, JWTClaimHeaders{"x-pomerium-claim-x": "x", "x-pomerium-claim-y": "y"}, hdrs)
+	})
+}
+
```
wasaga, line 34: yaml test?

wasaga (approved):
lgtm + optional comments

PR #1917, opened 2021-02-19
config: add metrics_basic_auth option

calebdoxsey:
## Summary
Add a new option to the config: `metrics_basic_auth`. When enabled the metrics endpoint will require basic auth.

## Checklist
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/httputil/handlers.go
```diff
@@ -64,3 +64,24 @@ func (f HandlerFunc) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		e.ErrorResponse(w, r)
 	}
 }
+
+// RequireBasicAuth creates a new handler that requires basic auth from the client before
+// calling the underlying handler.
+func RequireBasicAuth(handler http.Handler, username, password string) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("WWW-Authenticate", `Basic realm="Restricted"`)
```
desimone, line 72: Good catch.

internal/httputil/handlers.go
```diff
@@ -64,3 +64,24 @@ func (f HandlerFunc) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+		u, p, ok := r.BasicAuth()
+		if !ok {
+			http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+			return
+		}
+
+		if u != username || p != password {
```
desimone, line 80: probably complete overkill here but... we probably want `subtle.ConstantTimeCompare` to prevent any timing attack.

internal/httputil/handlers.go
```diff
@@ -64,3 +65,25 @@ func (f HandlerFunc) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("WWW-Authenticate", `Basic realm="Restricted"`)
+
+		u, p, ok := r.BasicAuth()
+		if !ok {
+			http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+			return
+		}
+
+		if subtle.ConstantTimeCompare([]byte(u), []byte(username)) != 1 ||
```
desimone, line 81: Sorry, I just realized another subtle thing about (no pun intended) `subtle.ConstantTimeCompare`.

> The time taken is a function of the length of the slices and is independent of the contents.

https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare

https://play.golang.org/p/Xga-wsZvhT

I think we have to hash (sha256) each value first to make sure the bytes are the same length.

wasaga (approved):

PR #1932, opened 2021-02-22
autocert: remove non-determinism

calebdoxsey:
## Summary
This test would fluctuate on test coverage because the ticker didn't always fire before completing. By making the ticker fire faster, it should always happen at least once before the test loop finishes.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (approved):
LGTM but I'm not sure it worked?

PR #1934, opened 2021-02-23
middleware: basic auth equalize lengths of input

### LINKED ISSUE #1933: metrics: hash before constant time compare
Sorry, I just realized another subtle thing about (no pun intended) `subtle.ConstantTimeCompare`.

> The time taken is a function of the length of the slices and is independent of the contents.

https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare

I think we have to hash (sha256) each value first to make sure the bytes are the same length.

_Originally posted by @desimone in https://github.com/pomerium/pomerium/pull/1917#r580409418_

desimone:
## Summary

- Equalize the the length of the inputs to avoid timing attack
- Move the basic auth handler into the middleware package
- Add more tests

## Related issues

- Fixes #1933 
- Related to #1917 

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/middleware/middleware.go
+++ b/internal/middleware/middleware.go
@@ -1,6 +1,8 @@
 package middleware
 
 import (
+	"crypto/sha256"
+	"crypto/subtle"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


internal/middleware/middleware.go
```diff
@@ -75,3 +77,31 @@ func TimeoutHandlerFunc(timeout time.Duration, timeoutError string) func(next ht
...
 }
+
+// RequireBasicAuth creates a new handler that requires basic auth from the client before
+// calling the underlying handler.
+func RequireBasicAuth(username, password string) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			u, p, ok := r.BasicAuth()
+			if !ok {
+				w.Header().Set("WWW-Authenticate", `Basic realm="Restricted"`)
```
desimone, line 88: Only set on error.

calebdoxsey (approved):

PR #1939, opened 2021-02-23
metrics: add TLS options

calebdoxsey:
## Summary
Add TLS options for the metrics endpoint and move it into envoy.

## Checklist
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


wasaga:
if you set `metrics_address` to non-IP address such as `host:port` then ingress would not be created.

internal/controlplane/http.go
```diff
@@ -46,4 +46,7 @@ func (srv *Server) addHTTPMiddleware() {
 	root.Path("/debug/pprof/symbol").HandlerFunc(pprof.Symbol)
 	root.Path("/debug/pprof/trace").HandlerFunc(pprof.Trace)
 	root.PathPrefix("/debug/pprof/").HandlerFunc(pprof.Index)
+
+	// metrics
+	root.Handle("/metrics", srv.metricsMgr)
```
wasaga, line 51: I'm not sure we want to also expose all the above special handlers for a metrics endpoint.

calebdoxsey:
@wasaga only using ip addresses is, as far as I know, an envoy limitation. It doesn't support names for listeners. I guess I could add code to convert `localhost` to `127.0.0.1`, but I'm not sure how else I could handle this. If this is a deal breaker, I guess I can not use envoy and revert to a go listener.

internal/controlplane/http.go
calebdoxsey: Do you want me to create a separate control plane HTTP server for metrics?

wasaga:
there's no need to revert to go listener, we can just have a format validation on config level allowing only `:port` format and elaborating in the docs and error message during config validation.

In my settings, I historically had an actual `hostname:port` set as `metrics_addr`, and then it took me some time (via envoy admin's `/config_dump`, which is probably very few people would know of) to figure out why the metrics endpoint is unavailable. 

as far as all the extra middleware endpoints are concerned, I think it's ok to leave as is if creating a separate cluster is too much of a trouble.

wasaga (approved):

PR #1947, opened 2021-02-25
config: expose viper policy hooks

calebdoxsey:
## Summary
So we can use these policy hooks in other projects to fill config options.

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/constants.go
+++ b/config/constants.go
@@ -29,8 +29,8 @@ var (
 
 var (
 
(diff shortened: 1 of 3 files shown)


desimone (approved):

wasaga (commented):
I'd suggest maybe only exposing `decodeJWTClaimHeadersHookFunc` instead of full set of hooks - functions are bit more stable then a composition.

PR #1948, opened 2021-02-26
telemetry: add process collector for envoy

calebdoxsey:
## Summary
Add a process collector to collect process metrics for envoy. This is based on the process collector in prometheus but compatible with our opencensus config.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/envoy/envoy.go
+++ b/internal/envoy/envoy.go
@@ -16,6 +16,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


internal/envoy/envoy.go
```diff
@@ -109,6 +113,7 @@ func NewServer(src config.Source, grpcPort, httpPort string) (*Server, error) {
 		httpPort:  httpPort,
 		envoyPath: envoyPath,
 	}
+	go srv.runProcessCollector()
```
desimone, line 116: How does this get cleaned up?

calebdoxsey: It does not. I can add code to fix this if we want, though AFAIK there's only ever one server.

internal/envoy/envoy.go
```diff
@@ -531,3 +536,35 @@ func (srv *Server) handleLogs(rc io.ReadCloser) {
...
+	if runtime.GOOS == "darwin" {
+		return
+	}
+
+	pc := metrics.NewProcessCollector("envoy")
+	if err := view.Register(pc.Views()...); err != nil {
+		log.Error().Err(err).Msg("failed to register envoy process metric views")
+	}
+
+	const collectInterval = time.Second * 20
```
wasaga, line 551: we scrape other metrics @ 10s

internal/telemetry/metrics/processes_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opencensus.io/stats/view"
+)
+
+func TestProcessCollector(t *testing.T) {
+	if runtime.GOOS == "darwin" {
```
wasaga, line 18: `!= "linux"`

wasaga (approved):
couple minor comments

PR #1957, opened 2021-03-02
config: multiple endpoints for authorize and databroker

calebdoxsey:
## Summary
Add new config options: `databroker_service_urls` and `authorize_service_urls` to support multiple endpoints. The existing config options continue to work.

## Checklist
- [x]  updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/controlplane/xds_clusters.go
```diff
@@ -31,12 +31,12 @@ type Endpoint struct {
 }
 
 // NewEndpoint creates a new Endpoint.
-func NewEndpoint(u url.URL, ts *envoy_config_core_v3.TransportSocket, weight uint32) Endpoint {
+func NewEndpoint(u *url.URL, ts *envoy_config_core_v3.TransportSocket, weight uint32) Endpoint {
```
desimone, line 34: @wasaga 😉

pkg/grpc/config/config.proto
```diff
@@ -159,7 +159,7 @@ message Settings {
   google.protobuf.Duration grpc_server_max_connection_age = 48;
   google.protobuf.Duration grpc_server_max_connection_age_grace = 49;
   optional string forward_auth_url = 50;
-  optional string databroker_service_url = 52;
+  repeated string databroker_service_urls = 52;
```
desimone, line 162: just double checking but it is safe to change to repeated without breaking change right?

desimone (approved):

pkg/grpc/config/config.proto
desimone: Saw referenced thread about this very topic :); ignore 👍

desimone: > Based on my understanding of protobuf it is safe to add repeated to the existing to field since it's not packed because its a string. In the encoded serialization format each field is prefixed with its number, so repeated fields simply re-use the same field number multiple times. Parsers are supposed to tolerate this and use the last value seen for the field according to https://developers.google.com/protocol-buffers/docs/encoding#optional

PR #1961, opened 2021-03-05
config: add rewrite_response_headers option

calebdoxsey:
## Summary
Add a new `rewrite_response_headers` option which allows users to modify response headers before they are returned to the user. Similar to http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect.

Only prefix matching is supported via lua. There's work being done in envoy to support regex and other matchers natively.

I also added a unit test for the lua code using a pure-go lua implementation to simulate a call from envoy.

## Checklist
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


wasaga (approved):

travisgroth (approved):

desimone (approved):
🎊

PR #1970, opened 2021-03-08
oidc: use groups claim from ID token if present

bonifaido:
Signed-off-by: Nandor Kracser <bonifaido@gmail.com>

## Summary

A lot of OIDC providers offer the de-facto "groups" claim, which lists a user's existing group memberships. These are (incomplete list):
- [OneLogin](https://developers.onelogin.com/openid-connect/scopes)
- [Dex](https://dexidp.io/docs/custom-scopes-claims-clients/#scopes)
- [okta](https://developer.okta.com/docs/guides/customize-tokens-groups-claim/add-groups-claim-org-as/)
- [Auth0](https://auth0.com/docs/extensions/authorization-extension?_ga=2.80662615.1996932338.1615212854-499469682.1614950441)

This way group information can be sent back to the backend service via `X-Pomerium-Claim-Groups`, grabbing the `groups` claim from the ID token.

A documentation update would be needed if this is accepted, but first I wanted to make sure this makes sense.

My use is currently with Dex, without this change I couldn't configure Pomerium in a way that it returned the group header properly.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/authorize/evaluator/opa/policy/authz.rego
+++ b/authorize/evaluator/opa/policy/authz.rego
@@ -205,6 +205,10 @@ jwt_payload_email = v {
 
 jwt_payload_groups = v {
 	v = array.concat(group_ids, get_databroker_group_names(group_ids))
+	v != []
+} else = v {
+	v = session.claims["groups"]
+	v != null
 } else = [] {
 	true
 }


tarokkk:
Hey, @desimone can you take a look at this PR? We use our own fork until upstream is merged but it would be nice to know if this is a viable solution. Thanks!

desimone:
@bonifaido @tarokkk  -- Thanks for the PR! At first glance, I think this makes sense, but want to review some potential edge cases (e.g. if both directory groups and id_token groups are available).

stefanhenseler:
@bonifaido Awesome! This was exactly what I needed! I'm currently using your branch rebased on 0.13.3, no problems so far. Only thing I noticed is that on the .pomerium page, the groups are not visible in the UI.

image

calebdoxsey (approved):
Thanks @bonifaido for this pull request. Apologies for the delay in response, we were having internal discussions around whether or not this was the best approach going forward.

We'd like to support some more general mechanism in the future for any arbitrary claim, as well as a more flexible way of implementing directory sync, since with this change although the groups are available in the headers and JWT, they're not actual directory groups and can't be used as part of policy (ie you can't use `allowed_groups` with them and have to use `allowed_idp_claims` instead). This is the same reason why they're displayed differently on the `.pomerium` dashboard.

After discussing this more we decided to go ahead and merge this change. We may have a different way of solving it in the future, so we may deprecate this approach eventually, but it's a good solution for now.

Thanks for your patience. LGTM, will merge.

PR #1974, opened 2021-03-09
docs: add JWT Verification w/Envoy guide

### LINKED ISSUE #1925: docs: demonstrate how to use envoy / sidecar to validate JWT assertion headers

 [Pomerium's assertion jwt](https://www.pomerium.com/docs/topics/getting-users-identity.html#verification) cannot be validated by legacy applications, or third party applications where oeprators do not have access to modify the source code.  

Ideally, it would be possible to document how to validate Pomerium's asertion JWT via a sidecar. 

Forunately, Envoy supports writing http filters based on incoming JWTs so long as the referenced JWKs endpoint is static. 

> The HTTP filter can be used to verify JSON Web Token (JWT). It will verify its signature, audiences and issuer. It will also check its time restrictions, such as expiration and nbf (not before) time. If the JWT verification fails, its request will be rejected. If the JWT verification succeeds, its payload can be forwarded to the upstream for further authorization if desired.

We should support this functionality today out of the box. However, it would be nice to provide example configurations to make it easier to setup envoy's configuration. 

- Using something like istio to enforce protocol mTLS, but that's not always possible. 

- Legacy applications
- Third-party applications

**Additional context**

- [Envoy JWT Authentication](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto.html?highlight=jwt%20authentication)
- [HTTP Filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/jwt_authn_filter#config-http-filters-jwt-authn)
- https://tools.ietf.org/html/rfc7517 
- https://github.com/salrashid123/envoy_iap

calebdoxsey:
## Summary
Add an example for how to verify the JWT header using Envoy.

## Related issues
Fixes #1925 

## Checklist
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/guides/jwt-verification.md
+++ b/docs/guides/jwt-verification.md
@@ -1,0 +1,227 @@
+---
+title: JWT Verification
+lang: en-US
+meta:
+  - name: keywords
+    content: pomerium identity-access-proxy envoy jwt
+description: >-
+  This example demonstrates how to verify the Pomerium JWT assertion header using Envoy.
+---
+
+# JWT Verification
+This example demonstrates how to verify the [Pomerium JWT assertion header](https://www.pomerium.io/reference/#pass-identity-headers) using [Envoy](https://www.envoyproxy.io/).
+
+## Requirements
+- [Docker](https://www.docker.com/)
+- [Docker Compose](https://docs.docker.com/compose/)
+- [mkcert](https://github.com/FiloSottile/mkcert)
+
+## Overview
+Two services are configured in a `docker-compose.yaml` file:
+
+- `pomerium` running an all-in-one deployment of Pomerium on `*.localhost.pomerium.io`
+- `envoy-jwt-checker` running envoy with a JWT Authn filter
+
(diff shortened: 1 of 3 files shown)


docs/guides/jwt-verification.md
```diff
@@ -0,0 +1,227 @@
...
+lang: en-US
+meta:
+  - name: keywords
+    content: pomerium identity-access-proxy envoy jwt
+description: >-
+  This example demonstrates how to verify the Pomerium JWT assertion header using Envoy.
+---
+
+# JWT Verification
+This example demonstrates how to verify the [Pomerium JWT assertion header](https://www.pomerium.io/reference/#pass-identity-headers) using [Envoy](https://www.envoyproxy.io/).
```
desimone, line 12: Maybe just add a short blurb about why this is useful? (e.g. a way to assert user went through pomerium without manual JWT verification, or mTLS).
- legacy applications
- third party apps

desimone (commented):
Looks great. Maybe just add a short blurb covering use-cases where this is useful.

travisgroth (approved):
👍

PR #1992, opened 2021-03-17
redis: add redis cluster support

calebdoxsey:
## Summary
Adds support for redis cluster URLs. Redis cluster doesn't seem to support transactions across multiple nodes, so we have to ensure all keys are stored in the same slot.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/testutil/redis.go
+++ b/internal/testutil/redis.go
@@ -85,6 +85,96 @@ func WithTestRedis(useTLS bool, handler func(rawURL string) error) error {
 	return e
 }
 
+// WithTestRedisCluster creates a new redis cluster 3 node cluster.
+func WithTestRedisCluster(handler func(rawURL string) error) error {
+	ctx, clearTimeout := context.WithTimeout(context.Background(), maxWait)
+	defer clearTimeout()
+
+	// uses a sensible default on windows (tcp/http) and linux/osx (socket)
+	pool, err := dockertest.NewPool("")
+	if err != nil {
+		return err
+	}
+
+	redises := make([]*dockertest.Resource, 3)
+	for i := range redises {
+		conf := "cluster-enabled yes\ncluster-config-file nodes.conf"
+		r, err := pool.RunWithOptions(&dockertest.RunOptions{
+			Hostname:   fmt.Sprintf("redis%d", i),
+			Repository: "redis",
+			Tag:        "6",
+			Entrypoint: []string{
+				"/bin/bash", "-c",
+				`echo "` + conf + `" >/tmp/redis.conf && chmod 0777 /tmp/redis.conf && exec docker-entrypoint.sh /tmp/redis.conf`,
+			},
+			ExposedPorts: []string{
+				"6379/tcp",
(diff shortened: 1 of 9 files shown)


internal/testutil/redis.go
```diff
@@ -85,6 +85,96 @@ func WithTestRedis(useTLS bool, handler func(rawURL string) error) error {
...
+		return err
+	}
+
+	e := handler(fmt.Sprintf("redis+cluster://%s", strings.Join(addrs, ",")))
+
+	for _, r := range redises {
+		if err := pool.Purge(r); err != nil {
+			return err
+		}
+	}
```
wasaga, line 173: maybe wrap this into `defer func()` ?

wasaga: idk how important is it to purge docker test pool; but also maybe try to clean up as many containers as possible and collect errors into `multierror.Error`

pkg/storage/redis/client.go
```diff
@@ -30,7 +61,17 @@ func newClientFromURL(rawurl string, tlsConfig *tls.Config) (*redis.Client, erro
...
 		return redis.NewClient(opts), nil
 
-	case "redis-sentinel", "rediss-sentinel", "redis-sentinels":
+	case clusterSchemes.Has(u.Scheme):
+		opts, err := ParseClusterURL(rawurl)
+		if err != nil {
+			return nil, err
+		}
+		if opts.TLSConfig != nil {
+			opts.TLSConfig = tlsConfig
```
wasaga, line 70: should we maybe complain when i.e. `redis` scheme is specified but TLS config is not null?

pkg/storage/redis/client_test.go
```diff
@@ -9,6 +9,43 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestParseClusterURL(t *testing.T) {
+	opts, err := ParseClusterURL("redis+cluster://CLUSTER_USERNAME:CLUSTER_PASSWORD@localhost:26379,otherhost:26479/?" + (&url.Values{
+		"read_only":            {"true"},
+		"username":             {"USERNAME"},
+		"password":             {"PASSWORD"},
```
wasaga, line 16: maybe we should complain if people provide unsupported parameters, instead of silently ignoring them?

wasaga (approved):
minor comments

wasaga:
REDIS seems to [discourage](https://redislabs.com/blog/redis-clustering-best-practices-with-keys/) the _same slot_ approach

internal/testutil/redis.go
calebdoxsey: we already defer a purge when the resource is created

pkg/storage/redis/client.go
calebdoxsey: The TLS config is optional. by default it would only validate the server name. I guess we could enforce requiring it, but that may break backwards compatibility.

PR #1993, opened 2021-03-17
envoy: upgrade to v1.17.1

calebdoxsey:
## Summary
Upgrade envoy to v1.17.1. The `v2` API is deprecated, so we switch the authorize service to the v3 API as well as several typed configs. Tracing also had to move to the control plane instead of the bootstrap.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/check_response.go
+++ b/authorize/check_response.go
@@ -7,9 +7,9 @@ import (
 	"sort"
 	"strings"
 
-	envoy_api_v2_core "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
-	envoy_service_auth_v2 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v2"
-	envoy_type "github.com/envoyproxy/go-control-plane/envoy/type"
+	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoy_service_auth_v3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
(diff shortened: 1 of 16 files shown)


wasaga (approved):
I didn't see much difference in behaviour between v2 and v3 auth; I wonder why did they call it v3 then.

desimone:
@wasaga  -- I'm not entirely clear either. 

```
The closest thing I could find was...

For the v3 xDS APIs, a brief list of the key improvements that were made with a clean break from v2:

- Packages organization was improved to reflect a more logical grouping of related APIs:
- The legacy envoy.api.v2 tree was eliminated, with protos moved to their logical groupings, e.g. envoy.config.core.v3, envoy.server.listener.v3.
- All packages are now versioned with a vN at the end. This allows for type-level identification of major version.
- xDS service endpoints/transport and configuration are split between envoy.service and envoy.config.
- Extensions now reflect the Envoy source tree layout under envoy.extensions.
- std::regex regular expressions were dropped from the API, in favor of RE2. The former have dangerous security implications.
- google.protobuf.Struct configuration of extensions was dropped from the API, in favor of typed configuration. This provides for better support for multiple instances of extensions, e.g. in filter chains, and more flexible naming of extension instances.

Over 60 deprecated fields were removed from the API.

Tooling and processes were established for API versioning support. This has now been reflected in the bootstrap Node, providing a long term notion of API support that control planes can depend upon for client negotiation.
```

https://www.envoyproxy.io/docs/envoy/latest/faq/api/why_versioning

PR #2001, opened 2021-03-19
config: use getters for certificates

calebdoxsey:
## Summary
As part of removing side-effects from config option validation, this PR uses `Get`... methods for certificates.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	"sort"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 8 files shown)


config/options.go
```diff
@@ -666,12 +661,7 @@ func (o *Options) Validate() error {
 	// strip quotes from redirect address (#811)
 	o.HTTPRedirectAddr = strings.Trim(o.HTTPRedirectAddr, `"'`)
 
-	// sort the certificates so we get a consistent hash
-	sort.Slice(o.Certificates, func(i, j int) bool {
```
desimone, line 670: Will we no longer get a consistent hash? Is that a problem?

internal/controlplane/xds_listeners.go
```diff
@@ -630,7 +630,13 @@ func (srv *Server) buildRouteConfiguration(name string, virtualHosts []*envoy_co
 }
 
 func (srv *Server) buildDownstreamTLSContext(cfg *config.Config, domain string) *envoy_extensions_transport_sockets_tls_v3.DownstreamTlsContext {
-	cert, err := cryptutil.GetCertificateForDomain(cfg.AllCertificates(), domain)
+	certs, err := cfg.AllCertificates()
+	if err != nil {
+		log.Warn().Str("domain", domain).Err(err).Msg("failed to get certificate for domain")
```
desimone, line 635: This warning looks incorrect?

config/options.go
calebdoxsey: It should be ok because it will only be hashing the strings/filenames and not the actual certificates anymore. Our file watcher should still trigger changes when files change.

desimone (approved):

PR #2016, opened 2021-03-23
deps: bundle all patch upgrades in a single group

travisgroth:
## Summary

Streamline our renovate upgrades so that all patch version bumps are in a single PR.

## Related issues

n/a

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


renovate.json
```diff
@@ -1,14 +1,27 @@
 {
   "extends": [
-    "config:base",
-    "schedule:earlyMondays"
+    "config:base"
   ],
   "automerge": true,
   "prHourlyLimit": 0,
+  "schedule": [
+    "before 10am on Monday"
```
desimone, line 8: Maybe sunday just because it can take a while? IDK.

desimone (approved):

renovate.json
travisgroth: I'm hoping with patches collapsed it won't be so bad.  Would rather it not be opening PRs on sunday to tempt people into looking 😄

PR #2017, opened 2021-03-23
telemetry: add installation id

calebdoxsey:
## Summary
Add a new `installation_id` to distinguish between multiple pomerium installs in metrics.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/telemetry/metrics/providers.go
```diff
@@ -73,32 +79,60 @@ func registerDefaultViews() error {
...
 			log.Error().Err(err).Msg("telemetry/metrics: fail to fetch proxy metrics")
 			return
 		}
 		defer resp.Body.Close()
 
-		envoyBody, err := ioutil.ReadAll(resp.Body)
+		err = writeMetricsWithInstallationID(w, resp.Body, installationID)
 		if err != nil {
-			log.Error().Err(err).Msg("telemetry/metric: failed to read proxy metrics")
+			log.Error().Err(err).Send()
```
desimone, line 111: Curious why `.Send` vs `Msg` here vs other places?

internal/telemetry/metrics/providers.go
```diff
@@ -73,32 +79,60 @@ func registerDefaultViews() error {
...
+		rec := httptest.NewRecorder()
+		exporter.ServeHTTP(rec, r)
+
+		err := writeMetricsWithInstallationID(w, rec.Body, installationID)
+		if err != nil {
+			log.Error().Err(err).Send()
+			return
+		}
+
+		req, err := http.NewRequestWithContext(r.Context(), "GET", envoyURL.String(), nil)
```
desimone, line 96: I suppose it makes sense to add the request context (`r.Context()`) to any log entry in this function as well?

internal/telemetry/metrics/providers.go
```diff
@@ -73,32 +79,60 @@ func registerDefaultViews() error {
...
-		w.Write(envoyBody)
+func writeMetricsWithInstallationID(w io.Writer, r io.Reader, installationID string) error {
+	var parser expfmt.TextParser
+	ms, err := parser.TextToMetricFamilies(r)
+	if err != nil {
+		return fmt.Errorf("telemetry/metric: failed to read prometheus metrics: %w", err)
+	}
+
+	for _, m := range ms {
+		for _, mm := range m.Metric {
```
desimone, line 125: Any potential performance concerns of this being O(M*N)? ; I don't have a good sense of the potential dimensionality of the `TextToMetricFamilies` struct.

internal/telemetry/metrics/providers.go
desimone: I guess this error doesn't need additional markup / context. TIL about Send.

desimone (approved):
LGTM, mostly questions for my own learnings :)

internal/telemetry/metrics/providers.go
calebdoxsey: Each metric is a separate line. Not sure how else we could solve this. I didn't find anything in the envoy docs about how to add arbitrary tags to metrics.

internal/telemetry/metrics/providers.go
calebdoxsey: I don't think I understand. What do we want out of the context in the log entry?

PR #2020, opened 2021-03-25
cryptutil: add envelope encryption w/key encryption key and data encryption key

calebdoxsey:
## Summary
Add `KeyEncryptionKey` and `DataEncryptionKey` for performing envelope encryption.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/cryptutil/dek.go
+++ b/pkg/cryptutil/dek.go
@@ -1,0 +1,113 @@
+package cryptutil
+
+import (
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"io"
+
+	lru "github.com/hashicorp/golang-lru"
+	"golang.org/x/crypto/chacha20poly1305"
+)
+
+const (
+	// DataEncryptionKeySize is the size of a data encryption key.
+	DataEncryptionKeySize = chacha20poly1305.KeySize
+	// DataEncryptionKeyCacheSize is the number of DEKs to keep in the LRU cache.
+	DataEncryptionKeyCacheSize = 20
+)
+
+// A DataEncryptionKey is an XChaCha20Poly1305 symmetric encryption key. For more details
+// see the documentation on KeyEncryptionKeys.
+type DataEncryptionKey struct {
+	data   [DataEncryptionKeySize]byte
+	cipher cipher.AEAD
+}
+
+// NewDataEncryptionKey returns a new DataEncryptionKey from existing bytes.
+func NewDataEncryptionKey(raw []byte) (*DataEncryptionKey, error) {
+	if len(raw) != DataEncryptionKeySize {
+		return nil, fmt.Errorf("cryptutil: invalid data encryption key, expected %d bytes, got %d",
+			DataEncryptionKeySize, len(raw))
+	}
+	dek := new(DataEncryptionKey)
+	copy(dek.data[:], raw)
+	dek.cipher, _ = chacha20poly1305.NewX(raw) // only errors on invalid size
+	return dek, nil
+}
+
+// GenerateDataEncryptionKey generates a new random data encryption key.
+func GenerateDataEncryptionKey() (*DataEncryptionKey, error) {
+	raw := make([]byte, DataEncryptionKeySize)
+	_, err := io.ReadFull(rand.Reader, raw)
+	if err != nil {
+		return nil, fmt.Errorf("cryptutil: error generating random data encryption key: %w", err)
+	}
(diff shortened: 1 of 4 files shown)


pkg/cryptutil/dek.go
```diff
@@ -0,0 +1,113 @@
...
+			DataEncryptionKeySize, len(raw))
+	}
+	dek := new(DataEncryptionKey)
+	copy(dek.data[:], raw)
+	dek.cipher, _ = chacha20poly1305.NewX(raw) // only errors on invalid size
+	return dek, nil
+}
+
+// GenerateDataEncryptionKey generates a new random data encryption key.
+func GenerateDataEncryptionKey() (*DataEncryptionKey, error) {
```
desimone, line 41: I suppose we could use the existing function `NewKey` which I think does the same thing?

desimone: or, `randomBytes(c)`

pkg/cryptutil/dek.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// EncryptString encrypts a string using the data encryption key and base64 encoding.
+func (dek *DataEncryptionKey) EncryptString(plaintext string) string {
+	bs := dek.Encrypt([]byte(plaintext))
+	return base64.StdEncoding.EncodeToString(bs)
+}
+
+// KeyBytes returns the private key encryption key's raw bytes.
+func (dek *DataEncryptionKey) KeyBytes() []byte {
```
desimone, line 80: Could we add tests for this func?

pkg/cryptutil/dek.go
```diff
@@ -0,0 +1,113 @@
...
+	return string(plaintextBytes), nil
+}
+
+// Encrypt encrypts data using the data encryption key.
+func (dek *DataEncryptionKey) Encrypt(plaintext []byte) []byte {
+	return Encrypt(dek.cipher, plaintext, nil)
+}
+
+// EncryptString encrypts a string using the data encryption key and base64 encoding.
+func (dek *DataEncryptionKey) EncryptString(plaintext string) string {
```
desimone, line 74: Could we add tests for this func?

pkg/cryptutil/dek.go
```diff
@@ -0,0 +1,113 @@
...
+	return NewDataEncryptionKey(raw)
+}
+
+// Decrypt decrypts encrypted data using the data encryption key.
+func (dek *DataEncryptionKey) Decrypt(ciphertext []byte) ([]byte, error) {
+	return Decrypt(dek.cipher, ciphertext, nil)
+}
+
+// DecryptString decrypts an encrypted string using the data encryption key and base64 encoding.
+func (dek *DataEncryptionKey) DecryptString(ciphertext string) (string, error) {
```
desimone, line 56: Could we add tests for this func?

pkg/cryptutil/kek.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	kek := new(PrivateKeyEncryptionKey)
+	kek.id = id
+	copy(kek.data[:], raw)
+	return kek, nil
+}
+
+// GenerateKeyEncryptionKey generates a new random key encryption key.
+func GenerateKeyEncryptionKey() (*PrivateKeyEncryptionKey, error) {
+	raw := make([]byte, KeyEncryptionKeySize)
```
desimone, line 60: can we reuse existing `randomBytes(c)` func?

pkg/cryptutil/kek.go
```diff
@@ -0,0 +1,165 @@
...
+	_, err := io.ReadFull(rand.Reader, raw)
+	if err != nil {
+		return nil, fmt.Errorf("cryptutil: error generating random data encryption key: %w", err)
+	}
+	return NewPrivateKeyEncryptionKey(uuid.New().String(), raw)
```
desimone, line 65: Thoughts on using the sha256-hmac of the key itself for this value to:
1. Limit depedencies
2. If, for whatever reason, you have just the private key, you can find out the associated key-id

pkg/cryptutil/kek.go
desimone: (the downside being that randomBytes does panic on lack of entropy, but I would argue that is one of the few times panic'ing makes sense.

pkg/cryptutil/kek.go
calebdoxsey: I used our existing `Hash` function and base58.

desimone (approved):

PR #2024, opened 2021-03-26
authorize: add databroker server and record version to result, force sync via polling

calebdoxsey:
## Summary
Add `databroker_server_version` and `databroker_record_version` to the rego policy and evaluator result. This is useful for auditing.

- Change `UpdateRecord` so that it takes in the server version as well, and updates all 3 values within a transaction so that the version is consistent
- Change how `forceSync` works so that it polls for the record to appear instead of overwriting it. This is necessary so that we don't apply changes out of order

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


authorize/evaluator/store.go
```diff
@@ -128,7 +128,25 @@ func (s *Store) UpdateRecord(record *databroker.Record) {
 		return
 	}
 
-	s.write(rawPath, msg)
+	err = storage.Txn(context.Background(), s.opaStore, storage.WriteParams, func(txn storage.Transaction) error {
+		err := s.writeTxn(txn, "/databroker_server_version", fmt.Sprint(serverVersion))
```
calebdoxsey, line 132: saved as a string to avoid round-trip integer precision loss

authorize/sync.go
```diff
@@ -29,13 +49,119 @@ func (syncer *dataBrokerSyncer) ClearRecords(ctx context.Context) {
...
+	return u
+}
+
+// waitForRecordSync waits for the first sync of a record to complete
+func (a *Authorize) waitForRecordSync(ctx context.Context,
+	recordTypeURL, recordID string,
+	maxWait, pollInterval time.Duration,
+) {
+	ctx, clearTimeout := context.WithTimeout(ctx, maxWait)
+	defer clearTimeout()
```
wasaga, line 130: I think it's customary to pass the context with timeout set on top, rather than passing a duration for a timeout.

authorize/sync.go
```diff
@@ -29,13 +49,119 @@ func (syncer *dataBrokerSyncer) ClearRecords(ctx context.Context) {
...
+
+	ticker := time.NewTicker(pollInterval)
```
wasaga, line 132: maybe use exponential backoff instead?

authorize/sync.go
```diff
@@ -29,13 +49,119 @@ func (syncer *dataBrokerSyncer) ClearRecords(ctx context.Context) {
...
+	defer ticker.Stop()
+
+	for {
+		current := a.store.GetRecordData(recordTypeURL, recordID)
+		if current != nil {
+			// record found, so it's already synced
+			return
```
wasaga, line 139: why not return a record? the way this method is used it's followed by requesting a record once again

authorize/sync.go
```diff
@@ -29,13 +49,119 @@ func (syncer *dataBrokerSyncer) ClearRecords(ctx context.Context) {
...
+		return nil
+	}
+	return u
+}
+
+// waitForRecordSync waits for the first sync of a record to complete
+func (a *Authorize) waitForRecordSync(ctx context.Context,
+	recordTypeURL, recordID string,
+	maxWait, pollInterval time.Duration,
+) {
```
wasaga, line 128: I'd suggest changing the signature of the function 
- return a record if found
- return an error

wasaga (commented):
I'm unsure about `waitForRecord`

calebdoxsey:
This conflicts with https://github.com/pomerium/pomerium/pull/2041. Will wait on that review before making changes.

authorize/sync.go
```diff
@@ -104,39 +108,37 @@ func (a *Authorize) forceSyncUser(ctx context.Context, userID string) *user.User
...
-	maxWait, pollInterval time.Duration,
-) {
-	ctx, clearTimeout := context.WithTimeout(ctx, maxWait)
-	defer clearTimeout()
-
-	ticker := time.NewTicker(pollInterval)
-	defer ticker.Stop()
+func (a *Authorize) waitForRecordSync(ctx context.Context, recordTypeURL, recordID string) (proto.Message, error) {
+	bo := backoff.NewExponentialBackOff()
+	bo.InitialInterval = time.Millisecond
```
wasaga, line 134: `bo.MaxElapsedTime = 0` so that it won't reach internal default value.

wasaga (approved):

PR #1777, opened 2021-01-15
add support for proxy protocol on HTTP listener

calebdoxsey:
## Summary
This PR adds a new config option for requiring incoming HTTP requests to use the [HAProxy proxy protocol](https://www.haproxy.org/download/1.9/doc/proxy-protocol.txt) (either version 1 or version 2).

**Checklist**:
- [ ] add related issues
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

PR #2043, opened 2021-03-31
change require_proxy_protocol to use_proxy_protocol

contrun:
I set `use_proxy_protocol` to be true in my yaml config. Envoy didn't use proxy protocol albeit. Both the documents and https://github.com/pomerium/pomerium/pull/1777 hint the name should be use_proxy_protocol.


calebdoxsey (approved):
Whoops... thanks for fix this. Definitely an oversight on my part.

PR #2044, opened 2021-03-31
add default gitlab url

contrun:
## Summary
The document about how to set gitlab.com as a idp provider https://www.pomerium.com/docs/identity-providers/gitlab.html#gitlab-com didn't mention we need to specify `idp_provider_url: "https://gitlab.com/"` in the config file.
I only set `id_provider`, `idp_client_id`, `idp_client_secret` and `idp_service_account` as in the document. I was baffled  by the error 
```
8:54PM WRN failed to refresh directory users and groups error="gitlab: error querying groups: Get \"/api/v4/groups\": unsupported protocol scheme \"\"" service=identity_manager
```

This error is due to pomerium actually tried to request url `/api/v4/groups` instead of `https://gitlab.com/api/v4/groups`. I think it is more reasonable to set a default endpoint `https://gitlab.com/` for gitlab.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/internal/directory/provider.go
+++ b/internal/directory/provider.go
@@ -59,7 +59,11 @@ func GetProvider(options Options) (provider Provider) {
 		globalProvider.options = options
 	}()
 
-	providerURL, _ := url.Parse(options.ProviderURL)
+	var providerURL *url.URL
+	// url.Parse will succeed even if we pass an empty string
+	if options.ProviderURL != "" {
+		providerURL, _ = url.Parse(options.ProviderURL)
+	}
 	switch options.Provider {
 	case auth0.Name:
 		serviceAccount, err := auth0.ParseServiceAccount(options)


calebdoxsey (changes requested):
Thanks for submitting this PR @contrun . I think it makes sense that we should default the provider URL when possible. I wonder if perhaps we accidentally removed this in another PR.

That said, it would make more sense for this code to exist in the gitlab package, and in fact it already does:

```go
var defaultURL = &url.URL{
	Scheme: "https",
	Host:   "gitlab.com",
}

func getConfig(options ...Option) *config {
	cfg := new(config)
	WithHTTPClient(http.DefaultClient)(cfg)
	WithURL(defaultURL)(cfg)
	for _, option := range options {
		option(cfg)
	}
	return cfg
}
```

I wonder if something else is going wrong? Maybe the `if providerURL == nil {` check isn't working? Indeed: https://play.golang.org/p/6qKrJxl3pmq

Can we change the check to:

```go
if options.ProviderURL == "" {
	return gitlab.New(gitlab.WithServiceAccount(serviceAccount))
}
```

then we can remove

```go
const (
	DefaultGitlabURL = "https://gitlab.com/"
)
```
and
```
if options.ProviderURL == "" {
	providerURL, _ = url.Parse(DefaultGitlabURL)
}
```

calebdoxsey (approved):

PR #2048, opened 2021-03-31
authenticate: validate signature on /.pomerium, /.pomerium/sign_in and /.pomerium/sign_out

travisgroth:
## Summary

Ensure redirect parameter signatures are validated for `/.pomerium/*` routes.

## Related issues

#2046 
Fixes https://github.com/pomerium/pomerium/security/advisories/GHSA-fv82-r8qv-ch4v

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/proxy/state.go
+++ b/proxy/state.go
@@ -60,7 +60,7 @@ func newProxyStateFromConfig(cfg *config.Config) (*proxyState, error) {
 	if err != nil {
 		return nil, err
 	}
-	state.authenticateDashboardURL = state.authenticateURL.ResolveReference(&url.URL{Path: dashboardPath})
+	state.authenticateDashboardURL = state.authenticateURL.ResolveReference(&url.URL{Path: "/.pomerium/"})
 	state.authenticateSigninURL = state.authenticateURL.ResolveReference(&url.URL{Path: signinURL})
 	state.authenticateRefreshURL = state.authenticateURL.ResolveReference(&url.URL{Path: refreshURL})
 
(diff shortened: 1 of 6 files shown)


proxy/state.go
```diff
@@ -60,7 +60,7 @@ func newProxyStateFromConfig(cfg *config.Config) (*proxyState, error) {
 	if err != nil {
 		return nil, err
 	}
-	state.authenticateDashboardURL = state.authenticateURL.ResolveReference(&url.URL{Path: dashboardPath})
+	state.authenticateDashboardURL = state.authenticateURL.ResolveReference(&url.URL{Path: "/.pomerium/"})
```
travisgroth, line 63: It looks like this was changed intentionally in the original fix.  Unsure if that should be retained.

calebdoxsey: Sorry I've lost track of these changes. I can't really review this for accuracy. I feel like this is the 7th time I've been asked for a code review on the same thing.

Are you saying we should use the variable name? My understanding is that the code in authenticate adds a trailing `/`, so our redirect URI fails HMAC validation because the url is changed. But these changes were made a while ago so maybe I'm misremembering.

travisgroth:
This is not a direct port of https://github.com/pomerium/pomerium/pull/2046, as we had some refactoring in #2000 that requires different option handling.

calebdoxsey (approved):

PR #2050, opened 2021-03-31
authorize: audit logging

calebdoxsey:
## Summary
Add audit logging to authorize service.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/protoutil/crypt.go
```diff
@@ -0,0 +1,169 @@
...
+
+// Encrypt encrypts a protobuf message.
+func (enc *Encryptor) Encrypt(msg proto.Message) (*cryptpb.SealedMessage, error) {
+	// get the data encryption key
+	dek, encryptedDEK, err := enc.getDataEncryptionKey()
+	if err != nil {
+		return nil, err
+	}
+
+	plaintext, err := protojson.Marshal(msg)
```
wasaga, line 92: why serialize protobuf to JSON?

pkg/protoutil/crypt.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return nil, err
+	}
+	ciphertext := dek.Encrypt(plaintext)
+
+	return &cryptpb.SealedMessage{
+		KeyId:             enc.kek.ID(),
+		DataEncryptionKey: encryptedDEK,
+		MessageType:       GetTypeURL(msg),
```
wasaga, line 101: also, if you keep it protobuf, every protobuf message has a [type identifier](https://pkg.go.dev/google.golang.org/protobuf@v1.26.0/reflect/protoreflect#Descriptor)

wasaga (approved):
LGTM but I'm unsure what benefits serialization to JSON instead of keeping protobuf provides.

pkg/protoutil/crypt.go
calebdoxsey: The intention here was that if someone were decrypting the payloads through some other mechanism, it'd be easier to read JSON than raw protobuf, since it's self-describing and most log systems seem to have native support for it.

pkg/protoutil/crypt.go
calebdoxsey: I think only when used with `Any`.

PR #2051, opened 2021-04-02
proxy: support re-proxying request through control plane for kubernetes

calebdoxsey:
## Summary
So here's one solution to our kubernetes `Impersonate-Group` issues. We can proxy from envoy to the control plane and then to the kubernetes API. By doing this we can rewrite the headers so they're split across multiple lines instead of coalesced.

This works, but there are a few sharp edges we probably need to fix if we want to actually move forward with this. Namely the TLS options and we need to verify things like port-forwarding work as expected.

### Context
(added by @desimone )

- kubernetes/kubernetes#94683
- kubernetes/kubernetes#82468
- envoyproxy/envoy#13053

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


internal/httputil/headers.go
```diff
@@ -17,6 +18,8 @@ const (
 	HeaderPomeriumResponse = "x-pomerium-intercepted-response"
 	// HeaderPomeriumJWTAssertion is the header key containing JWT signed user details.
 	HeaderPomeriumJWTAssertion = "x-pomerium-jwt-assertion"
+	// HeaderPomeriumReProxyPolicy is the header key containing the policy to reproxy a request to.
+	HeaderPomeriumReProxyPolicy = "x-pomerium-reproxy-policy"
```
desimone, line 22: nit: `ReProxy` implies two words but the header itself is `..-reproxy-...` vs `...re-proxy...`; would either change the header name or the variable name for consistently

internal/controlplane/server.go
```diff
@@ -167,6 +170,7 @@ func (srv *Server) Run(ctx context.Context) error {
 
 // OnConfigChange updates the pomerium config options.
 func (srv *Server) OnConfigChange(cfg *config.Config) error {
+	srv.reproxy.Update(cfg)
```
desimone, line 173: I seem to remember that order was important for this function. Is this correct?

internal/httputil/reproxy/reproxy.go
```diff
@@ -0,0 +1,144 @@
...
+	mu       sync.RWMutex
+	cipher   cipher.AEAD
+	options  *config.Options
+	policies map[uint64]*config.Policy
+}
+
+// New creates a new Handler.
+func New() *Handler {
+	h := new(Handler)
+	h.cipher, _ = cryptutil.NewAEADCipher(cryptutil.NewKey())
```
desimone, line 36: I think this can only fail if we give it the wrong key size or run out of entropy.

desimone: Could there be any issues by generating a new key for each for each instance? Presumably a request will only be re...proxied to itself but I want to be sure.

internal/httputil/reproxy/reproxy.go
```diff
@@ -0,0 +1,144 @@
...
+				r.Header.Add(httputil.HeaderImpersonateGroup, v)
+			}
+		}
+
+		var dsts []url.URL
+		for _, wu := range policy.To {
+			dsts = append(dsts, wu.URL)
+		}
+		if len(dsts) == 0 {
+			http.Error(w, "policy destination not found", http.StatusNotFound)
```
desimone, line 110: should we use the http error from the httputil package? It adds some nice helpful context, logging, and adapts the accepted payload type.

internal/httputil/reproxy/reproxy.go
```diff
@@ -0,0 +1,144 @@
...
+	defer h.mu.RUnlock()
+
+	encryptedPolicy := cryptutil.Encrypt(h.cipher, rawPolicy, nil)
+	encryptedPolicyStr := base64.StdEncoding.EncodeToString(encryptedPolicy)
+	return encryptedPolicyStr
+}
+
+// Middleware returns an HTTP middleware for handling reproxying.
+func (h *Handler) Middleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
desimone, line 73: Do we want to use `httputil.HandlerFunc`

internal/httputil/reproxy/reproxy.go
```diff
@@ -0,0 +1,144 @@
...
+		encryptedPolicyStr := r.Header.Get(httputil.HeaderPomeriumReProxyPolicy)
+		if encryptedPolicyStr == "" {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		policyID, err := h.DecryptPolicyID(encryptedPolicyStr)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
```
desimone, line 82: do we want to use `return httputil.NewError`

desimone (approved):
LGTM, some suggestions and questions.

desimone (changes requested):
I thought about this a bit more. 

I think we may need to HMAC (well, the AEAD construct is fine here -- it's less the `H` and more the `MAC` I'm concerned about).

 In the currently implementation, if upstream reflects or somehow otherwise leaks the response headers, could an attacker use that information replay a request, or get the value of the `HeaderPomeriumReProxyPolicy`?

We may need to `MAC` an issue, expiry, and underlying request metadata (host, headers, path, etc) to ensure the request is truly coming from who we think it is, and is not being replayed. 

This feels exceedingly unlikely, but want to be sure. Happy to discuss further.

internal/controlplane/server.go
calebdoxsey: Sorry I don't know.

internal/httputil/reproxy/reproxy.go
calebdoxsey: The shared key is used.

calebdoxsey:
@desimone I changed it to an HMAC but you are correct its vulnerable to replay. I can't add an expiration to the header because the header is a literal string set once when configuration is generated.

I would need to implement this in lua or via the authorize service for dynamic headers.

A better solution to this would be to require a client certificate for the connection from envoy so that only envoy can access the controlplane.

FWIW the controlplane is not accessible outside of the local machine and binds a random port on boot. To further restrict it I believe we could switch to a unix socket with a 0600 file so only the user running the pomerium process would be able to access it.

calebdoxsey:
@desimone as discussed, I now always strip the reproxy headers for incoming requests and also added a check that the reproxied policy is only for kubernetes. This should substantially reduce the potential for abuse.

PR #2062, opened 2021-04-06
databroker: remove unused installation id, close streams when backend is closed

calebdoxsey:
## Summary
Updating the installation id was causing the in-memory databroker store to get recreated. This would then close the backend, but corresponding streams weren't being closed. This change fixes it so they're closed explicitly for both the in-memory and redis (which was less prone to the issue because it also periodically polls). I also removed the installation id from the databroker server config, because we actually only use it in the telemetry package, so there's no need to use it directly.

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


wasaga (approved):
fix confirmed

PR #2065, opened 2021-04-06
config related metrics

wasaga:
## Summary

adds two new metrics: 

- `pomerium_config_db_errors{config_id="dashboard"}` - amount of errors, including routes skipped while parsing this config 
- `pomerium_config_db_version{config_id="dashboard"}` - current databroker version that was applied

## Related issues

https://github.com/pomerium/pomerium-console/issues/708

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,10 +7,7 @@
       "request": "launch",
       "mode": "debug",
       "program": "${workspaceRoot}/cmd/pomerium",
-      "args": [
-        "-config",
-        "${workspaceRoot}/.config.yaml"
-      ],
+      "args": ["-config", "${workspaceRoot}/.config.yaml"]
     },
     {
       "name": "Connect to server",
@@ -19,7 +16,7 @@
       "mode": "remote",
       "remotePath": "/go/src/github.com/pomerium/pomerium/",
       "port": 9999,
-      "host": "127.0.0.1",
+      "host": "127.0.0.1"
     }
   ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,10 @@
       "args": [
         "-config",
         "${workspaceRoot}/.config.yaml"
-      ]
+      ],
+      "env": {
+        "PATH": "${workspaceRoot}/bin/.getenvoy/builds/standard/1.17.1/darwin/bin/:${env:PATH}"
+      }
     },
     {
       "name": "Connect to server",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,9 +11,6 @@
         "-config",
         "${workspaceRoot}/.config.yaml"
       ],
-      "env": {
-        "PATH": "${workspaceRoot}/bin/.getenvoy/builds/standard/1.17.1/darwin/bin/:${env:PATH}"
-      }
     },
     {
       "name": "Connect to server",
--- a/internal/databroker/config_source.go
+++ b/internal/databroker/config_source.go
@@ -22,17 +22,22 @@ type ConfigSource struct {
 	mu               sync.RWMutex
 	computedConfig   *config.Config
 	underlyingConfig *config.Config
-	dbConfigs        map[string]*configpb.Config
+	dbConfigs        map[string]dbConfig
 	updaterHash      uint64
 	cancel           func()
 
 	config.ChangeDispatcher
 }
 
+type dbConfig struct {
+	*configpb.Config
+	version uint64
+}
+
 // NewConfigSource creates a new ConfigSource.
 func NewConfigSource(underlying config.Source, listeners ...config.ChangeListener) *ConfigSource {
 	src := &ConfigSource{
-		dbConfigs: map[string]*configpb.Config{},
+		dbConfigs: map[string]dbConfig{},
 	}
 	for _, li := range listeners {
 		src.OnConfigChange(li)
@@ -84,25 +89,31 @@ func (src *ConfigSource) rebuild(firstTime bool) {
 	var additionalPolicies []config.Policy
 
 	// add all the config policies to the list
-	for _, cfgpb := range src.dbConfigs {
+	for id, cfgpb := range src.dbConfigs {
 		cfg.Options.ApplySettings(cfgpb.Settings)
+		var errCount uint64
 
 		err := cfg.Options.Validate()
 		if err != nil {
-			log.Warn().Err(err).Msg("databroker: invalid config detected, ignoring")
-			return
+			metrics.SetDbConfigRejected(cfg.Options.Services, id, cfgpb.version, err)
+			continue
 		}
 
 		for _, routepb := range cfgpb.GetRoutes() {
 			policy, err := config.NewPolicyFromProto(routepb)
 			if err != nil {
-				log.Warn().Err(err).Msg("databroker: error converting protobuf into policy")
+				errCount++
+				log.Warn().Err(err).
+					Str("db_config_id", id).
+					Msg("databroker: error converting protobuf into policy")
 				continue
 			}
 
 			err = policy.Validate()
 			if err != nil {
+				errCount++
 				log.Warn().Err(err).
+					Str("db_config_id", id).
 					Str("policy", policy.String()).
 					Msg("databroker: invalid policy, ignoring")
 				continue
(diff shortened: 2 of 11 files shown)


.vscode/launch.json
```diff
@@ -10,7 +10,10 @@
       "args": [
         "-config",
         "${workspaceRoot}/.config.yaml"
-      ]
+      ],
+      "env": {
+        "PATH": "${workspaceRoot}/bin/.getenvoy/builds/standard/1.17.1/darwin/bin/:${env:PATH}"
+      }
```
calebdoxsey, line 16: I don't think we want this?

pkg/grpc/databroker/syncer.go
```diff
@@ -148,6 +155,12 @@ func (syncer *Syncer) sync(ctx context.Context) error {
 			return err
 		}
 
+		syncer.log().Info().
```
calebdoxsey, line 158: I think this should be `Debug`?

pkg/metrics/constants.go
```diff
@@ -19,6 +19,14 @@ const (
...
 	// ConfigChecksumDecimal should only be used to compare config on a single node, it will be different in multi-node environment
 	ConfigChecksumDecimal = "config_checksum_decimal"
+	// ConfigDbVersion sets currently loaded databroker config version config_db_version{service="service",id="config_id"}
+	ConfigDbVersion = "config_db_version"
+	// ConfigDbVersionHelp
+	ConfigDbVersionHelp = "databroker current config record version"
+	// ConfigDbErrors sets number of errors while parsing current config that were tolerated
+	ConfigDbErrors = "config_db_errors"
+	// ConfigDbErrorsHelp
+	ConfigDbErrorsHelp = "amount of errors observed while applying databroker config; -1 if validation failed and was rejected altogether"
```
calebdoxsey, line 29: I think the convention would be `ConfigDBErrorsHelp`

calebdoxsey (approved):

.vscode/launch.json
wasaga: otherwise it picks up random envoy from $PATH
what is your suggestion?

pkg/grpc/databroker/syncer.go
wasaga: I used the same log level as i.e. 
```log
{"level":"info","service":"databroker","peer":"127.0.0.1:53995","type":"type.googleapis.com/pomerium.config.Config","id":"dashboard","time":"2021-04-06T15:13:46-04:00","message":"get"}
```

calebdoxsey: Well its going to be a lot of log lines. In our case ~40k?

.vscode/launch.json
travisgroth: This has the OS and version hardcoded; I'm not sure if it will even work well on M1 chips since getenvoy doesn't support arm.

Would prefer something this opinionated isn't in a shared config.  Maybe it just shouldn't be in the repo, or we should make a more consistent location for the envoy binary.  We should tackle that separately.

travisgroth (changes requested):

.vscode/launch.json
```diff
@@ -10,7 +10,7 @@
       "args": [
         "-config",
         "${workspaceRoot}/.config.yaml"
-      ]
+      ],
```
desimone, line 13: nit: unrelated change.

pkg/grpc/databroker/syncer.go
desimone: I agree this is a better fit for debug. 

And in this case, we may want to even hide this under the verbose flag. 

And we can follow up with another PR to see that databroker set / gets are also debug.

pkg/metrics/constants.go
```diff
@@ -19,6 +19,14 @@ const (
 	PolicyCountTotal = "policy_count_total"
 	// ConfigChecksumDecimal should only be used to compare config on a single node, it will be different in multi-node environment
 	ConfigChecksumDecimal = "config_checksum_decimal"
+	// ConfigDBVersion sets currently loaded databroker config version config_db_version{service="service",id="config_id"}
+	ConfigDBVersion = "config_db_version"
+	// ConfigDBVersionHelp
```
desimone, line 24: missing doctring.

pkg/metrics/constants.go
```diff
@@ -19,6 +19,14 @@ const (
...
+	ConfigDBVersionHelp = "databroker current config record version"
+	// ConfigDBErrors sets number of errors while parsing current config that were tolerated
+	ConfigDBErrors = "config_db_errors"
+	// ConfigDBErrorsHelp
```
desimone, line 28: missing docstring

internal/databroker/config_source.go
```diff
@@ -84,40 +89,50 @@ func (src *ConfigSource) rebuild(firstTime bool) {
...
+	for id, cfgpb := range src.dbConfigs {
 		cfg.Options.ApplySettings(cfgpb.Settings)
+		var errCount uint64
 
 		err := cfg.Options.Validate()
 		if err != nil {
-			log.Warn().Err(err).Msg("databroker: invalid config detected, ignoring")
-			return
+			metrics.SetDBConfigRejected(cfg.Options.Services, id, cfgpb.version, err)
+			continue
```
desimone, line 99: do we not want to fail fast in this case?

desimone: Could this cause unexpected issues since we typically fail hard after failed validation?

desimone (changes requested):

pkg/metrics/constants.go
wasaga: a help text for a variable with help text ? :-)

internal/databroker/config_source.go
wasaga: it loops over available databroker configs. idk what other configs were originally planned to exist besides one supplied by the console, but as long as we're looping over and merging them, it should be more natural to just proceed to next one.

pkg/metrics/constants.go
desimone: lol, that's fair. Maybe delete the docstring title then?

pkg/metrics/constants.go
desimone: https://github.com/pomerium/pomerium/pull/2065#discussion_r608330400

pkg/metrics/constants.go
wasaga: then linter would be unhappy

travisgroth (approved):

internal/databroker/config_source.go
wasaga: by returning we are ignoring other configs that may be valid. your logic is valid in case we have just a single config, however by design we have multiple here that we merge. 

@calebdoxsey please advise

wasaga: also looping over map is random order that makes it even more non-deterministic. 
maybe we should eliminate the multiple databroker configs altogether here

calebdoxsey: I don't know what to do if config is invalid.

travisgroth: Was discussing with @wasaga - 

My understanding of the current code is that we're build a new config from the databroker by starting with a copy of current state, then mutating this copy.  It looks like we abort if the overall config doesn't validate, though we continue if individual routes have issues.  Is that right?

What I think I'd _want_ is to abort making changes if the new configuration is invalid for any reason.  If we proceed through a partial validation error, a previously functioning route or option could get dropped and cause an outage or otherwise undesirable state.  Preventing me from making new changes would retain whatever the current (likely functional) state is.  This does block new changes from happening, but I think that's the least surprising result and matches `HUP` behavior in a lot of other software.

pkg/metrics/constants.go
desimone, line 28:
```suggestion
	// ConfigDBErrorsHelp is the help text for ConfigDBErrors.
```

pkg/metrics/constants.go
desimone, line 24:
```suggestion
	// ConfigDBVersionHelp is the help text for ConfigDBVersion.
```

internal/databroker/config_source.go
wasaga: ok I'm reverting to original behaviour

internal/databroker/config_source.go
```diff
@@ -84,40 +89,50 @@ func (src *ConfigSource) rebuild(firstTime bool) {
 	var additionalPolicies []config.Policy
 
 	// add all the config policies to the list
-	for _, cfgpb := range src.dbConfigs {
+	for id, cfgpb := range src.dbConfigs {
 		cfg.Options.ApplySettings(cfgpb.Settings)
+		var errCount uint64
```
desimone, line 94: nit: `errCount` is now a ways away from its use. We might want to move this to just above the for loop

desimone (approved):

PR #2066, opened 2021-04-07
cryptutil: always use kek public id, add x509 support

calebdoxsey:
## Summary
Several changes:

1. Generate the key encryption key (KEK) id using the public KEK bytes instead of the private KEK bytes. This way given a public KEK, you can know the key id without needing the private KEK.
2. Add support for marshaling private KEKs via `MarshalPKCS8PrivateKey` similar to ed25519 keys, but with a different algorithm id.
3. Add support for marshaling public KEKs via `MarshalPKIXPublicKey` similar to ed25519 keys, but with a differnet algorithm id.

This should be interoperable with the standard format used by `openssl genpkey` and allows us to store KEKs in a predictable way.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/crypt.go
+++ b/config/crypt.go
@@ -22,5 +22,5 @@ func (o *Options) GetAuditKey() (*cryptutil.PublicKeyEncryptionKey, error) {
 	if err != nil {
 		return nil, err
 	}
-	return cryptutil.NewPublicKeyEncryptionKey(o.AuditKey.ID, raw)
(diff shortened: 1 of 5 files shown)


wasaga (commented):
LGTM but crypto is not my speciality

desimone (approved):

PR #2075, opened 2021-04-08
crypto: use actual bytes of shared secret, not the base64 encoded representation

calebdoxsey:
## Summary
We've been using the `SharedKey` directly from the options, which means we were actually HMACing with the base64 string, not the bytes themselves. This change fixes that.

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authenticate/state.go
+++ b/authenticate/state.go
@@ -69,12 +69,24 @@ func newAuthenticateStateFromConfig(cfg *config.Config) (*authenticateState, err
 	if err != nil {
 		return nil, err
 	}
-	state.redirectURL, _ = urlutil.DeepCopy(authenticateURL)
+
+	state.redirectURL, err = urlutil.DeepCopy(authenticateURL)
+	if err != nil {
+		return nil, err
+	}
+
 	state.redirectURL.Path = cfg.Options.AuthenticateCallbackPath
 
 	// shared cipher to encrypt data before passing data between services
-	state.sharedKey, _ = base64.StdEncoding.DecodeString(cfg.Options.SharedKey)
-	state.sharedCipher, _ = cryptutil.NewAEADCipher(state.sharedKey)
+	state.sharedKey, err = base64.StdEncoding.DecodeString(cfg.Options.SharedKey)
+	if err != nil {
+		return nil, err
+	}
+
+	state.sharedCipher, err = cryptutil.NewAEADCipher(state.sharedKey)
+	if err != nil {
+		return nil, err
+	}
 
 	// shared state encoder setup
 	state.sharedEncoder, err = jws.NewHS256Signer(state.sharedKey)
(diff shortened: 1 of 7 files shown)


authenticate/state.go
```diff
@@ -72,16 +72,16 @@ func newAuthenticateStateFromConfig(cfg *config.Config) (*authenticateState, err
 	state.redirectURL, _ = urlutil.DeepCopy(authenticateURL)
 	state.redirectURL.Path = cfg.Options.AuthenticateCallbackPath
 
+	// shared cipher to encrypt data before passing data between services
+	state.sharedKey, _ = base64.StdEncoding.DecodeString(cfg.Options.SharedKey)
+	state.sharedCipher, _ = cryptutil.NewAEADCipher(state.sharedKey)
```
wasaga, line 77: let's not ignore errors

authenticate/state.go
```diff
@@ -72,16 +72,16 @@ func newAuthenticateStateFromConfig(cfg *config.Config) (*authenticateState, err
...
 		return nil, err
 	}
 
-	// shared cipher to encrypt data before passing data between services
-	state.sharedSecret, _ = base64.StdEncoding.DecodeString(cfg.Options.SharedKey)
-	state.sharedCipher, _ = cryptutil.NewAEADCipher(state.sharedSecret)
-
 	// private state encoder setup, used to encrypt oauth2 tokens
 	state.cookieSecret, _ = base64.StdEncoding.DecodeString(cfg.Options.CookieSecret)
 	state.cookieCipher, _ = cryptutil.NewAEADCipher(state.cookieSecret)
```
wasaga, line 87: errors should not be shielded. if configuration if wrong, the function should fail.

proxy/state.go
```diff
@@ -44,12 +44,12 @@ func newProxyStateFromConfig(cfg *config.Config) (*proxyState, error) {
 	}
 
 	state := new(proxyState)
-	state.sharedKey = []byte(cfg.Options.SharedKey)
+	state.sharedKey, _ = base64.StdEncoding.DecodeString(cfg.Options.SharedKey)
```
wasaga, line 47: err

wasaga (changes requested):

authenticate/state.go
calebdoxsey: I guess we've been through this a few times, but errors are checked upstream. I didn't change this behavior.

I will update the code but it will hurt readability.

wasaga (approved):

authenticate/state.go
wasaga: I looked how this function is called, and there's no other upstream that checks and validates these parameters. This function is parsing and preparing the runtime object, and it has to return all errors. 

btw following that logic, you should also drop the `cfg.Options.Validate()` call at the very top of this function, as that is an example of a call that was supposed to be called upstream.

calebdoxsey: Shared key is checked with the call to Validate in config.

PR #2081, opened 2021-04-12
config: rename headers to set_response_headers

calebdoxsey:
## Summary
This config setting was ambiguous about what it was for. With this change we rename it to `set_response_headers` to make it clear that it sets the response headers of the HTTP response.

## Checklist
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -170,9 +170,9 @@ type Options struct {
 	SigningKey          string `mapstructure:"signing_key" yaml:"signing_key,omitempty"`
 	SigningKeyAlgorithm string `mapstructure:"signing_key_algorithm" yaml:"signing_key_algorithm,omitempty"`
 
-	// Headers to set on all proxied requests. Add a 'disable' key map to turn off.
-	HeadersEnv string            `yaml:",omitempty"`
-	Headers    map[string]string `yaml:",omitempty"`
+	HeadersEnv string `yaml:",omitempty"`
+	// SetResponseHeaders to set on all proxied requests. Add a 'disable' key map to turn off.
+	SetResponseHeaders map[string]string `yaml:",omitempty"`
 
 	// List of JWT claims to insert as x-pomerium-claim-* headers on proxied requests
 	JWTClaimsHeaders JWTClaimHeaders `mapstructure:"jwt_claims_headers" yaml:"jwt_claims_headers,omitempty"`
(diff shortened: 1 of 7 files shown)


wasaga (approved):

pkg/grpc/config/config.proto
```diff
@@ -149,7 +149,7 @@ message Settings {
   optional string certificate_authority_file = 35;
   optional string signing_key = 36;
   optional string signing_key_algorithm = 62;
-  map<string, string> headers = 69;
+  map<string, string> set_response_headers = 69;
```
desimone, line 152: Just double checking that this is fine to rename?

config/options.go
```diff
@@ -445,12 +445,23 @@ func (o *Options) parseHeaders() error {
...
+	}
+
+	if o.viperIsSet("headers") {
+		log.Warn().Msg("config: headers has been renamed to set_response_headers and may be removed in the future")
+	}
+
+	// option was renamed from `headers` to `set_response_headers`. Both config settings are supported.
+	headerKeys := []string{"headers", "set_response_headers"}
+	for _, headerKey := range headerKeys {
+		if o.viperIsSet(headerKey) {
```
desimone, line 459: I think we could also add a `header!=nil` check here to avoid looping again if values are found for a given key since we would overwrite the value anyway?

desimone (approved):
LGTM. One small suggestion, and a question. 

I'm also thinking we should follow up with the ability to set route local response headers similar to how we do for [request headers](https://www.pomerium.io/reference/#set-request-headers) which I imagine is where most folks want to apply this.

pkg/grpc/config/config.proto
calebdoxsey: yes it should be. protobuf is purely by the field id unless we use reflection

config/options.go
calebdoxsey: Sorry I don't understand. What check do I need to add?

desimone: I was just thinking it could be possible to break early if `headers` was already set vs the current behavior would check both, and then overwrite whatever was in `headers` with `set_response_headers` which probably what we want anyway.

config/options.go
```diff
@@ -445,12 +445,23 @@ func (o *Options) parseHeaders() error {
...
-		o.Headers = headers
-	} else if o.viperIsSet("headers") {
-		if err := o.viper.UnmarshalKey("headers", &headers); err != nil {
-			return fmt.Errorf("header %s failed to parse: %w", o.viper.Get("headers"), err)
+		o.SetResponseHeaders = headers
+		return nil
+	}
+
+	if o.viperIsSet("headers") {
+		log.Warn().Msg("config: headers has been renamed to set_response_headers and may be removed in the future")
```
desimone, line 453:
```suggestion
		log.Warn().Msg("config: headers has been renamed to set_response_headers will be removed in v0.16")
```

desimone (approved):
LGTM, added a comment specify when we plan to deprecate, and I created FUP issue to remind us to do just that :)

PR #2092, opened 2021-04-15
config: don't change address value on databroker or authorize

### LINKED ISSUE #2091: Databroker fails to associate correct TLS cert with listener downstream context
## What happened?

When providing multiple certificates to databroker, the name resolution fails to locate the correct certificate by DNS name.  We fall back to the "default" cert, whatever that happens to be.

If a cert with a name which matches the databroker URL is present, we should use that in the downstream context.

## How'd it happen?

1. 
```bash
mkcert databroker.localhost.pomerium.io
mkcert proxy.localhost.pomerium.io
```

2. Configure `certificates` block:
```yaml
certificates:
  - cert: proxy.localhost.pomerium.io.pem
     key: proxy.localhost.pomerium.io-key.pem
  - cert: databroker.localhost.pomerium.io.pem
     key: databroker.localhost.pomerium.io-key.pem  
```
3. 
```bash
curl --cacert "$(mkcert -CAROOT)/rootCA.pem" https://databroker.localhost.pomerium.io
curl: (60) SSL: no alternative certificate subject name matches target host name 'databroker.localhost.pomerium.io'
More details here: https://curl.haxx.se/docs/sslcerts.html
```

re-ordering the cert list so that `databroker.localhost.pomerium.io` is first results in a successful curl.

- Pomerium version (retrieve with `pomerium --version`): `116805acb3b59ced3ea78ea86de38dc49b9eb612`
- Server Operating System/Architecture/Cloud: OSX

## What's your config.yaml?

```config.yaml
address: :443
services: databroker
databroker_service_url: https://databroker.localhost.pomerium.io
certificates:
  - cert: proxy.localhost.pomerium.io.pem
     key: proxy.localhost.pomerium.io-key.pem
  - cert: databroker.localhost.pomerium.io.pem
     key: databroker.localhost.pomerium.io-key.pem
[SNIP]
```

n/a

## Additional context

Upon further digging, it seems like we [consider options.GRPCAddr](https://github.com/pomerium/pomerium/blob/116805acb3b59ced3ea78ea86de38dc49b9eb612/config/envoyconfig/listeners.go#L713-L714) while looking up certificates.  `addr` isn't equal to `options.GRPCAddr` due to [this mutation](https://github.com/pomerium/pomerium/blob/116805acb3b59ced3ea78ea86de38dc49b9eb612/config/options.go#L544-L552).

I _think_ this code can just be removed - we don't create the HTTP listener unless we're proxy or authenticate, so the field is unused on instances of databroker or authorize.  Commenting it out does result in the right cert being loaded, since `options.Addr` is no longer being changed.

travisgroth:
## Summary

Removes a legacy mutation on the `address` parameter which breaks GRPC TLS lookup.  See #2091 for additional context.

## Related issues

Closes #2091 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -541,16 +541,6 @@ func (o *Options) Validate() error {
 		return errors.New("config: unknown databroker storage backend type")
 	}
 
-	if IsAuthorize(o.Services) || IsDataBroker(o.Services) {
-		// if authorize is set, we don't really need a http server
-		// but we'll still set one up incase the user wants to use
-		// the HTTP health check api
-		if o.Addr == o.GRPCAddr {
-			o.Addr = DefaultAlternativeAddr
-			log.Warn().Str("Addr", o.Addr).Str("GRPCAddr", o.Addr).Msg("config: default http handler changed")
-		}
-	}
-
 	if o.SharedKey == "" {
 		return errors.New("config: shared-key cannot be empty")
 	}


calebdoxsey (approved):
Not sure what this will do, but if it works I guess it's ok.

desimone (changes requested):
I believe this will break people's current all in one configuration.

travisgroth:
> I believe this will break people's current all in one configuration.

I think that's all handled up here https://github.com/pomerium/pomerium/blob/55783d3d7c223fdd3129fa7eb39e15cef46ae9fc/config/options.go#L514-L532 but I might be mistaken.  

The code being removed changes the HTTP port, which is different than how we behave in all-in-one.  In the latter, we alter the GRPC port to avoid conflict.

desimone (approved):

PR #2095, opened 2021-04-16
databroker: add options for maximum capacity

calebdoxsey:
## Summary
Add options for controlling a maximum capacity based on record type. Both the in-memory and redis versions have been updated.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/storage/redis/redis.go
```diff
@@ -182,13 +257,68 @@ func (backend *Backend) Put(ctx context.Context, record *databroker.Record) (err
...
+	if err != nil {
+		return err
+	}
+
+	// nothing to do if capacity isn't set
+	if options.Capacity == nil {
+		return nil
+	}
+
+	key := getRecordTypeChangesKey(recordType)
```
wasaga, line 280: please add a trace span here as it's called on every `Put()`

pkg/storage/redis/redis.go
```diff
@@ -182,13 +257,68 @@ func (backend *Backend) Put(ctx context.Context, record *databroker.Record) (err
 				Score:  float64(version),
 				Member: bs,
 			})
+			p.ZAdd(ctx, getRecordTypeChangesKey(record.GetType()), &redis.Z{
+				Score:  float64(record.GetModifiedAt().GetSeconds()),
```
wasaga, line 261: is there a reason to limit resolution to just seconds?

wasaga (approved):

pkg/storage/redis/redis.go
calebdoxsey: Because it was a simple method call. I guess we can use something else, though I don't know what the advantage would be.

wasaga: there might be multiple records arriving at the very same second

Issue #2100, opened 2021-04-19
Would it be possible to drop tun.cfg.dstHost from jwtCacheKey


bl0m1:
**Is your feature request related to a problem? Please describe.**
Running Ansible through Pomerium-cli causes hundreds of popus a day, See #2082 

**Describe the solution you'd like**

Currently the `tcptunnel` `jwtCacheKey` using both destination host, proxyhost and tlsconfig. Would it be possible to remove destination host and only have proxyhost here? We have tested the following patch with great success.

```patch
diff --git a/internal/tcptunnel/tcptunnel.go b/internal/tcptunnel/tcptunnel.go
index b4919f9..782d06b 100644
--- a/internal/tcptunnel/tcptunnel.go
+++ b/internal/tcptunnel/tcptunnel.go
@@ -214,7 +214,7 @@ func (tun *Tunnel) run(ctx context.Context, local io.ReadWriter, rawJWT string,
 }
 
 func (tun *Tunnel) jwtCacheKey() string {
-       return fmt.Sprintf("%s|%s|%v", tun.cfg.dstHost, tun.cfg.proxyHost, tun.cfg.tlsConfig != nil)
+       return fmt.Sprintf("%s|%v", tun.cfg.proxyHost, tun.cfg.tlsConfig != nil)
 }
 
 func deBuffer(br *bufio.Reader, underlying io.Reader) io.Reader {
```

This allows us to use ssh helper script https://www.pomerium.io/guides/tcp.html#setup with a `--pomerium-url` set without getting a popup a day for every destination host you would like to proxy traffic towards.

We have considered the security perspective and since the `tun.cfg.proxyHost` is used for connections we are unable to come up with any negative impact.

**Describe alternatives you've considered**

- Running a browser inside the users shell to try "mute" the popups for each new host accessed thru the proxy.

- We have also discussed using a "jumphost" to proxy ssh thru meaning we only allow pomerium to talk to our jumphost and the jumphost to be an additional proxy jump inside our end users ssh configs. However we do not like this singel point of failure and there is a few security concerns since you need to forward your ssh key etc.

**Explain any additional use-cases**

The users would only need one authorisation popup per work day.

**Additional context**

This together with the feature/bug discussed in #2082 creates a synergy and a really seamless experience for end users.

calebdoxsey:
Implemented with the above PR.

PR #2115, opened 2021-04-22
Drop tun.cfg.dstHost from jwtCacheKey

bl0m1:
## Summary

Remove tun.cfg.dstHost from jwtCacheKey. 
Users that uses an `--pomerium-url` does only need one browser popup per auth cycle (default 14 hours)

This is handy when running an Ansible playbook or similar tools which makes you to connect to multiple hosts and now get one browser popup per server per auth cycle.

Mor information can be found in the related issue.

## Related issues

Implements #2100 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/tcptunnel/tcptunnel.go
+++ b/internal/tcptunnel/tcptunnel.go
@@ -214,7 +214,7 @@ func (tun *Tunnel) run(ctx context.Context, local io.ReadWriter, rawJWT string, 
 }
 
 func (tun *Tunnel) jwtCacheKey() string {
-	return fmt.Sprintf("%s|%s|%v", tun.cfg.dstHost, tun.cfg.proxyHost, tun.cfg.tlsConfig != nil)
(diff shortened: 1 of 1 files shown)


calebdoxsey (approved):
Thanks @bl0m1 for the PR. Although I implemented much of this functionality I still have a hard time wrapping my head around it.

My first though was this wasn't going to work because you could have two different policies:

```yaml
- from: tcp+https://redis.corp.example.com:6379
  allowed_users: ["user1"]
- from: tcp+https://mysql.corp.example.com:3306
  allowed_users: ["user2"]
```

And this same JWT would end up being used for both which doesn't seem right. But on further consideration, I think it's ok, because the JWT stores identity, not policy decisions. Regardless of which users are allowed the user that logged in would be the same, so re-using the JWT should be ok. The request will still pass through authorize and be checked against the policy.

LGTM!

PR #2124, opened 2021-04-23
envoyconfig: fix metrics ingress listener name

calebdoxsey:
## Summary
It appears that envoy doesn't allow you to update a listener address via LDS, so instead we'll name the ingress with a hash of the address. The `xdsmgr` will then remove the previous one and add the new one, thus allowing us to change the metrics address at runtime.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/envoyconfig/listeners.go
+++ b/config/envoyconfig/listeners.go
@@ -25,6 +25,8 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
+	"github.com/pomerium/pomerium/internal/hashutil"
+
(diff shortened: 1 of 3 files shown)


config/envoyconfig/listeners.go
```diff
@@ -25,6 +25,8 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
+	"github.com/pomerium/pomerium/internal/hashutil"
```
desimone: nit: formatting

desimone (approved):

wasaga (approved):

PR #2132, opened 2021-04-26
use cached envoy

wasaga:
## Summary

Currently envoy is re-downloaded on every `make build`, this changes it to use whatever was previously cached. 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey (commented):
Where is `_dst` set?

calebdoxsey (approved):

PR #2145, opened 2021-04-27
do not default to grpc.WaitForReady(true)

wasaga:
## Summary

Prefer fail fast in case GRPC upstream cannot be reached, which is default GRPC behaviour. Pomerium code is mostly cyclic and that operation would be retried. When this option is present, in case GRPC upstream cannot be reached for whatever reason, under current GRPC configuration the call would be stuck indefinitely. 

## Related issues

I suspect it to be root case of https://github.com/pomerium/internal/issues/376
The behaviour is very similar to the one observed in https://github.com/pomerium/pomerium-console/issues/1016

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/grpc/client.go
+++ b/pkg/grpc/client.go
@@ -99,7 +99,6 @@ func NewGRPCClientConn(opts *Options, other ...grpc.DialOption) (*grpc.ClientCon
 	dialOptions := []grpc.DialOption{
 		grpc.WithChainUnaryInterceptor(unaryClientInterceptors...),
 		grpc.WithChainStreamInterceptor(streamClientInterceptors...),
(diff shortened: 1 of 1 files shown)


calebdoxsey (approved):

travisgroth (changes requested):
As discussed yesterday, I think `WaitForReady` is useful in user request flows to avoid errors during brief disruptions.  We'd still want authz and authn to be able to set this flag when communicating with the databroker rather than removing it entirely.

travisgroth (approved):

PR #2151, opened 2021-04-28
authorize: refactor store locking

calebdoxsey:
## Summary
It looks like the way we're doing locking with the OPA evaluator store is causing deadlocks. I misunderstood how the `RWMutex` works, and it turns out it can't reliably be used recursively:

> If a goroutine holds a RWMutex for reading and another goroutine might
> call Lock, no goroutine should expect to be able to acquire a read lock
> until the initial read lock is released. In particular, this prohibits
> recursive read locking. This is to ensure that the lock eventually becomes
> available; a blocked Lock call excludes new readers from acquiring the
> lock.

I believe the deadlock happens like this:

1. We perform an evaluation which takes a read-only transaction to the store `RLock`
2. During sync we update a record `Lock`
3. During policy evaluation we attempt to retrieve a record `RLock`

Step (2) is blocked on (1) and step (3) is blocked on (2). But step (1) is blocked on (3), because the policy evaluation can't be completed until we get the record. This PR removes taking a read lock in the transaction, and instead only uses the lock for map access (which is very short and non-recursive).

The original purpose of the transaction lock was to prevent an update while we were evaluating a policy. This is so that the version numbers we get back from evaluation (server version / record version) reflect the data that was actually used during evaluation. So to preserve this behavior I introduce another lock in the authorize service itself. The syncer's update and evaluator will never run simultaneously, and since they don't directly depend on each other, we should have no issues with deadlocks.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/sync.go
+++ b/authorize/sync.go
@@ -47,13 +47,17 @@ func (syncer *dataBrokerSyncer) GetDataBrokerServiceClient() databroker.DataBrok
 }
 
 func (syncer *dataBrokerSyncer) ClearRecords(ctx context.Context) {
+	syncer.authorize.stateLock.Lock()
 	syncer.authorize.store.ClearRecords()
+	syncer.authorize.stateLock.Unlock()
 }
 
 func (syncer *dataBrokerSyncer) UpdateRecords(ctx context.Context, serverVersion uint64, records []*databroker.Record) {
+	syncer.authorize.stateLock.Lock()
 	for _, record := range records {
 		syncer.authorize.store.UpdateRecord(serverVersion, record)
 	}
+	syncer.authorize.stateLock.Unlock()
 
 	// the first time we update records we signal the initial sync
 	syncer.signalOnce.Do(func() {
--- a/authorize/authorize.go
+++ b/authorize/authorize.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"html/template"
+	"sync"
 
 	"github.com/pomerium/pomerium/authorize/evaluator"
 	"github.com/pomerium/pomerium/config"
@@ -24,6 +25,11 @@ type Authorize struct {
 	templates      *template.Template
 
 	dataBrokerInitialSync chan struct{}
+
+	// The stateLock prevents updating the evaluator store simultaneously with an evaluation.
+	// This should provide a consistent view of the data at a given server/record version and
+	// avoid partial updates.
+	stateLock sync.RWMutex
 }
 
 // New validates and creates a new Authorize service from a set of config options.
--- a/authorize/evaluator/store.go
+++ b/authorize/evaluator/store.go
@@ -23,19 +23,75 @@ import (
 	"github.com/pomerium/pomerium/pkg/grpc/databroker"
 )
 
+type dataBrokerData struct {
+	mu sync.RWMutex
+	m  map[string]map[string]proto.Message
+}
+
+func newDataBrokerData() *dataBrokerData {
+	return &dataBrokerData{
+		m: map[string]map[string]proto.Message{},
+	}
+}
+
+func (dbd *dataBrokerData) clear() {
+	dbd.mu.Lock()
+	defer dbd.mu.Unlock()
+
+	dbd.m = map[string]map[string]proto.Message{}
+}
+
+func (dbd *dataBrokerData) delete(typeURL, id string) {
+	dbd.mu.Lock()
+	defer dbd.mu.Unlock()
+
+	m, ok := dbd.m[typeURL]
+	if !ok {
+		return
+	}
+
+	delete(m, id)
+
+	if len(m) == 0 {
+		delete(dbd.m, typeURL)
+	}
+}
+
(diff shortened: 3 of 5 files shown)


authorize/sync.go
```diff
@@ -47,13 +47,17 @@ func (syncer *dataBrokerSyncer) GetDataBrokerServiceClient() databroker.DataBrok
...
 	syncer.authorize.store.ClearRecords()
+	syncer.authorize.stateLock.Unlock()
 }
 
 func (syncer *dataBrokerSyncer) UpdateRecords(ctx context.Context, serverVersion uint64, records []*databroker.Record) {
+	syncer.authorize.stateLock.Lock()
 	for _, record := range records {
 		syncer.authorize.store.UpdateRecord(serverVersion, record)
 	}
+	syncer.authorize.stateLock.Unlock()
```
wasaga, line 60: why not expand the interface of `store` and have UpdateRecords method there - then you do not seem to need this extra `stateLock` mutex and you won't need acquire the other mutex inside `dataBrokerData` for every `UpdateRecord()` ?

calebdoxsey: If we remove the state lock there's no guarantee we won't upgrade records in the middle of an evaluation. You would then have an inconsistent view of the data. For example you would get record version 1234 in the audit log, but will have actually used data from version 1235 in the evaluation.

The mutex around map access is necessary because maps are not thread safe in Go. If one goroutine reads a map while another writes it it will cause the program to crash.

If having two locks in unacceptable what should I do instead to preserve thread safety and consistency?

wasaga: I missed that you also acquire RLock in `Check()`; I see why you do need that extra mutex now. 

my comment re acquiring mutex in a loop 
```go
for _, record := range records {
  syncer.authorize.store.UpdateRecord(serverVersion, record)
}
```

was about acquiring mutex on each `UpdateRecord` call, while you probably can change the interface of it to accept multiple records, and change underlying `dataBrokerData` funcs to `setLocked` and `deleteLocked` so that you could acquire mutex once outside the loop. but it's not too important probably in terms of optimization.

travisgroth:
Was running some load against this branch and got a panic after some time:


```
│ panic: runtime error: invalid memory address or nil pointer dereference                                                                                                                                                                                    │
│ [signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0xf62689]                                                                                                                                                                                    │
│ goroutine 70 [running]:                                                                                                                                                                                                                                    │
│ github.com/pomerium/pomerium/pkg/grpc/databroker.(*Syncer).sync(0xc00067c420, 0x1afaba0, 0xc004c0bc80, 0x0, 0x2)                                                                                                                                           │
│     /go/src/github.com/pomerium/pomerium/pkg/grpc/databroker/syncer.go:160 +0x369                                                                                                                                                                          │
│ github.com/pomerium/pomerium/pkg/grpc/databroker.(*Syncer).Run(0xc00067c420, 0x1afaba0, 0xc004c0bc80, 0xc004c759a0, 0x0)                                                                                                                                   │
│     /go/src/github.com/pomerium/pomerium/pkg/grpc/databroker/syncer.go:100 +0x297                                                                                                                                                                          │
│ github.com/pomerium/pomerium/authorize.(*Authorize).Run(0xc000cc0b80, 0x1afaba0, 0xc004c0bc00, 0x0, 0x0)                                                                                                                                                   │
│     /go/src/github.com/pomerium/pomerium/authorize/authorize.go:55 +0xc5                                                                                                                                                                                   │
│ github.com/pomerium/pomerium/internal/cmd/pomerium.Run.func4(0x0, 0x0)                                                                                                                                                                                     │
│     /go/src/github.com/pomerium/pomerium/internal/cmd/pomerium/pomerium.go:145 +0x3c                                                                                                                                                                       │
│ golang.org/x/sync/errgroup.(*Group).Go.func1(0xc004c7ae40, 0xc004c75980)                                                                                                                                                                                   │
│     /go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57 +0x59                                                                                                                                                         │
│ created by golang.org/x/sync/errgroup.(*Group).Go                                                                                                                                                                                                          │
│     /go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:54 +0x66
```

calebdoxsey:
Based on the line number, this is a log message and appears unrelated. I'll fix it though.

travisgroth (approved):
Haven't been able to reproduce the deadlock on this branch.  👍

LGTM unless there's other feedback from @wasaga.

authorize/sync.go
calebdoxsey: This violates separation of concerns. The reason I created this type was to avoid having to think about the lock from the store, since locks are easy to misuse. If you're really concerned about performance I can make the change, but I think its negligible. FWIW during sync the method is only called with a single record.

wasaga (approved):

PR #2157, opened 2021-04-30
config: add client_crl

### LINKED ISSUE #2104: mTLS client certificate revocation

The client certificate auth today does not check if the certificate that is presented is revoked before allowing access which poses a security issue as the certificate can be used after it has been revoked.

A way for pomerium to do a certificate revocation check for each client certificate that is presented, either through OCSP or CRL.

calebdoxsey:
## Summary
This PR adds new `client_crl` and `client_crl_file` options to support certificate revocation lists for client certificates.

## Related issues
Fixes #2104 

## Checklist
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/cryptutil/certificates.go
```diff
@@ -37,6 +38,24 @@ func CertificateFromFile(certFile, keyFile string) (*tls.Certificate, error) {
...
+		return nil, fmt.Errorf("failed to decode base64 crl: %w", err)
+	}
+	return x509.ParseCRL(bs)
+}
+
+// CRLFromFile parses a certificate revocation list from a file.
+func CRLFromFile(fileName string) (*pkix.CertificateList, error) {
+	bs, err := ioutil.ReadFile(fileName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read crl file (%s): %w", fileName, err)
```
desimone, line 54:
```suggestion
		return nil, fmt.Errorf("cryptutil: failed to read crl file (%s): %w", fileName, err)
```

pkg/cryptutil/certificates.go
```diff
@@ -37,6 +38,24 @@ func CertificateFromFile(certFile, keyFile string) (*tls.Certificate, error) {
 	return &cert, err
 }
 
+// CRLFromBase64 parses a certificate revocation list from a base64 encoded blob.
+func CRLFromBase64(rawCRL string) (*pkix.CertificateList, error) {
+	bs, err := base64.StdEncoding.DecodeString(rawCRL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode base64 crl: %w", err)
```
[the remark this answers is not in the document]
wasaga: is it standard to add package name to error text? I think each error should just adequately identify the line where it all happened, and it would be further annotated by the calling code.

pkg/cryptutil/certificates.go
```diff
@@ -37,6 +38,24 @@ func CertificateFromFile(certFile, keyFile string) (*tls.Certificate, error) {
...
+	}
+	return x509.ParseCRL(bs)
```
wasaga, line 47: _ParseCRL parses a CRL from the given bytes. It's often the case that PEM encoded CRLs will appear where they should be DER encoded, so this function will transparently handle PEM encoding as long as there isn't any leading garbage._

maybe we should explicitly decode PEM block and complain if it's not the right format. i.e. if we would allow uploading it from console.

calebdoxsey: I updated the function to only allow PEM encoded CRLs.

desimone (approved):
LGTM!

PR #1908, opened 2021-02-17
envoy: validate binary checksum

calebdoxsey:
## Summary
Validate that the embedded envoy binary matches the expected checksum of the one we downloaded. I added a new `get-envoy.bash` script which downloads envoy to `./bin/envoy` and updated `embed-envoy.bash` so that it uses that binary. The checksum of the binary is written to `./bin/envoy.sha256` which is then injected into the binary at `github.com/pomerium/pomerium/internal/envoy.Checksum`. The checksum is only checked if that variable is set, which means we only run this verification on a production release.

## Checklist
- [ ] reference any related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] add appropriate tag (`improvement` / `bug` / etc)
- [x] ready for review

PR #2160, opened 2021-04-30
internal/envoy: always extract envoy

travisgroth:
## Summary

We currently optimize for startup time and attempt to cache the envoy binary once we've extracted it on a system.  This creates starting security conditions that are hard to validate even with checksumming of the binary and atomic creation.

Changes:

- switch to a randomized subdirectory created on-demand, to help ensure we are running the same binary we checksum
- clean the entire base directory before extraction, so that we don't consume additional disk space every time we start on a system
- always extract a fresh copy of envoy

## Related issues

https://github.com/pomerium/pomerium/pull/1908

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/envoy/embed.go
+++ b/internal/envoy/embed.go
@@ -1,6 +1,7 @@
 package envoy
 
 import (
+	"context"
 	"fmt"
 	"io/fs"
 	"os"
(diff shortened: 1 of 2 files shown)


internal/envoy/embed.go
```diff
@@ -31,24 +35,23 @@ func extractEmbeddedEnvoy() (outPath string, err error) {
...
+	err = os.MkdirAll(embeddedFilesBaseDirectory, embeddedDirectoryPermissions)
+	if err != nil {
+		return "", fmt.Errorf("error creating embedded file directory: (directory=%s): %w", embeddedFilesBaseDirectory, err)
 	}
 
+	// build a random temp directory inside our base directory to guarantee permissions
+	tmpDir, err := os.MkdirTemp(embeddedFilesBaseDirectory, "envoy-")
+	outPath = filepath.Join(tmpDir, "envoy")
+
+	log.Info(ctx).Str("path", outPath).Msg("extracting envoy binary")
```
travisgroth, line 54: Just in case the operation takes time, let people know why we're "paused".

calebdoxsey (approved):

PR #2172, opened 2021-05-03
databroker: implement leases

calebdoxsey:
## Summary
This is a distributed mutex algorithm implementing using leases with TTLs. This allows multiple databrokers with only one identity manager loop.

The flow is:

1. acquire a lease
2. call the handler with a cancelable context
3. periodically renew the lease. If this fails, cancel the context.
4. If the parent context hasn't been canceled, try to acquire the lease

If the lease can't be acquired the `Leaser` will just keep retrying every `ttl/2`.

The backend implements leasing via a single method that can be called in 3 ways:

1. The initial lease is created by using a new UUID.
2. Renewal is implemented by re-using the UUID for the lease id. The TTL will then be updated.
3. Release is implemented by using a TTL of 0.

The redis implementation uses a watch+transaction to ensure atomicity.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


wasaga (approved):
looks great

desimone (approved):

pkg/grpc/databroker/leaser.go
```diff
@@ -0,0 +1,130 @@
...
+	defer retryTicker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-retryTicker.C:
+		}
+
+		res, err := locker.handler.GetDataBrokerServiceClient().AcquireLease(ctx, &AcquireLeaseRequest{
```
calebdoxsey, line 56: I think for the databroker calls we actually want to retry with backoff. We still need to kill the `RunLeased` ctx immediately, but getting disconnected from the databroker seems like something we should handle here. (as opposed to crashing out and relying on the process manager to restart)

PR #2190, opened 2021-05-11
Update programmatic-access.md

yyolk:
## Summary
tries to give a bit more context to the curl snippet to get a token for programmatic access

can further be improved, and open to suggestions from the maintainers.

## Related issues
none

## Checklist

- [X] updated docs
- [X] add appropriate tag (`improvement` / `bug` / etc)
- [x] ready for review

### DIFF
--- a/docs/docs/topics/programmatic-access.md
+++ b/docs/docs/topics/programmatic-access.md
@@ -15,15 +15,78 @@ This page describes how to obtain Pomerium access credentials programmatically v
 
 The API returns a cryptographically signed sign-in url that can be used to complete a user-driven login process with Pomerium and your identity provider. The login API endpoint takes a `pomerium_redirect_uri` query parameter as an argument which points to the location of the callback server to be called following a successful login.
 
-For example:
+Here's a full example. 
 
 ```bash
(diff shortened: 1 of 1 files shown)


yyolk:
I refactored my thoughts on this a bit and want to update this to be more accurate with when to use `redirect_uri` and want to note that the `MY_SERVICE_URL` should actually be verify.example.com, or anything fronted by pomerium.

I'll mark this as ready for review again and will submit my CLA along with my update

yyolk:
I have gone ahead and included some more context that I thought would be useful, including linking to the callback handler section on how to use the `pomerium_jwt`. I wanted to indicate `ANY_POMERIUM_PROXIED_SERVICE`, and include mentioning of the `/.pomerium/...` path if for some reason you land here and don't know anything about Pomerium (yet)

I also included an alternative that was best suited for my needs where I describe standing up a bastion host and forwarding the remote port from that bastion to localhost over a ssh tunnel. I am open to moving these instructions if there is a more appropriate place, but figured I'd include it as an alternative to the primary workflow described in programmatic_access.md.

desimone (approved):
Thank you @yyolk  -- this is great. And thanks for including the section on alternatives using a bastion-type setup. 

I've added a small formatting commit but the content is the same.

PR #2193, opened 2021-05-11
deployment: fix empty version on master builds

travisgroth:
## Summary

It looks like our metadata ldflags were getting overridden during `make build`, resulting in `v0.0.0` for the version in non-release builds.

## Related issues

n/a

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,8 @@ CTIMEVAR=-X $(PKG)/internal/version.GitCommit=$(GITCOMMIT) \
 	-X $(PKG)/internal/version.Version=$(VERSION) \
 	-X $(PKG)/internal/version.BuildMeta=$(BUILDMETA) \
 	-X $(PKG)/internal/version.ProjectName=$(NAME) \
-	-X $(PKG)/internal/version.ProjectURL=$(PKG)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,7 +24,8 @@ CTIMEVAR=-X $(PKG)/internal/version.GitCommit=$(GITCOMMIT) \
 	-X $(PKG)/internal/version.Version=$(VERSION) \
 	-X $(PKG)/internal/version.BuildMeta=$(BUILDMETA) \
 	-X $(PKG)/internal/version.ProjectName=$(NAME) \
-	-X $(PKG)/internal/version.ProjectURL=$(PKG)
+	-X $(PKG)/internal/version.ProjectURL=$(PKG) \
+	-X $(PKG)/internal/envoy.Checksum=$$(cat ./bin/envoy.sha256 | tr -d '\n')
```
desimone, line 28: Maybe we should add the envoy version to the FullVersion() build as well.... thoughts?

desimone (approved):

Makefile
travisgroth: That's probably not a bad idea

PR #2202, opened 2021-05-14
Pomerium Policy Language

calebdoxsey:
## Summary
Implement the Pomerium Policy Language. The Pomerium Policy Language is a DSL for constructing rego scripts. A subsequent PR will convert our existing policy into PPL and then into Rego.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/policy/rules/rules.go
+++ b/pkg/policy/rules/rules.go
@@ -1,0 +1,117 @@
+// Package rules contains useful pre-defined rego AST rules.
+package rules
+
+import "github.com/open-policy-agent/opa/ast"
+
+// GetSession the session for the given id.
+func GetSession() *ast.Rule {
+	return ast.MustParseRule(`
+get_session(id) = v {
+	v := get_databroker_record("type.googleapis.com/user.ServiceAccount", id)
+} else = v {
+	v := get_databroker_record("type.googleapis.com/session.Session", id)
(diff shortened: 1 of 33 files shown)


pkg/policy/rules/rules.go
```diff
@@ -0,0 +1,117 @@
+// Package rules contains useful pre-defined rego AST rules.
+package rules
+
+import "github.com/open-policy-agent/opa/ast"
+
```
wasaga, line 5: given that any of this functions may panic, and also is not really customizing any of the outputs, maybe a better approach would be to have those snippets in `var` section so that they'll all be evaluated on start. 

```go 
var (
    Session = ast.MustParseRule(`
get_session(id) = v {
	v := get_databroker_record("type.googleapis.com/user.ServiceAccount", id)
} else = v {
	v := get_databroker_record("type.googleapis.com/session.Session", id)
} else = v {
	v := {}
}
`)
)

wasaga (approved):

pkg/policy/rules/rules.go
calebdoxsey: We modify the rules sometimes, so this would result in inadvertent data sharing.

PR #2223, opened 2021-05-19
fix(docs): use correct name for code-server

### LINKED ISSUE #2222: Fix docs to use code-server
I noticed the docs call it "VS Code Server" but the correct name is "code-server."

jsjoeio:
## Summary

Updates documentation to use the correct name for code-server

## Related issues

Fixes #2222 

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review
- [x] update [config](https://github.com/pomerium/pomerium/blob/master/docs/.vuepress/config.js#L170)
- [x] add [redirect](https://github.com/pomerium/pomerium/blob/master/docs/.vuepress/public/_redirects#L33)

### DIFF
--- a/docs/guides/code-server.md
+++ b/docs/guides/code-server.md
@@ -1,5 +1,5 @@
 ---
-title: VS Code Server
+title: code-server
 lang: en-US
 meta:
   - name: keywords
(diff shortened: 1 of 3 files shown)


desimone (changes requested):
Thanks for the PR @jsjoeio  -- Could you also update the links?

jsjoeio:
Updated! Let me know if I missed any.

docs/guides/code-server.md
```diff
@@ -124,4 +124,4 @@ Happy remote hacking!!!😁
 [quick start]: ../docs/quick-start
 [synology nas]: ../docs/quick-start/synology.md
 [visual studio code]: https://code.visualstudio.com/
-[visual studio code server]: https://github.com/cdr/code-server
```
desimone, line 127: I believe this needs a dash.

desimone (changes requested):
One small issue. 

Also, I just saw that you are a maintainer of code-server. It has been a while since I tried this guide with the software. Have you tested it? If so, would you recommend any other general changes?

Also, let me know if we can be helpful in being part of the auth story for code. I think the assertion header + code could be a great way to pass along user identity similar to how we work with grafana.

docs/guides/code-server.md
jsjoeio: Whoops! Missed that. Updated just now

jsjoeio:
> Also, I just saw that you are a maintainer of code-server. It has been a while since I tried this guide with the software. Have you tested it? 

I have not. The first time I heard about it was this week when a community member mentioned it [here](https://github.com/cdr/code-server/issues/2636#issuecomment-844109621). I can try and find some time to test it out soon though.  

> If so, would you recommend any other general changes?

Nothing specifically but let me see if @bpmct has any feedback. 

> Also, let me know if we can be helpful in being part of the auth story for code

Thanks for offering that! We really appreciate it ♥️

bpmct:
Thanks for looping me in @jsjoeio. I can also test this approach, as I know our architecture has changed a bit from when the initial guide was created. I can also record a short video on how it works and share on our YouTube :)

jsjoeio:
Updated the config and added a redirect (because I changed the file name)

desimone (approved):
LGTM! Thanks @jsjoeio !

PR #2224, opened 2021-05-19
ppl: refactor authorize to evaluate PPL

calebdoxsey:
## Summary
Update the authorize evaluator to use PPL instead of `authz.rego`.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


authorize/grpc.go
```diff
@@ -58,26 +57,29 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v3.CheckRe
...
 
-	switch {
-	case reply.Status == http.StatusOK:
-		return a.okResponse(reply), nil
-	case reply.Status == http.StatusUnauthorized:
-		if isForwardAuth && hreq.URL.Path == "/verify" {
-			return a.deniedResponse(ctx, in, http.StatusUnauthorized, "Unauthenticated", nil)
-		}
-		return a.redirectResponse(ctx, in)
+	if res.Deny != nil {
```
desimone, line 70: This is easier to reason about

authorize/check_response.go
```diff
@@ -25,15 +26,15 @@ import (
 
 func (a *Authorize) okResponse(reply *evaluator.Result) *envoy_service_auth_v3.CheckResponse {
 	var requestHeaders []*envoy_config_core_v3.HeaderValueOption
-	for k, v := range reply.Headers {
-		requestHeaders = append(requestHeaders, mkHeader(k, v, false))
+	for k, vs := range reply.Headers {
+		requestHeaders = append(requestHeaders, mkHeader(k, strings.Join(vs, ","), false))
```
desimone, line 30: Should we also sort the header key values to ensure determinism?

authorize/check_response.go
```diff
@@ -25,15 +26,15 @@ import (
...
 	}
 	// ensure request headers are sorted by key for deterministic output
 	sort.Slice(requestHeaders, func(i, j int) bool {
 		return requestHeaders[i].Header.Key < requestHeaders[j].Header.Value
 	})
 	return &envoy_service_auth_v3.CheckResponse{
-		Status: &status.Status{Code: int32(codes.OK), Message: reply.Message},
+		Status: &status.Status{Code: int32(codes.OK), Message: "OK"},
```
desimone, line 37: I wonder if this is where it would be nice to reply with "AllowReason" or something similar.

authorize/evaluator/policy_evaluator.go
```diff
@@ -0,0 +1,177 @@
...
+			)
+			q, err = r.PrepareForEval(ctx)
+		}
+		if err != nil {
+			return nil, err
+		}
+		e.queries = append(e.queries, q)
+	}
+
+	return e, nil
```
desimone, line 105: It might be useful to add a debug log statement here which prints out the entirety of the configured policy evaluator.

authorize/evaluator/policy_evaluator.go
```diff
@@ -0,0 +1,177 @@
...
+			rego.Store(store),
+			rego.Module("pomerium.policy", script),
+			rego.Query("result = data.pomerium.policy"),
+			getGoogleCloudServerlessHeadersRegoOption,
+			store.GetDataBrokerRecordOption(),
+		)
+
+		q, err := r.PrepareForEval(ctx)
+		// if no package is in the src, add it
+		if err != nil && strings.Contains(err.Error(), "package expected") {
```
desimone, line 89: This is a good catch.

authorize/grpc.go
```diff
@@ -134,38 +136,12 @@ func (a *Authorize) getEvaluatorRequestFromCheckRequest(
...
 	if p != nil {
-		for _, sp := range p.SubPolicies {
-			req.CustomPolicies = append(req.CustomPolicies, sp.Rego...)
-		}
-	}
-
-	ca, err := a.getDownstreamClientCA(p)
-	if err != nil {
-		return nil, err
+		req.Policy = p
```
desimone, line 139: if p is not nil, what would policy be? Could we instead do `req.Policy = a.getMatchingPolicy(requestURL)` above?

desimone (commented):
Looks great. A few questions and comments. 

It looks like half of my review comments were sent as vanilla comments. Sorry about that.

authorize/check_response.go
calebdoxsey: we already sort the headers after this loop.

authorize/check_response.go
calebdoxsey: We would need to refactor the PPL code to bubble up allow reasons instead of just true.

authorize/grpc.go
```diff
@@ -58,26 +57,29 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v3.CheckRe
...
+	if res.Allow {
+		return a.okResponse(res), nil
 	}
-	return a.deniedResponse(ctx, in, int32(reply.Status), reply.Message, nil)
+
+	if isForwardAuth && hreq.URL.Path == "/verify" {
+		return a.deniedResponse(ctx, in, http.StatusUnauthorized, "Unauthenticated", nil)
+	}
+
+	return a.redirectResponse(ctx, in)
```
calebdoxsey, line 82: I think this isn't right. We need to make sure we only redirect if the user doesn't have a session.

authorize/evaluator/headers_evaluator.go
```diff
@@ -0,0 +1,112 @@
...
+	return &HeadersResponse{
+		Headers: e.getHeader(rs[0].Bindings),
+	}, nil
+}
+
+func (e *HeadersEvaluator) getHeader(vars rego.Vars) http.Header {
+	h := make(http.Header)
+
+	m, ok := vars["result"].(map[string]interface{})
+	if !ok {
```
wasaga, line 93: shouldn't this an others below be an error that should bubble up from `Evaluate`?

authorize/evaluator/policy_evaluator.go
```diff
@@ -0,0 +1,183 @@
...
+
+// Merge merges another PolicyResponse into this PolicyResponse. Access is allowed if either is allowed. Access is denied if
+// either is denied. (and denials take precedence)
+func (res *PolicyResponse) Merge(other *PolicyResponse) *PolicyResponse {
+	merged := &PolicyResponse{
+		Allow: res.Allow || other.Allow,
+		Deny:  res.Deny,
+	}
+	if other.Deny != nil {
+		merged.Deny = other.Deny
```
wasaga, line 37: should it be only if other.Deny is false?

wasaga (approved):

authorize/check_response.go
desimone: 👍

authorize/evaluator/evaluator.go
```diff
@@ -8,132 +7,178 @@ import (
...
 	"github.com/pomerium/pomerium/pkg/cryptutil"
 )
 
-// Evaluator specifies the interface for a policy engine.
+// notFoundOutput is what's returned if a route isn't found for a policy.
+var notFoundOutput = &Result{
+	Allow: false,
+	Deny: &Denial{
+		Status:  http.StatusNotFound,
+		Message: "route not found",
```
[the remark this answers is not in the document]
calebdoxsey: this is the error if there is no route

authorize/evaluator/headers_evaluator.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// A HeadersEvaluator evaluates the headers.rego script.
+type HeadersEvaluator struct {
+	q rego.PreparedEvalQuery
+}
+
+// NewHeadersEvaluator creates a new HeadersEvaluator.
+func NewHeadersEvaluator(ctx context.Context, store *Store) (*HeadersEvaluator, error) {
+	headersSrc, err := opa.FS.ReadFile("policy/headers.rego")
```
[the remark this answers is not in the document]
calebdoxsey: We are using go:embed. What would you like me to use instead?

authorize/evaluator/policy_evaluator.go
calebdoxsey: Deny is not a boolean. It includes an http status code and reason. A non-nil deny means the request should be denied.

authorize/evaluator/headers_evaluator.go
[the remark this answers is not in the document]
wasaga: go:embed has two modes - one makes a filesystem, another directly puts data into byte array compile time. https://golang.org/pkg/embed/#hdr-Strings_and_Bytes

authorize/evaluator/evaluator.go
```diff
@@ -16,124 +15,171 @@ import (
...
-	e := &Evaluator{
-		custom:   NewCustomEvaluator(store),
-		policies: options.GetAllPolicies(),
-		store:    store,
-	}
-	jwk, err := getJWK(options)
-	if err != nil {
-		return nil, fmt.Errorf("authorize: couldn't create signer: %w", err)
-	}
+func New(ctx context.Context, store *Store, options *config.Options) (*Evaluator, error) {
```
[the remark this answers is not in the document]
calebdoxsey: There are now evaluator-specific options.

authorize/evaluator/headers_evaluator.go
calebdoxsey: I don't think so. I can make this change if you want but I suspect we'll have to roll it back. There may be valid reasons why no `allow` or `deny` block occurs in the result.

PR #2235, opened 2021-05-24
envoyconfig: fallback to global custom ca when no policy ca is defined

calebdoxsey:
## Summary
Fallback to the custom ca defined in the global options when no policy ca is defined. 

There's a non-trivial chance this may break existing users, since previously we only validated internal connections between pomerium services using this certificate authority, and now we will be validating all upstream services.

## Related issues
Fixes https://github.com/pomerium/internal/issues/417

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/envoyconfig/clusters.go
+++ b/config/envoyconfig/clusters.go
@@ -118,7 +118,7 @@ func (b *Builder) buildPolicyCluster(ctx context.Context, options *config.Option
 	cluster.AltStatName = getClusterStatsName(policy)
 
 	name := getClusterID(policy)
-	endpoints, err := b.buildPolicyEndpoints(ctx, policy)
+	endpoints, err := b.buildPolicyEndpoints(ctx, options, policy)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


wasaga (approved):
The code LGTM.

- Please add an explicit warning log entry during that route X will have global CA applied, so that users who run into upstream unavailability could figure out why without reading docs.

travisgroth (changes requested):
Discussed with @desimone yesterday and augmenting the system cert pool rather than replacing it makes the most sense for this option.  That's how the internal golang grpc and http clients currently function.  

Doing something like https://github.com/pomerium/pomerium/blob/a1061c5c03705aa69333540fad2792acb916b53c/pkg/cryptutil/tls.go#L17 for building the envoy upstream tls context would be the desired behavior.

calebdoxsey:
We can't use that function. It has no way of serializing back to PEM.

I guess I'll concatenate the PEM files, though I don't know enough about the format to know if that actually works or not.

travisgroth (approved):

PR #2240, opened 2021-05-26
envoy: exit if envoy exits

calebdoxsey:
## Summary
If envoy exits unexpectedly, we should use `log.Fatal` to exit as well. This PR also fixes a bug with hot-restart. The `restart-epoch` was supposed to start at 1 for the first restart, not 0.

## Related issues
Fixes https://github.com/pomerium/internal/issues/414

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/envoy/envoy.go
+++ b/internal/envoy/envoy.go
@@ -15,7 +15,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
-	"runtime"
 	"strconv"
 	"strings"
 	"sync"
@@ -28,13 +27,12 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/natefinch/atomic"
 	"github.com/rs/zerolog"
-	"go.opencensus.io/stats/view"
+	"github.com/shirou/gopsutil/process"
 	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/config/envoyconfig"
 	"github.com/pomerium/pomerium/internal/log"
-	"github.com/pomerium/pomerium/internal/telemetry/metrics"
 	"github.com/pomerium/pomerium/internal/telemetry/trace"
 )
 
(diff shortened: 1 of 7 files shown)


internal/envoy/envoy.go
```diff
@@ -213,6 +218,13 @@ func (srv *Server) run(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("error starting envoy: %w", err)
 	}
+	// call Wait to avoid zombie processes
+	go func() { _ = cmd.Wait() }()
```
wasaga, line 222: I'm not sure why `monitorProcessCancel` has to be inside `Server` - you can just wait for the process completion in this goroutine, and invoke `monitorProcessCancel` there.

wasaga (approved):
lgtm; 

I'd remove `monitorProcessCancel` from `Server` but it's a style preference.

internal/envoy/envoy.go
calebdoxsey: When we update envoy config we start a new envoy process using hot-reload. Once hot-reload completes it will kill the previous envoy process. If we don't cancel the previous process monitor, we will immediately exit when this happens. 
We have to cancel the process monitor **before** the process exits.

internal/envoy/envoy.go
```diff
@@ -186,10 +191,10 @@ func (srv *Server) run(ctx context.Context) error {
 
 	if baseID, ok := readBaseID(); ok {
 		args = append(args, "--base-id", strconv.Itoa(baseID), "--restart-epoch", strconv.Itoa(srv.restartEpoch))
-		srv.restartEpoch++ // start with epoch zero when we're a fresh pomerium process
 	} else {
 		args = append(args, "--use-dynamic-base-id", "--base-id-path", baseIDPath)
 	}
+	srv.restartEpoch++
 
 	cmd := exec.Command(srv.envoyPath, args...) // #nosec
```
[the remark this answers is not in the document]
calebdoxsey: We want the background context. The context passed to run comes from the on change handler. But envoy needs to survive past the end of this function.

PR #2246, opened 2021-05-28
darwin: use x86 envoy build for arm64

calebdoxsey:
## Summary
We don't currently have a build of envoy for darwin/arm64 (ie m1), but we can run the x86 build using rosetta. This updates the envoy script and run command.

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/.github/goreleaser.yaml
+++ b/.github/goreleaser.yaml
@@ -26,9 +26,6 @@ builds:
     goos:
       - linux
       - darwin
-    ignore:
-      - goos: darwin
-        goarch: arm64
 
     ldflags:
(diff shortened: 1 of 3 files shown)


travisgroth (commented):
@calebdoxsey can you update goreleaser config too?  I think we're blacklisting darwin/arm right now. 

https://github.com/pomerium/pomerium/blob/master/.github/goreleaser.yaml#L29-L31

desimone (approved):

PR #2261, opened 2021-06-01
envoy: always set jwt claim headers even if no value is available

calebdoxsey:
## Summary
If no value was defined for a given claim it would be possible for an incoming request to set a value and have it still go through. This makes it so we always set a header even if there is no value, thus overwriting the original request header.

The reason why we only check `PassIdentityHeaders` and don't remove the headers there is that the remove header code is called *after* the extauthz code, which means it removes the headers we added with rego.

## Related issues
Fixes https://github.com/pomerium/internal/issues/423
Fixes https://github.com/pomerium/internal/issues/422

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


wasaga (approved):
it would be great to have a unit test to avoid regressions in the future.

PR #2267, opened 2021-06-03
redis: enforce capacity via ZREVRANGE to avoid race

calebdoxsey:
## Summary
Changes the way we enforce capacity in redis. There were 3 issues previously:

1. Because we used `uint64`s, the check `removeCnt <= 0` didn't make sense. It often resulted in a very large number.
2. If two databrokers simultaneously enforced options, there was a race on `ZCARD` vs `ZPOPMIN`. 
3. We were adding the removed record back to the same list of records.

By using `ZREVRANGE` we can find the excess records and remove them one at a time. If two databrokers attempt to enforce the options simultaneously we may end up removing the same record twice, but that should be ok.

## Related issues
Fixes https://github.com/pomerium/internal/issues/428

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/storage/redis/redis.go
+++ b/pkg/storage/redis/redis.go
@@ -312,14 +312,14 @@ func (backend *Backend) put(ctx context.Context, record *databroker.Record) erro
 				p.HDel(ctx, key, field)
 			} else {
 				p.HSet(ctx, key, field, bs)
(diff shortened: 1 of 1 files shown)


wasaga (approved):
it would be great to have a unit test with multiple databrokers if that's not too much of a trouble

Issue #1052, opened 2020-07-09
docs: google service account ui changed


desimone:
## What happened?

Google changed their service account process. 

## What did you expect to happen?

That's a loaded question :) 

## Additional context

![image (3)](https://user-images.githubusercontent.com/1544881/86998101-84f8a700-c164-11ea-84a5-b8a6c541f2d2.png)
![image (2)](https://user-images.githubusercontent.com/1544881/86998103-8629d400-c164-11ea-9baa-3d8736ee4be6.png)
![image (1)](https://user-images.githubusercontent.com/1544881/86998106-86c26a80-c164-11ea-8fb2-6dcbca347184.png)
![image](https://user-images.githubusercontent.com/1544881/86998108-875b0100-c164-11ea-8348-3bdef41f47ae.png)

desimone:
Thank you @shreyaskarnik for reporting

PR #2272, opened 2021-06-06
docs:  google gcp / workspace instructions

desimone:
## Summary

Updated the google guide to be much more deferential to Google's own documentation as their screens have been changing rather constantly. Tried to be more explicit about the structure of the steps. Overall, there's not at ton we can do since the process is split between GCP based OAuth2 credential and service account creation, manually adding an impersonated user for the SA, then adding permissions from GSuite to that domain wide service account. Despite the updates on Google's side, these settings are just as hard to find as before. 

## Related issues

- Related to #1052 
- Related to https://github.com/pomerium/pomerium/pull/1738

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone:
For preview: https://deploy-preview-2272--pomerium-docs.netlify.app/docs/identity-providers/google.html

docs/docs/identity-providers/google.md
```diff
@@ -54,26 +93,34 @@ Then, you'll need to manually open an editor and add an `impersonate_user` field
...
-Next we'll delegate G-suite group membership access to the service account we just created .
+Next, we need to give that service account permissionson the GSuite / Workspace side of the house.
 
-1. Go to your G Suite domain's [Admin console](http://admin.google.com/).
-2. Select **Security** from the list of controls. If you don't see Security listed, select More controls 1\. from the gray bar at the bottom of the page, then select Security from the list of controls.
-3. Select **Advanced settings** from the list of options.
-4. Select **Manage API client** access in the Authentication section.
-5. In the **Client name** field enter the service account's **Client ID**.
-6. In the **One or More API Scopes** field enter the following list of scopes: `https://www.googleapis.com/auth/admin.directory.group.readonly` `https://www.googleapis.com/auth/admin.directory.user.readonly`
+1. From your Google Workspace domain's [Admin console](http://admin.google.com/), go to **Main menu** menu**> Security > API controls**.
```
calebdoxsey, line 115: formatting looks off in the preview. Maybe a duplicate `menu**`?

calebdoxsey (changes requested):

wasaga:
They have command line equivalents for every UI, maybe we should just use them?

desimone:
> They have command line equivalents for every UI, maybe we should just use them?

I think we continue with this format because:

1. It's consistent with our other guides
2. The barrier for entry is lower
3. I'm not sure all of these settings (especially the bizzar-o gsuite scopes) are available. 
4. I'm unconvinced the API churn is any less than the docs.

docs/docs/identity-providers/google.md
desimone: Thank you, good catch. I think I fixed it.

desimone: Thank you, good catch. I think I fixed it.

calebdoxsey (approved):

PR #2280, opened 2021-06-09
Currently there is no way use device state in policies

redbaron:
## Summary

I believe current readme is misleading as device state is not supported in policies.

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone:
Fair point -- I'll update the readme with more explicit coverage of the device state characteristics.  Thanks @redbaron  (It's coming, soon 😄 !)

PR #2282, opened 2021-06-09
envoy: use go:embed

calebdoxsey:
## Summary
Use go embed instead of attaching envoy via zip.

## Related issues
Fixes https://github.com/pomerium/internal/issues/435

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


scripts/get-envoy.bash
```diff
@@ -42,6 +42,7 @@ if [ -f "$_dir/envoy" ]; then
 fi
 
 mkdir -p "$_dir"
-curl -L --compressed -o "$_dir/envoy" "https://github.com/pomerium/envoy-binaries/releases/download/v${_envoy_version}/envoy-${_target}"
+curl -L --compressed -o "$_dir/envoy" \
+    "https://github.com/pomerium/envoy-binaries/releases/download/v${_envoy_version}/envoy-${_target}"
 
 hash_sha256 "$_dir/envoy" >"$_dir/envoy.sha256"
```
wasaga, line 48: use sha from the remote repo

scripts/get-envoy.bash
```diff
@@ -42,6 +42,7 @@ if [ -f "$_dir/envoy" ]; then
 fi
 
 mkdir -p "$_dir"
-curl -L --compressed -o "$_dir/envoy" "https://github.com/pomerium/envoy-binaries/releases/download/v${_envoy_version}/envoy-${_target}"
+curl -L --compressed -o "$_dir/envoy" \
```
wasaga, line 45: can we please do arch-platform the way console does? 
when developing on mac I need switch between local and docker environments and it helps to avoid errors like wrong binary being embedded

wasaga (changes requested):

scripts/get-envoy.bash
calebdoxsey: There is no sha in the remote repo.

scripts/get-envoy.bash
calebdoxsey: I don't know what that means. If you want a linux binary set `GOOS=linux`.

wasaga, line 45:
```suggestion
curl -L --compressed -o "$_dir/envoy-${_target}" \
```

wasaga: I mean save target as `"$_dir/envoy-${_target}"`

travisgroth: Yes, I think the binaries need to retain their platform name upon download so they don't overlap during build.  Previously we were able to put them in the build directory so it wasn't an issue, but embed requires the file to be inline with the code.  Since goreleaser builds the platforms concurrently, it can wind up picking up the wrong binary.  

Concretely, I tried a release snapshot and I'm getting an execv error for envoy when I build on OSX and then run the linux binary:

`
{"level":"error","error":"error starting envoy: fork/exec /tmp/.pomerium-envoy3444331264/envoy: exec format error","service":"envoy","time":"2021-06-10T12:00:51Z","message":"envoy: failed to run envoy process"}
`

I think the name overlap is the cause.

travisgroth (changes requested):
See comment about envoy binary names.  It looks like building cross platform can pick up the wrong envoy binary arch.

scripts/get-envoy.bash
calebdoxsey: that won't work because the extraction will fail. I updated the script to use an etag check to see if content was modified.

calebdoxsey:
:man_shrugging: I implemented what the ticket asked for. I mostly copied this code verbatim from the other repo. I'll close this and move that issue back to the backlog as clearly I didn't understand what was being asked for.

PR #2297, opened 2021-06-16
context propagation

wasaga:
## Summary

Cleaned up most of `context.TODO` calls. 

- a function is performing some kind of fallback, and it emits a log statement about it. 
- a function is performing network or 3rd party calls that use context 
- changed error handling to be more explicit in few places, primarily while constructing envoy objects

## Related issues

Implements https://github.com/pomerium/internal/issues/443

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey (commented):
Use of context on `New` functions can be problematic. There are subtle bugs from inadvertently using the wrong context. (For example you start a long running goroutine, pass the context, and everything appears fine, but then you add a cancel to the context and suddenly it breaks)

In other words, does the context last the lifetime of the object from `New` to `Close`, or just the `New` function?

It's ambiguous and I suspect we use both. It will take me some time to confirm we aren't misusing them.

internal/directory/provider.go
```diff
@@ -68,101 +76,66 @@ func GetProvider(options Options) (provider Provider) {
 	switch options.Provider {
 	case auth0.Name:
 		serviceAccount, err := auth0.ParseServiceAccount(options)
-		if err == nil {
-			return auth0.New(
-				auth0.WithDomain(options.ProviderURL),
-				auth0.WithServiceAccount(serviceAccount))
+		if err != nil {
```
calebdoxsey, line 79: So we now require the use of a directory provider? This is a regression in functionality.

wasaga:
Many New functions indeed spawn background goroutines.

I believe in that case, context should be used for the entire lifecycle of the object created.

Unfortunately we do indeed have objects that reload themselves that may complicate things. 

There's no rush with this PR though. It indeed became rather large. Maybe it need be broken down into few smaller ones.

config/envoyconfig/listeners.go
```diff
@@ -652,15 +659,18 @@ func (b *Builder) buildDownstreamTLSContext(ctx context.Context,
 		alpnProtocols = []string{"h2", "http/1.1"}
 	}
 
-	envoyCert := b.envoyTLSCertificateFromGoTLSCertificate(ctx, cert)
+	envoyCert, err := b.envoyTLSCertificateFromGoTLSCertificate(ctx, cert)
+	if err != nil {
+
```
calebdoxsey, line 664: empty if

config/envoyconfig/listeners.go
```diff
@@ -629,17 +638,15 @@ func (b *Builder) buildRouteConfiguration(name string, virtualHosts []*envoy_con
...
-		log.Warn(ctx).Str("domain", domain).Err(err).Msg("failed to get all certificates from config")
-		return nil
+		return nil, fmt.Errorf("get all certificates from config: %w", err)
 	}
 
 	cert, err := cryptutil.GetCertificateForDomain(certs, domain)
 	if err != nil {
-		log.Warn(ctx).Str("domain", domain).Err(err).Msg("failed to get certificate for domain")
-		return nil
+		return nil, fmt.Errorf("failed to get certificate for domain: %w", err)
```
calebdoxsey, line 649: This makes the code more restrictive. It may break for existing users.

pkg/storage/redis/redis.go
```diff
@@ -69,8 +69,7 @@ type Backend struct {
 }
 
 // New creates a new redis storage backend.
-func New(rawURL string, options ...Option) (*Backend, error) {
-	ctx := context.TODO()
+func New(ctx context.Context, rawURL string, options ...Option) (*Backend, error) {
```
calebdoxsey, line 72: This is causing `ctx` to be persisted. I believe since we call this from `OnConfigChange` this will result in hanging on to changes inadvertently. We need to use a context that persists across the lifetime of the databroker.

internal/directory/provider.go
wasaga: no I just rewrote the upper level function that if there's an error then `provider = nullProvider{}` and removed the defer() logic of setting values

wasaga:
we will only merge this post-GA.

config/envoyconfig/listeners.go
wasaga: currently it never happens because of 

https://github.com/pomerium/pomerium/blob/9bce8314ba7935df46e7ca4a3a1a958498cc61e3/pkg/cryptutil/tls.go#L58-L61

which I believe is not a correct behaviour, as it anyway results in unusable endpoint. if we would change it, it would just return 404 because cluster would not be created.

calebdoxsey (approved):

PR #2316, opened 2021-06-28
cli: use proxy from environment

tskinn:
## Summary

The existing implementation of pomerium-cli does not respect the http proxy env vars so here we add the default http.Transport behavior back with [ProxyFromEnvironment](https://golang.org/pkg/net/http/#ProxyFromEnvironment).

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/authclient/authclient.go
+++ b/internal/authclient/authclient.go
@@ -119,9 +119,9 @@ func (client *AuthClient) runOpenBrowser(ctx context.Context, li net.Listener, s
 		return err
 	}
 
-	transport := &http.Transport{
-		TLSClientConfig: client.cfg.tlsConfig,
-	}
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.TLSClientConfig = client.cfg.tlsConfig
+
 	hc := &http.Client{
 		Transport: transport,
 	}


tskinn:
Hey @calebdoxsey what needs to happen to get this PR approved? Pomerium is an awesome project that we would love to use but not being able to use a proxy is a deal breaker for us.

internal/authclient/authclient.go
```diff
@@ -120,6 +120,7 @@ func (client *AuthClient) runOpenBrowser(ctx context.Context, li net.Listener, s
 	}
 
 	transport := &http.Transport{
+		Proxy:           http.ProxyFromEnvironment,
```
desimone, line 123: This makes sense to me. I wonder if a broader fix would be to use a Clone'd version of the default http transport `http.DefaultTransport` which also has some likely sane default timeouts and settings beyond Proxy?

https://golang.org/pkg/net/http/#Transport.Clone

desimone:
@tskinn -- This makes sense to me. Had one suggestion, but otherwise I think we can move forward.

internal/authclient/authclient.go
tskinn: Yeah that sounds good to me. You mean something like this right?

```
transport := http.DefaultTransport.Clone()
transport.TLSClientConfig = client.cfg.tlsConfig
```

desimone: Yep!

desimone (approved):

desimone:
Thanks @tskinn !

PR #2338, opened 2021-07-06
authorize: support boolean deny results

calebdoxsey:
## Summary
PPL rules generally create simple boolean values. For example `user: x` returns true if the user is `x`. Historically our policy rego script used an array with `[status, reason]` for the deny block. This PR updates the policy evaluator so that it supports either. If the deny block returns true it will set `[403,""]`.

## Related issues
Fixes https://github.com/pomerium/pomerium-console/issues/1473

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/evaluator/policy_evaluator.go
+++ b/authorize/evaluator/policy_evaluator.go
@@ -3,6 +3,7 @@ package evaluator
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


authorize/evaluator/policy_evaluator.go
```diff
@@ -164,20 +165,30 @@ func (e *PolicyEvaluator) getDeny(ctx context.Context, vars rego.Vars) *Denial {
...
+	case bool:
+		if t {
+			status = http.StatusForbidden
+			reason = ""
+		} else {
+			return nil
+		}
+	case []interface{}:
+		var err error
+		status, err = strconv.Atoi(fmt.Sprint(t[0]))
```
wasaga, line 180: prevent panic in case of empty array

wasaga (approved):

PR #1489, opened 2020-10-05
hadolint Dockerfile

stephengroat:
## Summary

start using [`hadolint`](https://github.com/hadolint/hadolint) on `Dockerfile`s to improve syntax and performance.

## Related issues

N/A

**Checklist**:
- [ ] add related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

PR #2363, opened 2021-07-13
Hadolint

stephengroat:
## Summary

adds hadolint to dockerfile builds

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,10 @@
+# hadolint ignore=DL3007
 FROM golang:latest as build
 WORKDIR /go/src/github.com/pomerium/pomerium
 
+# hadolint ignore=DL3008
 RUN apt-get update \
-    && apt-get -y install zip
+    && apt-get -y --no-install-recommends install zip
 
 # cache depedency downloads
 COPY go.mod go.sum ./
@@ -10,10 +12,10 @@ RUN go mod download
 COPY . .
 
 # build
-RUN make build-deps
-RUN make build NAME=pomerium
-RUN make build NAME=pomerium-cli
-RUN touch /config.yaml
+RUN make build-deps && \
+    make build NAME=pomerium && \
+    make build NAME=pomerium-cli && \
+    touch /config.yaml
 
 FROM gcr.io/distroless/base:debug
 ENV AUTOCERT_DIR /data/autocert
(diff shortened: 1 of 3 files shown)


desimone:
@stephengroat  -- Thanks for the contribution. What's hadolint do?

desimone:
Previously https://github.com/pomerium/pomerium/pull/1489

stephengroat:
@desimone hadolint is a dockerfile linter, gives specific suggestions to help optimize dockerfiles for reproducibility, size, performance, etc.

for example, the `--no-install-recommends` helps to create more reproducible builds, the single `RUN` helps to create less layers and is (usually) more efficient for pulling

Dockerfile
```diff
@@ -1,19 +1,21 @@
...
 COPY go.mod go.sum ./
 RUN go mod download
 COPY . .
 
 # build
-RUN make build-deps
-RUN make build NAME=pomerium
-RUN make build NAME=pomerium-cli
-RUN touch /config.yaml
+RUN make build-deps && \
```
travisgroth, line 15: In this context, multiple layers are actually better.  If one of the three make commands is failing, I'd prefer to re-use the other layers rather than start from scratch at `make build-deps` every time.  This is a transient build stage so it doesn't make sense to optimize for space or pull performance.

Dockerfile.debug
```diff
@@ -1,24 +1,28 @@
...
 RUN go mod download
 COPY . .
 
 # build
-RUN make build-deps
-RUN make build-debug NAME=pomerium
-RUN make build-debug NAME=pomerium-cli
-RUN touch /config.yaml
-RUN go get github.com/go-delve/delve/cmd/dlv
+RUN make build-deps && \
```
travisgroth, line 15: See other comment re: build stage

travisgroth (changes requested):
Seems reasonable but I disagree with collapsing the build stage `RUN` layers.  It's useful to keep them separate when debugging and has no practical impact on the resulting image.

Dockerfile.debug
stephengroat: reverted in 3bb958f92366163fae3c64bf45149fa892cde0ab

Dockerfile
stephengroat: reverted in 3bb958f92366163fae3c64bf45149fa892cde0ab

travisgroth (approved):

PR #2379, opened 2021-07-20
k8s: add flush-credentials command

### LINKED ISSUE #2378: pomerium-cli: subcommand to flush k8s credentials

If a user logs out their browser's Pomerium session after using the `k8s` integration, their cached session token is invalidated.

- It isn't apparent where the token was stored or how to clear it without deeper digging
- From the auth provider's perspective, we have no way to tell if the credentials are invalid.  We're only aware of the token expiration, which may be some distance in the future.

A k8s subcommand which clears stored k8s session credentials.  This command could simply clear all the credentials or accept an optional URL to clear the token for a specific API server.

- Detection of invalid credentials.  Without making extra requests directly from our auth plugin, there's no reliable way to do this that I can think of.  

n/a

**Additional context**

Originally discussed in https://github.com/pomerium/pomerium/issues/2263

calebdoxsey:
## Summary
Add subcommand to flush credentials.

## Related issues
Fixes #2378 

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/cmd/pomerium-cli/kubernetes.go
+++ b/cmd/pomerium-cli/kubernetes.go
@@ -27,7 +27,7 @@ var kubernetesCmd = &cobra.Command{
 }
 
 var kubernetesFlushCredentialsCmd = &cobra.Command{
-	Use: "flush-credentials",
+	Use: "flush-credentials [API Server URL]",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) == 0 {
 			clearAllCachedCredentials()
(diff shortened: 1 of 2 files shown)


cmd/pomerium-cli/kubernetes.go
```diff
@@ -18,13 +18,26 @@ import (
...
+	kubernetesCmd.AddCommand(kubernetesFlushCredentialsCmd)
 	rootCmd.AddCommand(kubernetesCmd)
 }
 
 var kubernetesCmd = &cobra.Command{
 	Use: "k8s",
 }
 
+var kubernetesFlushCredentialsCmd = &cobra.Command{
+	Use: "flush-credentials",
```
travisgroth, line 30:
```suggestion
	Use: "flush-credentials [API Server URL]",
```

Issue #2388, opened 2021-07-23
Websocket not working if the upstream is HTTPS


xlejo:
## What happened?

Pomerium fails when try upgrade the connection if the upstream is HTTPS. Pomerium return me a HTTP 400.

If the upstream is HTTP Pomerium upgrade the connection without problems.

## What did you expect to happen?

Upgrade the connection successfully.

## How'd it happen?

1. Install Pomerium with the Helm chart.
2. Config the policy in Pomerium:

```yaml
 - from: https://cryptpad.example.com
   to: https://cryptpad.inside.example.com
   allow_any_authenticated_user: true
   preserve_host_header: true
   allow_websockets: true
   tls_client_cert_file: /pomerium/certs/tls.crt
   tls_client_key_file: /pomerium/certs/tls.key
   tls_custom_ca_file: /pomerium/certs/ca.crt
   tls_server_name: cryptpad.example.com
 - from: https://cryptpad-sandbox.example.com
   to: https://cryptpad-sandbox.inside.example.com
   allow_any_authenticated_user: true
   preserve_host_header: true
   allow_websockets: true
   tls_client_cert_file: /pomerium/certs/tls.crt
   tls_client_key_file: /pomerium/certs/tls.key
   tls_custom_ca_file: /pomerium/certs/ca.crt
   tls_server_name: cryptpad-sandbox.example.com
```
3. Setup a Ingress with HTTPS in the upstream.
4. Expose the 443/TCP of the Pomerium Proxy
5. Connect to  `https://cryptpad.example.com`
6. Try to register

## What's your environment like?

- Pomerium version 0.14.7
- Kubernetes v1.21.2
- Nginx Ingress Controller v0.47.0

## What's your config.yaml?

```config.yaml
address: ":443"
grpc_address: ":443"
certificate_file: "/pomerium/tls/tls.crt"
certificate_key_file: "/pomerium/tls/tls.key"
certificate_authority_file: "/pomerium/ca/ca.crt"
certificates:
  - cert: /etc/pomerium/tls/pomerium-tls/tls.crt
    key: /etc/pomerium/tls/pomerium-tls/tls.key
authenticate_service_url: https://authenticate.example.com
authorize_service_url: https://pomerium-authorize.pomerium.svc.cluster.local
databroker_service_url: https://pomerium-databroker.pomerium.svc.cluster.local
idp_provider: oidc
idp_scopes: [openid profile email]
[...]

PR #2399, opened 2021-07-26
disable http/2 for websockets

calebdoxsey:
## Summary
I guess you can't use http/2 with websockets because they use the same header field. This PR updates the envoy config to explicitly disable http/2 when using web sockets.

## Related issues
- #2388 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/envoyconfig/clusters.go
+++ b/config/envoyconfig/clusters.go
@@ -40,15 +40,15 @@ func (b *Builder) BuildClusters(ctx context.Context, cfg *config.Config) ([]*env
 		return nil, err
 	}
 
-	controlGRPC, err := b.buildInternalCluster(ctx, cfg.Options, "pomerium-control-plane-grpc", []*url.URL{grpcURL}, true)
+	controlGRPC, err := b.buildInternalCluster(ctx, cfg.Options, "pomerium-control-plane-grpc", []*url.URL{grpcURL}, upstreamProtocolHTTP2)
 	if err != nil {
 		return nil, err
 	}
-	controlHTTP, err := b.buildInternalCluster(ctx, cfg.Options, "pomerium-control-plane-http", []*url.URL{httpURL}, false)
+	controlHTTP, err := b.buildInternalCluster(ctx, cfg.Options, "pomerium-control-plane-http", []*url.URL{httpURL}, upstreamProtocolAuto)
 	if err != nil {
 		return nil, err
 	}
-	authZ, err := b.buildInternalCluster(ctx, cfg.Options, "pomerium-authorize", authzURLs, true)
+	authZ, err := b.buildInternalCluster(ctx, cfg.Options, "pomerium-authorize", authzURLs, upstreamProtocolHTTP2)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


wasaga:
I tried that; it did not fix my test using streamlit.io as websocket server 

```
policy:
  - name: ws-1
    from: https://ws1.localhost.pomerium.io
    to: http://localhost:8501
    allow_websockets: true
  - name: ws-2
    from: https://ws2.localhost.pomerium.io
    to: https://ws1.localhost.pomerium.io
    allow_websockets: true
    tls_custom_ca_file: "/path/to/mkcert/rootCA.pem"
```

calebdoxsey:
It works for me. Can you provide concrete feedback. Or should I just close this PR too?

calebdoxsey:
FWIW https://gist.github.com/calebdoxsey/b2040f758c91fa306178fa3fff3e2162

wasaga (approved):
your test environment is different, and yes this PR fixes it for that case. 

the reason my configuration still does not work is different, I'll file a separate ticket about it.

config/envoyconfig/protocols.go
```diff
@@ -33,23 +41,25 @@ func buildUpstreamProtocolOptions(endpoints []Endpoint, forceHTTP2 bool) *envoy_
...
 			},
 		}
-	}
-
-	// when using TLS use ALPN auto config
-	tlsCount := 0
-	for _, e := range endpoints {
-		if e.transportSocket != nil {
-			tlsCount++
+	case upstreamProtocolAuto:
```
desimone, line 44: Could you add some contextualizing comments to this case? The fallthrough and empty default is uncommon enough where I'm not sure what it does without searching.

config/envoyconfig/clusters.go
```diff
@@ -124,8 +124,14 @@ func (b *Builder) buildPolicyCluster(ctx context.Context, options *config.Option
 
 	cluster.AltStatName = getClusterStatsName(policy)
 
+	upstreamProtocol := upstreamProtocolAuto
+	if policy.AllowWebsockets {
+		// #2388, force http/1 when using web sockets
+		upstreamProtocol = upstreamProtocolHTTP1
```
desimone, line 130: Can we add a log message to make clear we are using HTTP1 on this policy because websockets was set?

desimone: To that point, I think we could abstract this little loop into a small function and call it directly in `buildPolicyEndpoints` and `buildCluster`?

config/envoyconfig/clusters.go
```diff
@@ -201,6 +216,7 @@ func (b *Builder) buildPolicyTransportSocket(
 	options *config.Options,
 	policy *config.Policy,
 	dst url.URL,
+	upstreamProtocol upstreamProtocolConfig,
```
desimone, line 219: Having an additional function argument for upstreamProtocol here feels weird to me since we already have the info we need in `policy *config.Policy` to determine the upstreamProtocol.

desimone (changes requested):

desimone:
Welp, guess I missed the bus on this PR. 😆

config/envoyconfig/protocols.go
calebdoxsey: there are already comments. I don't know what else to add. I will remove the default case.

config/envoyconfig/clusters.go
calebdoxsey: This will be very noisy, but ok.

PR #2407, opened 2021-07-29
sessions: add impersonate_session_id, remove legacy impersonation

calebdoxsey:
## Summary
Add impersonate_session_id, remove legacy impersonation.

## Related issues
- https://github.com/pomerium/pomerium-console/issues/1714

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (changes requested):

authenticate/handlers.go
```diff
@@ -451,12 +451,18 @@ func (a *Authenticate) userInfo(w http.ResponseWriter, r *http.Request) error {
 		s.ID = uuid.New().String()
 	}
 
+	isImpersonated := false
 	pbSession, err := session.Get(ctx, state.dataBrokerClient, s.ID)
```
desimone, line 455: This error looks unchecked? What happens if `err!=nil` here?

desimone: e.g. could `pbSession` be `nil` and cause a nil pointer? Probably worth adding a test.

calebdoxsey: no it can't. `Get` methods in protobuf are nil-safe. But sure I can change the code.

desimone: If the answer is no we don't need to update the code.

desimone (approved):

wasaga (approved):
👍 

Side note - I don't understand why we do `headers.rego` instead of writing it in Go.

calebdoxsey:
@wasaga we used to do it in go. It was moved to rego with https://github.com/pomerium/internal/issues/278.

PR #2412, opened 2021-07-30
telemetry: try guess hostname or external IP addr for metrics

wasaga:
## Summary

Try guess hostname or external IP addr for metrics.

## Related issues

Fixes https://github.com/pomerium/pomerium-console/issues/1490

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey (approved):
Trying to guess a FQDN seems like a hopeless task.

If we really wanted to fix this we could use the pion library with ICE. But I still think the whole telemetry stack is way out of scope, so wouldn't advocate building our own UDP service mesh to get it to work.

I guess if we're ok with a half broken solution this is fine.

travisgroth:
We will discuss on monday.

travisgroth:
It's understood that fqdn detection isn't totally reliable; the desire is to have telemetry work out of the box for users if/when possible.  We will likely get feedback on common scenarios where it doesn't work and can fine tune from there.

It does look like there's an issue with tests currently, though.

travisgroth (approved):

PR #2431, opened 2021-08-04
docs/reference: Clarify use of idp_service_account

the-maldridge:
This PR clarifies the use of the `idp_service_account` value which is
not used in all cases, though the warning callout suggests it is
required in all cases.


the-maldridge:
For context, this came out of this discussion in slack: https://pomerium-io.slack.com/archives/CK92MUAES/p1628014051013200

desimone (approved):

desimone:
Thanks @the-maldridge !

PR #2433, opened 2021-08-04
doc updates

calebdoxsey:
## Summary
Various doc updates. In PPL the `groups` criterion supports a single string value, not an array. I think I updated all of them.

## Checklist
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/docs/topics/getting-users-identity.md
+++ b/docs/docs/topics/getting-users-identity.md
@@ -85,17 +85,17 @@ Copy the base64 encoded value of your private key to `pomerium-proxy`'s environm
 SIGNING_KEY=ZxqyyIPPX0oWrrOwsxXgl0hHnTx3mBVhQ2kvW1YB4MM=
 ```
 
-2. Reload `pomerium-proxy`. Navigate to httpbin (by default, `https://httpbin.corp.${YOUR-DOMAIN}.com`), and login as usual. Click **request inspection**. Select `/headers`. Click **try it out** and then **execute**. You should see something like the following.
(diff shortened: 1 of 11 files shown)


docs/docs/topics/getting-users-identity.md
```diff
@@ -85,17 +85,17 @@ Copy the base64 encoded value of your private key to `pomerium-proxy`'s environm
 SIGNING_KEY=ZxqyyIPPX0oWrrOwsxXgl0hHnTx3mBVhQ2kvW1YB4MM=
 ```
 
-1. Reload `pomerium-proxy`. Navigate to httpbin (by default, `https://httpbin.corp.${YOUR-DOMAIN}.com`), and login as usual. Click **request inspection**. Select `/headers'. Click **try it out** and then **execute**. You should see something like the following.
+2. Reload `pomerium-proxy`. Navigate to httpbin (by default, `https://httpbin.corp.${YOUR-DOMAIN}.com`), and login as usual. Click **request inspection**. Select `/headers`. Click **try it out** and then **execute**. You should see something like the following.
```
alexfornuto, line 88: @calebdoxsey I actually prefer using `1.` for each numbered step. Most markdown processors (including GitHub) will render them sequentially, and this way allows steps to be added or deleted without having to adjust all the others.

docs/docs/topics/tcp-support.md
```diff
@@ -39,7 +39,7 @@ routes:
           - email:
               is: contractor@not-example.com
           - groups:
-              has: ["datascience@example.com"]
+              has: "datascience@example.com"
```
alexfornuto, line 42: I thought you said groups.has took an array, or is that only required when there are multiple values?

alexfornuto (approved):
Thanks @calebdoxsey, though I'd still like to get a better understanding of when we can and can't use arrays for group keys.

desimone (approved):

PR #2460, opened 2021-08-09
Document moving routes

alexfornuto:
## Summary

Adds a new section to Concepts describing how associated policies could prevent routes from being moved.

## Related issues

https://github.com/pomerium/pomerium-console/pull/1865

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/docs/enterprise/concepts.md
+++ b/docs/enterprise/concepts.md
@@ -108,7 +108,7 @@ This term refers to the system or service the route provides or restricts access
 
 ### Moving Routes
 
-When moving a Route from one [Namespace](#namespace) to another, enforced policies will automatically be removed or applied. Optional policies available in the source Namespace but not the target will prevent the move. This is intentional, to prevent unassociated policies.
+When moving a Route from one [Namespace](#namespace) to another, enforced policies will automatically be removed or applied. Optional policies available in the source Namespace but not the target will prevent the move. This is intentional to prevent unassociated policies.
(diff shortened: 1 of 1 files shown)


docs/enterprise/concepts.md
```diff
@@ -106,6 +106,10 @@ Routes define the connection pathway and configuration from the internet to your
 
 This term refers to the system or service the route provides or restricts access to.
 
+### Moving Routes
+
+When moving a Route from one [Namespace](#namespace) to another, enforced policies will automatically be removed or applied. Optional policies available in the source Namespace but not the target will prevent the move. This is intentional, to prevent unassociated policies.
```
desimone, line 111:
```suggestion
When moving a Route from one [Namespace](#namespace) to another, enforced policies will automatically be removed or applied. Optional policies available in the source Namespace but not the target will prevent the move. This is intentional to prevent unassociated policies.
```

I think there is not comma, but I am not sure. 

> If the independent clause comes first, no punctuation separates the two.

desimone (approved):

docs/enterprise/concepts.md
alexfornuto: I think it's a style choice in this example, but I tend to over-comma in my style so happy to adjust. Thanks!

Issue #2429, opened 2021-08-04
Multiple databroker urls returns x509 validation error


ventris:
## What happened?

Pomerium seems to incorrectly validate certificates for databroker_service_urls when you add multiple items to the list.

## What did you expect to happen?

Pomerium to be able to validate each FQDN separately when it does its connection.

## How'd it happen?

1. Use `databroker_service_urls` with multiple hosts
2. Start Pomerium with `log_level: debug` and `debug: true`

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`):
```
{"level":"info","version":"0.14.7-1624903390+e5d4c824","time":"2021-08-04T07:30:17Z","message":"cmd/pomerium"}
```
- Server Operating System/Architecture/Cloud:
Docker container running on Ubuntu 18.04

## What's your config.yaml?

```config.yaml
address: :443

databroker_service_urls:
        - "https://databroker1.example.com:5001"
        - "https://databroker2.example.com:5001"
        - "https://databroker3.example.com:5001"

databroker_storage_tls_skip_verify: true

authorize_service_url: "https://authorize.example.com"

authenticate_service_url: "https://authenticate.example.com"

shared_secret: "redacted"
cookie_secret: "redacted"
```

## What did you see in the logs?

```logs
{"level":"error","error":"rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = \"transport: authentication handshake failed: x509: certificate is valid for databroker1.example.com, 10.10.10.10, 127.0.0.1, localhost, not databroker1.example.com:5001,databroker2.example.com:5001,databroker3.example.com:5001\"","time":"2021-08-02T09:02:41Z","message":"controlplane: error storing configuration event"}
```

## Additional context

If I add all hostnames to the same certificate I still get this error.

From the looks of the error log it seems like Pomerium is concatenating all items in the list before trying to validate the certificate when it does the connection.

If I use `databroker_service_urls` with just one item it works fine but when I have more than 1 item I get the above error every time. Same goes when trying to use a wildcard certificate.

@travisgroth

calebdoxsey:
The issue here is caused by our custom resolver. It supports URLs like:

```
pomerium:///databroker1.example.com,databroker2.example.com
```

Although it establishes the connections properly, the TLS verification is done using the server name:

```
databroker1.example.com,databroker2.example.com
```

[...]

PR #2469, opened 2021-08-12
grpc: send client traffic through envoy

calebdoxsey:
## Summary
This PR updates our gRPC clients to send gRPC traffic to a local address that envoy forwards to the appropriate backend service. This should fix the issue with multiple databroker endpoints, but also makes the code more consistent (extauthz traffic from the proxy service was already being handled by Envoy).

## Related issues
- #2429 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


authorize/grpc.go
```diff
@@ -23,6 +23,10 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v3.CheckRe
 	ctx, span := trace.StartSpan(ctx, "authorize.grpc.Check")
 	defer span.End()
 
+	if err := a.WaitForInitialSync(ctx); err != nil {
```
calebdoxsey, line 26: this was moved here because the initial sync can't begin until the outbound gRPC listener starts - i.e. the initial sync was waiting for the control plane to start, and the control plane was waiting for the initial sync to finish, so they were deadlocked.

config/envoyconfig/tls.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"net/url"
+	"regexp"
+	"strings"
+
+	envoy_type_matcher_v3 "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
+)
+
+func (b *Builder) buildSubjectAlternativeNameMatcher(
```
calebdoxsey, line 11: Our docs say:

```
Example: *.corp.example.com if wild card or authenticate.corp.example.com/authorize.corp.example.com
```

I'm not sure if that actually worked before, but this code should fix it by converting it into a regular expression in Envoy.

internal/registry/reporter.go
```diff
@@ -39,23 +39,11 @@ func (r *Reporter) OnConfigChange(ctx context.Context, cfg *config.Config) {
...
-		OverrideCertificateName: cfg.Options.OverrideCertificateName,
-		CA:                      cfg.Options.CA,
-		CAFile:                  cfg.Options.CAFile,
-		RequestTimeout:          cfg.Options.GRPCClientTimeout,
-		ClientDNSRoundRobin:     cfg.Options.GRPCClientDNSRoundRobin,
-		WithInsecure:            cfg.Options.GetGRPCInsecure(),
-		InstallationID:          cfg.Options.InstallationID,
-		ServiceName:             cfg.Options.Services,
-		SignedJWTKey:            sharedKey,
+	registryConn, err := grpc.GetOutboundGRPCClientConn(context.Background(), &grpc.OutboundOptions{
```
desimone, line 42: Do we want to send backgroud context here or the existing?

internal/netutil/netutil.go
```diff
@@ -0,0 +1,22 @@
+// Package netutil contains various functions that help with networking.
+package netutil
+
+import "net"
+
+// AllocatePorts allocates random ports suitable for listening.
+func AllocatePorts(count int) ([]string, error) {
```
desimone, line 7: My understanding is that `:0` gets a list of system-chosen ports. Does `net.Listen` hold these ports until they are used? 

> If the port in the address parameter is empty or "0", as in "127.0.0.1:" or "[::1]:0", a port number is automatically chosen.

Is there a real risk of another service grabbing one of those ports prior to us using them?

internal/cmd/pomerium/pomerium.go
```diff
@@ -143,13 +142,6 @@ func Run(ctx context.Context, configFile string) error {
 		eg.Go(func() error {
 			return authorizeServer.Run(ctx)
 		})
```
desimone, line 144: Mostly just curios what this was removed

authorize/grpc.go
desimone: Do you mind adding adding a comment to that effect? (e.g. not the change, but why we are waiting)

authenticate/authenticate_test.go
```diff
@@ -110,7 +110,6 @@ func TestNew(t *testing.T) {
 		{"empty opts", &config.Options{}, true},
 		{"fails to validate", badRedirectURL, true},
 		{"bad provider", badProvider, true},
```
desimone, line 112: Can we update this test? Or remove the build up if no longer relevant?

config/envoyconfig/tls.go
desimone: nice!

internal/cmd/pomerium/pomerium.go
desimone: Got it. Saw your other comment.

desimone (commented):
This feels much more consistent and is a nice improvement. Left some minor comments, and suggestions.

Not sure why the tests are timing out, however. Maybe unrelated.

internal/netutil/netutil.go
calebdoxsey: It does not hold them. Yes there's a risk of another service grabbing them. I don't know of a way to prevent this.

internal/registry/reporter.go
wasaga: our grpc dial is non-blocking, thus it doesn't matter for GRPC purposes
however specifically OnConfigChange context carries the configuration source thus it might be beneficial to do `ctx`

wasaga (approved):

config/envoyconfig/outbound.go
```diff
@@ -0,0 +1,134 @@
...
+		ConfigType: &envoy_config_listener_v3.Filter_TypedConfig{
+			TypedConfig: tc,
+		},
+	}, nil
+}
+
+func (b *Builder) buildOutboundRouteConfiguration() (*envoy_config_route_v3.RouteConfiguration, error) {
+	return b.buildRouteConfiguration("grpc", []*envoy_config_route_v3.VirtualHost{{
+		Name:    "grpc",
+		Domains: []string{"*"},
```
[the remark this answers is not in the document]
calebdoxsey: The connection won't have a domain name set since its going to `127.0.0.1`. We're using paths to do the routing.

PR #2509, opened 2021-08-23
fix forward-auth, logging

calebdoxsey:
## Summary
Discovered while working on integration tests, there are circumstances where forward auth with nginx can get into an infinite redirect loop. Moving the `isForwardAuth` check to after the session id check seems to fix it. It also makes more sense in term of the flow.

Basically we want to Redirect to authenticate if there's no session, but if there is a session we want to return Forbidden. In forward auth mode that means returning a `401 Unauthorized` with no session and a `403 Forbidden` when there is a session.

I also moved a noisy log and added error logging for google cloud serverless.

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/evaluator/google_cloud_serverless.go
+++ b/authorize/evaluator/google_cloud_serverless.go
@@ -18,6 +18,8 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/sync/singleflight"
 	"google.golang.org/api/idtoken"
+
+	"github.com/pomerium/pomerium/internal/log"
 )
 
 // GCP pre-defined values.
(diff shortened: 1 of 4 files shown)


authorize/evaluator/google_cloud_serverless.go
```diff
@@ -160,11 +162,13 @@ func getGoogleCloudServerlessTokenSource(serviceAccount, audience string) (oauth
 func getGoogleCloudServerlessHeaders(serviceAccount, audience string) (map[string]string, error) {
 	src, err := getGoogleCloudServerlessTokenSource(serviceAccount, audience)
 	if err != nil {
+		log.Error(context.Background()).Err(err).Msg("error retrieving google cloud serverless token source")
```
desimone, line 165: Can we log this in the calling function, and wrap the error instead here?

wasaga (approved):

desimone (approved):

PR #2518, opened 2021-08-24
Remove api from GitLab defaultScope

alexfornuto:
## Summary

Currently, we require a GitLab application with the `api` scope on GitLab (SAAS) to authenticate using it as an IdP. This gives the application full read/write permissions for the entire API, much more than we need.

This change restricts the scope to `email` ~and `openid`~ (already defined). Including `profile` was also discussed, ~but the values therein are already provided by `openid`~ and put back in to match OIDC spec.:

image

I've built and tested this locally, with no problems (other than those caused by my own stupidity :smile: ).

## Related issues

https://github.com/pomerium/docs/issues/25

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/internal/identity/oidc/gitlab/gitlab.go
+++ b/internal/identity/oidc/gitlab/gitlab.go
@@ -16,7 +16,7 @@ import (
 // Name identifies the GitLab identity provider.
 const Name = "gitlab"
 
-var defaultScopes = []string{oidc.ScopeOpenID, "profile", "email", "openid"}
+var defaultScopes = []string{oidc.ScopeOpenID, "profile", "email"}
 
 const (
 	defaultProviderURL = "https://gitlab.com"
--- a/internal/identity/oidc/gitlab/gitlab.go
+++ b/internal/identity/oidc/gitlab/gitlab.go
@@ -16,7 +16,7 @@ import (
 // Name identifies the GitLab identity provider.
 const Name = "gitlab"
 
-var defaultScopes = []string{oidc.ScopeOpenID, "profile", "email", "api"}
+var defaultScopes = []string{oidc.ScopeOpenID, "profile", "email", "openid"}
 
 const (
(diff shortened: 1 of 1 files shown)


internal/identity/oidc/gitlab/gitlab.go
```diff
@@ -16,7 +16,7 @@ import (
 // Name identifies the GitLab identity provider.
 const Name = "gitlab"
 
-var defaultScopes = []string{oidc.ScopeOpenID, "profile", "email", "api"}
+var defaultScopes = []string{oidc.ScopeOpenID, "email", "openid"}
```
desimone, line 19:
```suggestion
var defaultScopes = []string{oidc.ScopeOpenID, "email", "profile"}
```
oidc.ScopeOpenID is `openid`

desimone: I still feel like `profile` should be included to be in compliance with the openid spec. Unless it doesn't work?

alexfornuto: Dropped the commit.

calebdoxsey (approved):

internal/identity/oidc/gitlab/gitlab.go
```diff
@@ -16,7 +16,7 @@ import (
 // Name identifies the GitLab identity provider.
 const Name = "gitlab"
 
-var defaultScopes = []string{oidc.ScopeOpenID, "profile", "email", "api"}
+var defaultScopes = []string{oidc.ScopeOpenID, "profile", "email", "openid"}
```
desimone:
```suggestion
var defaultScopes = []string{oidc.ScopeOpenID, "profile", "email"}
```

desimone (approved):

PR #2557, opened 2021-09-01
config: remove signature_key_algorithm

calebdoxsey:
## Summary
Remove `signing_key_algorithm` and infer it from `signing_key`.

## Related issues
Fixes https://github.com/pomerium/internal/issues/521

## Checklist
- [x] updated docs
- [x] updated unit tests
- [x] updated UPGRADING.md
- [x] ready for review


pkg/cryptutil/jose.go
```diff
@@ -93,3 +97,15 @@ func loadPublicKey(b []byte) (interface{}, error) {
 
 	return nil, fmt.Errorf("couldn't load public key: %w", wrappedErr)
 }
+
+// SignatureAlgorithmForKey returns the signature algorithm for the given key.
+func SignatureAlgorithmForKey(key interface{}) (jose.SignatureAlgorithm, error) {
```
desimone, line 102: Could we add a few unit tests that explicitly test this function for both private and public keys for both algo types?

desimone (changes requested):
LGTM. Only suggestion would be to directly test the exported `SignatureAlgorithmForKey` function.

docs/docs/upgrading.md
```diff
@@ -10,6 +10,8 @@ description: >-
 ### Removed options
 The deprecated `headers` option has been removed. Use [`set_response_headers`](/reference/readme.md#set-response-headers) instead.
 
+The `signing_key_algorithm` option has been removed and will now be inferred from `signing_key`.
```
alexfornuto, line 13: Is this being backported to 15.0?

calebdoxsey: No it should not be backported.

alexfornuto (approved):

desimone (approved):

PR #2561, opened 2021-09-02
TCP Client Doc

alexfornuto:
## Summary

Documents using the pomerium-cli and desktop app

## Related issues

https://github.com/pomerium/docs/issues/37

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


docs/docs/client.md
```diff
@@ -0,0 +1,113 @@
...
+description: Learn how to create secured TCP connections to services behind Pomerium.
+---
+
+# Pomerium Desktop and CLI Clients
+
+Pomerium is capable of creating secure connections to services like SSH, Redis, and more by creating a TCP tunnel to the service with a local client. This article describes configuring a route to accept TCP connections, and using either the CLI or GUI client to connect to it.
+
+## Create a TCP Route
+
+1. Specify it as a TCP Route by prefixing `tcp+` in the **From** field, along with a port suffix.
```
desimone, line 12: > Specify it as

What is `it` in this sentence?

docs/docs/client.md
```diff
@@ -0,0 +1,113 @@
...
+
+   ::: tip
+   The port is not used in the connection to the Pomerium gateway; this will always be port 443. Rather, the port defined in **From** is part of the mapping to the individual route. In this way, you can create multiple routes that share a DNS entry, differentiated by the port to determine which route they use.
```
desimone, line 15: > the Pomerium gateway

Can we define what the gateway is in this context?

docs/docs/client.md
alexfornuto:
```suggestion
1. Specify this new Route as a TCP Route by prefixing `tcp+` in the **From** field, along with a port suffix.
```

docs/docs/client.md
alexfornuto:
```suggestion
   The port is not used to connect to the Pomerium proxy service from the internet; this will always be port 443 (unless otherwise defined in `config.yaml`). Rather, the port defined in **From** is part of the mapping to the individual route. In this way, you can create multiple routes that share a DNS entry, differentiated by the port to determine which route they use.
```

desimone: SGTM

desimone: SGTM

desimone (approved):

docs/docs/client.md
```diff
@@ -0,0 +1,119 @@
...
+## TCP Client Software
+
+You can connect to this route with either the Pomerium CLI or Pomerium Desktop cient.
+
+::::: tabs
+
+:::: tab Pomerium Desktop
+### Install
+
+TBD
```
alexfornuto, line 50: CC @travisgroth for details on Client release methods.

docs/docs/client.md
```diff
@@ -0,0 +1,123 @@
...
+# Pomerium Desktop and CLI Clients
+
+Pomerium is capable of creating secure connections to services like SSH, Redis, and more by creating a TCP tunnel to the service with a local client. This article describes configuring a route to accept TCP connections, and using either the CLI or GUI client to connect to it.
+
+## Create a TCP Route
+
+1. Specify this new Route as a TCP Route by prefixing `tcp+` in the **From** field, along with a port suffix.
+
+   ::: tip
+   The port is not used to connect to the Pomerium proxy service from the internet; this will always be port 443 (unless otherwise defined in `config.yaml`). Rather, the port defined in **From** is part of the mapping to the individual route. In this way, you can create multiple routes that share a DNS entry, differentiated by the port to determine which route they use.
```
travisgroth, line 15: Might be long winded for a `tip` section but should an example be included for clarity?  My go-to is always accessing SSH and PSQL on the same host without requiring different DNS entires.

travisgroth (approved):
LGTM; one question about how expansive to make the tip about `from` port.

docs/docs/client.md
alexfornuto: We can do that, and we'll just make it part of the step and not a tip.

alexfornuto: How does 08cc640be32739917861f2b7cfd87f8001201ad3 grab ya?

travisgroth: 👍

PR #2567, opened 2021-09-03
google: support provider URL

calebdoxsey:
## Summary
Add support for a provider URL with the google provider. This is so we can mock directory calls in integration tests. If no provider URL is supplied the current default will be used.

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/directory/google/google.go
+++ b/internal/directory/google/google.go
@@ -49,6 +49,10 @@ func WithServiceAccount(serviceAccount *ServiceAccount) Option {
 
 // WithURL sets the provider url to use.
 func WithURL(url string) Option {
+	// if the empty string is sent in, use the default provider URL.
+	if url == "" {
+		url = defaultProviderURL
+	}
 	return func(cfg *config) {
 		cfg.url = url
 	}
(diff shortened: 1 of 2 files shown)


internal/directory/google/google.go
```diff
@@ -49,14 +49,18 @@ func WithServiceAccount(serviceAccount *ServiceAccount) Option {
 
 // WithURL sets the provider url to use.
 func WithURL(url string) Option {
+	// if the empty string is sent in, use the default provider URL.
+	if url == "" {
+		url = defaultProviderURL
+	}
```
wasaga, line 55: imho, that makes `WithURL` less readable. 

I'd make this defaulting part of `New` instead, so that if you do not supply WithURL option, then a default would be used.

internal/directory/google/google.go
```diff
@@ -49,14 +49,18 @@ func WithServiceAccount(serviceAccount *ServiceAccount) Option {
...
 	return func(cfg *config) {
 		cfg.url = url
 	}
 }
 
 func getConfig(opts ...Option) *config {
 	cfg := new(config)
-	WithURL(defaultProviderURL)(cfg)
+	WithURL("")(cfg)
```
wasaga, line 63: this would require me to peek to WithURL implementation to understand what is the magic behind empty argument.

wasaga (approved):

internal/directory/google/google.go
desimone: I agree.

desimone (approved):

PR #2579, opened 2021-09-07
Document recovery token generation

alexfornuto:
## Summary

Adds a partial file to both current Enterprise installation pages, detailing how to generate a recovery token.

## Related issues

Closes https://github.com/pomerium/pomerium-console/issues/2009

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


docs/enterprise/install/helm.md
```diff
@@ -162,3 +162,9 @@ kubectl delete svc pomerium-proxy
 
 [Pomerium using Helm]: /docs/install/helm.md
 [cert-manager]: https://cert-manager.io/docs/
+
+### Generate Recovery Token
+
+!!!include(generate-recovery-token.md)!!!
+
+Note that you will need to connect to a terminal environment for the same container running the Enterprise console you're connected to to run this command.
```
alexfornuto, line 170: @travisgroth this is how I would think to run this in a k8s environment, but I'd bet you have a better idea.

travisgroth: Actually, you can run it anywhere.  It will require access to the enterprise console's config and network access to the database.  I believe only the database parameter from the config is required.

travisgroth (approved):

desimone (approved):

PR #2584, opened 2021-09-09
autocert: remove log

### LINKED ISSUE #2559: get cert errors every 10 minutes
## What happened?

We are getting get cert errors consistently for every route. 

We don't use autocert and use a wildcard, so I wouldn't expect an error at all.

## How'd it happen?

Every 10 minutes, this appears for all routes from Enterpise.

```
pomerium: 0.15.0-1628130593+7692b2ac
envoy: 1.19.0+6869b5e5bc8111b4554ec3c565f2f8661516efa4c7dfa4f55c588b1c79cead42
```
    Centos 8
    Two Cores and Consoles with shared Postgres and Redis

## What's your config.yaml?

```config.yaml
certificates:
- cert: /PATH/TO/CERT.cer
  key: /PATH/TO/KEY.key   
```

```logs

{
    "level": "error",
    "error": "open /PATH/TO/certificates/acme-v02.api.letsencrypt.org-directory/ROUTE/ROUTE.crt: no such file or directory",
    "domain": "ROUTE",
    "time": "2021-09-01T08:15:03-04:00",
    "message": "get cert"
}
```

## Additional context

We do not notice any problems associated with the error messages, and all routes are managed by Enterprise console.

calebdoxsey:
## Summary
Remove a noisy error log. This error is returned for domains not managed by certmagic. It's not a real error.

It was introduced in #2286 

## Related issues
Fixes #2559 

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/autocert/manager.go
+++ b/internal/autocert/manager.go
@@ -166,7 +166,7 @@ func (mgr *Manager) renewConfigCerts(ctx context.Context) error {
 	for _, domain := range sourceHostnames(cfg) {
 		cert, err := cm.CacheManagedCertificate(domain)
 		if err != nil {
-			log.Error(ctx).Err(err).Str("domain", domain).Msg("get cert")
+			// this happens for unmanaged certificates
 			continue
 		}
 		if cert.NeedsRenewal(cm) {


internal/autocert/manager.go
```diff
@@ -166,7 +166,7 @@ func (mgr *Manager) renewConfigCerts(ctx context.Context) error {
 	for _, domain := range sourceHostnames(cfg) {
 		cert, err := cm.CacheManagedCertificate(domain)
 		if err != nil {
-			log.Error(ctx).Err(err).Str("domain", domain).Msg("get cert")
+			// this happens for unmanaged certificates
```
desimone, line 169: Should we handle the error if `cert` is nil just to be safe? Or is that impossible?

https://github.com/caddyserver/certmagic/blob/b74a9ce8088fde76facb32d545d2494f52b227fe/certificates.go#L110

calebdoxsey: If the library follows Go conventions `cert` should never be nil. But I don't know if it does or not, so I guess I can add code to handle this.

calebdoxsey: Nevermind. `cert` is not a pointer type.

wasaga (approved):

desimone (approved):

PR #2595, opened 2021-09-14
Document Enterprise API

alexfornuto:
## Summary

Documents the Enterprise API

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/docs/enterprise/api.md
+++ b/docs/enterprise/api.md
@@ -50,4 +50,138 @@ go get github.com:pomerium/enterprise-client-go.git
 ::::
 ## Test the API Connection
 
+The repositories for our [Python][client-py] and [Go][client-go] implementations include example scripts:
+
+:::: tabs
+::: tab Python
+```python
+#!/usr/bin/env python
+
+import os
+from pomerium.client import Client
+from pomerium.pb.policy_pb2 import ListPoliciesRequest
+from pomerium.pb.namespaces_pb2 import ListNamespacesRequest
+from pomerium.pb.routes_pb2 import SetRouteRequest, Route
+
+# get custom CA and service account credentials from environment
+ca_cert = os.getenv('CA_CERT', '').encode('utf-8')
+sa = os.getenv('SERVICE_ACCOUNT', '')
+console_api = 'console-api.localhost.pomerium.io'
+
+client = Client(console_api, sa, root_certificates=ca_cert)
+
+# get id for namespace 'Production'
+resp = client.NamespaceService.ListNamespaces(ListNamespacesRequest())
+ns = [n for n in resp.namespaces if n.name == 'Production'][0]
+
+# find policy named 'my policy' in namespace 'Production'
+resp = client.PolicyService.ListPolicies(
+    ListPoliciesRequest(query='my policy', namespace=ns.id)
+)
+policy = resp.policies[0]
+
+# set route in namespace 'Production', associated to 'my policy'
+route = Route(**{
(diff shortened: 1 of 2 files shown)


docs/enterprise/api.md
```diff
@@ -0,0 +1,203 @@
...
+title: API
+lang: en-US
+meta:
+  - name: keywords
+    content: pomerium identity-access-proxy oidc reverse-proxy enterprise console api python go
+---
+
+# Enterprise Console API
+
+The Pantheon Enterprise Console supports programmatic interaction through an API. This page covers enabling and authenticating to the API.
```
travisgroth, line 11:
```suggestion
The Pomerium Enterprise Console supports programmatic interaction through an API. This page covers enabling and authenticating to the API.
```

docs/enterprise/api.md
```diff
@@ -0,0 +1,203 @@
...
+	```
+
+	If you're running Pomerium Enterprise as a system service, restart the daemon.
+
+
+## Create a Service Account
+
+1. In the enterprise Console under **Configure -> Service Accounts**, Click **+ Add Service Account**. You can choose an existing user for the service account to impersonate, or create a new user. Note that a new user will not be synced to your IdP.
+
+1. The Enterprise Console will display the service account token. Be sure to store it securely not, as you cannot view it again after this point.
```
travisgroth, line 44:
```suggestion
1. The Enterprise Console will display the service account token. Be sure to store it securely now, as you cannot view it again after this point.
```

docs/enterprise/api.md
```diff
@@ -0,0 +1,203 @@
...
+
+1. Add the service account to the Namespaces you want it to access, with the correct role based on the actions it will perform.
```
travisgroth, line 46:
```suggestion
1. Grant the service account the appropriate role for access to various Namespaces
```

travisgroth: Soft suggestion that I don't love.  But I think we should be careful about phrasing and the difference between creating a service account in a namespace and granting it permissions on that namespace.  "add to a namespace" is a bit ambiguous.

docs/enterprise/api.md
```diff
@@ -0,0 +1,203 @@
...
+
+:::: tabs
+::: tab Python
+```bash
+pip3 install git+ssh://git@github.com/pomerium/enterprise-client-python.git
+```
+:::
+::: tab Go
+```bash
+go get github.com:pomerium/enterprise-client-go.git
```
travisgroth, line 62:
```suggestion
go get github.com/pomerium/enterprise-client-go
```

travisgroth (changes requested):

docs/enterprise/api.md
alexfornuto: Whoooopsie!

docs/enterprise/api.md
alexfornuto: Agreed that my wording was not great. How about:

```suggestion
1. Grant the service account the appropriate [role](/enterprise/concepts.md#rbac-for-enterprise-console-users) in the Namespace(s) it will operate in.
```

alexfornuto:
```suggestion
1. Grant the service account the appropriate [role](/enterprise/concepts.md#rbac-for-enterprise-console-users) on the Namespace(s) it will operate against.

```

travisgroth (approved):

PR #2663, opened 2021-10-05
ppl: add version to generator

calebdoxsey:
## Summary
The PPL generator converts PPL into rego. That rego code is then evaluated by Authorize which expects data in a certain format. That format may change in the future, so this PR adds a version comment that specifies the expected version to work properly.

## Related issues
- https://github.com/pomerium/internal/issues/602

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/evaluator/policy_evaluator.go
+++ b/authorize/evaluator/policy_evaluator.go
@@ -2,11 +2,14 @@ package evaluator
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/open-policy-agent/opa/rego"
 	octrace "go.opencensus.io/trace"
+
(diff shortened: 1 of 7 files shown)


authorize/evaluator/policy_evaluator.go
```diff
@@ -2,12 +2,15 @@ package evaluator
...
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/open-policy-agent/opa/rego"
 	octrace "go.opencensus.io/trace"
 
+	"github.com/pomerium/pomerium/pkg/policy/generator"
+
```
desimone, line 13:
```suggestion
```

nit fmt

wasaga (changes requested):

PR #2758, opened 2021-11-09
Docs: Update Securing Kubernetes Guide

alexfornuto:
## Summary

Updates [Securing Kubernetes](https://pomerium.com/guides/kubernetes.html) to continue from our Helm-based install instructions, using the service account created in the chart.

## Related issues

Closes https://github.com/pomerium/docs/issues/7
Supersedes #2738 

## Checklist

- [ ] reference any related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] add appropriate tag (`improvement` / `bug` / etc)
- [ ] ready for review

PR #2759, opened 2021-11-09
Docs: Update Kubernetes Dashboard Guide

### LINKED ISSUE #1498: update kubernetes dashboard guide
Hi,

I am reading kubernetes dashboard guide, https://www.pomerium.com/guides/kubernetes-dashboard.html.

and you used stable/kubernetes-dashboard chart in there.
that's too old.

Please update using kubernetes-dashboard/kubernetes-dashboard way.
https://github.com/kubernetes/dashboard/tree/master/aio/deploy/helm-chart/kubernetes-dashboard

and that would be useful that you can control ingress, without `kubectl apply -f docs/recipes/yml/dashboard-forwardauth.ingress.yaml`

Thanks,

alexfornuto:
## Summary

This new version takes advantage of the RBAC options the Pomerium Helm chart now makes available. Since we can define user permissions this way, I've also removed the forward-auth instructions.

## Related issues

Fixes #1498
Fixes https://github.com/pomerium/docs/issues/8

## Checklist

- [x] https://github.com/pomerium/pomerium/pull/2758 must be merged first! This PR builds off of new content there.
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


docs/guides/kubernetes-dashboard.md
```diff
@@ -14,359 +14,100 @@ The following guide covers how to secure [Kubernetes Dashboard] using Pomerium.
...
-token:      eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9.......
+Following the configuration defined in [Install Pomerium using Helm], add a route for the Kubernetes Dashboard.
 
-```
+1. Modify `pomerium-values.yaml` with the following route:
 
-The above token then needs to be assigned to our route configuration and policy.
+    ```yaml
+        - from: https://dashboard.localhost.pomerium.io
+          to: https://kubernetes-dashboard.default.svc
```
travisgroth, line 58:
```suggestion
          to: https://kubernetes-dashboard.default.svc.cluster.local
```

This chart could be installed anywhere but the fqdn to default namespace is a good guess

docs/guides/kubernetes-dashboard.md
```diff
@@ -14,359 +14,100 @@ The following guide covers how to secure [Kubernetes Dashboard] using Pomerium.
```
travisgroth, line 14: [copy] intro blurb is no longer accurate.  We don't cover forward auth.  It's also worth calling out that we're able to pass the user's identity to the API server through the dashboard.  This is in contrast to most deployments that use static tokens.

travisgroth (approved):
One copy suggestion and one config suggestion but LGTM from tech standpoint.

docs/guides/kubernetes-dashboard.md
alexfornuto: Adjusted in https://github.com/pomerium/pomerium/pull/2759/commits/55f216691b1af0726fa74b3b0ad1bfd4c8f81257

docs/guides/kubernetes-dashboard.md
```diff
@@ -10,363 +10,104 @@ description: >-
...
-      to: https://helm-dashboard-kubernetes-dashboard
-      policy:
-        - allow:
-            or:
-              - email:
-                  is: user@domain.example
-      tls_skip_verify: true # dashboard uses self-signed certificates in its default configuration
-      set_request_headers:
-        Authorization: Bearer eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9.....
+    Apply the permissions with `kubectl apply -f rbac-someuser.yaml`
```
cmo-pomerium, line 89:
```suggestion
    Apply the permissions with `kubectl apply -f rbac-someuser.yaml`.
```

docs/guides/kubernetes-dashboard.md
```diff
@@ -10,363 +10,104 @@ description: >-
...
-  </video>
+Because we've defined RBAC for our users, they can authenticate with Pomerium and Kubernetes will recognize that user in the Dashboard:
 
-In the end, you should choose whichever option makes the most sense for your use-case and environment.
+<video controls  muted="" playsinline="" width="100%" height="600" control=""><source src="./img/k8s-dashboard-user.mp4" type="video/mp4">
+Your browser does not support the video tag.
+</video>
 
-Whichever option you choose to go with, 🎉🍾🎊 **congratulations** 🎉🍾🎊! You now have a single-sign-on enabled [Kubernetes Dashboard] protected by Pomerium and automatically renewing [LetsEncrypt] certificates.
+🎉🍾🎊 **Congratulations** 🎉🍾🎊! You now have a single-sign-on enabled [Kubernetes Dashboard] protected by Pomerium.
```
cmo-pomerium, line 105:
```suggestion
🎉🍾🎊 **Congratulations!** 🎉🍾🎊 You now have a single-sign-on enabled [Kubernetes Dashboard] protected by Pomerium.
```
- Propose moving the ! mark into the bold so it's closer. The triple emojis seem to be a good enough spacer anyways.

cmo-pomerium (commented):
Comments and proposed changes added.

docs/guides/kubernetes-dashboard.md
```diff
@@ -10,363 +10,104 @@ description: >-
...
-- Installing [Helm] a package manager for Kubernetes
-- Deploying [NGINX Ingress Controller]
-- Install and configure [Cert-Manager] to issue [LetsEncrypt] certificates
-- Deploying Pomerium
-- Deploying [Kubernetes Dashboard]
-- Secure Kubernetes Dashboard access:
-  - _directly_, using Pomerium's proxy component
-  - _indirectly_, using Pomerium as a [forward-auth] provider
+- Deploying [Kubernetes Dashboard] using [Helm]
+- Secure Kubernetes Dashboard access through Pomerium
```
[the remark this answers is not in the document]
alexfornuto: I'm torn on this. I agree that the proposed change would make it more consistent with above, but as a standalone sentence I prefer "Secure" as an adjective versus "Securing" as a verb.

The verb form implies that access exists already and we're modifying it, when actually there's no access until we create an Ingress for it, which will be Secure because it's using our Ingress Controller.

docs/guides/kubernetes-dashboard.md
```diff
@@ -10,363 +10,104 @@ description: >-
...
-## Pomerium
-
-Pomerium is an identity-aware access proxy that can used to serve as an identity-aware reverse proxy, or as a forward-auth provider.
-
-### Configure
-
-Before installing, we will configure Pomerium's configuration settings in `values.yaml`. Other than the typical configuration settings covered in the quick-start guides, we will add a few settings that will make working with Kubernetes Dashboard easier.
-
-We can retrieve the token to add to our proxied policy's authorization header as follows.
+1. Use [Helm] to install a new instance of [Kubernetes Dashboard].
```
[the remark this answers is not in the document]
alexfornuto: Agree that it doesn't need to be numbered, disagree on an action being a header. Will adjust.

docs/guides/kubernetes-dashboard.md
```diff
@@ -10,363 +10,104 @@ description: >-
...
+    --set ingress.enabled="false"
+  ```
 
-```bash
-Name:         dashboard-kubernetes-dashboard-token-bv9jq
-Namespace:    default
-Labels:       <none>
-Annotations:  kubernetes.io/service-account.name: dashboard-kubernetes-dashboard
-              kubernetes.io/service-account.uid: 18ab35ee-eca1-11e9-8c75-025000000001
+That's it. We've now configured kubernetes dashboard in our cluster. We've also explicitly told Helm that we are going to deploy our own custom access to the service through Pomerium, instead of a standard ingress.
```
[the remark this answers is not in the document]
alexfornuto: Proposed diff is :+1: 

re last point: This is "an ingress" of our own type, which is handled by our "ingress controller", opposed to a standard ingress handled by a standard ingress controller.

docs/guides/kubernetes-dashboard.md
alexfornuto: IMHO I would get rid of all the emoji and whatnot, it has no place in documentation... But it was there when I got here and I don't want to be the killjoy.

docs/guides/kubernetes-dashboard.md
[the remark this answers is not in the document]
alexfornuto: Adjusted in 2def395267f750e8021008830d82c074125623b9

docs/guides/kubernetes-dashboard.md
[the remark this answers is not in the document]
travisgroth: Contextually it needs a verb, but maybe...

```suggestion
- Establishing secure Kubernetes Dashboard access through Pomerium
```

PR #2767, opened 2021-11-15
config: detect changes to the kubernetes service account token file

### LINKED ISSUE #2722: kubernetes_service_account_token_file not reloaded after rotation by k8s
## What happened?

Following the kubernetes/kubectl guide at https://www.pomerium.io/guides/kubernetes.html
We have a working setup, and I can properly use `kubectl` to interact with the running cluster.

We are mounting the serviceAccountToken in as a projected volume, with a set experationSeconds. And then specifying it's location with `kubernetes_service_account_token_file`
When this gets rotated by k8s before expiration, pomerium does not update its token and the kubernetes API will return unauthenticated.

Pomerium should pick up the changed file, and reload the token and configuration.

I can see in the code that there's already functionality set up to hot-reload the configuration in case some of the other files have changed, so I expect it should be easy to also include the kubernetes service account token files for this. 
https://github.com/pomerium/pomerium/blob/master/config/config_source.go

## How'd it happen?

1. Deploy Pomerium in kubernetes, mounting a serviceAccountToken into the container.
2. kubectl calls via-pomerium will be successfull.
3. Wait `experationSeconds`-seconds until serviceAccountToken has expired.
4. kubectl calls via-pomerium will fail with an unauthenticated error.
   - `error: You must be logged in to the server (Unauthorized)`

- Pomerium version (retrieve with `pomerium --version`): 0.15.0
- Server Operating System/Architecture/Cloud: On-Prem OpenShift Kubernetes

## What's your config.yaml?

**config.yaml**
```config.yaml
  policy:
    - from: https://our-external-k8s-domain.com
      to: https://kubernetes.default.svc
      tls_skip_verify: true
      allow_spdy: true
      kubernetes_service_account_token_file: /var/run/secrets/tokens/pomerium-token
```

**deployment.yaml**
```deployment.yaml
      volumes:
        - name: pomerium-token
          projected:
               sources:
               - serviceAccountToken:
                   path: pomerium-token
                   expirationSeconds: 7200
                   audience: api
```

Calls to pomerium are authenticated, logs are clear from that side.

```logs
# Paste your logs here.
# Be sure to scrub any sensitive values
```

## Additional Context

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection

> The kubelet will request and store the token on behalf of the pod, make the token available to the pod at a configurable file path, and refresh the token as it approaches expiration. Kubelet proactively rotates the token if it is older than 80% of its total TTL, or if the token is older than 24 hours.
>
> The application is responsible for reloading the token when it rotates. Periodic reloading (e.g. once every 5 minutes) is sufficient for most use cases.

https://kubernetes.io/docs/concepts/storage/volumes/#example-configuration-secrets-nondefault-permission-mode

> The expirationSeconds is the expected duration of validity of the service account token. It defaults to 1 hour and must be at least 10 minutes (600 seconds).

calebdoxsey:
## Summary
Update the config source to watch policy files in addition to config changes.

## Related issues
Fixes #2722 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/config_source.go
+++ b/config/config_source.go
@@ -238,6 +238,16 @@ func (src *FileWatcherSource) check(ctx context.Context, cfg *Config) {
 		fs = append(fs, pair.CertFile, pair.KeyFile)
 	}
 
+	for _, policy := range cfg.Options.Policies {
+		fs = append(fs,
+			policy.KubernetesServiceAccountTokenFile,
+			policy.TLSClientCertFile,
+			policy.TLSClientKeyFile,
+			policy.TLSCustomCAFile,
+			policy.TLSDownstreamClientCAFile,
+		)
+	}
+
 	for _, f := range fs {
 		_, _ = h.Write([]byte{0})
 		bs, err := ioutil.ReadFile(f)
--- a/config/config_source_test.go
+++ b/config/config_source_test.go
@@ -27,9 +27,17 @@ func TestFileWatcherSource(t *testing.T) {
 		return
 	}
 
+	err = ioutil.WriteFile(filepath.Join(tmpdir, "kubernetes-example.txt"), []byte{1, 2, 3, 4}, 0o600)
+	if !assert.NoError(t, err) {
+		return
+	}
+
 	ssrc := NewStaticSource(&Config{
 		Options: &Options{
 			CAFile: filepath.Join(tmpdir, "example.txt"),
+			Policies: []Policy{{
+				KubernetesServiceAccountTokenFile: filepath.Join(tmpdir, "kubernetes-example.txt"),
+			}},
 		},
 	})
 
@@ -51,6 +59,17 @@ func TestFileWatcherSource(t *testing.T) {
 	case <-ch:
 	case <-time.After(time.Second):
 		t.Error("expected OnConfigChange to be fired after modifying a file")
+	}
+
+	err = ioutil.WriteFile(filepath.Join(tmpdir, "kubernetes-example.txt"), []byte{5, 6, 7, 8}, 0o600)
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	select {
+	case <-ch:
+	case <-time.After(time.Second):
+		t.Error("expected OnConfigChange to be fired after modifying a policy file")
 	}
 
 	ssrc.SetConfig(ctx, &Config{


gernest:
gentle ping @calebdoxsey 

Interesting, I was  working on fixing the same issue. Took similar approach, however I noticed that viper only watches the main configuration.

This means files referenced inside the main configuration (like in this case from `Policy`) will not trigger the reload when they change and happen to be outside the main configuration  directory.

Basically, while this will detect the changes ,it will not fix the issue when  only `kubernetes_service_account_token_file` file changed (Since we aren't watching this file so no reload will  trigger).

so, I had to port `Viper.WatchConfig` to support this scenario.

`watch_source.go`

```go
package config

import (
	"errors"
	"io"
	"log"
	"path/filepath"

)

var errDeletedMainConfig = errors.New("watcher: main configuration file was deleted")

// WatchConfig this is a port of (*Viper).WatchConfig modified to allow watching
// main configuration file and extra files that might be referenced by the main
// configuration file.
//
// The watching logic is:
//  - watch the mainConfigFile directory
//  - for every file in files
//    - ignore files sharing same directory with main configuration
//    - watch the file if it is outside the mainConfigFile directory
//  - exit the watching loop when mainConfigFile is deleted.
//
// This takes into account symlinks. onExit will be called with an error when
// the watching loop exits.
func WatchConfig(
	mainConfigFile string,
	files []string,
	onChange func(e fsnotify.Event),
	onExit func(err error),
) (io.Closer, error) {
	watcher, err := fsnotify.NewWatcher()
	if err != nil {
		return nil, err
	}

	// we have to watch the entire directory to pick up renames/atomic saves in a
	// cross-platform way.
	configFile := filepath.Clean(mainConfigFile)
	configDir, _ := filepath.Split(configFile)

	realConfigFile, _ := filepath.EvalSymlinks(mainConfigFile)
	realConfigDir, _ := filepath.Split(realConfigFile)

	// Keep track of mapping between watched/real files
	fileMaping := make(map[string]string)
	fileMaping[configFile] = realConfigFile
	watching := []string{configDir}
	for _, file := range files {
		file = filepath.Clean(file)
		realFile, _ := filepath.EvalSymlinks(file)
		if filepath.Dir(realFile) == realConfigDir {
			// This file is in the same directory as main configuration, so we skip
			// watching it.
			continue
		}
		fileMaping[file] = realFile
		watching = append(watching, file)
	}
	for _, file := range watching {
		err := watcher.Add(file)
		if err != nil {
			return nil, err
		}
	}
	go func() {
		for {
			select {
			case event, ok := <-watcher.Events:
				if !ok { // 'Events' channel is closed
					return
				}
				log.Println(event)
				currentFile := filepath.Clean(event.Name)
				originalRealFile := fileMaping[currentFile]

				currentRealFile, _ := filepath.EvalSymlinks(mainConfigFile)
				// we only care about the config file with the following cases:
				// 1 - if the config file was modified or created
				// 2 - if the real path to the config file changed (eg: k8s ConfigMap replacement)
				const writeOrCreateMask = fsnotify.Write | fsnotify.Create
				if (currentFile == configFile &&
					event.Op&writeOrCreateMask != 0) ||
					(currentRealFile != "" && currentRealFile != originalRealFile) {
					fileMaping[currentFile] = currentRealFile
					onChange(event)
				} else if currentFile == configFile &&
					event.Op&fsnotify.Remove&fsnotify.Remove != 0 {
					// The main configuration must never be deleted. For the event when it is
					// deleted from the system we exit this watching loop.
					if onExit != nil {
						onExit(errDeletedMainConfig)
					}
					return
				}
			case err, ok := <-watcher.Errors:
				if ok { // 'Errors' channel is not closed
					if onExit != nil {
						onExit(err)
					}
				}
				return
			}
		}
	}()
	return watcher, nil
}
```

And pair it with what you have done already.

Sorry for dumping this here, I was hoping to get the PR ready by tomorrow (I was writing unit tests), I hope this will be useful.

__some background__

I was looking for Go projects to contribute to (I'm into api gateways/reverse  proxies etc) and stumbled upon this project. Since I'm into k8s as well the issue was ticking all the boxes. 

I'm impressed with the quality of the code base. I'm a big fan.

NB: I can be completely wrong, I installed pomerium today.

calebdoxsey:
@gernest thanks for chiming in. We already have code to handle detecting changes to files referenced in the config file here: https://github.com/pomerium/pomerium/blob/master/config/config_source.go#L211

We have a layered set of config sources which are merged together to build the final options. The first source is the `FileOrEnvironmentSource` which uses viper. The second source is the `databroker` source so options can come directly from the databroker (this is primarily used as part of our enterprise product). The file change detection is added as the third source so we can pick up files from either source.

The change here was just to add a few more config settings to monitor since I had overlooked files referenced in the policy options. The kubernetes token option is different from the others in that it's not handled by Envoy, rather we construct the special header in the authorize service as part of policy enforcement. Given that, I wasn't sure our change detection would work properly, but as it turns out we rebuild the authorize evaluator whenever config is changed, so if we detect a file change we'll reload the token from disk at that point and the header should get updated properly.

gernest:
Thanks for clarifying.

desimone (approved):

PR #2772, opened 2021-11-17
envoy: add support for bind_config bootstrap options

calebdoxsey:
## Summary
Add support for Envoy `upstream_bind_config` options in the bootstrap: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto.html?highlight=bindconfig#config-bootstrap-v3-clustermanager

I'm not sure what these options are supposed to be used for but I tried setting `127.0.0.1` and it allowed access to local services, but not to services on the internet. I believe this is the intended functionality.

## Related issues
Fixes https://github.com/pomerium/internal/issues/628

## Checklist
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/reference/settings.yaml
+++ b/docs/reference/settings.yaml
@@ -764,11 +764,11 @@ settings:
           - Optional
         doc: |
           Setting `use_proxy_protocol` will configure Pomerium to require the [HAProxy proxy protocol](https://www.haproxy.org/download/1.9/doc/proxy-protocol.txt) on incoming connections. Versions 1 and 2 of the protocol are supported.
-      - name: "Envoy Admin Options"
-        keys: ["envoy_admin_options"]
+      - name: "Envoy Bootstrap Options"
+        keys: ["envoy_admin_address","envoy_admin_access_log_path","envoy_admin_profile_path","envoy_bind_config_freebind","envoy_bind_config_source_address"]
         attributes: |
           - Environment Variable: `ENVOY_ADMIN_ADDRESS`, `ENVOY_ADMIN_ACCESS_LOG_PATH`, `ENVOY_ADMIN_PROFILE_PATH`
-          - Config File Keys: `envoy_admin_address`, `envoy_admin_access_log_path`, `envoy_admin_profile_path`
+          - Config File Keys: `envoy_admin_address`, `envoy_admin_access_log_path`, `envoy_admin_profile_path`, `envoy_bind_config_freebind`, `envoy_bind_config_source_address`
           - Type: `string`
           - Optional
         doc: |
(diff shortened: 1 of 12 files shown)


travisgroth:
> I'm not sure what these options are supposed to be used for but I tried setting `127.0.0.1` and it allowed access to local services, but not to services on the internet. I believe this is the intended functionality.

👍  that behavior sounds correct for that configuration.

For reference/searchability - Beyond just offering more precise control over source address, it is also used to alter the outbound routing table when a system has multiple [vrfs](https://www.kernel.org/doc/Documentation/networking/vrf.txt).

docs/reference/settings.yaml
```diff
@@ -764,11 +764,11 @@ settings:
...
-        keys: ["envoy_admin_options"]
+      - name: "Envoy Bootstrap Options"
+        keys: ["envoy_admin_address","envoy_admin_access_log_path","envoy_admin_profile_path","envoy_bind_config_freebind","envoy_bind_config_source_address"]
         attributes: |
           - Environment Variable: `ENVOY_ADMIN_ADDRESS`, `ENVOY_ADMIN_ACCESS_LOG_PATH`, `ENVOY_ADMIN_PROFILE_PATH`
-          - Config File Keys: `envoy_admin_address`, `envoy_admin_access_log_path`, `envoy_admin_profile_path`
+          - Config File Keys: `envoy_admin_address`, `envoy_admin_access_log_path`, `envoy_admin_profile_path`, `envoy_bind_config_freebind`, `envoy_bind_config_source_address`
           - Type: `string`
           - Optional
         doc: |
```
travisgroth, line 774: The `doc` section should probably work in a link to the [cluster manager proto](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto.html#config-bootstrap-v3-clustermanager) for these options.

/cc @alexfornuto

travisgroth:
Looks like there's more to this.  Our control plane and communication to authorize and databroker are _also_ subject to this setting since they're represented by clusters internally.

I think we at least need an exception for control plane to always be `127.0.0.1` in the cluster specific settings, but I'm following up on some use cases to see what else might be in the details before we go further.

calebdoxsey:
I updated the config so that the `upstream_bind_config` is only set for route clusters and not globally (thus excluding the control-plane route).

alexfornuto:
I updated the docs. For review: https://github.com/pomerium/pomerium/pull/2772/commits/309a6b56f0229eeadff53a8fdeab7d57a35e61d9

docs/reference/settings.yaml
alexfornuto: @travisgroth: Updated, should be reviewed by an SME.

alexfornuto (approved):
👍, pending a 👍 from an SME on my docs additions.

travisgroth (approved):

PR #2788, opened 2021-11-29
Update mTLS guide

alexfornuto:
## Summary

Update with copy edits and formatted. Tech required no changes.

## Related issues

Resolves https://github.com/pomerium/docs/issues/52
Resolves https://github.com/pomerium/docs/issues/10

## Checklist

- [x] reference any related issues
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] add appropriate tag (`improvement` / `bug` / etc)
- [x] ready for review

PR #2797, opened 2021-12-01
DOCS: Create Consolidated Troubleshooting Guide and Replace FAQ

alexfornuto:
## Summary

- Removed FAQ.html (There were no FAQ only a couple troubleshooting items).
- Created troubleshooting.html.
  -  Moved troubleshooting content from old FAQ page here.
  -  Included troubleshooting from Partial files (in that form because they are non-contextual
  -  Added troubleshooting from specific pages that could be de-contextualized or are of a important enough context to warrant.

## Related issues

Resolves https://github.com/pomerium/docs/issues/57

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,144 @@
+---
+title: Troubleshooting
+description: >-
+  Learn how to troubleshoot common configuration issues or work around any outstanding bugs.
+sidebarDepth: 0
+lang: en-US
+meta:
+  - name: keywords
+    content: pomerium troubleshooting
```
desimone, line 9:
```suggestion
    content: pomerium troubleshooting faq frequently asked questions
```

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,144 @@
...
+---
+
+# Troubleshooting
+
+This article provides troubleshooting information for various tools and features from Pomerium.
```
desimone, line 14:
```suggestion
This article provides troubleshooting information for various tools and features in Pomerium.
```

defer to @cmo-pomerium tho

alexfornuto:
Depending on which PR is merged first, either this or #2788 will want to add [this section](https://github.com/pomerium/pomerium/pull/2788/files#diff-70e9de814a62daf00d6df168b376adbc23e7cabb2a371084e0da1e58291c6a89R1-R7) to this page.

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,144 @@
...
+::::
+:::: tab wget
+
+```bash
+wget -O /etc/upstream-service/jwks.json https://authenticate.localhost.pomerium.io/.well-known/pomerium/jwks.json
+```
+::::
+:::::
+
+Edit the upstream service configuration to use the local key to to verify tokens.
```
desimone, line 45:
```suggestion
Edit the upstream service configuration to use the local key to verify tokens.
```

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,144 @@
...
+
+```log
+Events:
+  Type     Reason       Age                 From              Message
+  ----     ------       ----                ----              -------
+  Normal   Updated      5m53s               pomerium-ingress  updated pomerium configuration
+  Warning  UpdateError  3s                  pomerium-ingress  upsert routes: parsing ingress: annotations: applying policy annotations: parsing policy: invalid rules in policy: unsupported conditional "maybe", only and, or, not, nor and action are allowed
+```
+
+#### HSTS
```
desimone, line 74: - We can pull up HSTS to a higher level vs just being k8s based
- Reference the full header name here too `Strict-Transport-Security` too for search ability
- We should probably update the reference docs where this is set (https://www.pomerium.io/reference/#set-response-headers) to mention and link to (I think Mozilla does a really nice job here) each of the default security headers we set by default. (e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,144 @@
...
+
+This article provides troubleshooting information for various tools and features from Pomerium.
+
+[[toc]]
+
+## Pomerium Core
+
+### JWT Authentication with Let's Encrypt
+
+Wildcard certificates signed by LetsEncrypt may still be cross-signed by the [expired DST R3 root](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/). While many browsers still trust these certificates (as long as they are also signed by a valid root), some upstream applications reject them, including Grafana as an example:
```
desimone, line 22: - Can we provide some additional context here around the different certificates at play? (JWKS / signing keys VS the lets encrypt certificates )
-  Maybe a little more explanation on why the LetsEncrypt issue results in the upstream client not being able to pull down and cache the jwks keys

desimone (changes requested):

docs/docs/troubleshooting.md
alexfornuto: Adjusted in https://github.com/pomerium/pomerium/pull/2797/commits/414d1cd70d80d03ec22f6947f31da44bb36530db

docs/docs/troubleshooting.md
alexfornuto: Troubleshooting page adjuster in https://github.com/pomerium/pomerium/pull/2797/commits/58101e1bda4b42e22b6d944e8d0da8f9b577a6bb
Reference page adjusted in https://github.com/pomerium/pomerium/pull/2797/commits/6cc4df64fb5272d0804e3aafe5ea0049e5c1c590

desimone (approved):

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+
+### JWT Authentication
+
+When securing the Pomerium Authenticate service with a certificate signed by Let's Encrypt, your upstream applications may reject the certificate when attempting to access the JWT signing key. Here's an example log line from [Grafana](/guides/grafana.md):
+
+```log
+logger=context error=Get "https://authenticate.localhost.pomerium.io/.well-known/pomerium/jwks.json": x509: certificate signed by unknown authority
+```
+
+This is often due to the recent expiration of the [DST Root CA X3](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/) certificate. Many default keystores used by docker images and less-frequently updated distributions still carry this expired certificate. Even though LE certs are cross-signed with the [ISRG Root X1](https://letsencrypt.org/certificates/) CA certificate, some applications will still reject them.
```
cmo-pomerium, line 38:
```suggestion
This is often due to the recent expiration of the [DST Root CA X3](https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/) certificate. Many default keystores used by docker images and less-frequently updated distributions still carry this expired certificate. Even though Let's Encrypt (LE) certs are cross-signed with the [ISRG Root X1](https://letsencrypt.org/certificates/) CA certificate, some applications will still reject them.
```

If we don't re-insert Let's Encrypt in the above, let's define it before using the LE acronym.

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+
+To clarify; this does not mean that the upstream service is rejecting the JWT singing key. Rather, it doesn't trust the Let's Encrypt certificate used by the Authorize service for TLS, and so it will not read the JWKS file.
```
cmo-pomerium, line 40:
```suggestion
To clarify; this does not mean that the upstream service is rejecting the JWT signing key. Rather, it doesn't trust the Let's Encrypt certificate used by the Authorize service for TLS, and so it will not read the JWKS file.
```
Assuming it's signing, not singing. Disregard if this is some specific jargon I don't know.

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+Events:
+  Type     Reason       Age                 From              Message
+  ----     ------       ----                ----              -------
+  Normal   Updated      5m53s               pomerium-ingress  updated pomerium configuration
+  Warning  UpdateError  3s                  pomerium-ingress  upsert routes: parsing ingress: annotations: applying policy annotations: parsing policy: invalid rules in policy: unsupported conditional "maybe", only and, or, not, nor and action are allowed
+```
+
+### Redirect Loop with Redis Databroker
+
+When using Redis, the [shared secret](/reference/readme.md#shared-secret) is used to encrypt data in Redis itself. if you change the configured shared secret, data from Redis can no longer be decrypted. This results in errant behavior, including redirect loops when a user session cannot be retrieved from the databroker.
```
cmo-pomerium, line 101:
```suggestion
When using Redis, the [shared secret](/reference/readme.md#shared-secret) is used to encrypt data in Redis itself. If you change the configured shared secret, data from Redis can no longer be decrypted. This results in errant behavior, including redirect loops when a user session cannot be retrieved from the databroker.
```

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+
+The resolution is to flush the Redis database with [`FLUSHDB`](https://redis.io/commands/flushdb) or [`FLUSHALL`](https://redis.io/commands/FLUSHALL).
+
+### RPC Errors
+
+#### certificate signed by unknown authority
```
cmo-pomerium, line 107: Should this all be in lower case? Not sure, just pointing it out

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+
+When authenticating and authorizing a user for the first time, you may see the following in your Pomerium logs:
+
+```log
+ERR http-error error="401 Unauthorized: ..... rpc error: code = DeadlineExceeded desc = latest connection error: connection error: desc = "transport: authentication handshake failed: x509: certificate signed by unknown authority...."
+```
+
+**Why**
```
cmo-pomerium, line 115:
```suggestion
**Why You Are Seeing This**
```

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+
+This error means that the proxy is rejecting the authorize service's supplied certificate (used to establish a secure connection) because it doesn't know or trust the certificate authority that is associated with the supplied certificate. This can happen for a few reasons.
```
cmo-pomerium, line 117:
```suggestion
This error means that the proxy is rejecting the authorized service's supplied certificate (used to establish a secure connection) because it doesn't know or trust the certificate authority that signed it. This can happen for a few reasons.
```
Regarding "This can happen for a few reasons" I propose explaining at least some of those reasons or just cutting it out.

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+
+#### rpc error: code = DeadlineExceeded
+
+When authenticating and authorizing a user for the first time, you may get the following in your Pomerium logs.
+
+```log
+{"level":"error",..."error":"rpc error: code = DeadlineExceeded desc = context deadline exceeded","http-message":"rpc error: code = DeadlineExceeded desc = context deadline exceeded","http-code":500,"message":"http-error"}
+```
+
+**Why**
```
cmo-pomerium, line 147:
```suggestion
**Why You Are Seeing This**
```

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+
+The proxy service is not able to create a connection with the authorization service to validate a user.
+
+**Solution**
+
+Usually, this is the result of either a routing issue or a configuration error. Make sure that you are using the _internally_ routable URL for authorize service. Many cloud loud balancers _do not_ yet support gRPC transposing the ingress. So while your authenticate service url will probably look like `https://authenticate.corp.example.com`, your authorizer service url will likely be more like `https://pomerium-authorize-service.default.svc.cluster.local` or `https://localhost:5443`.
```
cmo-pomerium, line 153:
```suggestion
Usually, this is the result of either a routing issue or a configuration error. Make sure that you are using the _internally_ routable URL for authorizing the service. Many cloud loud balancers _do not_ yet support gRPC transposing the ingress. So while your authenticate service url will probably look like `https://authenticate.corp.example.com`, your authorizer service url will likely be more like `https://pomerium-authorize-service.default.svc.cluster.local` or `https://localhost:5443`.
```
Should "ingress" be capitalized in this?

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Generate Recovery Token
+
+!!!include(generate-recovery-token.md)!!!
+
+## Miscellaneous
+
+### Invalid Certificates from Command Line Tools
+
+When using Let's Encrypt certificates, you must use the `fullchain.pem` file, not `cert.pem` in order to include intermediate certs. Browsers like Chrome store intermediate certs for LE but other tools (like `curl`) don't, which is why your route might look fine in a web browser, but not when curl'd or used for TCP tunneling.
```
cmo-pomerium, line 165:
```suggestion
When using Let's Encrypt certificates, you must use the `fullchain.pem` file, not `cert.pem` in order to include intermediate certs. Browsers like Chrome will store intermediate certs for LE but other tools (like `curl`) don't, which is why your route might look fine in a web browser, but not when curl'd or used for TCP tunneling.
```
Added "will" because my brain conjugated "Chrome store" as one noun and I had to re-read it again, so I doubt I'm the only one.

cmo-pomerium (commented):
Finished copy review. Proposed some changes and made some comments. Please review and ask me if there's any questions.

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,165 @@
...
+
+[[toc]]
+
+## Pomerium Core
+
+### HSTS
+
+By default, Pomerium sends the [`Strict-Transport-Security`](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) response header to the browser, which pins the certificate to our browser for one year. This is common best practice to help prevent man-in-the-middle attacks, but can create issues while a new Pomerium configuration is in development.
+
+When you visit an endpoint while Pomerium is using the self-signed bootstrap certificate or a Let's Encrypt staging certificate (before switching to a production certificate), the untrusted certificate may be pinned in your browser. It then must be reset once Pomerium is switched to production certificates.
```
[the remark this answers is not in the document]
alexfornuto: I'm not a fan of the "as yet" tense in a technical doc, and I do want to cite examples of untrusted certs. How about:

```suggestion
When you visit an endpoint while Pomerium is using an untrusted certificate (like the self-signed bootstrap certificate or a Let's Encrypt staging certificate), that certificate may be pinned in your browser. Once Pomerium is switched to a trusted production certificate, the untrusted cert must reset in the browser.
```

docs/docs/troubleshooting.md
alexfornuto: Yes, because in this case the header is a direct copy of an error message. This should only occur in H4s, and when it does we copy the case of the error message for optimum search.

docs/docs/troubleshooting.md
alexfornuto: This makes sense in this instance, but thinking ahead: as we create more troubleshooting entries related to specific errors and/or log output, I want to have a standard format. "Why are you seeing this" will not always make sense in each context, so the more generic "why" is preferred.

docs/docs/troubleshooting.md
alexfornuto: :+1:  on reducing to "signed it", but the "authorize service" is correct. Pomerium is broken down into components: authorize, proxy, and the data broker. So in this case this isn't an adjective but a noun. It would be more obvious if it were a proper noun "Authorize service".... CC @desimone to validate that idea.

```suggestion
This error means that the proxy is rejecting the Authorize service's supplied certificate (used to establish a secure connection) because it doesn't know or trust the certificate authority that signed it. This can happen for a few reasons.
```

> I propose explaining at least some of those reasons

I feel like this concern is addressed by the **Solution** section, which covers possible resolutions for several possible reasons. But if we don't have consensus on that idea, then cutting it out works for me.

```suggestion
This error means that the proxy is rejecting the authorize service's supplied certificate (used to establish a secure connection) because it doesn't know or trust the certificate authority that signed it.
```

docs/docs/troubleshooting.md
alexfornuto: As discussed above, authorize is a noun not an adjective in this context. But the grammar was off either way. I copied this from the old FAQ page and did not review/rewrite it.
```suggestion
Usually, this is the result of either a routing issue or a configuration error. Make sure that you are using the _internally_ routable URL for the authorize service. Many cloud loud balancers _do not_ yet support gRPC transposing the ingress. So while your authenticate service url will probably look like `https://authenticate.corp.example.com`, your authorizer service url will likely be more like `https://pomerium-authorize-service.default.svc.cluster.local` or `https://localhost:5443`.
```

re "ingress": in this context we're not explicitly talking about a Kubernetes object type, so we're not bound by their... questionable choices in capitalization.

alexfornuto:
I'll merge review suggestions after hearing back on the idea to capitalize Pomerium service names.

Issue #2794, opened 2021-11-30
Per-Route mTLS Fails with Non-Standard Ports


alexfornuto:
## What happened?

When using alternate ports for routes (not `443`), mTLS works when set globally with `client_ca_file`, but not when set per route with `tls_downstream_client_ca` or `tls_downstream_client_ca_file`. Removing the alternate port from the `from` key allowed for the client certificate to be confirmed, but of course breaks the route afterwards.

## What did you expect to happen?

I expected to be able to set mTLS on a per-route basis while using alternate ports for Pomerium routes.

## How'd it happen?

1. Went to a route protected by mTLS
2. Got HTTP error 495

## What's your environment like?

- Pomerium version: both 15.6 and 15.7
- Server Operating System/Architecture/Cloud: Running in docker (`pomerium/pomerium:latest) on unraid (Slackware 14.2) 

## What's your config.yaml?

```config.yaml
routes:
  - from: https://myroute.myDomain.tld:8444
    to: http://my.internal.ip:80
    tls_downstream_client_ca_file: "/pomerium/clientCA.pem" # or tls_downstream_client_ca: "base64-encoded string of clientCA.pem"
    policy:
      - allow:
          or:
            - domain:
                is: myDomain.com
            - email:
                is: myGmailAccount@gmail.com

```

## What did you see in the logs?

```logs
{"level":"info","service":"authorize","request-id":"4debe501-...","check-request-id":"860f446e-...","method":"GET","path":"/","host":"myroute.myDomain.tld:8444","query":"","session-id":"dcd011c6-...","allow":true,"deny":{"Status":495,"Message":"invalid client certificate"},"user":"110175062518023122375","email":"myUser@myDomain.com","databroker_server_version":11399670210472874068,"databroker_record_version":6750,"time":"2021-11-30T15:33:03-06:00","message":"authorize check"}
{"level":"info","service":"envoy","upstream-cluster":"","method":"GET","authority":"myroute.myDomain.tld:8444","path":"/","user-agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.45 Safari/537.36","referer":"","forwarded-for":"my.router.ip","request-id":"860f446e-...","duration":10.887377,"size":11219,"response-code":495,"response-code-details":"ext_authz_denied","time":"2021-11-30T15:33:03-06:00","message":"http-request"}
[...]

PR #2802, opened 2021-12-03
envoyconfig: fix tls_downstream_client_ca for non-standard ports

calebdoxsey:
## Summary
Fix route-specific client certificates when using a non-standard `from` port.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/2794

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/envoyconfig/listeners.go
+++ b/config/envoyconfig/listeners.go
@@ -853,7 +853,7 @@ func hostMatchesDomain(u *url.URL, host string) bool {
 func getPoliciesForDomain(options *config.Options, domain string) []config.Policy {
 	var policies []config.Policy
 	for _, p := range options.GetAllPolicies() {
-		if p.Source != nil && hostMatchesDomain(p.Source.URL, domain) {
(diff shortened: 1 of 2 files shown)


config/envoyconfig/listeners.go
```diff
@@ -853,7 +853,7 @@ func hostMatchesDomain(u *url.URL, host string) bool {
 func getPoliciesForDomain(options *config.Options, domain string) []config.Policy {
 	var policies []config.Policy
 	for _, p := range options.GetAllPolicies() {
-		if p.Source != nil && hostMatchesDomain(p.Source.URL, domain) {
```
calebdoxsey, line 856: The `hostMatchesDomain` function was intended to be used with a URL and a host, not a domain (e.g. `example.com:1234` not `example.com`). Confusingly the `Domain` in the name refers to how the matching is done not that it expects a domain.

wasaga (approved):

PR #2814, opened 2021-12-10
DOCS: New Guide: Upstream mTLS

alexfornuto:
## Summary

This new guide uses OpenSSL server and mkcert to demonstrate configuring Pomerium to provide a client certificate to an upstream service for mTLS authentication.

## Related issues

Resolves https://github.com/pomerium/docs/issues/51

## Checklist
- [x] https://github.com/pomerium/pomerium/pull/2788 **Must be merged** and this branch rebased, to avoid conflicts.
- [x] reference any related issues
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] add appropriate tag (`improvement` / `bug` / etc)
- [x] ready for review

PR #2746, opened 2021-11-04
Docs: Update JWT Verification Guide

alexfornuto:
## Summary

- Updates the JWT Verification guide.
- ~Creates a new topic page on mutual authentication~ Moved to a separate branch.

## Related issues

Closes https://github.com/pomerium/docs/issues/6

## Checklist

- [x] reference any related issues
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] add appropriate tag (`improvement` / `bug` / etc)
- [x] ready for review

PR #2820, opened 2021-12-13
Docs: Mutual Auth Topic page

alexfornuto:
## Summary

This new topics doc breaks down mutual authenticating into the components of JWT and mTLS for both the upstream and downstream connection.

[Preview](https://deploy-preview-2820--pomerium-docs.netlify.app/docs/topics/mutual-auth.html)

## Related issues

https://github.com/pomerium/pomerium/pull/2788
https://github.com/pomerium/pomerium/pull/2814
https://github.com/pomerium/pomerium/pull/2746

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/.vuepress/styles/index.styl
+++ b/docs/.vuepress/styles/index.styl
@@ -1,0 +1,7 @@
+hr {
+  border-top: 1px solid #000000;
+}
+blockquote {
+  color: #363636;
+  border-left: 0.25rem solid #363636;
+}
--- a/docs/.vuepress/styles/index.styl
+++ b/docs/.vuepress/styles/index.styl
@@ -1,7 +1,3 @@
 hr {
   border-top: 1px solid #000000;
 }
-blockquote {
-  color: #363636;
-  border-left: 0.25rem solid #363636;
-}
--- a/docs/docs/topics/mutual-auth.md
+++ b/docs/docs/topics/mutual-auth.md
@@ -72,7 +72,7 @@ flowchart TD
 
 While your network *should* be secured to only allow traffic at specified ports and directed to specified services, this creates a single point of failure. A hacker need only bypass your firewall to get direct access to your service.
 
-## JWT Verification: Internal Mutual Auth
+## JWT Verification: Application-based Mutual Authentication
 
 > - **Security**: Strong
 > - **Difficulty**: Moderate
@@ -130,7 +130,7 @@ flowchart TD
 
 In this way, we've applied a zero-trust security model to the application layer of our infrastructure's network model'.
 
-## mTLS
+## mTLS: Protocol-based Mutual Authentication
 
 > - **Security**: Excellent
 > - **Difficulty**: Hard
--- a/docs/docs/topics/mutual-auth.md
+++ b/docs/docs/topics/mutual-auth.md
@@ -11,15 +11,15 @@ description: >-
 
 # Mutual Authentication: A Component of Zero-Trust
 
-Pomerium provides a good layer of security out of the box, but it's not (and can't be) configured for complete [zero trust] right out of the box. This page explains several methods of achieving mutual authentication, a big part of the zero-trust model, with practical examples.
+Pomerium provides a good layer of security out of the box, but it's not (and can't be) configured for complete [zero trust] right out of the box. This page explains several methods of achieving mutual authentication — a big part of the zero-trust model — with practical examples.
 
-This is a long page that dives into several specific security practices that provide mutual authentication. You can use the table of contents below to narrow down to the specific tools you're interested in, or read the entire doc for a deeper understanding of how these tools work together to support strong infrastructure security.
+This is a nuanced topic that dives into several specific security practices that provide mutual authentication. You can use the table of contents below to narrow down to the specific tools you're interested in or read the entire doc for a deeper understanding of how these tools work together to support strong infrastructure security.
 
 [[toc]]
 
 ## What Is Mutual Authentication?
 
-The concept of mutual authentication is deceptively simple. It means that both sides of a connection can validate the identity of the other. The application of this concept however, can be varied and complex. Pomerium provides the features and capability to provide mutual authentication between itself and end users, as well, as between itself and upstream services, but configuring those external endpoints can vary depending on each service's features.
+The concept of mutual authentication is deceptively simple. It means that both sides of a connection can validate the identity of the other. The application of this concept, however, can be varied and complex. Pomerium provides the features and capability to provide mutual authentication between itself and end users, as well as between itself and upstream services, but configuring those external endpoints can vary depending on each service's features.
 
 ## Pomerium Defaults
 
@@ -28,7 +28,7 @@ The concept of mutual authentication is deceptively simple. It means that both s
 > - **Encrypted**: Yes (from the end user to Pomerium. It's up to the service to provide a TLS endpoint for Pomerium to use.)
 > - **Mutual Authentication**: None
 
-Let's look at a basic installation of Pomerium on a local network, with a single downstream service. This service contains sensitive data that we only want the right people to access.
+Let's look at a basic installation of Pomerium on a local network, with a single downstream service. This service contains sensitive data that we want to ensure is confidential and cannot be tampered with.
 
 ::: tip Note
 This section describes multiple services of Pomerium like Authenticate and Proxy as separate services. If you are running Pomerium in all-in-one mode, these will be a single process/container in your environment.
@@ -79,7 +79,7 @@ While your network *should* be secured to only allow traffic at specified ports 
 > - **Encrypted**: Yes (from the end user to Pomerium. It's up to the service to provide a TLS endpoint for Pomerium to use.)
 > - **Mutual Authentication**: Application Layer
 
-Many, but not all, modern web applications support json web tokens (**JWTs**). These tokens are provided by Pomerium (with the [`pass_identity_headers`] key) to the downstream service so that it can independently verify that the traffic it receives is authorized.
+Many, but not all, modern web applications support [json web tokens](https://datatracker.ietf.org/doc/html/rfc7519) (**JWTs**). These tokens can be provided by Pomerium (with the [`pass_identity_headers`] key) so that an upstream service can independently verify that the traffic has properly gone through Pomerium, and that the incoming user request is authorized.
 
 Let's look at an example, modeled from our [Grafana] integration guide:
 
@@ -153,7 +153,7 @@ flowchart RL
 
 1. The browser initiates a connection to `example.com` over the `http` protocol.
 1. The server sends its public certificate to the browser.
-1. The browser reads the certificate chain to find the CA, and checks against the computers keystore to see if the CA is one that it trusts.
+1. The browser reads the certificate chain to find the CA, and checks against the computer's keystore to see if the CA is one that it trusts.
 1. After confirming the CA is trusted the browser reconnects to the server, this time using the `https` protocol and encrypting the traffic using the public certificate.
 
 The process above confirms the identity of the *server*, protecting the client. Mutual TLS (**mTLS**) allows the server to confirm the identity of the *client* using a client certificate.
@@ -172,7 +172,7 @@ The exchange diagrammed here occurs *after* the initial connection and confirmat
 :::
 
 1. After the server certificate is trusted and an `HTTPS` connection is established, the server requests a client certificate.
-1. The user is prompted to use one of the certificates previously imported into thr browser. This certificate is send to the server
+1. The user is prompted to use one of the certificates previously imported into the browser. This certificate is sent to the server
 1. The server validates the client certificate signing authority against its trusted keystore or authorized client CA. Once authorized, the server resumes normal encrypted communication with the client.
 
 mTLS can also be configured between Pomerium and an upstream service. This secures the service itself from bad actors in your network by only allowing communication with the Pomerium proxy.
--- a/docs/docs/topics/mutual-auth.md
+++ b/docs/docs/topics/mutual-auth.md
@@ -50,7 +50,7 @@ flowchart LR
 
 1. When the user connects to the route `service.example.com` the traffic is received by Pomerium.
 1. Pomerium redirects the user to the identity provider (**IdP**) to sign in and validate their identity.
-1. After signing in to the IdP, Pomerium verifies that the authenticated (**authn**) user is authorized (**authz**) to access the service, and begins communication with it.
+1. After signing in to the IdP, Pomerium verifies that the authenticated (**AuthN**) user is authorized (**AuthZ**) to access the service, and begins communication with it.
 
 This is a good start, only the users that are supposed to have access to the service get through. But this model is dependent on the security of your network perimeter. If a bad actor gains access to your internal network, they can now communicate with the service directly:
 
@@ -131,6 +131,8 @@ In this way, we've applied a zero-trust security model to the application layer 
 > - **Encrypted**: Yes
 > - **Mutual Authentication**: Protocol Layer
 
+### [North-south mTLS](https://en.wikipedia.org/wiki/North-south_traffic)
+
 Most tech professionals are familiar with [Transport Layer Security] (**TLS**). The majority of traffic on the web today is sent using TLS. In addition to encrypting data using the server's TLS certificate, the server identity is validated by the certificate and the Certificate Authority (**CA**) that signed it.
 
 ```mermaid
@@ -171,7 +173,7 @@ flowchart RL
 1. After the server certificate is trusted and an `HTTPS` connection is established, the server requests a client certificate.
 1. The user is prompted to use one of the certificates previously imported into the browser. This certificate is sent to the server
 1. The server validates the client certificate signing authority against its trusted keystore or authorized client CA. Once authorized, the server resumes normal encrypted communication with the client.
-
+### [East-west mTLS](https://en.wikipedia.org/wiki/East-west_traffic)
 mTLS can also be configured between Pomerium and the end user, and/or with an upstream service. Configuring mTLS for many end users is cumbersome, and new technologies like [device identity verification] provide verification of the user and their hardware. But mTLS between Pomerium and an upstream service need only be configured once per service, and minimally maintained by updating certificates.
 
 ```mermaid
--- a/docs/docs/topics/mutual-auth.md
+++ b/docs/docs/topics/mutual-auth.md
@@ -108,7 +108,7 @@ flowchart TD
     C -.-> B
 ```
 
-In this way, we've applied a zero-trust security model to the authZ aspect of our infrastructure security.
+In this way, we've applied a zero-trust security model to the application layer of our infrastructure's network model'.
 
 ## mTLS
 
@@ -167,7 +167,7 @@ flowchart TD
     B---xE
 ```
 
-In this way, we've now applied a zero-trust security model to the authN aspect of our infrastructure security.
+In this way, we've applied a zero-trust security model to the protocol layer of our infrastructure's network model'.
 
 ## Mutual Authentication With a Sidecar
 
--- a/docs/docs/topics/mutual-auth.md
+++ b/docs/docs/topics/mutual-auth.md
@@ -1,0 +1,175 @@
+---
+title: Mutual Authentication
+lang: en-US
+meta:
+  - name: keywords
+    content: pomerium identity-access-proxy mutual authentication jwt jwks mtls
+description: >-
+  This page describes the concept of mutual authentication and why it's important.
+---
+
+# Mutual Authentication: A Component of Zero-Trust
+
+Pomerium provides a good layer of security out of the box, but it's not (and can't be) configured for complete [zero trust] right out of the box. This page explains the concepts of zero trust with practical examples.
+
+## Pomerium Default
+
+Let's look at a basic installation of Pomerium on a local network, with a single downstream service. This service contains sensitive data that we only want the right people to access.
+
+::: tip Note
+This section describes multiple services of Pomerium like Authenticate and Proxy as separate services. If you are running Pomerium in all-in-one mode, these will be a single process/container in your environment.
+:::
+
+```mermaid
+flowchart TD
+    D[End User]
+        D <--> A
+    subgraph LAN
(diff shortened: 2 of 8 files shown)


docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,234 @@
...
+    B(Pomerium Authenticate Service)
+    A <-.-> B
+    C(Service)
+    A <--> C
+    end
+```
+
+While your network *should* be secured to only allow traffic at specified ports and directed to specified services, this creates a single point of failure. A hacker need only bypass your firewall to get direct access to your service.
+
+## JWT Verification: Internal Mutual Auth
```
desimone, line 75:
```suggestion
## JWT Verification: Application-based Mutual Authentication
```

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,259 @@
...
+            C(Grafana)
+    end
+    A <-.-> B
+    A <--> C
+    C -.-> B
+```
+
+In this way, we've applied a zero-trust security model to the application layer of our infrastructure's network model'.
+
+## mTLS
```
desimone, line 133:
```suggestion
## Mutual TLS (mTLS)
```

desimone, line 133:
```suggestion
## mTLS: Protocol-based Mutual Authentication
```
or  (trying to make it match the jwt one)

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+meta:
+  - name: keywords
+    content: pomerium identity-access-proxy mutual authentication jwt jwks mtls
+description: >-
+  This page describes the concept of mutual authentication and why it's important.
+---
+
+# Mutual Authentication: A Component of Zero-Trust
+
+Pomerium provides a good layer of security out of the box, but it's not (and can't be) configured for complete [zero trust] right out of the box. This page explains several methods of achieving mutual authentication, a big part of the zero-trust model, with practical examples.
```
desimone, line 14:
```suggestion
Pomerium provides a good layer of security out of the box, but it's not (and can't be) configured for complete [zero trust] right out of the box. This page explains several methods of achieving mutual authentication — a big part of the zero-trust model — with practical examples.
```

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+
+This is a long page that dives into several specific security practices that provide mutual authentication. You can use the table of contents below to narrow down to the specific tools you're interested in, or read the entire doc for a deeper understanding of how these tools work together to support strong infrastructure security.
```
desimone, line 16:
```suggestion
This is a nuanced topic that dives into several specific security practices that provide mutual authentication. You can use the table of contents below to narrow down to the specific tools you're interested in or read the entire doc for a deeper understanding of how these tools work together to support strong infrastructure security.
```

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+    style LAN stroke-dasharray: 5 5
+    A(Pomerium Proxy Service)
+    B(Pomerium Authenticate Service)
+    A <-.-> B
+    C(Service)
+    A <--> C
+    end
+```
+
+While your network *should* be secured to only allow traffic at specified ports and directed to specified services, this creates a single point of failure. A hacker need only bypass your firewall to get direct access to your service.
```
desimone, line 73: >> A hacker **need** only bypass your firewall to get direct access to your service.

I think the verb is wrong here.

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+
+::: tip Note
+This section describes multiple services of Pomerium like Authenticate and Proxy as separate services. If you are running Pomerium in all-in-one mode, these will be a single process/container in your environment.
+:::
+
+```mermaid
+flowchart TD
+    D[End User]
+        D <--> A
+    subgraph LAN
```
desimone, line 41:
```suggestion
    subgraph LAN
```

Can we update all the the diagrams to call this "internal network" vs LAN?

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+The concept of mutual authentication is deceptively simple. It means that both sides of a connection can validate the identity of the other. The application of this concept however, can be varied and complex. Pomerium provides the features and capability to provide mutual authentication between itself and end users, as well, as between itself and upstream services, but configuring those external endpoints can vary depending on each service's features.
+
+## Pomerium Defaults
+
+> - **Security**: Moderate
+> - **Difficulty**: Easy
+> - **Encrypted**: Yes (from the end user to Pomerium. It's up to the service to provide a TLS endpoint for Pomerium to use.)
+> - **Mutual Authentication**: None
+
+Let's look at a basic installation of Pomerium on a local network, with a single downstream service. This service contains sensitive data that we only want the right people to access.
```
desimone, line 31:
```suggestion
Let's look at a basic installation of Pomerium on a local network, with a single upstream service. This service contains sensitive data that we only want the right people to access.
```

We should add a terminology page like envoy. It's so confusing.

https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/intro/terminology

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+While your network *should* be secured to only allow traffic at specified ports and directed to specified services, this creates a single point of failure. A hacker need only bypass your firewall to get direct access to your service.
+
+## JWT Verification: Application-based Mutual Authentication
+
+> - **Security**: Strong
+> - **Difficulty**: Moderate
+> - **Encrypted**: Yes (from the end user to Pomerium. It's up to the service to provide a TLS endpoint for Pomerium to use.)
+> - **Mutual Authentication**: Application Layer
+
+Many, but not all, modern web applications support json web tokens (**JWTs**). These tokens are provided by Pomerium (with the [`pass_identity_headers`] key) to the downstream service so that it can independently verify that the traffic it receives is authorized.
```
desimone, line 82:
```suggestion
Many, but not all, modern web applications support [json web tokens](https://datatracker.ietf.org/doc/html/rfc7519) (**JWTs**). These tokens can be provided by Pomerium (with the [`pass_identity_headers`] key) so that an upstream service can independently verify that the traffic has properly gone through Pomerium, and that the incoming user request is authorized.
```

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+        end
+            C(Grafana)
+    end
+    A <-.-> B
+    A <--> C
+    C -.-> B
+```
+
+1. When the user connects to the route `grafana.example.com` the traffic is received by the Pomerium Proxy service.
+1. The proxy redirects to the Authenticate service to validate the user by having them sign in with the identity provider (**IdP**).
```
desimone, line 105:
```suggestion
1. The proxy redirects to the `Pomerium Authenticate` service to validate the user by having them sign in with the identity provider (**IdP**).
```

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+            B(Pomerium Authenticate Service)
+        end
+            C(Grafana)
+    end
+    A <-.-> B
+    A <--> C
+    C -.-> B
+```
+
+1. When the user connects to the route `grafana.example.com` the traffic is received by the Pomerium Proxy service.
```
desimone, line 104:
```suggestion
1. When the user connects to the route `grafana.example.com` the traffic is received by the `Pomerium Proxy` service.
```

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+    E---xC
+    D<-->A
+    subgraph LAN
+    style LAN stroke-dasharray: 5 5
+        subgraph Pomerium[ ];
+        style Pomerium stroke-width: 0
+            A(Pomerium Proxy Service)
+            B(Pomerium Authenticate Service)
+        end
+            C(Grafana)
```
desimone, line 124:
```suggestion
            C(Grafana)
```

It's confusing to jump between examples because in some places the upstream service is called `Service` and in others `Grafana`. We should be consistent.

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+    subgraph Server
+    D-. http .-> A[Service]
+    A-. Certificate .->D
+    end
+    D<-- https -->A
+```
+
+1. The browser initiates a connection to `example.com` over the `http` protocol.
+1. The server sends its public certificate to the browser.
+1. The browser reads the certificate chain to find the CA, and checks against the computers keystore to see if the CA is one that it trusts.
```
desimone, line 156:
```suggestion
1. The browser reads the certificate chain to find the CA, and checks against the computer's keystore to see if the CA is one that it trusts.
```

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+
+Pomerium provides a good layer of security out of the box, but it's not (and can't be) configured for complete [zero trust] right out of the box. This page explains several methods of achieving mutual authentication, a big part of the zero-trust model, with practical examples.
+
+This is a long page that dives into several specific security practices that provide mutual authentication. You can use the table of contents below to narrow down to the specific tools you're interested in, or read the entire doc for a deeper understanding of how these tools work together to support strong infrastructure security.
+
+[[toc]]
+
+## What Is Mutual Authentication?
+
+The concept of mutual authentication is deceptively simple. It means that both sides of a connection can validate the identity of the other. The application of this concept however, can be varied and complex. Pomerium provides the features and capability to provide mutual authentication between itself and end users, as well, as between itself and upstream services, but configuring those external endpoints can vary depending on each service's features.
```
desimone, line 22:
```suggestion
The concept of mutual authentication is deceptively simple. It means that both sides of a connection can validate the identity of the other. The application of this concept, however, can be varied and complex. Pomerium provides the features and capability to provide mutual authentication between itself and end users, as well as between itself and upstream services, but configuring those external endpoints can vary depending on each service's features.
```

docs/docs/topics/mutual-auth.md
desimone, line 31:
```suggestion
Let's look at a basic installation of Pomerium on a local network, with a single downstream service. This service contains sensitive data that we want to ensure is confidential and cannot be tampered with.
```

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+
+> - **Security**: Moderate
+> - **Difficulty**: Easy
+> - **Encrypted**: Yes (from the end user to Pomerium. It's up to the service to provide a TLS endpoint for Pomerium to use.)
+> - **Mutual Authentication**: None
+
+Let's look at a basic installation of Pomerium on a local network, with a single downstream service. This service contains sensitive data that we only want the right people to access.
+
+::: tip Note
+This section describes multiple services of Pomerium like Authenticate and Proxy as separate services. If you are running Pomerium in all-in-one mode, these will be a single process/container in your environment.
```
desimone, line 34: See my combined PR suggestion.

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,306 @@
...
+    D(Browser)-- Client Certificate -->A
+    A-- Client Certificate Request --->D
+```
+
+::: tip
+The exchange diagrammed here occurs *after* the initial connection and confirmation of the server-side certificate shown above.
+:::
+
+1. After the server certificate is trusted and an `HTTPS` connection is established, the server requests a client certificate.
+1. The user is prompted to use one of the certificates previously imported into thr browser. This certificate is send to the server
```
desimone, line 175:
```suggestion
1. The user is prompted to use one of the certificates previously imported into the browser. This certificate is sent to the server
```

desimone (commented):
Looks good overall. 

1. I suggest we remove references to the authenticate service because I think the focus of this doc is on mutual auth between Pomerium and Upstream services, and not between pomerium services. I think we can also skip discussing the overall auth flow / relationship between the services which is already described in the architecture page (https://deploy-preview-2820--pomerium-docs.netlify.app/docs/architecture.html#authentication-flow). For example, there are dotted lines between both grafana and authN and authN and proxy. Does this mean the traffic is in plaintext? Or that it's a redirect flow? Or something else? I think it's better we simplify the model if possible and focus on the communication between Pomerium (as a combined product) and the upstream services. 
2. Can we update the diagrams to call `LAN` --> `Internal network`? 
3. I think we should standardize on the generic term "Upstream service A/B/C/etc.." in place of Grafana, API, User App, admin App and so on. I understand that those apps reference other guides, which we can mention in text ("Here the upstream service would be Grafana from the Grafana guide") but I think it makes sense to have the diagrams themselves be internally consistent to each-other. 

Thoughts?


I'm halfway through the review but wanted to stop there to not overwhelm / if further discussion makes sense.

desimone (changes requested):
Left comment, not request changes by accident.

docs/docs/topics/mutual-auth.md
alexfornuto: > However, in a rather formal style of English, there also exists an auxiliary verb ("helping verb") need, which is followed by a bare infinitive, and is only used in negative polarity contexts: with not, with only, in questions, and so on. Additionally, it only exists in the present tense. Aside from these two restrictions, it's very much like other auxiliary verbs (can, may, should, etc.): "We need not discuss it."
> 
> So your first example ("[…] one need only look […]") is correct — using the auxiliary verb need — whereas your second example ("[…] one needs only look […]") is wrong — it should either be "one need only look" (using the auxiliary verb), or "one only needs to look" (using the regular verb).

From [english.stackexchange](https://english.stackexchange.com/questions/297233/is-one-needs-only-or-one-need-only-correct).

docs/docs/topics/mutual-auth.md
alexfornuto: I can do that.

docs/docs/topics/mutual-auth.md
alexfornuto: I get the addition of Pomerium, but why make it an inline code snippet?

docs/docs/topics/mutual-auth.md
desimone: @alexfornuto  should we track that effort separately? I'm sure there's other terminology that would fit the bill. Some that come to mind are:

- upstream / downstream
- identity provider
- secure enclave / TEE / security keys
- peremeter / zero trust
- least user privilege
- Namespaces
- mutual authentication
- authentication
- authorization


and so on. Even if they are just links to other resources and our own docs.

alexfornuto: Yea, I'm thinking of a separate page, linking the first instance of each term in a page to it with anchors. 

In a future iteration of docs, we can make reusable tooltips to define concepts. See [this example](https://pantheon.io/docs/style-guide#tooltips).

docs/.vuepress/styles/index.styl
```diff
@@ -0,0 +1,7 @@
+hr {
+  border-top: 1px solid #000000;
+}
+blockquote {
```
desimone, line 4: Do we want to adjust _all_ of the block quotes? For example, now the actual quotes look a little funky. https://deploy-preview-2820--pomerium-docs.netlify.app/docs/background.html

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,390 @@
...
+While your network *should* be secured to only allow traffic at specified ports and directed to specified services, this creates a single point of failure. A hacker need only bypass your firewall to get direct access to your service.
+
+## JWT Verification: Application-based Mutual Authentication
+
+> - **Security**: Strong
+> - **Difficulty**: Moderate
+> - **Encrypted**: Yes (from the end user to Pomerium. It's up to the service to provide a TLS endpoint for Pomerium to use.)
+> - **Mutual Authentication**: Application Layer
+
+Many, but not all, modern web applications support [json web tokens][jwt-rfc] (**JWTs**). These tokens can be provided by Pomerium (with the [`pass_identity_headers`] key) so that an upstream service can independently verify that the traffic has properly gone through Pomerium, and that the incoming user request is authorized.
```
desimone, line 79:
```suggestion
Many, but not all, modern web applications support [JSON web tokens][jwt-rfc] (**JWTs**). These tokens can be provided by Pomerium (with the [`pass_identity_headers`] key) so that an upstream service can independently verify that the traffic has properly gone through Pomerium, and that the incoming user request is authorized.
```

I believe JSON should be capitalized. 

https://en.wikipedia.org/wiki/JSON

desimone: There may be other instances I missed.

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,390 @@
...
+    style lan stroke-dasharray: 5 5
+    A(Pomerium)
+    B(Upstream Service)
+    A --- B
+  end
+```
+
+1. When the user connects to the route `service.example.com` the traffic is received by Pomerium.
+1. Pomerium redirects the user to the identity provider (**IdP**) to sign in and validate their identity.
+1. After signing in to the IdP, Pomerium verifies that the authenticated (**authn**) user is authorized (**authz**) to access the service, and begins communication with it.
```
desimone, line 53:
```suggestion
1. After signing in to the IdP, Pomerium verifies that the authenticated (**AuthN**) user is authorized (**AuthZ**) to access the service, and begins communication with it.
```

More of a convention but I typically see these mixed cased. There may be other instances in this docs but I'm just flagging the first.

desimone: https://en.wikipedia.org/wiki/Web_API_security

> Along with the ease of API integrations come the difficulties of ensuring proper authentication (AuthN) and authorization (AuthZ).

docs/docs/topics/mutual-auth.md
desimone: TIL thanks for the explanation.

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,390 @@
...
+In this way, we've applied a zero-trust security model to the application layer of our infrastructure's network model. You can see JWT verification in practice with our [Grafana] integration guide.
+
+## mTLS: Protocol-based Mutual Authentication
+
+> - **Security**: Excellent
+> - **Difficulty**: Hard
+> - **Encrypted**: Yes
+> - **Mutual Authentication**: Protocol Layer
+
+Most tech professionals are familiar with [Transport Layer Security] (**TLS**). The majority of traffic on the web today is sent using TLS. In addition to encrypting data using the server's TLS certificate, the server identity is validated by the certificate and the Certificate Authority (**CA**) that signed it.
```
desimone, line 134: Maybe we should add a subheading clarifying mTLS between browser and pomerium and pomerium and upstreams?

desimone, line 134:
```suggestion
### [North-south mTLS](https://en.wikipedia.org/wiki/North-south_traffic)

Most tech professionals are familiar with [Transport Layer Security] (**TLS**). The majority of traffic on the web today is sent using TLS. In addition to encrypting data using the server's TLS certificate, the server identity is validated by the certificate and the Certificate Authority (**CA**) that signed it.
```

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,390 @@
...
+  A-. Server Certificate ..->D
+  D<==>A
+  style pc fill: white, stroke: black
+  style Server fill: white, stroke: black
+```
+
+1. After the server certificate is trusted and an `HTTPS` connection is established, the server requests a client certificate.
+1. The user is prompted to use one of the certificates previously imported into the browser. This certificate is sent to the server
+1. The server validates the client certificate signing authority against its trusted keystore or authorized client CA. Once authorized, the server resumes normal encrypted communication with the client.
+
```
desimone, line 174:
```suggestion
### [East-west mTLS](https://en.wikipedia.org/wiki/East-west_traffic)
```

https://en.wikipedia.org/wiki/East-west_traffic

desimone: Though we should probably add this to the terminology doc too.

docs/docs/topics/mutual-auth.md
```diff
@@ -0,0 +1,390 @@
...
+- In this example, End users authenticate with Pomerium to access the user app:
+  - Their connection to Pomerium is authenticated after signing in to their identity provider, and encrypted with TLS.
+  - The communication between Pomerium and the user app is mutually authenticated with mTLS, and the user's JWT is passed from Pomerium to the app to confirm the user identity to the service. The service validates the JWT against Pomerium's signing Key.
+- The admin also authenticates with Pomerium to access their services:
+  - The connection is authenticated and encrypted the same as the user's, but the route to the admin app also requires a client certificate for mutual authentication at the protocol layer.
+  - The connections from the admin to the admin services are now mutually authenticated both upstream and downstream from Pomerium.
+- The API server is required by both the user and admin apps. Rather than build authentication into the API service, it is accessible only through a sidecar. The sidecar only accepts mTLS-authenticated connections from the proxy, so both apps connect to the API through Pomerium.
+
+## Legend
```
desimone, line 302: Could we move this to the top of the doc?

docs/.vuepress/styles/index.styl
desimone: It might just be easier to not have those items in a quote block. It looks fine to me just as bullets. 

Screen Shot 2021-12-30 at 11 49 31 AM

desimone (changes requested):
LGTM pending a few small changes.

docs/docs/topics/mutual-auth.md
alexfornuto: Discussed in Slack.

docs/docs/topics/mutual-auth.md
alexfornuto: Resolved by review changes.

desimone (approved):

PR #2826, opened 2021-12-16
move NewGRPCClientConn to public package

wasaga:
## Summary

`NewGRPCClientConn` snippet exists with minor variations in core, console, ingress-controller and cli. Moving to public package so that other repos can just use it directly.  

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/grpcutil/client.go
+++ b/pkg/grpcutil/client.go
@@ -85,6 +85,7 @@ func NewGRPCClientConn(ctx context.Context, opts *Options, other ...grpc.DialOpt
 		}
 
 		cert := credentials.NewTLS(&tls.Config{
+			// nolint: gosec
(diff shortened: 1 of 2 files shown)


desimone (approved):

pkg/grpcutil/client.go
```diff
@@ -90,7 +85,6 @@ func NewGRPCClientConn(ctx context.Context, opts *Options, other ...grpc.DialOpt
 		}
 
 		cert := credentials.NewTLS(&tls.Config{
-			// nolint: gosec
```
calebdoxsey, line 93: I think this is required to placate linting

calebdoxsey (approved):

PR #2156, opened 2021-04-29
config: add support for codec_type

calebdoxsey:
## Summary
This PR adds a new configuration option to specify the codec type. Options are:

- `auto`: codec is chosen based on ALPN or protocol inference
- `http1`: use HTTP/1.1 explicitly
- `http2`: use HTTP/2 explicitly

In all-in-one mode `http1` is the default. Otherwise `auto` is the default.

## Related issues
- #2150 

## Checklist
- [ ] reference any related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] add appropriate tag (`improvement` / `bug` / etc)
- [ ] ready for review

PR #2839, opened 2021-12-21
set default codec type to auto/http1

wasaga:
## Summary

If left blank, `codec_type` would not be set to `auto`, and `http1` would not be enforced,
which caused issues with i.e. Safari.

## Related issues

Fixes https://github.com/pomerium/internal/issues/707

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
Why are we making this change? I don't remember what the default behavior was supposed to be so I can't reason about what will happen if we do this. I would guess it would break something, but maybe this is what we want?

wasaga:
without this fix, https://github.com/pomerium/pomerium/pull/2156 doesn't work unless you explicitly set `codec_type: auto` in the config file. If you just omit this setting, it would be set to empty string and the following condition wouldn't trigger. 

https://github.com/pomerium/pomerium/pull/2156/files#diff-643b22ef1a79b955d4198f94014ce060da1e0c28279396877e5eaca9f5247253R948-R950

this PR fixes it by explicitly setting `codec_type: auto` by default.

calebdoxsey (approved):

calebdoxsey:
According to the docs the default is `AUTO` if not set: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-enum-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-codectype

wasaga:
I think previously when it wasn't set, then `default` clause would be executed here: 

https://github.com/pomerium/pomerium/blob/70e0e866fc5cac54978b05ead03d38117e8af991/config/envoyconfig/listeners.go#L651-L659

PR #2892, opened 2022-01-04
authenticate: add device-enrolled page

### LINKED ISSUE #2891: feat: display enrollment confirmation page following device registration

It's not clear if a user has successfully enrolled their device with Pomerium. 

- We should have successful registration endpoint (a under `.pomerium`) on the authenticate service that notifies the user that they've successfully enrolled their device. 
- We should show the user the device enrollment id (maybe a short hash which they can compare to admin logs? for any debugging needs) on the thank you screen.
- We should show any errors during enrollment to the end user if they can occur. 

In the enterprise version of Pomerium, this would serve as a reasonable default redirect url following TOFU registration. 

**Additional context**

- https://github.com/pomerium/pomerium-console/issues/2232

calebdoxsey:
## Summary
Add a new device-enrolled page that can be shown after a device is enrolled successfully.

## Related issues
Fixes #2891 

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authenticate/handlers.go
+++ b/authenticate/handlers.go
@@ -18,6 +18,7 @@ import (
 	"golang.org/x/oauth2"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	"github.com/pomerium/pomerium/authenticate/handlers"
 	"github.com/pomerium/pomerium/authenticate/handlers/webauthn"
 	"github.com/pomerium/pomerium/internal/httputil"
 	"github.com/pomerium/pomerium/internal/identity"
@@ -98,6 +99,7 @@ func (a *Authenticate) mountDashboard(r *mux.Router) {
 	sr.Path("/sign_in").Handler(a.requireValidSignature(a.SignIn))
 	sr.Path("/sign_out").Handler(a.requireValidSignature(a.SignOut))
 	sr.Path("/webauthn").Handler(webauthn.New(a.getWebauthnState))
+	sr.Path("/device-enrolled").Handler(handlers.DeviceEnrolled())
 }
 
 func (a *Authenticate) mountWellKnown(r *mux.Router) {
--- a/authenticate/handlers/device-enrolled.go
+++ b/authenticate/handlers/device-enrolled.go
@@ -6,18 +6,13 @@ import (
 
 	"github.com/pomerium/pomerium/internal/frontend"
 	"github.com/pomerium/pomerium/internal/httputil"
-	"github.com/pomerium/pomerium/internal/urlutil"
 )
 
 // DeviceEnrolled displays an HTML page informing the user that they've successfully enrolled a device.
 func DeviceEnrolled() http.Handler {
 	tpl := template.Must(frontend.NewTemplates())
-	type TemplateData struct {
-		DeviceCredentialID string
-	}
+	type TemplateData struct{}
 	return httputil.HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
-		return tpl.ExecuteTemplate(w, "device-enrolled.html", TemplateData{
-			DeviceCredentialID: r.FormValue(urlutil.QueryDeviceCredentialID),
-		})
+		return tpl.ExecuteTemplate(w, "device-enrolled.html", TemplateData{})
 	})
 }
--- a/authenticate/handlers/device-enrolled.go
+++ b/authenticate/handlers/device-enrolled.go
@@ -1,0 +1,23 @@
+package handlers
+
+import (
+	"html/template"
+	"net/http"
+
+	"github.com/pomerium/pomerium/internal/frontend"
+	"github.com/pomerium/pomerium/internal/httputil"
+	"github.com/pomerium/pomerium/internal/urlutil"
+)
+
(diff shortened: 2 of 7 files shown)


wasaga (approved):

authenticate/handlers.go
```diff
@@ -98,6 +99,7 @@ func (a *Authenticate) mountDashboard(r *mux.Router) {
 	sr.Path("/sign_in").Handler(a.requireValidSignature(a.SignIn))
 	sr.Path("/sign_out").Handler(a.requireValidSignature(a.SignOut))
 	sr.Path("/webauthn").Handler(webauthn.New(a.getWebauthnState))
+	sr.Path("/device-enrolled").Handler(handlers.DeviceEnrolled())
```
desimone, line 102: Should we ensure that the origin of the request is Pomerium (e.g. is valid signed / hmac'd?)

authenticate/handlers/webauthn/webauthn.go
```diff
@@ -302,7 +302,17 @@ func (h *Handler) handleRegister(w http.ResponseWriter, r *http.Request, state *
...
+	// add the device credential ID to the redirect uri
+	redirectURI, err := urlutil.ParseAndValidateURL(redirectURIParam)
+	if err != nil {
+		return err
+	}
+	q := redirectURI.Query()
+	q.Set(urlutil.QueryDeviceCredentialID, deviceCredentialID)
+	redirectURI.RawQuery = q.Encode()
+
+	return h.saveSessionAndRedirect(w, r, state, redirectURI.String())
```
desimone, line 315: I believe this creates a signed url right?

authenticate/handlers.go
desimone: It's becoming a little hard to follow all the different handlers getting mounted here. 

We have handlers in this directory, `webauthn` and now /handlers. Thoughts on adding these handler to the existing `webauthn` package?

authenticate/handlers/device-enrolled.go
```diff
@@ -0,0 +1,23 @@
...
+	"net/http"
+
+	"github.com/pomerium/pomerium/internal/frontend"
+	"github.com/pomerium/pomerium/internal/httputil"
+	"github.com/pomerium/pomerium/internal/urlutil"
+)
+
+// DeviceEnrolled displays an HTML page informing the user that they've successfully enrolled a device.
+func DeviceEnrolled() http.Handler {
+	tpl := template.Must(frontend.NewTemplates())
```
desimone, line 14: Does this get evaluated on each call?

authenticate/handlers/device-enrolled.go
```diff
@@ -0,0 +1,23 @@
...
+	type TemplateData struct {
+		DeviceCredentialID string
+	}
+	return httputil.HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
+		return tpl.ExecuteTemplate(w, "device-enrolled.html", TemplateData{
+			DeviceCredentialID: r.FormValue(urlutil.QueryDeviceCredentialID),
```
desimone, line 20: I might have missed it but it doesn't look like this url's HMAC is checked. It's probably not a huge deal but it could mean this value could be spoofed, and this url could be redirected to from another source besides pomerium.

desimone (changes requested):

authenticate/handlers/webauthn/webauthn.go
calebdoxsey: I did not change the behavior of this.

authenticate/handlers/device-enrolled.go
calebdoxsey: no

authenticate/handlers/device-enrolled.go
calebdoxsey: It is not. If I HMAC it we will have a double-HMAC'd URL that will expire almost immediately. I can do this but it won't be useful for what this page was supposed to be since we're generating links which last a very long time. (the redirect URI is a component of the generated link)

authenticate/handlers.go
calebdoxsey: I don't think it belongs there, but sure I can move it.

authenticate/handlers/device-enrolled.go
desimone: Understood. I was thinking that the webauthn handler would create a signed url on redirect. I get why that doesn't make sense if the endpoint is not Pomerium (and therefore doesn't want or need the hmac'd overhead). 

OK. Can an attacker craft a url that results in this page displaying an arbitrary crafted device credential id? Do we care? I guess we can accomplish the same thing by providing the user a link back to their user info page (`.pomerium`).

authenticate/handlers.go
desimone: It's okay.

desimone (changes requested):
let's just KISS and get rid of displaying the `QueryDeviceCredentialID` in the go HTML template.

PR #2909, opened 2022-01-07
envoy: check certificates for must-staple flag and drop them if they are missing the response

### LINKED ISSUE #2852: 0.16.0: crashloop when cert request was rate limited by Let's Encrypt
## What happened?
One of my certificates is due to refresh but due to my misconfig on other parts of the system, the `/.well-known/` URL was returning 404, this failed the cert renew process and apparently caused crashloop of Pomerium. Eventually the crashloop triggered the rate limiting on Let's Encrypt.

1. A temporary failure to get certificate should not crashloop Pomerium
2. A crashlooping Pomerium should not cause excessive requests to renew certificate (exponential back off of retrying the renewal of certificate should be followed even after Pomerium restarts).

## How'd it happen?

1. Update to 0.16.0
2. Wait for one cert to expire, but misconfig your system so the renew request never finishes
3. Now Pomerium should start crashlooping and eventually you can see rate limit errors from let's Encrypt.

- Pomerium version (retrieve with `pomerium --version`): 0.16.0
- Server Operating System/Architecture/Cloud: docker on Debian

## What's your config.yaml?

I guess the only important bit is I have environment `AUTOCERT_MUST_STAPLE` set to 1.

```logs
ERR delta config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) https-ingress: OCSP response is required for must-staple certificate name=config service=envoy
```

Rate limit error from let's Encrypt:

```log
WRN renew client certificated failed, use existing cert error="[dns.apkay.com] Renew: [foo.example.com] creating new order: attempt 1: https://acme-v02.api.letsencrypt.org/acme/new-order: HTTP 429 urn:ietf:params:acme:error:rateLimited - Error creating new order :: too many failed authorizations recently: see https://letsencrypt.org/docs/rate-limits/ (ca=https://acme-v02.api.letsencrypt.org/directory)" service=autocert-manager
```

## Additional context

Reverting to 0.15.8 will stop the crashlooping. The crash behavior seems to be related to #2777

calebdoxsey:
## Summary
If a TLS certificate has a must-staple extension set, envoy will reject the certificate if doesn't also have an OCSP response. This results in an invalid configuration and a fatal error in Pomerium. When Pomerium restarts it picks the certificate up again and crashes, resulting in rate-limiting from the Let's Encrypt side.

This PR adds a validation step to reject invalid certificates before we hand them to envoy. This allows us to recover if the OCSP response subsequently gets added to the certificate.

The bug can be quite difficult to reproduce: it requires a server accessible from the internet and I had to modify the code to erase the OCSP response to simulate it missing. I wasn't able to come up with a way to build a real integration test, but I did add a unit test.

## Related issues
Fixes #2852 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/envoyconfig/tls_test.go
+++ b/config/envoyconfig/tls_test.go
@@ -1,12 +1,16 @@
 package envoyconfig
 
 import (
+	"crypto/x509"
+	"crypto/x509/pkix"
 	"net/url"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/pomerium/pomerium/internal/testutil"
+	"github.com/pomerium/pomerium/pkg/cryptutil"
 )
 
 func TestBuildSubjectAlternativeNameMatcher(t *testing.T) {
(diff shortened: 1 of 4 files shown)


config/envoyconfig/tls_test.go
```diff
@@ -31,3 +35,15 @@ func TestBuildSubjectNameIndication(t *testing.T) {
 	assert.Equal(t, "example.org", b.buildSubjectNameIndication(&url.URL{Host: "example.com:1234"}, "example.org"))
 	assert.Equal(t, "example.example.org", b.buildSubjectNameIndication(&url.URL{Host: "example.com:1234"}, "*.example.org"))
 }
+
+func TestValidateCertificatge(t *testing.T) {
```
wasaga, line 39:
```suggestion
func TestValidateCertificate(t *testing.T) {
```

wasaga (approved):

PR #2919, opened 2022-01-11
handle device states in deny block, fix default device type

calebdoxsey:
## Summary
For device handling we rely on `Reasons` returned by PPL. There are two device-specific reasons: `device-unauthenticated` and `device-unauthorized`. The first triggers the device registration/authentication flow. The second results in a special access denied error.

Currently we are only checking these reasons when they are returned by an `allow` block and not a `deny` block. This means that if you were to use a `device` criterion in a `deny` block the user would only ever see forbidden.

This PR updates the code so that both the "Not Allowed" and "Denied" conditions are handled in similar ways so that the user will go through the device authentication flow before (possibly) seeing a forbidden error.

The other issue is that we were still using `default` as the default device type from within the PPL criterion. This meant that if no device type was specified we were triggering registration for a device type that didn't exist, resulting in an error.

## Related issues
- https://github.com/pomerium/pomerium-console/issues/2274

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/authorize/check_response.go
+++ b/authorize/check_response.go
@@ -39,45 +39,33 @@ func (a *Authorize) handleResultDenied(
 	ctx context.Context,
 	in *envoy_service_auth_v3.CheckRequest,
 	result *evaluator.Result,
+	isForwardAuthVerify bool,
+	reasons criteria.Reasons,
 ) (*envoy_service_auth_v3.CheckResponse, error) {
 	denyStatusCode := int32(http.StatusForbidden)
 	denyStatusText := http.StatusText(http.StatusForbidden)
 
 	switch {
-	case result.Deny.Reasons.Has(criteria.ReasonRouteNotFound):
+	case reasons.Has(criteria.ReasonUserUnauthenticated):
+		// when the user is unauthenticated it means they haven't
+		// logged in yet, so redirect to authenticate
+		return a.requireLoginResponse(ctx, in, isForwardAuthVerify)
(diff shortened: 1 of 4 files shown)


wasaga:
the following policy does not work for me 

```yaml
    to: http://httpbin.org
    prefix: /
    policy:
      allow:
        and:
          - groups:
              has: 00gsi7m9cIhm3pLFE4x6
      deny:
        nor:
          - device: PELoEqJukACnKEfmrdrpcjs7uBhsHdxii52ThcpNv7A7RVWeQVJ8oUhVjJ6aSTf5S2svjVotPCsWhUY4rQA6LQeGdvaZCeR2z9snzUZG9FuyoqG
```

```
{"level":"fatal","error":"error creating authorize service: authorize: failed to update policy with options: error generating criterion rules: expected object for device criterion, got: parser.String","time":"2022-01-11T11:20:41-05:00","message":"cmd/pomerium"}
```

calebdoxsey:
That's not valid PPL.  https://master.docs.pomerium.io/docs/topics/ppl.html#device-matcher

wasaga (approved):
works for me with 

```yaml
    to: http://httpbin.org
    prefix: /
    policy:
      allow:
        and:
          - groups:
              has: admins
      deny:
        not:
          - device: 
              is: A
          - device: 
              is: B
          - device:
              is: C
          - device:
              is: D
```

PR #2925, opened 2022-01-11
deployment: fix distroless base arch

travisgroth:
## Summary

It seems like we're using the wrong env var to detect the target platform when building the image from master.  We should be using `TARGETARCH` rather than `ARCH` per [the docs](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope).

- Update the top level `Dockerfile` to use the correct variable name
- Update our release pipeline and `Dockerfile` to use the same variable name for consistency
- Set target platform explicitly in goreleaser config

## Related issues

#2896

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG ARCH=amd64
+ARG TARGETARCH=amd64
 
 FROM golang:latest as build
 WORKDIR /go/src/github.com/pomerium/pomerium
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,5 +1,3 @@
-ARG ARCH=amd64
```
travisgroth, line 1: This is intentional.  It seems TARGETARCH is specially handled in the `FROM` during build time

wasaga (approved):

PR #2929, opened 2022-01-12
device: attempt to recover from resource exhausted error

calebdoxsey:
## Summary
Add code to saving the device credential to remove the responses if we exceed the max message size in gRPC.

## Related issues
Fixes https://github.com/pomerium/pomerium-console/issues/2281

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/grpc/device/device.go
+++ b/pkg/grpc/device/device.go
@@ -9,6 +9,7 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	"github.com/pomerium/pomerium/internal/log"
 	"github.com/pomerium/pomerium/pkg/encoding/base58"
 	"github.com/pomerium/pomerium/pkg/grpc/databroker"
 	"github.com/pomerium/pomerium/pkg/protoutil"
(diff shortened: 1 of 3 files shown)


wasaga (commented):
I'm not sure about that approach.  

Why not just increase limits https://pkg.go.dev/google.golang.org/grpc#MaxSendMsgSizeCallOption ?

calebdoxsey:
I believe the error is coming from envoy. Though I don't know how to reproduce the issue, so I'm just guessing.

wasaga:
it looks like if it would be Envoy, it'll be HTTP 431 and not 500? 

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#:~:text=scheme%20handling%20applies.-,max_request_headers_kb,Requests%20that%20exceed%20this%20limit%20will%20receive%20a%20431%20response.,-stream_idle_timeout

wasaga:
BTW it is reproducing for me on `master-pgsql` environment 

image

calebdoxsey:
It does not reproduce for me. I guess this is blocked then. I'll close and put the issue back in the backlog.

PR #2936, opened 2022-01-14
authenticate: support webauthn redirects to non-pomerium domains

calebdoxsey:
## Summary
When redirecting to a Pomerium domain from authenticate we typically use the `/.pomerium/callback` handler so that we can update the cookie for that route. However if we have a non-Pomerium redirect URI, that domain won't have that path, so we should just redirect without modifying it.

The `GetAllRouteableHTTPDomains` function basically already existed in the envoy config subpackage, so I pulled it so that webauthn could use it.

## Related issues
Fixes https://github.com/pomerium/pomerium-console/issues/2262

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authenticate/handlers/webauthn/webauthn.go
+++ b/authenticate/handlers/webauthn/webauthn.go
@@ -50,12 +50,13 @@ var (
 
 // State is the state needed by the Handler to handle requests.
 type State struct {
-	SharedKey    []byte
-	Client       databroker.DataBrokerServiceClient
-	Session      *session.Session
-	SessionState *sessions.State
-	SessionStore sessions.SessionStore
-	RelyingParty *webauthn.RelyingParty
+	SharedKey       []byte
+	Client          databroker.DataBrokerServiceClient
+	PomeriumDomains []string
+	Session         *session.Session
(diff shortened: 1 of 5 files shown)


desimone:
Code LGTM but it would be nice to have a set of tests for `GetAllRouteableGRPCDomains` and `GetAllRouteableHTTPDomains`

authenticate/handlers/webauthn/webauthn.go
```diff
@@ -50,12 +50,13 @@ var (
...
 type State struct {
-	SharedKey    []byte
-	Client       databroker.DataBrokerServiceClient
-	Session      *session.Session
-	SessionState *sessions.State
-	SessionStore sessions.SessionStore
-	RelyingParty *webauthn.RelyingParty
+	SharedKey       []byte
+	Client          databroker.DataBrokerServiceClient
+	PomeriumDomains []string
```
wasaga, line 55: maybe []url.URL ?

wasaga (approved):

PR #2949, opened 2022-01-19
return explicit error when directory sync is disabled

wasaga:
## Summary

Persist and return a root cause when IDP directory sync is disabled. 

## Related issues

Fixes https://github.com/pomerium/internal/issues/722

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/directory/provider.go
+++ b/internal/directory/provider.go
@@ -3,6 +3,7 @@ package directory
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


internal/directory/provider.go
```diff
@@ -73,6 +75,7 @@ func GetProvider(options Options) (provider Provider) {
 				auth0.WithDomain(options.ProviderURL),
 				auth0.WithServiceAccount(serviceAccount))
 		}
+		errSyncDisabled = fmt.Errorf("invalid auth0 service account: %w", err)
```
desimone, line 78: We could use `multierror` package here?

calebdoxsey (approved):

internal/directory/provider.go
wasaga: there aren't multiple errors here. it's still one error.

desimone (approved):

PR #2950, opened 2022-01-20
google: support groups for users outside of the organization

### LINKED ISSUE #2945: Feature Request: Match Google Group Members Outside of Org

This applies specifically to Pomerium configurations using Google as an IdP, because it's the one IdP that could have users both in and outside of the org.

Scenario: An org creates a Google group around which policies are written. They add a non-domain user (i.e. a `@gmail.com` address) to that group. 

Google will not provide Pomerium with group membership information about that email, since it's outside of the org that we're authorized to get user info on.

In order to solve this issue, we can match users without group membership information against the known list of members of the group. I'm sorry if you have to read that sentence multiple times, language is imperfect.

**Additional context**

This issue is known to affect users in the wild as well as one in-house user (me).
Slack: https://pomerium-io.slack.com/archives/CK92MUAES/p1642190260002900
Internal Discussion: https://pomerium-com.slack.com/archives/C012D1Q96TX/p1642520824068000

calebdoxsey:
## Summary
Google groups can contain user members which aren't members of the Group's organization. Currently when this happens we don't create `directory.User` records for these members. This PR updates the code to create `directory.User` records, which in turn makes it possible to support group permission checks for user that aren't also a member of the organization.

For example, previously if we had data like this:

```yaml
organization:
  users:
  - email: user1@inside.test
  - email: user2@inside.test
  groups:
  - email: group1@inside.test
    members:
    - user1@inside.test
    - user2@inside.test
    - user3@outside.test
```

And a policy that required membership in the `group1@inside.test` group, `user3@outside.test` would still not have access to the route, even though they were a member of that group. With this PR this should now work as expected.

To support this I also had to add a new error type: `ErrPreferExistingInformation`, which signals the databroker not to erase the existing `directory.User` entry when refreshing the user. This is necessary because we always refresh the user details on login and the service account for the organization won't have access to the outside user information, resulting in a code path that re-created the `directory.User` with no groups. It should now keep the groups as expected.

## Related issues
Fixes #2945 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/databroker/directory.go
+++ b/databroker/directory.go
@@ -3,6 +3,7 @@ package databroker
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -36,7 +37,7 @@ func (c *DataBroker) RefreshUser(ctx context.Context, req *directory.RefreshUser
 			return new(emptypb.Empty), nil
 		case codes.NotFound: // go ahead and save the user that was returned
 		default:
-			return nil, err
+			return nil, fmt.Errorf("databroker: error retrieving existing user record for refresh: %w", err)
 		}
 	} else if err != nil {
 		return nil, err
--- a/databroker/directory.go
+++ b/databroker/directory.go
@@ -4,8 +4,11 @@ import (
 	"context"
 	"errors"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/emptypb"
 
+	"github.com/pomerium/pomerium/internal/directory/directoryerrors"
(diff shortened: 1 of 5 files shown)


databroker/directory.go
```diff
@@ -22,7 +25,20 @@ func (c *DataBroker) RefreshUser(ctx context.Context, req *directory.RefreshUser
...
+		})
+		switch status.Code(err) {
+		case codes.OK:
+			return new(emptypb.Empty), nil
+		case codes.NotFound: // go ahead and save the user that was returned
+		default:
+			return nil, err
+		}
+	} else if err != nil {
 		return nil, err
```
wasaga, line 42: wrap the err?

wasaga (approved):

PR #2951, opened 2022-01-20
devices: shrink credentials by removing unnecessary data

calebdoxsey:
## Summary
Device credentials contain three properties intended for debugging, but not essential for functionality:

```proto
  message WebAuthn {
    // ...
    // the options that were used to do initial registration
    bytes register_options = 3;
    // the response returned from initial registration
    bytes register_response = 4;
    // subsequent authenticate responses
    repeated bytes authenticate_response = 5;
    // ...
}
```

Sometimes these properties can be quite large, which can cause issues when attempting to store them in a database or send them over the wire via gRPC. This PR adds code to shrink credentials by removing data until the entire payload is less than 256KB.

## Related issues
Fixes https://github.com/pomerium/pomerium-console/issues/2288

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/grpc/device/device.go
+++ b/pkg/grpc/device/device.go
@@ -170,6 +170,8 @@ func PutCredential(
 	client databroker.DataBrokerServiceClient,
 	credential *Credential,
 ) error {
+	shrinkCredential(credential)
+
 	any := protoutil.NewAny(credential)
 	_, err := client.Put(ctx, &databroker.PutRequest{
 		Record: &databroker.Record{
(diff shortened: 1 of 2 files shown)


desimone (approved):

pkg/grpc/device/device.go
```diff
@@ -214,3 +216,33 @@ func PutOwnerCredentialRecord(
 	})
 	return err
 }
+
+var maxCredentialSize = 256 * 1024
+
+// shrinkCredential shrinks a credential object by removing unnecessary responses and options
+// until its within the max credential size
+func shrinkCredential(credential *Credential) {
+	for len(protoutil.NewAny(credential).GetValue()) > maxCredentialSize {
```
wasaga, line 225: can there be a situation that this loop will never quit? maybe i.e. limit it to 3 iterations ?

calebdoxsey: No that should not be possible. The loop always breaks if the 3 conditions aren't met, so even if there were nothing left to clear we would not enter an infinite loop. And the 3 conditions all eventually terminate.

PR #2966, opened 2022-01-24
config: fix policy matching for regular expressions

### LINKED ISSUE #2952: regex key does not accept ^ char
## What happened?

When creating routes in core with the regex key or in Enterprise with the regex matcher, values that start with `^` do not work.

I expected the key to accept `^`, needed to ensure the pattern doesn't match when the path has other content before the match.

## How'd it happen?

Example routes:

```
  - from: https://route.example.com
    to: https://local-address
    regex: ^\/foo\/bar\/[0-9a-f]\/{0,1}$
    policy:
      - deny:
          or: 
            - email:
                is: afornuto@pomerium.com
  - from: /route.example.com
    to: https://local-address
    policy:
      - allow:
          or:
            - email:
                is: afornuto@pomerium.com
```

In the example above, going to a matching path is allowed for `afornuto@pomerium.com`. When `^` is removed the matching paths are denied.

  ```
  $ docker exec -it pomerium /bin/pomerium --version
  pomerium: 0.16.0-1640214477+45a348de
  envoy: 1.19.1+c5ecc1e167273edac76f6c07f980ed5a728f4b7e077679d74cfbc59c087933e5
  ```

- Server Operating System/Architecture/Cloud: Docker

## What's your config.yaml?

See above

```logs
# Without ^
2022-01-21T02:06:14.813271889Z {"level":"info","service":"authorize","request-id":"42e07f0b-e543-446b-bc89-7e5b15ace549","check-request-id":"cb49e914-765e-4474-be87-5810e34e6c68","method":"GET","path":"/foo/bar/123adb","host":"route.example.com","query":"","session-id":"350f9cfc-ba56-4ea4-8626-65d3127a325e","allow":false,"allow-why-false":["non-pomerium-route"],"deny":true,"deny-why-true":["email-ok"],"user":"108445263666202956835","email":"afornuto@pomerium.com","databroker_server_version":2697703461692766345,"databroker_record_version":13848,"time":"2022-01-21T02:06:14Z","message":"authorize check"}

# With ^
2022-01-21T02:07:54.392366461Z {"level":"info","service":"authorize","request-id":"dd45ebec-eb53-4100-82ec-9a43e7158f4b","check-request-id":"c9b80a14-5b56-41f6-926b-a4099d6caef6","method":"GET","path":"/foo/bar/123adb","host":"route.example.com","query":"","session-id":"350f9cfc-ba56-4ea4-8626-65d3127a325e","allow":true,"allow-why-true":["email-ok"],"deny":false,"deny-why-false":["valid-client-certificate-or-none-required"],"user":"108445263666202956835","email":"afornuto@pomerium.com","databroker_server_version":2697703461692766345,"databroker_record_version":14061,"time":"2022-01-21T02:07:54Z","message":"authorize check"}

```

calebdoxsey:
## Summary
The regular expression route matching should only be matching the path part of the URL. It should also always match a full string and not a partial submatch. This PR fixes both issues.

## Related issues
Fixes #2952 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/policy.go
+++ b/config/policy.go
@@ -565,8 +565,16 @@ func (p *Policy) Matches(requestURL url.URL) bool {
 	}
 
 	if p.Regex != "" {
-		re, err := regexp.Compile(p.Regex)
-		if err == nil && !re.MatchString(requestURL.String()) {
+		rawRE := p.Regex
+		if !strings.HasPrefix(rawRE, "^") {
+			rawRE = "^" + rawRE
+		}
+		if !strings.HasSuffix(rawRE, "$") {
+			rawRE += "$"
+		}
+
+		re, err := regexp.Compile(rawRE)
+		if err == nil && !re.MatchString(requestURL.Path) {
 			return false
 		}
 	}
--- a/config/policy.go
+++ b/config/policy.go
@@ -44,9 +44,10 @@ type Policy struct {
 	Source *StringURL `yaml:",omitempty" json:"source,omitempty" hash:"ignore"`
 
 	// Additional route matching options
-	Prefix string `mapstructure:"prefix" yaml:"prefix,omitempty" json:"prefix,omitempty"`
-	Path   string `mapstructure:"path" yaml:"path,omitempty" json:"path,omitempty"`
-	Regex  string `mapstructure:"regex" yaml:"regex,omitempty" json:"regex,omitempty"`
+	Prefix        string `mapstructure:"prefix" yaml:"prefix,omitempty" json:"prefix,omitempty"`
+	Path          string `mapstructure:"path" yaml:"path,omitempty" json:"path,omitempty"`
+	Regex         string `mapstructure:"regex" yaml:"regex,omitempty" json:"regex,omitempty"`
+	compiledRegex *regexp.Regexp
 
 	// Path Rewrite Options
 	PrefixRewrite            string `mapstructure:"prefix_rewrite" yaml:"prefix_rewrite,omitempty" json:"prefix_rewrite,omitempty"`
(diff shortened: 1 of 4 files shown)


config/policy.go
```diff
@@ -565,8 +565,16 @@ func (p *Policy) Matches(requestURL url.URL) bool {
...
-		if err == nil && !re.MatchString(requestURL.String()) {
+		rawRE := p.Regex
+		if !strings.HasPrefix(rawRE, "^") {
+			rawRE = "^" + rawRE
+		}
+		if !strings.HasSuffix(rawRE, "$") {
+			rawRE += "$"
+		}
+
+		re, err := regexp.Compile(rawRE)
```
wasaga, line 576: can we move regex compilation out of hot path? it seems to be called every time authz is invoked.

wasaga (changes requested):

sherifkayad:
How about that test case?
```golang
t.Run("full", func(t *testing.T) {
		p := &Policy{
			From:  "https://www.example.com",
			To:    mustParseWeightedURLs(t, "https://localhost"),
			Regex: `/admin/.*`,
		}
		assert.NoError(t, p.Validate())

		assert.True(t, p.Matches(urlutil.MustParseAndValidateURL(`https://www.example.com/admin/foo`)))
                assert.True(t, p.Matches(urlutil.MustParseAndValidateURL(`https://www.example.com/admin/bar`)))
	})
```

Would that be a valid test case? Should it be added as well?

config/policy.go
calebdoxsey: I don't like that Validate modifies the state of Policy, but I updated the code as you requested.

wasaga: for the record, I was never fan of how we have so many side effects in `Validate` but we would need to factor all of them out to be consistent.

calebdoxsey:
@sherifkayad yes. I added the test.

PR #2976, opened 2022-01-27
authenticate: fix expiring user info endpoint

### LINKED ISSUE #2967: User Info page HMAC errors
## What happened?

The User Info page (`/.pomerium`) will sometimes result in an HMAC error when refreshed.

Screen Shot 2022-01-24 at 11 17 21 AM

This is because the query string contains an HMAC that expires.

Refresh to work.

## How'd it happen?

1. Go to a pomerium-hosted `/.pomerium` endpoint.
2. Waited a while
3. Refreshed
3. Saw HMAC error

- Pomerium version: any
- Server Operating System/Architecture/Cloud: any

calebdoxsey:
## Summary
When we redirect to the `/.pomerium` endpoint we include a `?pomerium_redirect_url` parameter on the query string. This is HMAC'd and expires after a few minutes, which means if you refresh the page you sometimes get an error. This PR updates the code so that if we see the redirect url on the userinfo page we save it to a cookie and strip it from the query string.

## Related issues
Fixes #2967 

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authenticate/authenticate.go
+++ b/authenticate/authenticate.go
@@ -16,7 +16,6 @@ import (
 	"github.com/pomerium/pomerium/internal/log"
 	"github.com/pomerium/pomerium/internal/urlutil"
 	"github.com/pomerium/pomerium/pkg/cryptutil"
-	"github.com/pomerium/pomerium/pkg/webauthnutil"
 )
 
 // ValidateOptions checks that configuration are complete and valid.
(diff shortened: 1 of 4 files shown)


desimone (commented):
Could we add a test to confirm?

desimone (approved):

PR #2980, opened 2022-01-28
auth0: support explicit domains in the service account

### LINKED ISSUE #2978: auth0: can't use groups with custom auth0 domain
If I configure my auth0 idp url to https://login.example.com/ instead of https://example.auth0.com/ then groups don't work. The data broker logs:

```
{"level":"warn","service":"identity_manager","error":"auth0: could not list roles: 401 Unauthorized: Bad audience: https://login.example.com/api/v2/","time":"2022-01-28T10:18:18Z","message":"failed to refresh directory users and groups"}
```

According to the auth0 documents, the management API has to be example.auth0.com even if login.example.com is used for logins.

calebdoxsey:
## Summary
Add an optional explicit `domain` property to the service account configuration for auth0. This is necessary when a custom domain is used for login, but the management api is still configured with the `*.auth0.com` domain.

## Related issues
Fixes #2978 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/directory/auth0/auth0.go
+++ b/internal/directory/auth0/auth0.go
@@ -250,6 +250,15 @@ func parseServiceAccountFromOptions(clientID, clientSecret string) (*ServiceAcco
 	if serviceAccount.ClientID == "" {
 		return nil, fmt.Errorf("auth0: client_id is required")
 	}
+
+	// for backwards compatibility we support secret and client_secret
+	if serviceAccount.Secret == "" {
+		serviceAccount.Secret = serviceAccount.ClientSecret
+	}
+	if serviceAccount.ClientSecret == "" {
+		serviceAccount.ClientSecret = serviceAccount.Secret
+	}
+
 	if serviceAccount.Secret == "" {
 		return nil, fmt.Errorf("auth0: client_secret is required")
 	}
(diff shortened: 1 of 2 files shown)


internal/directory/auth0/auth0.go
```diff
@@ -253,20 +258,23 @@ func parseServiceAccountFromOptions(clientID, clientSecret string) (*ServiceAcco
...
 	if serviceAccount.ClientID == "" {
 		return nil, errors.New("auth0: client_id is required")
 	}
 
+	// for backwards compatibility we support secret and client_secret
+	if serviceAccount.Secret == "" {
+		serviceAccount.Secret = serviceAccount.ClientSecret
+	}
+	if serviceAccount.ClientSecret == "" {
+		serviceAccount.ClientSecret = serviceAccount.Secret
```
abh, line 275: It's a little confusing both ends up being set. I think it'd leave less room for future ambiguity if the backwards compatibility only extends to reading the data, but the code always has to use one.

abh (approved):
Doesn't parseServiceAccountFromOptions need the same secret vs client secret treatment?

Otherwise looks good to me; I too prefer this over adding the ProviderAPI option to the idp generally.

abh (approved):
Yay.

If you are still fussing and waiting for a code review maybe some of this could be consolidated; though hopefully it'll never be touched again so ... 🤷

wasaga (approved):

PR #3006, opened 2022-02-04
Docs: Add License Notice

alexfornuto:
## Summary

This PR overrides the PageEdit component from the default theme, so that we can add in our license disclaimer.

Input welcome on how to best style the changes.

## Related issues

Resolves https://github.com/pomerium/docs/issues/59

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/docs/.vuepress/theme/components/PageEdit.vue
+++ b/docs/.vuepress/theme/components/PageEdit.vue
@@ -1,0 +1,166 @@
+<template>
+  <footer class="page-edit">
+    <div
+      v-if="editLink"
+      class="edit-link"
+    >
+      <a
+        :href="editLink"
+        target="_blank"
+        rel="noopener noreferrer"
+      >{{ editLinkText }}</a>
+      <OutboundLink />
+    </div>
+
+    <div
+      v-if="lastUpdated"
+      class="last-updated"
+    >
+      <span class="prefix">{{ lastUpdatedText }}:</span>
+      <span class="time">{{ lastUpdated }}</span>
+    </div>
+    <br/>
+    <br/>
+    <hr/>
+    <div class="license">
+    Except as otherwise noted, the content of this page is licensed under the Creative Commons Attribution 4.0 License, and code samples are licensed under the Apache 2.0 License. Pomerium is a registered trademark.
+    </div>
+  </footer>
+</template>
+
+<script>
+import isNil from 'lodash/isNil'
+import { endingSlashRE, outboundRE } from '../util'
+
+export default {
+  name: 'PageEdit',
+
+  computed: {
+    lastUpdated () {
+      return this.$page.lastUpdated
+    },
+
+    lastUpdatedText () {
+      if (typeof this.$themeLocaleConfig.lastUpdated === 'string') {
+        return this.$themeLocaleConfig.lastUpdated
+      }
+      if (typeof this.$site.themeConfig.lastUpdated === 'string') {
+        return this.$site.themeConfig.lastUpdated
+      }
+      return 'Last Updated'
+    },
+
+    editLink () {
+      const showEditLink = isNil(this.$page.frontmatter.editLink)
+        ? this.$site.themeConfig.editLinks
+        : this.$page.frontmatter.editLink
+
+      const {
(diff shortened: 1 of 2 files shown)


desimone (approved):

desimone (approved):

docs/.vuepress/theme/components/PageEdit.vue
```diff
@@ -0,0 +1,166 @@
...
+        return this.$themeLocaleConfig.lastUpdated
+      }
+      if (typeof this.$site.themeConfig.lastUpdated === 'string') {
+        return this.$site.themeConfig.lastUpdated
+      }
+      return 'Last Updated'
+    },
+
+    editLink () {
+      const showEditLink = isNil(this.$page.frontmatter.editLink)
```
nhayfield, line 54: pretty sure this would still error out if for instance this.$page was undefined. In the console we have access to the ? operator. Not sure if that is present here but it is used to access nested data without erroring out.

`const showEditLink = this?.$page?.frontmatter?.edtiLink || ''`

nhayfield: i could be wrong and lodash could handle that though. best to test. there is also the _.get function which is safe to use. 
More info here:
https://stackoverflow.com/questions/2631001/test-for-existence-of-nested-javascript-object-key

docs/.vuepress/theme/components/PageEdit.vue
```diff
@@ -0,0 +1,166 @@
...
+        ? this.$site.themeConfig.editLinks
+        : this.$page.frontmatter.editLink
+
+      const {
+        repo,
+        docsDir = '',
+        docsBranch = 'master',
+        docsRepo = repo
+      } = this.$site.themeConfig
```
nhayfield, line 63: these can be unsafe if the this.$site is undefined

docs/.vuepress/theme/components/PageEdit.vue
```diff
@@ -0,0 +1,166 @@
...
+
+      if (showEditLink && docsRepo && this.$page.relativePath) {
+        return this.createEditLink(
+          repo,
+          docsRepo,
+          docsDir,
+          docsBranch,
+          this.$page.relativePath
```
nhayfield, line 71: unsafe object access possible. for course if these variables are guaranteed then maybe not an issue

docs/.vuepress/theme/components/PageEdit.vue
nhayfield: this is one of the biggest pains of javascript. accessing inner properties that might not exist. If you don't have access to the ? operator then my recommendation would be to use const test = _.get(testObject, 'nested.path.dot.separated')  || 'default'  as a way to make sure it is safe.

nhayfield (changes requested):

docs/.vuepress/theme/components/PageEdit.vue
alexfornuto: Changing to `const showEditLink = this?.$page?.frontmatter?.editLink || ''` seems to work with no difference in console output, so I'm happy to go with that if it's safer.

docs/.vuepress/theme/components/PageEdit.vue
alexfornuto: I believe `$site` should always be defined: https://vuepress.vuejs.org/guide/global-computed.html#site

docs/.vuepress/theme/components/PageEdit.vue
alexfornuto: But as discussed below, `$page` should always be defined, and `editLink` is set as `true` in our `config.js` file: https://github.com/pomerium/pomerium/blob/master/docs/.vuepress/config.js#L52

With this in mind, do you still want to see this updated @nhayfield?

nhayfield: seems fine

nhayfield (approved):
this is fine

PR #3025, opened 2022-02-09
sessions: remove unused session state properties

calebdoxsey:
## Summary
Remove the following properties from the `sessions.State` object:

* `Issuer`
* `Subject`
* `Audience`
* `IssuedAt`
* `OID`

The properties are stored in `session.Session` object in the databroker so are redundant and unneeded. They were only being used to create the initial `session.Session` object, which will now be created via the decoded ID token. I also ported over the logic to handle `oid`, which I believe is only used by Azure.

This is in preparation for multi-user sessions.

## Related issues
- https://github.com/pomerium/internal/issues/762

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
These appear to be used in ways I don't understand.

PR #3030, opened 2022-02-09
authenticate: support for per-route client id and client secret

calebdoxsey:
## Summary
Add two new options to a route: `idp_client_id` and `idp_client_secret`. When set these options will override the global client id and client secret for the route. This is implemented by passing a new query string parameter `pomerium_idp_id` which the authenticate service references to discover which provider settings to use for authentication.

Once login is complete a separate session is created for the route, so the session state also stores which `idp_id` its associated with.

This seems to work when I tested it locally.

## Related issues
- https://github.com/pomerium/internal/issues/762

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


config/policy.go
```diff
@@ -162,6 +162,11 @@ type Policy struct {
 	// SetResponseHeaders sets response headers.
 	SetResponseHeaders map[string]string `mapstructure:"set_response_headers" yaml:"set_response_headers,omitempty"`
 
+	// IDPClientID is the client id used for the identity provider.
+	IDPClientID string `mapstructure:"idp_client_id" yaml:"idp_client_id,omitempty"`
```
desimone, line 166: Flagging @alexfornuto for docs requirement.

pkg/grpc/identity/identity.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/pomerium/pomerium/pkg/encoding/base58"
+)
+
+// Clone clones the Provider.
+func (x *Provider) Clone() *Provider {
+	return proto.Clone(x).(*Provider)
+}
+
+// Hash computes a sha256 hash of the provider's fields. It excludes the Id field.
```
desimone, line 17: Why is it important to exclude the Id field?

internal/sessions/state.go
```diff
@@ -15,34 +15,6 @@ var ErrMissingID = errors.New("invalid session: missing id")
 // timeNow is time.Now but pulled out as a variable for tests.
 var timeNow = time.Now
 
-// Version represents "ver" field in JWT public claims.
-//
-// The field is not specified by RFC 7519, so providers can
-// return either string or number (like okta).
-type Version string
```
desimone, line 22: I guess this was not being used anymore?

authenticate/authenticate.go
```diff
@@ -28,15 +25,6 @@ func ValidateOptions(o *config.Options) error {
 	if _, err := cryptutil.NewAEADCipherFromBase64(o.CookieSecret); err != nil {
 		return fmt.Errorf("authenticate: 'COOKIE_SECRET' invalid %w", err)
 	}
-	if o.Provider == "" {
```
desimone, line 31: To confirm, this means that the authenticate service can now run without a global idp credential set, right?

desimone: Yeah -- it appears so. This get's wonky fast, but I don't think it breaks anything. For example, I could have a configuration like follows. 

```yaml
idp_provider: google
#notice, no other client idp configuration details for global?

policy:
    to: https://httpbin.org
    idp_client_id: bar.apps.googleusercontent.com
    idp_client_secret: foo
    pass_identity_headers: true
    allowed_users:
      - bdd@pomerium.com
      - bobbydesimone@gmail.com
  - from: https://httpbin2.localhost.pomerium.io
    to: https://httpbin.org
    pass_identity_headers: true
    allowed_users:
      - bdd@pomerium.com
      - bobbydesimone@gmail.com
```

httpbin works but httpbin2 hits a redirect flow error. I'm not sure this is worth guarding against. 

Screen Shot 2022-02-09 at 4 18 32 PM

desimone (changes requested):
Needs docs updates, but otherwise LGTM. I tested on Google and Okta.

desimone:
I'm getting a peculiar refresh error now. It looks like my session is not being refreshed, but the session is not actually being deleted despite the errors.

It looks like the initial call is missing the user key?

https://gist.github.com/desimone/27b7c751bd855817e79f61c3750dd82f#file-pr303-json-L4

pkg/grpc/identity/identity.go
calebdoxsey: Because the ID is the hash. If it were included re-computing the Hash would result in the ID changing every time.

config/policy.go
calebdoxsey: There is another issue for updating docs, but sure I can include those changes in this PR.

internal/sessions/state.go
calebdoxsey: yes AFAIK it is not used

authenticate/authenticate.go
calebdoxsey: I don't know the right requirements, but I can re-add the restriction

calebdoxsey:
This solution will not work for path, prefix or regex differentiation between routes since the cookie is per-host. I'm not sure how to fix this but I guess we need to add code to make a single session aware of all the other sessions and switch accordingly. It probably won't work for forward auth, but I don't really understand how forward auth works, so I can't predict what will happen.

I didn't change the refresh code, but I guess there's a bug with that or something. I'll try to reproduce and include a fix in these changes, though I don't know what a "user key" is.

It sounds like we also want to delete sessions when refresh fails, so I'll also update this PR with changes for that.

authenticate/authenticate.go
desimone: I think this is fine behavior.

config/policy.go
desimone: I missed that.

desimone:
> This solution will not work for path, prefix or regex differentiation between routes since the cookie is per-host. 

👍 Not sure a way around that.


> I didn't change the refresh code, but I guess there's a bug with that or something. I'll try to reproduce and include a fix in these changes, though I don't know what a "user key" is.

```
rpc error: code = Unknown desc = google: error getting user: googleapi: Error 404: Resource Not Found: userKey, notFound
```

I investigated a bit more. These errors are definitely not related to this pull request. This occurs in google only when a non-directory user (e.g. a gmail user) logs in and the directory code tries to fetch that user's details. But that service account does not have permissions to look up that user. So it fails. 

https://github.com/pomerium/pomerium/blob/ed6c3e50871478c8fdade4ab23af56ab3048eae3/internal/directory/google/google.go#L106

It sounds like we also want to delete sessions when refresh fails, so I'll also update this PR with changes for that.

> It sounds like we also want to delete sessions when refresh fails, so I'll also update this PR with changes for that.

```
	"message": "failed to update user info, deleting session"
```

That's just what the error message says. The code seems to suggest the session would be deleted though.

https://github.com/pomerium/pomerium/blob/ac9e08669180f3bd411adea5e14fe76fc75a7163/internal/identity/manager/manager.go#L488-L490

https://github.com/pomerium/pomerium/blob/ac9e08669180f3bd411adea5e14fe76fc75a7163/internal/identity/manager/manager.go#L436-L438

My findings seem entirely unrelated to this PR.

calebdoxsey:
After further investigation I believe it would be possible to make `path`, `prefix` and `regex` work by adding another property to the session state, but doing so ends up changing a fair amount of code, and its not clear it would actually be a very useful feature. I suspect most people who would want per-application client IDs and client secrets would want them per-hostname anyway.

As an alternative we could additional verification steps and warn if we see behavior that's unlikely to work.

To be clear it does actually work, it just requires you to login again, which if you were using both apps simultaneously would be pretty annoying.

calebdoxsey:
Status of this PR: @wasaga to take a look. We will update docs in a separate PR, and will not implement a fix for path/prefix/regex unless requested since it seems like an unlikely use case.

authenticate/handlers.go
```diff
@@ -161,18 +161,34 @@ func (a *Authenticate) VerifySession(next http.Handler) http.Handler {
 		defer span.End()
 
 		state := a.state.Load()
+		idpID := r.FormValue(urlutil.QueryIdentityProviderID)
 
 		sessionState, err := a.getSessionFromCtx(ctx)
 		if err != nil {
-			log.FromRequest(r).Info().Err(err).Msg("authenticate: session load error")
+			log.FromRequest(r).Info().
```
wasaga, line 168: why is it `Info` level?

wasaga (approved):
LGTM; tested with Auth0 👍

PR #3047, opened 2022-02-14
authorize: add support for passing access or id token upstream

calebdoxsey:
## Summary
Add support for passing the OAuth access token or the OIDC id token upstream.

## Related issues
- https://github.com/pomerium/internal/issues/781

## Checklist
- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (commented):
What do you think about making configuration setting for this more generic like `PassTokenType` which takes an enum or string where such that the available options could start with `access` /  `id`  but could be expanded if required? 

If we do end up wanting to support more types of tokens to pass along (e.g. `refresh` / some other sensitive value we want to pass through) it could give us more flexibility to do so without introducing new config options. Thoughts?

Also, though weird -- does OPA / rego restrict us from setting multiple values for header keys per spec?

calebdoxsey:
Sure that makes sense. I can change the types to be an enum.

We convert the rego value into an envoy header array, so yes we could duplicate the header values, though if I remember right Envoy will re-combine them as `,` separated values (which is why we have the reproxy package to fix this problem for Kubernetes)

calebdoxsey:
I updated the option to a single enum (`set_authorization_header`) with 3 possible values: `pass_through`, `access_token`, `id_token`. `pass_through` means we just leave the Authorization header as-is, which is the current behavior.

desimone (approved):

PR #3053, opened 2022-02-16
Extract email for active directory users that don't have access to exchange

JBodkin-Amphora:
## Summary

Users that are created in active directory that have not been granted access to Microsoft Exchange do not have an email address however the upn claim is present in the JWT. Currently, the email is transformed incorrectly as it assumes all users without an email are guest accounts, which isn't always the case. Instead we should check if the user is an internal or external user and only apply the transformation on external (guest) users.

## Checklist

- [x] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/internal/directory/azure/api.go
+++ b/internal/directory/azure/api.go
@@ -37,7 +37,7 @@ func (obj apiUser) getEmail() string {
 	email := obj.UserPrincipalName
 	if idx := strings.Index(email, "#EXT"); idx > 0 {
 		email = email[:idx]
-		
+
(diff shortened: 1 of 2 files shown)


desimone (approved):
Thanks for the contribution @JBodkin-Amphora

desimone:
@JBodkin-Amphora do you mind rebasing your changes on top of `main`? I think we had an unrelated issue where our unit tests were failing due to a change in Go's test cert but want to be sure before merging.

PR #3094, opened 2022-03-02
Docs: Add 'more' key for 'View Docs' value

alexfornuto:
## Summary

Proof-of-concept solution for https://github.com/pomerium/pomerium-console/issues/2375 - I add a key with an appropriate "View Docs" link. If the key is there, "View Docs" points to the value. If not, use existing logic.

This also allows us to link to docs other than the source material generating the tooltip, which improves the user experience by bringing them to relevant new information instead of the same content in a different place.

### To Do

- [x] Review/Agree on solution
- [x] ~Update Scripts to include new key when creating JSON~ Not required, import script includes all values.
- [ ] Update core deps

## Related issues

https://github.com/pomerium/pomerium-console/issues/2375

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/docs/enterprise/console-settings.yaml
+++ b/docs/enterprise/console-settings.yaml
@@ -44,12 +44,14 @@ settings:
             settings:
               - name: "Name"
                 doc: This value is only visible in the Console UI.
+                more: '/enterprise/reference/manage.html#name'
               - name: "From"
               - name: "To"
               - name: "Redirect"
               - name: "Pass Identity Headers"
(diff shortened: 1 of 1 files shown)


nhayfield (commented):
if we start using this then it makes more sense to use it for everything instead of an algorithm on the front end based on name attribute

travisgroth (approved):

PR #3100, opened 2022-03-07
Correct domain name

shapeshed:
hello.localhost.pomerium.com has no dns record associated with it so won't work without hacking `/etc/hosts`
The walkthrough also issues certificates to `*.localhost.pomerium.io`

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


alexfornuto (approved):
Thanks @shapeshed! I took the opportunity to add some clarification in this doc and another guide on how we make that domain available for testing.

PR #3185, opened 2022-03-23
try pinning docker dependency

calebdoxsey:
## Summary
Pinning docker dependencies. I thought this wouldn't work because the digest manifests were platform-specific, but it turns out there's a hash of the manifest file which should work.

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
So apparently this does work, based on the actions passing.

wasaga (approved):

PR #3210, opened 2022-03-29
databroker: add support for field masks on Put

calebdoxsey:
## Summary
A `FieldMask` is part of the well-known-types in protobuf and is commonly used in various Google APIs. This PR allows us to use it to limit which fields are updated with a `Put` operation.

The immediate use case for this feature is to support updating a session's last access date without having to worry about data races on session updates. As an example of what could happen if we didn't have this: 

Imagine a request comes in to the Authenticate service to add a new device to a session. The Authenticate service will make a `Put` call to the databroker updating the session with the new device credential. Simultaneously the Authorize service does an ExtAuthz check on an unrelated request and updates the last access date of the session. This will also make a `Put` call to the databroker. If the device update came in first, then the authorize update would erase the device. By using a field mask we can prevent this from happening since we only update a single field and the device would be preserved.

## Related issues
- https://github.com/pomerium/pomerium-console/issues/2445

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/protoutil/mask.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/mennanov/fmutils"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+)
+
+// MergeAnyWithFieldMask merges the data in src with the data in dst,
+// but only the fields identified by the given mask.
+func MergeAnyWithFieldMask(dst, src *anypb.Any, mask *fieldmaskpb.FieldMask) *anypb.Any {
```
wasaga, line 12: should we really mask errors? it should probably be OK for a Put request to fail instead.

wasaga (approved):

pkg/protoutil/mask.go
calebdoxsey: I prefer it the way it was but I'll return the errors.

Issue #3216, opened 2022-03-30
Add name claim to JWT payload


alexfornuto:
**Is your feature request related to a problem? Please describe.**

Some systems (Like Grafana) require a name claim in order to auto-create new users based on the JWT payload. While we're not intending to be an Identity Provider for upstream services, there's no harm and only benefit from it.

**Describe the solution you'd like**

Include the 'name' claim in the JWT payload when provided by the IdP.

**Describe alternatives you've considered**

In the example situation of Grafana, the workaround is to use `X-Pomerium-Claim-` headers to provide additional data, but then you also have to entirely switch away from using JWT auth. [This blog post](https://linuxblog.xyz/posts/grafana-auth-proxy/) details what that would look like.

**Explain any additional use-cases**

I doubt this will be the last upstream service that would benefit from a slightly more expanded JWT payload.

sarasensible:
For the record the name claim is not required it will still create a user without the Name field populated.  The workaround is to set extraEnv.JWT_CLAIMS_HEADERS equal to email,name as discussed in https://discuss.pomerium.com/t/grafana-auto-sign-up-with-pomerium-identity/97/3

travisgroth:
@alexfornuto it looks like setting `name` in https://docs.pomerium.io/reference/#jwt-claim-headers should also add the claim in to the assertion JWT.  Could you test that in your setup?

alexfornuto:
I have tested and confirmed this. It was not documented as doing so, which will be resolved by #3226. Up to you if it should be included by default or not.

alexfornuto:
From discussion, worth adding by default to reduce complexity for user config.

PR #3238, opened 2022-04-04
authorize: add name claim

calebdoxsey:
## Summary
Add a name claim to the assertion JWT.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/3216

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/evaluator/headers_evaluator_test.go
+++ b/authorize/evaluator/headers_evaluator_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/structpb"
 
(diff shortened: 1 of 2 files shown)


calebdoxsey:
I'm unsure what will happen if we change that function. Currently if it receives null `get_header_string_value` produces nothing, which is an example where Rego is very different from typical programming languages, because it causes variables to be undefined too:

```rego
get_header_string_value(obj) = s {
	is_array(obj)
	s := concat(",", obj)
} else = s {
	s := concat(",", [obj])
}

# by passing null it's like this next line is never even written
# which also causes anything using `x` to be undefined too
x := get_header_string_value(null)

# after this line `y` is also undefined
y := sprintf("X: %s", [x])
```

wasaga (approved):

PR #3243, opened 2022-04-05
config: add support for downstream TLS server name

calebdoxsey:
## Summary
Add support for a `tls_downstream_server_name` option on the route config. This option overrides the hostname in the `from` part of a policy for the purpose of the establishing the TLS connection. For example if we have:

```yaml
from: https://internal.example.com
tls_downstream_server_name: www.example.com
```

The client should create a TLS connection using `www.example.com` as the TLS server name, and then make requests using `internal.example.com` as the `Host` (or `:authority`) header. This can be done with curl:

```bash
curl -v -k -H "Host: internal.example.com" https://www.example.com
```

## Related issues
- https://github.com/pomerium/internal/issues/800

## Checklist
- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


docs/reference/settings.yaml
```diff
@@ -1875,14 +1875,22 @@ settings:
...
+      - name: "TLS Upstream Server Name"
+        keys: ["tls_upstream_server_name"]
         attributes: |
-          - Config File Key: `tls_server_name`
+          - Config File Key: `tls_upstream_server_name`
           - Type: `string`
           - Optional
         doc: |
-          TLS Server Name overrides the hostname specified in the `to` field. If set, this server name will be used to verify the certificate name. This is useful when the backend of your service is an TLS server with a valid certificate, but mismatched name.
+          TLS Upstream Server Name overrides the hostname specified in the `to` field. If set, this server name will be used to verify the certificate name. This is useful when the backend of your service is an TLS server with a valid certificate, but mismatched name.
```
alexfornuto, line 1885:
```suggestion
          TLS Upstream Server Name overrides the hostname specified in the `to` field. If set, this server name will be used to verify the certificate name. This is useful when the backend of your service is a TLS server with a valid certificate, but mismatched name.
```

docs/reference/readme.md
```diff
@@ -1720,12 +1720,20 @@ See [Client-Side mTLS With Pomerium](/guides/mtls.md) for more information.
...
 
-### TLS Server Name
-- Config File Key: `tls_server_name`
+### TLS Upstream Server Name
+- Config File Key: `tls_upstream_server_name`
 - Type: `string`
 - Optional
 
-TLS Server Name overrides the hostname specified in the `to` field. If set, this server name will be used to verify the certificate name. This is useful when the backend of your service is an TLS server with a valid certificate, but mismatched name.
+TLS Upstream Server Name overrides the hostname specified in the `to` field. If set, this server name will be used to verify the certificate name. This is useful when the backend of your service is an TLS server with a valid certificate, but mismatched name.
```
alexfornuto, line 1728:
```suggestion
TLS Upstream Server Name overrides the hostname specified in the `to` field. If set, this server name will be used to verify the certificate name. This is useful when the backend of your service is a TLS server with a valid certificate, but mismatched name.
```

alexfornuto (commented):
Small edit, otherwise the docs are 👍

alexfornuto:
Added the `docs` tag since this changes docs and required review.

wasaga (approved):

Issue #3256, opened 2022-04-08
rego: using http.send in a rego policy results in an http.Transport error


calebdoxsey:
## What happened?
Any rego policy that includes an `http.send` will result in an interface conversion error. For example:

```rego
response := http.send({ ... })
```

```json
{
  "level": "error",
  "error": "authorize: error evaluating policy.rego: interface conversion: http.RoundTripper is *config.httpTransport, not *http.Transport",
  "time": "2022-04-08T07:35:16-06:00",
  "message": "error during OPA evaluation"
}
```

## What did you expect to happen?
Not to see this error and for the http request to be made.

## How'd it happen?
Created a route using custom rego and tried to access it.

## What's your environment like?
- git commit: 5e836913ae53e70c0db91693631d576ae755326f
- MacOS, arm64

PR #3257, opened 2022-04-08
config: fix DefaultTransport so it is still a *http.Transport

calebdoxsey:
## Summary
Some libraries (for example OPA) expect the `http.DefaultTransport` to be a `*http.Transport`. However when we override the `DefaultTransport` we replace it with a custom `RoundTripper`, so this interface conversion can't succeed.

This PR changes `NewHTTPTransport` so that it returns a `*http.Transport`. This is done by cloning the default transport and overriding the `DialTLSContext` function to use a custom `tls.Config` when necessary. We also set `ForceAttemptHTTP2` because the default behavior when the `DialTLSContext` is overridden is to disable http/2 entirely:

> ForceAttemptHTTP2 controls whether HTTP/2 is enabled when a non-zero Dial, DialTLS, or DialContext func or TLSClientConfig is provided. By default, use of any those fields conservatively disables HTTP/2. To use a custom dialer or TLS config and still attempt HTTP/2 upgrades, set this to true.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/3256

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/http.go
+++ b/config/http.go
@@ -20,22 +20,16 @@ func NewHTTPTransport(src Source) *http.Transport {
 		tlsConfig *tls.Config
 	)
 	update := func(ctx context.Context, cfg *Config) {
-		if cfg.Options.CA != "" || cfg.Options.CAFile != "" {
-			rootCAs, err := cryptutil.GetCertPool(cfg.Options.CA, cfg.Options.CAFile)
-			if err == nil {
-				lock.Lock()
-				tlsConfig = &tls.Config{
-					RootCAs:    rootCAs,
-					MinVersion: tls.VersionTLS12,
-				}
-				lock.Unlock()
-			} else {
-				log.Error(context.Background()).Err(err).Msg("config: error getting cert pool")
-			}
-		} else {
+		rootCAs, err := cryptutil.GetCertPool(cfg.Options.CA, cfg.Options.CAFile)
+		if err == nil {
 			lock.Lock()
-			tlsConfig = nil
+			tlsConfig = &tls.Config{
+				RootCAs:    rootCAs,
+				MinVersion: tls.VersionTLS12,
+			}
 			lock.Unlock()
+		} else {
+			log.Error(context.Background()).Err(err).Msg("config: error getting cert pool")
 		}
 	}
 	src.OnConfigChange(context.Background(), update)
--- a/config/http.go
+++ b/config/http.go
@@ -3,63 +3,55 @@ package config
 import (
 	"context"
 	"crypto/tls"
+	"net"
 	"net/http"
-	"sync/atomic"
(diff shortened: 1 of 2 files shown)


config/http.go
```diff
@@ -3,63 +3,55 @@ package config
...
-	if options.CA != "" || options.CAFile != "" {
-		rootCAs, err := cryptutil.GetCertPool(options.CA, options.CAFile)
-		if err == nil {
-			nt.TLSClientConfig = &tls.Config{
-				RootCAs:    rootCAs,
-				MinVersion: tls.VersionTLS12,
+func NewHTTPTransport(src Source) *http.Transport {
+	var (
+		lock      sync.Mutex
+		tlsConfig *tls.Config
```
calebdoxsey, line 20: `atomic.Value` was not used here because it doesn't support setting a `nil` value.

config/http.go
```diff
@@ -3,63 +3,55 @@ package config
...
+			rootCAs, err := cryptutil.GetCertPool(cfg.Options.CA, cfg.Options.CAFile)
+			if err == nil {
+				lock.Lock()
+				tlsConfig = &tls.Config{
+					RootCAs:    rootCAs,
+					MinVersion: tls.VersionTLS12,
+				}
+				lock.Unlock()
+			} else {
+				log.Error(context.Background()).Err(err).Msg("config: error getting cert pool")
```
wasaga, line 33:
```suggestion
				log.Error(ctx).Err(err).Msg("config: error getting cert pool")
```

as it'll carry information which config change caused it.

config/http.go
```diff
@@ -3,63 +3,55 @@ package config
...
-			nt.TLSClientConfig = &tls.Config{
-				RootCAs:    rootCAs,
-				MinVersion: tls.VersionTLS12,
+func NewHTTPTransport(src Source) *http.Transport {
+	var (
+		lock      sync.Mutex
+		tlsConfig *tls.Config
+	)
+	update := func(ctx context.Context, cfg *Config) {
+		if cfg.Options.CA != "" || cfg.Options.CAFile != "" {
```
wasaga, line 23: Is this check necessary? `cryptutil.GetCertPool` returns `x509.SystemCertPool()` CA is empty; we do pretty much the same below by setting config to nil

```go
        // RootCAs defines the set of root certificate authorities
	// that clients use when verifying server certificates.
	// If RootCAs is nil, TLS uses the host's root CA set.
```

wasaga (approved):

config/http.go
calebdoxsey: You're right. It's no longer needed.

PR #3263, opened 2022-04-11
ppl: support . in object_get paths

calebdoxsey:
## Summary
Our custom `object_get` function in rego is used to handle nested path look ups for JSON objects. Currently it uses `/` to indicate we should look inside an object. This PR also adds support for `.`.

## Related issues
- https://github.com/pomerium/pomerium-console/issues/2501

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/policy_ppl_test.go
+++ b/config/policy_ppl_test.go
@@ -751,14 +751,14 @@ else = [] {
 }
 
 object_get(obj, key, def) = value {
-	segments := split(key, "/")
+	segments := split(replace(key, ".", "/"), "/")
 	count(segments) == 2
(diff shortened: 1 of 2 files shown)


wasaga (approved):
could you please also update docs? 
+ @alexfornuto

calebdoxsey:
I don't think we expose this directly to end users. It's used by various criteria.

alexfornuto:
> could you please also update docs?
> 
> * @alexfornuto

Thanks for the callout. Caleb and I are working on documenting a REGO example for the Policy builder, but there's currently no documentation that covers this.

PR #3277, opened 2022-04-19
authorize: add support for cidr lookups

calebdoxsey:
## Summary
Add support for indexing databroker records by CIDR blocks in rego.

If a record is a struct type and contains an `$index` field with a `cidr` subfield it will be inserted into a patricia tree and can then be retrieved via an IP address that is within the block.

For example this policy:

```rego
package pomerium.policy

allow {
  record := get_databroker_record("type.googleapis.com/google.protobuf.Struct", "192.168.0.7")
  record.country == "US"
}
```

Will match this record:

```json
{
  "$index": { "cidr": "192.168.0.0/16" },
  "country": "US"
}
```

## Related issues
- https://github.com/pomerium/pomerium-console/issues/2522

## Checklist
- [] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/authorize.go
+++ b/authorize/authorize.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/pomerium/pomerium/authorize/evaluator"
+	"github.com/pomerium/pomerium/authorize/internal/store"
 	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/internal/log"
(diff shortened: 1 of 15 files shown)


calebdoxsey:
@alexfornuto If we've documented rego policies to a sufficient we may want to note this new capability or it may be something we want to tackle later.

wasaga (approved):

PR #3282, opened 2022-04-20
authorize: pass idp id for webauthn url, allow unauthenticated access to static files

calebdoxsey:
## Summary
We need to pass the idp id for webauthn registration so the proper session gets updated. Also update the dashboard subrouter on the authenticate service so that static files don't have middleware added to them (they were redirecting to the auth flow when not logged in leading to a blank screen)

## Related issues
- https://github.com/pomerium/internal/issues/830

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (approved):

desimone:
Should this be back-ported?

calebdoxsey:
I think this feature is not in 0.17.

PR #3285, opened 2022-04-22
integration: fix memory leakage

cfanbo:
fix memory leakage


desimone:
Thank you for the contribution @cfanbo !

PR #3300, opened 2022-04-27
databroker: fix in-memory backend deadlock

calebdoxsey:
## Summary
The in-memory storage backend will deadlock if multiple records are updated in a single call because we call `.Lock` multiple times. This PR fixes that and adds a test so it doesn't happen again.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (commented):
LGTM

wasaga (approved):

PR #3320, opened 2022-05-03
databroker: support record type for sync latest in storage backend

calebdoxsey:
## Summary
Currently the `SyncLatest` storage backend call always returns all records and filters them in the server. This PR updates the code to support filtering by record type in the storage backend.

Basically this allows us to use the `MATCH xyz*` argument to `HSCAN` in redis, which is much faster than doing the filtering later.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
Something about this doesn't work. Closing to investigate.

PR #3341, opened 2022-05-10
Docs: Document TCP routes in Kubernetes

alexfornuto:
## Summary

- Documents the annotations to create a TCP route.
- Includes an example route

## Related issues

- Documents https://github.com/pomerium/ingress-controller/pull/196
- Resolves https://github.com/pomerium/internal/issues/840
- Depends on https://github.com/pomerium/pomerium/pull/3340 to avoid merge conflicts.

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


docs/docs/k8s/ingress.md
```diff
@@ -350,6 +352,41 @@ spec:
     secretName: example-tls
 ```
 
+## TCP Endpoints
+
+The example route below defines a route providing a [tunneled TCP connection](/docs/tcp/readme.md) to an upstream service listening for non-web traffic:
```
wasaga, line 357: maybe `non-HTTP` ?

docs/docs/k8s/ingress.md
```diff
@@ -350,6 +352,41 @@ spec:
...
+                name: tcp-service
+                port:
+                  name: app
+```
+
+The important points to note in this example:
+
+- The annotation `ingress.pomerium.io/tcp_upstream:` is set to `"true"`,
+- `spec.rules.[].http.paths.[].path` is omitted,
+- `spec.rules.[].http.paths.[].pathType` is set to `ImplementationSpecific`,
```
wasaga, line 384: do we need change `is` into `should` ? 
cause all three conditions should be met for the route to be considered a TCP route.

wasaga (approved):

docs/docs/k8s/ingress.md
alexfornuto: Why?

docs/docs/k8s/ingress.md
alexfornuto: I see where you're coming from, but since we're referring to the specific example above, these things is/are.

PR #3351, opened 2022-05-12
fix: rename rego rule function param name from in to ins

cfanbo:
when call rego function `ast.Member.Expr()` occur conflict

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/config/policy_ppl_test.go
+++ b/config/policy_ppl_test.go
@@ -623,14 +623,14 @@ deny = v {
 	v := merge_with_or(normalized)
 }
 
-invert_criterion_result(in) = out {
-	in[0]
-	out = array.concat([false], array.slice(in, 1, count(in)))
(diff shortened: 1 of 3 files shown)


cfanbo:
I don't think `ins` is appropriate for variables, maybe `arr` is better, but the problem here is conflict, see example https://play.openpolicyagent.org/p/DyCJjkkvoi

desimone:
@cfanbo could you elaborate on what a better variable would be and why? As you note, `arr` is a reserved keyword in rego so I don't think that works. Thanks!

cfanbo:
`in` is a reserved word, but `arr` is not a reserved word, Here I think `arr` is more better.

PR #3365, opened 2022-05-16
storage: add filter expressions, upgrade go to 1.18.1

calebdoxsey:
## Summary
This PR adds a `FilterExpression` type to the storage package. This type represents a query filter derived from a JSON object similar to this:

```json
{
  "$and": [
    { "id": "example1" },
    { "$index": "10.0.0.1" }
  ]
}
```

Supported expressions are:

- `$and`: an array of sub expressions, each of which must be true for `$and` to be true
- `$or`: an array of sub expressions, at least one of which must be true for `$or` to be true
- `id`: for filtering a list of records by the record id
- `$index`: for filtering a list of records by an index value. Only CIDR indexing is currently supported.

The upgrade to 1.18 is necessary to get the new `net/netip` package, which has handy types and functions for implementing the CIDR indexing.

## Related issues
Fixes https://github.com/pomerium/internal/issues/855

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/storage/filter.go
+++ b/pkg/storage/filter.go
@@ -1,0 +1,143 @@
+package storage
+
+import (
+	"fmt"
+	"sort"
+
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// A FilterExpression describes an AST for record stream filters.
+type FilterExpression interface {
+	isFilterExpression()
+}
+
+// FilterExpressionFromStruct creates a FilterExpression from a protobuf struct.
+func FilterExpressionFromStruct(s *structpb.Struct) (FilterExpression, error) {
+	if s == nil {
+		return nil, nil
+	}
+	return filterExpressionFromStruct(nil, s)
+}
+
+func filterExpressionFromValue(path []string, v *structpb.Value) (FilterExpression, error) {
+	switch vv := v.GetKind().(type) {
+	case *structpb.Value_ListValue:
+		var or OrFilterExpression
+		for _, vvv := range vv.ListValue.Values {
+			e, err := filterExpressionFromValue(path, vvv)
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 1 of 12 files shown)


pkg/storage/filter.go
```diff
@@ -0,0 +1,143 @@
...
+		return EqualsFilterExpression{
+			Fields: path,
+			Value:  vv.StringValue,
+		}, nil
+	}
+	return nil, fmt.Errorf("unsupported struct value type for eq: %T", v.GetKind())
+}
+
+// An OrFilterExpression represents a logical-or comparison operator.
+type OrFilterExpression []FilterExpression
```
wasaga, line 128: Is that the right name for this type? It's not used just for $or.
Maybe FilterExpressionList?

calebdoxsey: It conveys meaning. If we used FilterExpressionList we wouldn't be able to distinguish between `$and` and `$or`.

pkg/storage/filter.go
```diff
@@ -0,0 +1,143 @@
...
+
+	for _, f := range fs {
+		v := s.GetFields()[f]
+		switch f {
+		case "$and":
+			expr, err := filterExpressionFromValue(path, v)
+			if err != nil {
+				return nil, err
+			}
+			or, ok := expr.(OrFilterExpression)
```
wasaga, line 57: should it be `AndFilterExpression` here?

calebdoxsey: yes, good catch

calebdoxsey: nevermind, it was an `or`. By default we're treating a list as an `or`. The `$and` transforms that to an `and`.

wasaga (approved):

PR #3368, opened 2022-05-17
storage: add filtering to SyncLatest

calebdoxsey:
## Summary
Add `recordType` and `FilterExpression` to the arguments for `SyncLatest`. Also updates the in-memory and redis backends to support filtering. Filtering is implemented via a function applied to all the results.

## Related issues
Fixes https://github.com/pomerium/internal/issues/854
Fixes https://github.com/pomerium/internal/issues/853

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/storage/inmemory/stream.go
+++ b/pkg/storage/inmemory/stream.go
@@ -17,6 +17,11 @@ func newSyncLatestRecordStream(
 	if err != nil {
 		return nil, err
 	}
+	if recordType != "" {
+		filter = filter.And(func(record *databroker.Record) (keep bool) {
+			return record.GetType() == recordType
+		})
+	}
 
 	var ready []*databroker.Record
 	generator := storage.FilteredRecordStreamGenerator(
--- a/pkg/storage/inmemory/stream.go
+++ b/pkg/storage/inmemory/stream.go
@@ -10,9 +10,16 @@ import (
 func newSyncLatestRecordStream(
 	ctx context.Context,
 	backend *Backend,
-) storage.RecordStream {
+	recordType string,
+	expr storage.FilterExpression,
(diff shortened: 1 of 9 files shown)


pkg/storage/storage.go
```diff
@@ -40,7 +40,7 @@ type Backend interface {
 	// Sync syncs record changes after the specified version.
 	Sync(ctx context.Context, serverVersion, recordVersion uint64) (RecordStream, error)
 	// SyncLatest syncs all the records.
-	SyncLatest(ctx context.Context) (serverVersion uint64, stream RecordStream, err error)
+	SyncLatest(ctx context.Context, recordType string, filter FilterExpression) (serverVersion, recordVersion uint64, stream RecordStream, err error)
```
wasaga, line 43: maybe it's time to make returned value a struct ?

pkg/storage/inmemory/stream.go
```diff
@@ -10,17 +10,35 @@ import (
...
 	backend *Backend,
-) storage.RecordStream {
+	recordType string,
+	expr storage.FilterExpression,
+) (storage.RecordStream, error) {
+	filter, err := storage.RecordStreamFilterFromFilterExpression(expr)
+	if err != nil {
+		return nil, err
+	}
+	if recordType != "" {
```
wasaga, line 20: do we have records w/o type?

wasaga (approved):

pkg/storage/inmemory/stream.go
calebdoxsey: the empty string means return all records

PR #3369, opened 2022-05-17
databroker: add support for query filtering

calebdoxsey:
## Summary
Add support to the databroker for filtering on a query call. The filter is structured as a protobuf struct, mirroring a JSON object.

## Related issues
Fixes https://github.com/pomerium/internal/issues/857

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/databroker/server_test.go
```diff
@@ -135,21 +137,45 @@ func TestServer_Query(t *testing.T) {
...
+					}}),
+				}}),
+			},
+		},
+		Limit: 10,
 	})
 	assert.NoError(t, err)
+
+	if assert.Len(t, res.Records, 4) {
+	}
```
wasaga, line 178: ?

internal/databroker/server.go
```diff
@@ -145,7 +146,12 @@ func (srv *Server) Query(ctx context.Context, req *databroker.QueryRequest) (*da
 		return nil, err
 	}
 
-	_, _, stream, err := db.SyncLatest(ctx, req.GetType(), nil)
+	expr, err := storage.FilterExpressionFromStruct(req.GetFilter())
+	if err != nil {
+		return nil, err
```
wasaga, line 151: Maybe wrap an error as there are multiple error conditions in that func

wasaga (approved):

internal/databroker/server_test.go
calebdoxsey: meant to add some checks there. Nice catch.

internal/databroker/server.go
calebdoxsey: Yes good idea. I made it return `InvalidArgument` in that case.

Issue #3355, opened 2022-05-12
authenticate_service_url change requires restart


wasaga:
## What happened?

1. set `authenticate_service_url` to`https://auth.localhost.pomerium.io`
2. start pomerium
3. change `authenticate_service_url` to `https://login.localhost.pomerium.io`
4. navigate in browser to some route that requires authentication, you will get redirected to the `https://login.localhost.pomerium.io`, however the route would be 404. 
5. you have to restart pomerium for the change to take effect. 

## What did you expect to happen?

Change applied without need for restart.

## What's your environment like?

- Pomerium version 0.17.3
- all-in-one

PR #3378, opened 2022-05-26
authenticate: allow changing the authenticate service URL at runtime

calebdoxsey:
## Summary
Update the way HTTP handlers are added to the control plane so that the authenticate service URL can be changed at runtime.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/3355 

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/config_source.go
+++ b/config/config_source.go
@@ -139,9 +139,6 @@ func NewFileOrEnvironmentSource(
 		config:     cfg,
 	}
 	src.watcher.Add(configFile)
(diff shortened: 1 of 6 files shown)


wasaga:
- authenticate URL update works now
- health check URIs `/ping`, `/healthz` stopped to work

calebdoxsey:
I think this may be broken due to the order the handlers were added. Will try re-ordering.

wasaga (approved):

PR #3389, opened 2022-06-02
postgres: fix CIDR query

calebdoxsey:
## Summary
Currently when we try to use an `$index: ...` query, we don't check the value to make sure that it's a valid CIDR before attempting to use the GIST index. This results in an error: 

```
{"level":"error","error":"rpc error: code = Unknown desc = ERROR: invalid input syntax for type inet: \"\" (SQLSTATE 22P02)","time":"2022-06-02T16:23:07Z","message":"authorize/store: error retrieving record"}
```

This PR changes the code so that we only use the CIDR index when `$index` is set to a valid IP address or prefix.

## Related issues
Fixes https://github.com/pomerium/internal/issues/872

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/storage/postgres/filter.go
+++ b/pkg/storage/postgres/filter.go
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"fmt"
+	"net/netip"
 	"strings"
 
 	"github.com/pomerium/pomerium/pkg/storage"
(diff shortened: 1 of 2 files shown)


pkg/storage/postgres/filter.go
```diff
@@ -39,8 +40,12 @@ func addFilterExpressionToQuery(query *string, args *[]interface{}, expr storage
 			*args = append(*args, expr.Value)
 			return nil
 		case "$index":
-			*query += schemaName + "." + recordsTableName + ".index_cidr >>= " + fmt.Sprintf("$%d", len(*args)+1)
-			*args = append(*args, expr.Value)
+			if isCIDR(expr.Value) {
+				*query += schemaName + "." + recordsTableName + ".index_cidr >>= " + fmt.Sprintf("$%d", len(*args)+1)
+				*args = append(*args, expr.Value)
+			} else {
+				*query += " false "
```
wasaga, line 47: why do we set query to `false` and not return an error instead?

calebdoxsey: It's not technically an error. Like you may want to search on an indexed field that's not a CIDR. We don't currently support other types of indexes, but might in the future. We use `false` because we have to put something otherwise postgres is going to complain about an invalid query, but we actually just don't want to perform the search on that field. (in other words if I return an error here it will still be broken because we always do a search by id and index, and we can't always control the input to make sure its valid)

wasaga: I'm confused - per RFC the index was strongly typed

```json
"$index": { "cidr": "192.168.0.0/16" }
```

wasaga (approved):

PR #3406, opened 2022-06-06
postgres: only use serializable for record updates, retry serialization errors

calebdoxsey:
## Summary
We are occasionally seeing errors like:

> postgres_1                         | 2022-06-06 18:37:40.636 UTC [62] ERROR:  could not serialize access due to read/write dependencies among transactions
> postgres_1                         | 2022-06-06 18:37:40.636 UTC [62] DETAIL:  Reason code: Canceled on identification as a pivot, during write.
> postgres_1                         | 2022-06-06 18:37:40.636 UTC [62] HINT:  The transaction might succeed if retried.
> postgres_1                         | 2022-06-06 18:37:40.636 UTC [62] STATEMENT:  
> postgres_1                         | 			INSERT INTO pomerium.record_changes (type, id, version, data, modified_at, deleted_at)
> postgres_1                         | 			VALUES ($1, $2, $3, $4, $5, $6)
> postgres_1                         |

This PR makes it so we make less SQL queries in the serializable transaction to reduce the likelihood of a serialization error (previously we enforced type limits in the transaction, now we will do that after updating the records in a normal transaction). We will also retry up to 5 times.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/storage/postgres/postgres.go
+++ b/pkg/storage/postgres/postgres.go
@@ -261,8 +261,8 @@ func putRecord(ctx context.Context, q querier, record *databroker.Record) error 
 	} else {
 		_, err = q.Exec(ctx, `
 			DELETE FROM `+schemaName+`.`+recordsTableName+`
-			WHERE type=$1 AND id=$2 AND version<$3
-		`, record.GetType(), record.GetId(), record.GetVersion())
+			WHERE type=$1 AND id=$2
+		`, record.GetType(), record.GetId())
 	}
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/storage/postgres/postgres.go
```diff
@@ -261,8 +261,8 @@ func putRecord(ctx context.Context, q querier, record *databroker.Record) error
 	} else {
 		_, err = q.Exec(ctx, `
 			DELETE FROM `+schemaName+`.`+recordsTableName+`
-			WHERE type=$1 AND id=$2 AND version<$3
```
calebdoxsey, line 264: checking the version here was unnecessary because now the updates are serialized, so a row won't get updated concurrently

calebdoxsey:
Superseded by https://github.com/pomerium/pomerium/pull/3408

PR #3412, opened 2022-06-09
databroker: add support for syncing by type

calebdoxsey:
## Summary
Now that we no longer require record versions to be contiguous, we can support filtering records by type in the call to `Sync`. This should reduce unnecessary network traffic. And in the case of Postgres we will never even receive the records in the databroker since they will be filtered by SQL.

## Related issues
- https://github.com/pomerium/pomerium-console/issues/2590

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/grpc/databroker/fast_forward.go
```diff
@@ -33,18 +32,7 @@ func newFastForwardHandler(ctx context.Context, handler SyncerHandler) SyncerHan
...
 
 func (ff *fastForwardHandler) update(ctx context.Context, c *ffCmd) {
-	versions := make([]uint64, len(c.records))
-	for i, r := range c.records {
-		versions[i] = r.Version
-	}
-
-	now := time.Now()
 	ff.handler.UpdateRecords(ctx, c.serverVersion, c.records)
-	log.Info(ctx).
```
calebdoxsey, line 43: These logs are extremely noisy. I don't think they are useful. Code using the syncer should log records if they want to, but we shouldn't do it here.

wasaga (approved):

Issue #3299, opened 2022-04-26
double URL encoding in pomerium_redirect_uri in forwardauth mode


krasnican:
Hello,

we're using pomerium as forwardauth server. We are seeing unwanted behavior where, if URL contains ? char, user is redirected (pomerium_redirect_uri) to malformed/double-encoded URL when IDP authentication is passed.

It also seems that encoding of character "?" (and possibly others) in pomerium_redirect_uri differs between deployments pomerium as proxy and pomerium as forwardauth service (char "?" seems to be encoded twice for forwardauth setup).

(please note that underlying pomerium instance and configuration is the same for both requests bellow).

### request via forwardauth service:
request: `curl -v https://<pomerium-ip> -H "X-Forwarded-Host: protected-host.tld" -H "X-Forwarded-Uri: /asdf?" -H "Host: forwardauth.tld"`

response:
```
< HTTP/1.1 302 Found
< strict-transport-security: max-age=31536000; includeSubDomains; preload
< x-frame-options: SAMEORIGIN
< x-xss-protection: 1; mode=block
< content-type: text/html; charset=UTF-8
< x-pomerium-intercepted-response: true
< location: https://auth.tld/.pomerium/sign_in?pomerium_expiry=1650972041&pomerium_idp_id=redacted&pomerium_issued=1650971741&pomerium_redirect_uri=https%3A%2F%2Fprotected-host.tld%2Fasdf%253F&pomerium_signature=OibO1Utn24dDLms-bAEjB9kb-IG3aF9ZQ_MJg8uFWzU%3D
< content-length: 1271
< date: Tue, 26 Apr 2022 11:15:41 GMT
< server: envoy
```

`...pomerium_redirect_uri=https%3A%2F%2Fprotected-host.tld%2Fasdf%253F...`
decoded URL:
`pomerium_redirect_uri=https://protected-host.tld/asdf%3F`

Please note that char "?" is encoded twice (? > %3F > %253F) in pomerium_redirect_uri.

### request via pomerium proxy
request: `curl -v https://<pomerium-ip>/asdf? -H "Host: protected-host.tld"`

response:
```
< HTTP/1.1 302 Found
< strict-transport-security: max-age=31536000; includeSubDomains; preload
< x-frame-options: SAMEORIGIN
< x-xss-protection: 1; mode=block
< content-type: text/html; charset=UTF-8
< x-pomerium-intercepted-response: true
[...]

PR #3479, opened 2022-07-13
authorize: fix x-forwarded-uri

calebdoxsey:
## Summary
Traefik sends the path and query string for the original URL as a header named `X-Forwarded-Uri`. Previously we were setting this as the path, but we should've been parsing out the `?` and setting that as the RawQuery.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/3299

## User Explanation
Fixes an escaping bug with Traefik.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/grpc.go
+++ b/authorize/grpc.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/pomerium/pomerium/authorize/evaluator"
 	"github.com/pomerium/pomerium/config"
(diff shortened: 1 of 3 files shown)


desimone (approved):

calebdoxsey:
The issue was using `RawPath` instead of `Path`.

PR #3492, opened 2022-07-19
options: fix overlapping certificate test

calebdoxsey:
## Summary
When testing for overlapping certificates we previously only supported testing files. We now support files or base64 encoded bytes.

## Related issues
Fixes https://github.com/pomerium/internal/issues/915

## User Explanation
I don't know how to explain this fix to a regular user. We fixed a bug in the internal machinations of pomerium.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -1237,7 +1237,7 @@ func (o *Options) Checksum() uint64 {
 	return hashutil.MustHash(o)
 }
 
-func (o Options) indexCerts(ctx context.Context) certsIndex {
+func (o *Options) indexCerts(ctx context.Context) certsIndex {
 	idx := make(certsIndex)
 
 	if o.CertFile != "" {
@@ -1248,9 +1248,7 @@ func (o Options) indexCerts(ctx context.Context) certsIndex {
 			idx.addCert(cert)
 		}
 	} else if o.Cert != "" {
-		if data, err := base64.StdEncoding.DecodeString(o.Cert); err != nil {
-			log.Error(ctx).Err(err).Msg("bad base64 for local cert: skipped")
-		} else if cert, err := cryptutil.ParsePEMCertificate(data); err != nil {
+		if cert, err := cryptutil.ParsePEMCertificateFromBase64(o.Cert); err != nil {
 			log.Error(ctx).Err(err).Msg("parsing local cert: skipped")
 		} else {
 			idx.addCert(cert)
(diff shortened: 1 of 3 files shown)


config/options.go
```diff
@@ -1248,19 +1248,20 @@ func (o Options) indexCerts(ctx context.Context) certsIndex {
...
+		if cert, err := cryptutil.ParsePEMCertificateFromBase64(o.Cert); err != nil {
 			log.Error(ctx).Err(err).Msg("parsing local cert: skipped")
 		} else {
 			idx.addCert(cert)
 		}
 	}
 
 	for _, c := range o.CertificateFiles {
-		cert, err := cryptutil.ParsePEMCertificateFromFile(c.CertFile)
+		cert, err := cryptutil.ParsePEMCertificateFromBase64(c.CertFile)
```
wasaga, line 1259: Why a B64-encoded cert may end up in `CertFile`?

calebdoxsey: The current settings store raw certificates as a base64 encoded string. I either don't know why or don't remember why. Changing this setting would be a breaking change.

calebdoxsey: We do not have separate keys for this: https://www.pomerium.com/docs/reference/certificates

```yaml
certificates:
  - cert: "$HOME/.acme.sh/authenticate.example.com_ecc/fullchain.cer"
    key: "$HOME/.acme.sh/authenticate.example.com_ecc/authenticate.example.com.key"
  - cert: "$HOME/.acme.sh/verify.example.com_ecc/fullchain.cer"
    key: "$HOME/.acme.sh/verify.example.com_ecc/verify.example.com.key"
  - cert: "$HOME/.acme.sh/prometheus.example.com_ecc/fullchain.cer"
    key: "$HOME/.acme.sh/prometheus.example.com_ecc/prometheus.example.com.key"
```

wasaga: oh I see we accept both files and b64 strings there. 

may I suggest we resolve the confusion by 
- renaming `CertificateFiles` var into `Certificates` 
- renaming `certificateFilePair` type into `certificatePair`
- adding `certBytes` and `keyBytes` to `certificatePair`
- writing a custom mapstructure func to parse the `certificates` field and decide there if it's a b64 string or a file path, and fill `certFile`/`certBytes` and `keyFile`/`keyBytes` accordingly. As I understand, it may only happen while parsing a config file, and is not the case for converting Protobuf `Settings` into `Options`, as the former only has `CertBytes`/`KeyBytes` fields.

desimone: Can we pull the above into a separate issue / proposal because -- unless I am missing something -- this would be a breaking change?

wasaga: Sure, we can eject refactoring into separate issue. This PR fixes the original issue with spillover of the cert data into logs. 

NB: what I suggested above would not constitute a breaking change.

wasaga (approved):

PR #3498, opened 2022-07-25
add lease name to the log

wasaga:
## Summary

Adds `lease_name` to the databroker leaser log to differentiate between various lease holders. 

## User Explanation

Only internal logs are affected. 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone (approved):
LGTM so long as `locker` can not be nil here.

PR #3517, opened 2022-07-28
atomicutil: use atomicutil.Value wherever possible

calebdoxsey:
## Summary
We now have an `atomicutil.Value` type which is a strongly-typed generic `atomic.Value`. This PR looks for anywhere we used `atomic.Value` and replaces it with `atomicutil.Value`. In many places we had custom types like `atomicAuthenticateState` which are now unneeded and were removed.

## Related issues
- https://github.com/pomerium/internal/issues/934

## User Explanation
Not user facing

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/atomicutil/value.go
+++ b/internal/atomicutil/value.go
@@ -17,7 +17,16 @@ func NewValue[T any](init T) *Value[T] {
 
 // Load loads the value atomically.
 func (v *Value[T]) Load() T {
-	return v.value.Load().(T)
+	var def T
+	if v == nil {
+		return def
+	}
+
+	cur := v.value.Load()
+	if cur == nil {
+		return def
+	}
+	return cur.(T)
 }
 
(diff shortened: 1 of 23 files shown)


internal/atomicutil/value.go
```diff
@@ -17,7 +17,16 @@ func NewValue[T any](init T) *Value[T] {
...
+	var def T
+	if v == nil {
+		return def
+	}
+
+	cur := v.value.Load()
+	if cur == nil {
+		return def
+	}
+	return cur.(T)
```
calebdoxsey, line 29: The extra code here was to make it a little more defensive about misuse. The type should always be created with `atomicutil.NewValue` but if its not at least with this added code `.Load` won't panic.

wasaga (approved):

Issue #3505, opened 2022-07-28
Istio mTLS doesn't work in strict mode due to incorrect outbound authority


bennycooly:
## What happened?

I've configured Istio mTLS in strict mode and deployed pomerium with `insecure_server` and `grpc_insecure` on. However, Istio is unable to match the outbound gRPC traffic coming from the pods since the authority is set to `127.0.0.1:<outbound_port>`.

Is there a way we can set the outbound authority when constructing the local Envoy gRPC outbound instance?

## What did you expect to happen?

mTLS should succeed from pod to pod communication with gRPC.

## How'd it happen?

1. Install pomerium using the official helm chart, setting `insecure` on
2. Pods are unable to fully initialize, since the call to databroker pod fails from the other pods

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`): 0.17.3-1651784655+f4bad550
- Server Operating System/Architecture/Cloud: Kubernetes + Istio

## What's your config.yaml?

```config.yaml
autocert: false
dns_lookup_family: V4_ONLY
address: :80
grpc_address: :80
authenticate_service_url: https://authenticate.my.domain
authorize_service_url: http://pomerium-authorize.pomerium.svc.cluster.local
databroker_service_url: http://pomerium-databroker.pomerium.svc.cluster.local
insecure_server: true
grpc_insecure: true

pomerium_debug: "true"
forward_auth_url: http://pomerium-proxy.pomerium
databroker_storage_tls_skip_verify: false
```

## What did you see in the logs?

In this scenario, `10.11.3.29` is the address of the databroker pod.

Logs for `pomerium-authenticate`:
```logs
6:02AM ERR controlplane: error storing configuration event, retrying error="rpc error: code = Unavailable desc = upstream connect error or disconnect/reset before headers. reset reason: connection termination"
{"level":"info","syncer_id":"databroker","syncer_type":"type.googleapis.com/pomerium.config.Config","time":"2022-07-28T06:03:41Z","message":"initial sync"}
{"level":"error","syncer_id":"databroker","syncer_type":"type.googleapis.com/pomerium.config.Config","error":"rpc error: code = Unavailable desc = upstream connect error or disconnect/reset before headers. reset reason: connection termination","time":"2022-07-28T06:03:41Z","message":"error during initial sync"}
[...]

PR #3545, opened 2022-08-10
envoyconfig: add authority header to outbound gRPC requests

calebdoxsey:
## Summary
Update the envoy config for outbound gRPC requests so that the authority header gets rewritten to match the cluster's destination address.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/3505

## User Explanation
Fixes an issue with route matching for Istio.

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


desimone:
Marking as blocked until we can confirm the fix

desimone (approved):
Works, per https://github.com/pomerium/pomerium/issues/3505#issuecomment-1226085011

PR #3590, opened 2022-08-29
autocert: add support for ACME TLS-ALPN

calebdoxsey:
## Summary
Add support for the ACME TLS-ALPN challenge. This makes it so that autocert can provision certificates using the existing HTTPS listener on port 443 without requiring another HTTP server on port 80.

This is implemented by adding config to envoy that causes any requests for the application protocol `acme-tls/1` to be forwarded to a local TLS server. All other requests will be handled as before.

## Related issues
Fixes https://github.com/pomerium/internal/issues/710

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -26,6 +26,8 @@ type Config struct {
 	MetricsPort string
 	// DebugPort is the port the debug listener is running on.
 	DebugPort string
+	// ACMETLSPort is the port that handles the ACME TLS-ALPN challenge.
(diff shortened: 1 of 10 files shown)


desimone (approved):

desimone:
LGTM, do you want to address the docs in a separate PR?

calebdoxsey:
Sure

PR #3593, opened 2022-08-30
envoyconfig: add virtual host domains for certificates in addition to routes

### LINKED ISSUE #3577: Self-signed certificate surfaced on unknown routes
## What happened?
Non-routes are exposed using self-signed certificate

404

## How'd it happen?

1. Wildcard DNS for domain
2. Browse to https://test3.example.com

- pomerium: 0.18.0-1658889797+89a105c8
- envoy: 1.21.3+4861429dfffb599f28b9399c34ea2a2c268bfb6d10aca0a53bc9b67d847a4595
- Pomerium console version: 0.18.0-1658971952 + 7b8e18a8 + 2022-07-27T21:30:12-04:00
- Server Operating System/Architecture/Cloud: 2x CentOS Stream on premise virtual machines
- postgres VMs for console and databroker

## What's your config.yaml?

```config.yaml
authenticate_service_url: https://example.com

certificates:
  - cert: /path/to/cert.pem
    key: /path/to/key.pem

signing_key: SECRET
metrics_address: localhost:9999

http_redirect_addr: :80

idp_provider: google
idp_client_id: clientid
idp_client_secret: secret
idp_service_account: serviceaccount

idp_refresh_directory_timeout: 10m
idp_refresh_directory_internal: 20m

cookie_secret: cookiesecret

shared_secret: sharedsecret

databroker_storage_type: postgres
databroker_storage_connection_string: postgres://:@postgres:5432

policy:
  - from: https://console.example.com
    to: https://127.0.0.1:8701
    pass_identity_headers: true
    allowed_groups:
      - group@example.com
    allowed_users:
      - example@example.com
```

```logs
{
  "level": "info",
  "service": "envoy",
  "upstream-cluster": "",
  "method": "GET",
  "authority": "test3.example.com",
  "path": "/",
  "user-agent": "useragent",
  "referer": "",
  "forwarded-for": "x.x.x.x",
  "request-id": "xxx",
  "duration": 0.331926,
  "size": 0,
  "response-code": 404,
  "response-code-details": "route_not_found",
  "time": "2022-08-29T12:04:54-04:00",
  "message": "http-request"
}
```

## Additional context

Wildcard certificate is specified in config.yaml, seems related to this change:
https://docs.pomerium.com/docs/overview/upgrading#self-signed-fallback-certificates

User gets an invalid cert browser splash, if they proceed through they get the 404.

calebdoxsey:
## Summary
Currently we only add domains to the envoy config based on routes. This means if we have a route `from.example.com` and a wildcard certificate `*.example.com`, that certificate will only be served to requests from `from.example.com`. So `unknown.example.com` would end up receiving a self-signed certificate. This PR updates the code to also add the `*.example.com` as a handled domain. This domain will have no routes attached to it, but it will serve the proper certificate, so the user would no longer receive a self-signed certificate.

Unlike route matching envoy matches filter chains based on most specific to least specific. So if there is a filter chain for `*.example.com` and another one for `from.example.com`, if a request comes in for `from.example.com` it will be sent to that filter chain regardless of the ordering of the filter chains.  (so it's ok if `*.example.com` occurs first in the list) 

## Related issues
Fixes #3577 

 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/envoyconfig/listeners.go
+++ b/config/envoyconfig/listeners.go
@@ -342,6 +342,8 @@ func (b *Builder) buildMainHTTPConnectionManagerFilter(
 		LuaFilter(luascripts.CleanUpstream),
 		LuaFilter(luascripts.RewriteHeaders),
 	}
+	// only return 421s for non-wildcard domains because the lua script doesn't understand how to
+	// parse wildcards properly
 	if tlsDomain != "" && !strings.Contains(tlsDomain, "*") {
 		filters = append(filters, LuaFilter(fmt.Sprintf(luascripts.FixMisdirected, tlsDomain)))
 	}
--- a/config/envoyconfig/listeners.go
+++ b/config/envoyconfig/listeners.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"strings"
 	"time"
 
 	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
@@ -118,7 +119,7 @@ func (b *Builder) buildMainListener(ctx context.Context, cfg *config.Config) (*e
 	}
 	listenerFilters = append(listenerFilters, TLSInspectorFilter())
 
-	chains, err := b.buildFilterChains(cfg.Options, cfg.Options.Addr,
+	chains, err := b.buildFilterChains(cfg, cfg.Options.Addr,
 		func(tlsDomain string, httpDomains []string) (*envoy_config_listener_v3.FilterChain, error) {
 			filter, err := b.buildMainHTTPConnectionManagerFilter(cfg.Options, httpDomains, tlsDomain)
 			if err != nil {
@@ -235,15 +236,15 @@ func (b *Builder) buildMetricsListener(cfg *config.Config) (*envoy_config_listen
 }
 
 func (b *Builder) buildFilterChains(
-	options *config.Options, addr string,
+	cfg *config.Config, addr string,
 	callback func(tlsDomain string, httpDomains []string) (*envoy_config_listener_v3.FilterChain, error),
 ) ([]*envoy_config_listener_v3.FilterChain, error) {
-	allDomains, err := getAllRouteableDomains(options, addr)
+	allDomains, err := getAllRouteableDomains(cfg.Options, addr)
 	if err != nil {
 		return nil, err
 	}
 
-	tlsDomains, err := getAllTLSDomains(options, addr)
+	tlsDomains, err := getAllTLSDomains(cfg, addr)
 	if err != nil {
 		return nil, err
 	}
@@ -251,7 +252,7 @@ func (b *Builder) buildFilterChains(
 	var chains []*envoy_config_listener_v3.FilterChain
 	chains = append(chains, b.buildACMETLSALPNFilterChain())
 	for _, domain := range tlsDomains {
-		routeableDomains, err := getRouteableDomainsForTLSServerName(options, addr, domain)
+		routeableDomains, err := getRouteableDomainsForTLSServerName(cfg.Options, addr, domain)
 		if err != nil {
 			return nil, err
 		}
@@ -341,7 +342,7 @@ func (b *Builder) buildMainHTTPConnectionManagerFilter(
 		LuaFilter(luascripts.CleanUpstream),
 		LuaFilter(luascripts.RewriteHeaders),
 	}
-	if tlsDomain != "" && tlsDomain != "*" {
+	if tlsDomain != "" && !strings.Contains(tlsDomain, "*") {
 		filters = append(filters, LuaFilter(fmt.Sprintf(luascripts.FixMisdirected, tlsDomain)))
 	}
 	filters = append(filters, HTTPRouterFilter())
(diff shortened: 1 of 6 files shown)


pkg/cryptutil/certificates.go
```diff
@@ -219,6 +219,17 @@ func GenerateSelfSignedCertificate(domain string, configure ...func(*x509.Certif
 	return &cert, nil
 }
 
+// EncodeCertificate encodes a TLS certificate into PEM compatible byte slices.
+func EncodeCertificate(cert *tls.Certificate) (pemCertificateBytes, pemKeyBytes []byte) {
+	publicKeyBytes := cert.Certificate[0]
```
desimone, line 224: Might be worth checking the length here

pkg/cryptutil/certificates.go
```diff
@@ -219,6 +219,17 @@ func GenerateSelfSignedCertificate(domain string, configure ...func(*x509.Certif
 	return &cert, nil
 }
 
+// EncodeCertificate encodes a TLS certificate into PEM compatible byte slices.
+func EncodeCertificate(cert *tls.Certificate) (pemCertificateBytes, pemKeyBytes []byte) {
```
desimone, line 223: To be on the safe side, do we want to return an error from this function since it's possible `x509.MarshalPKCS8PrivateKey` could fail?

config/envoyconfig/listeners.go
```diff
@@ -341,7 +342,7 @@ func (b *Builder) buildMainHTTPConnectionManagerFilter(
 		LuaFilter(luascripts.CleanUpstream),
 		LuaFilter(luascripts.RewriteHeaders),
 	}
-	if tlsDomain != "" && tlsDomain != "*" {
+	if tlsDomain != "" && !strings.Contains(tlsDomain, "*") {
```
desimone, line 345: This function is rather complicated to follow. A comment here (and above?) would be helpful.

pkg/cryptutil/certificates.go
```diff
@@ -219,6 +219,17 @@ func GenerateSelfSignedCertificate(domain string, configure ...func(*x509.Certif
 	return &cert, nil
 }
 
+// EncodeCertificate encodes a TLS certificate into PEM compatible byte slices.
```
desimone, line 222:
```suggestion
// EncodeCertificate encodes a TLS certificate into PEM compatible byte slices. 
// Returns `nil`, `nil` if there is an error marshalling the PKCS8 private key. 
```

pkg/cryptutil/certificates.go
```diff
@@ -219,6 +219,17 @@ func GenerateSelfSignedCertificate(domain string, configure ...func(*x509.Certif
...
+func EncodeCertificate(cert *tls.Certificate) (pemCertificateBytes, pemKeyBytes []byte) {
+	publicKeyBytes := cert.Certificate[0]
+	privateKeyBytes, err := x509.MarshalPKCS8PrivateKey(cert.PrivateKey)
+	if err != nil {
+		return
```
desimone, line 227: Are `pemCertificateBytes` and `pemKeyBytes` guaranteed to be `nil` here?

pkg/cryptutil/certificates.go
desimone, line 223: Could we add a few unit tests to cover ` EncodeCertificate`?

1. Empty certificate
2. Nil `*tls.Certificate`
3. Malformed `tls.Certificate` `PrivateKey` ? Maybe not possible?
4. Others?

pkg/cryptutil/tls.go
```diff
@@ -63,6 +63,29 @@ func GetCertificateForDomain(certificates []tls.Certificate, domain string) (*tl
 	return GenerateSelfSignedCertificate(domain)
 }
 
+// GetCertificateDomains gets all the certificate's matching domain names.
```
desimone, line 66:
```suggestion
// GetCertificateDomains gets all the certificate's matching domain names.
// Will return `nil` if certificate is nil, empty, or x509 parsing fails. 
```

pkg/cryptutil/tls.go
```diff
@@ -63,6 +63,29 @@ func GetCertificateForDomain(certificates []tls.Certificate, domain string) (*tl
...
+func GetCertificateDomains(cert *tls.Certificate) []string {
```
desimone, line 67: Unit test coverage would be nice here.

config/envoyconfig/listeners.go
```diff
@@ -658,21 +659,31 @@ func getAllRouteableDomains(options *config.Options, addr string) ([]string, err
 	return allDomains.ToSlice(), nil
 }
 
-func getAllTLSDomains(options *config.Options, addr string) ([]string, error) {
-	allDomains, err := getAllRouteableDomains(options, addr)
+func getAllTLSDomains(cfg *config.Config, addr string) ([]string, error) {
+	domains := sets.NewSorted[string]()
```
desimone, line 663: ✨ generics ✨

desimone (changes requested):
Few suggestions, mostly around unit tests but otherwise LGTM!

pkg/cryptutil/certificates.go
calebdoxsey: Go does not have an easy way of generating certificates which aren't ed25519, ECDSA, or RSA. It's possible but I will have to import additional libraries and write functions to generate the certificates.

desimone (approved):

Issue #2150, opened 2021-04-28
safari does not respect `421` status for HTTP/2 connection reuse / coalescing 


desimone:
## What happened?

When using wildcard certificates, HTTP/2, Pomerium will respond with status `421` for different sites being proxied. Because of the way request coalescing works with HTTP/2, when we change the domain (authority) of the site we are proxying, (e.g. `authenticate.x.y.z` / `verify.x.y.z` / `httpbin.x.y.z`) Pomerium needs to tell the browser to reset the connection. However, it appears that Safari is not respecting that `421` and fails silently. 

## What did you expect to happen?

I expect to be able to use HTTP/2 and wildcard certificates across multiple domains for all major evergreen browsers. 

Note, this only seems to effect safari. 

<img width="1519" alt="safari" src="https://user-images.githubusercontent.com/1544881/116481119-6e2c5980-a837-11eb-8bd8-040c96800fee.png">
<img width="1392" alt="firefox" src="https://user-images.githubusercontent.com/1544881/116481121-6ec4f000-a837-11eb-9564-a6dc4f68bd6d.png">

## How'd it happen?

1. Start with wildcard certs with no fronting proxy
2. Use safari 14.1
2. Go to `verify.x.y.z` 
3. The callback to `authenticate.x.y.z` will fail as safari tries re-use the connection
4. Pomerium / envoy responds with a `421`
5. Safari returns a blank empty response

## What's your environment like?

- Safari Version 14.1 (16611.1.21.161.3)
- HTTP/2 enabled
- Wildcard certificates need to be used.
- Pomerium version (retrieve with `pomerium --version`): g91c7dc7
- Server Operating System/Architecture/Cloud: N/A (all forms)

## What's your config.yaml?

N/A

## What did you see in the logs?

```logs
{
  "level": "info",
  "service": "envoy",
  "upstream-cluster": "",
  "method": "GET",
  "authority": "authenticate.localhost.pomerium.io",
  "path": "/oauth2/callback",
  "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.85 Safari/537.36",
  "referer": "https://accounts.google.com/",
  "forwarded-for": "192.168.1.63",
  "request-id": "4cd598e4-929f-4540-ab9f-c1359e6ac174",
  "duration": 0.471583,
  "size": 0,
  "response-code": 421,
[...]

PR #3596, opened 2022-08-31
envoyconfig: add all routes to all filter chains

calebdoxsey:
## Summary
With HTTP/2 browsers will re-use a connection for the same IP address and TLS certificate. For example if `a.example.com` and `b.example.com` map to the same IP address and use a wildcard certificate (`*.example.com`) the browser may use the same connection for requests to `a.example.com` and `b.example.com`. As a consequence it's possible for an initial request to come on a TLS connection made to `a.example.com` and a subsequent request for `b.example.com` to come on the same connection.

In Envoy we currently only add the routes that are applicable to that domain name. This was done to save memory. But it means HTTP/2 connection coalescing breaks. We have a lua script that returns a 421 if we detect this happens, and *some* browsers detect this and create a new connection, but not all browsers. In particular iOS devices will not re-establish the connection and instead users will see a 404 page.

This PR reverts the changes made for https://github.com/pomerium/internal/issues/234, and adds all the routes to every filter chain.

## Related issues
Fixes https://github.com/pomerium/internal/issues/972
Fixes https://github.com/pomerium/pomerium/issues/2150

 

## Checklist
- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
In terms of memory impact, if the version of envoy we're using has similar characteristics, 100 routes currently uses about 600MB of ram. With this change it will use 3GB of ram. These numbers may be obsolete. I also don't know how realistic this use case is. Or if 3GB of ram is unacceptable. I would say it's not, I would expect to need a large server for this kind of use case, and 3GB of ram is not a lot for a large server.

desimone (approved):

PR #3688, opened 2022-10-20
authenticate: remove ecjson

calebdoxsey:
## Summary
Currently we support two session loaders in the authenticate service:

```go
state.sessionLoaders = []sessions.SessionLoader{headerStore, cookieStore}
```

The cookie store loads signed sessions. The header store loads encrypted sessions using the `ecjson` package. However, as far as I could tell, we never actually create `ecjson` encrypted payloads anywhere. 

Although the authenticate service can theoretically load a session from a header, it's in a format that nothing can produce. So it makes more sense to simply remove this capability.

## Related issues

 

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


wasaga (approved):

desimone:
Yes, this is outdated from when we used to store encrypted session data in the cookie.

PR #3720, opened 2022-11-01
Allow basic-auth for programmatic access

### LINKED ISSUE #3687: Allow Basic Auth to be used for programmatic access

We have several cli applications that communicate over HTTP, and we would love to use pomerium for these as pomerium provides loads of logging, auditing, AAA, and also allows for injection of mTLS and headers. It is amazing and we love it!

One of the features I recently learned about was the "programmatic access" function of pomerium whereby you can add a specific Authorization header and it will authenticate past pomerium for non browser environments. This is exactly what we need! But unfortunately these CLI applications only allow certain things to be added (for example they allow for a basic auth username and password to be supplied). Would it be possible to also allow for "basic auth" to be used instead of just the

"Authorization: Pomerium a.real.jwt"

For example the username could be a static "pomerium" and the password could be the JWT. This is also how some other sites do it with API keys (i.e. the API key is the password of a basic auth and the username is some static thing like "api")

We thought about writing our own proxy but I think it would be nicer to have in Pomerium itself

Basically, for "legacy" or static applications that do not support adding any additional headers past basic auth (as basic auth is very well adopted)

**Additional context**

Seros:
## Summary

Allow basic-auth for a more feasible way to authenticate for a wide range of web apps

## Related issues

Fixes #3687

## User Explanation

Check if basic-auth headers are available and if so use them to authenticate your app in a programmatic way, using `pomerium` as default user name and the JWT as password.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/internal/sessions/header/header_store.go
+++ b/internal/sessions/header/header_store.go
@@ -3,6 +3,7 @@
 package header
 
 import (
+	"encoding/base64"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


calebdoxsey:
The challenge with this is that other applications may be expecting to use Basic Auth for themselves and we'd be intercepting it. So this is a potentially breaking change.

For the `Authorization` header we had `Authorization: Pomerium` and `Authorization: Bearer Pomerium-` to help prevent that from happening. Is there a way we could do something similar for Basic Auth?

Seros:
Thanks for you comment! I see your point about the interception but I think this has already been worked around. As this checks for the specific username `pomerium` it should only interfere in such cases which probably are edge cases if people configure their apps to use the same name. So in most cases the apps should not have problems using their configured basic auth IMO. In case you don't agree we have to investigate for a different solution which then will probably not as portable as allowing basic auth.

Seros:
Hey,
just wanted to ask if there is anything else you need from my side or if we can continue with this PR being merged?

PR #3734, opened 2022-11-07
authorize: fix user caching

calebdoxsey:
## Summary
The `user.User` object was always being re-queried because we were relying on the `dataBrokerRecordVersion` which is intended for a `session.Session`. So if a `session.Session` were updated after a `user.User` we'd always end up requerying the `user.User`, because its record version number wouldn't change.

This will mean that `user.User` objects will now be cached for 60s.

## Related issues
- https://github.com/pomerium/internal/issues/1019#issuecomment-1304397383
 

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/databroker.go
+++ b/authorize/databroker.go
@@ -90,12 +90,11 @@ func (a *Authorize) getDataBrokerSessionOrServiceAccount(
 func (a *Authorize) getDataBrokerUser(
 	ctx context.Context,
 	userID string,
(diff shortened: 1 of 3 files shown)


desimone (commented):
code LGTM

wasaga (approved):

PR #3742, opened 2022-11-10
config: generate cookie secret if not set in all-in-one mode

### LINKED ISSUE #3740: quickstart: generate cookie secret

Currently the `cookie_secret` config setting is required:

```
{"level":"fatal","error":"error creating authenticate service: authenticate: 'COOKIE_SECRET' invalid cryptutil: got 0 bytes but want 32","time":"2022-11-10T19:18:05Z","message":"cmd/pomerium"}
```

To make it easier to get started, we should generate it like we do the signing key.

Generate the cookie secret if its not set.

calebdoxsey:
## Summary
Make the `cookie_secret` optional. When not set (in all-in-one mode), we will set it to the randomly generated shared key. This will mean on restart that all users will be logged out and have to re-authenticate.

## Related issues
Fixes #3740 

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -1190,6 +1190,7 @@ func (o *Options) GetCookieSecret() ([]byte, error) {
 	}
 
 	if IsAll(o.Services) && cookieSecret == "" {
+		log.WarnCookieSecret()
 		cookieSecret = randomSharedKey
 	}
 	if cookieSecret == "" {
--- a/config/options.go
+++ b/config/options.go
@@ -985,7 +985,7 @@ func (o *Options) GetSharedKey() ([]byte, error) {
 		sharedKey = string(bs)
 	}
 	// mutual auth between services on the same host can be generated at runtime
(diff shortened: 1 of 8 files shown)


wasaga (approved):

config/options.go
```diff
@@ -1188,6 +1188,14 @@ func (o *Options) GetCookieSecret() ([]byte, error) {
 		}
 		cookieSecret = string(bs)
 	}
+
+	if IsAll(o.Services) && cookieSecret == "" {
+		cookieSecret = randomSharedKey
```
desimone, line 1193: Maybe warn saying if the user reload without this being set; this will break all active sessions and link to our docs on how to set this. Because this will be a footgun / users will trigger this without realizing why it’s happening.

desimone (changes requested):

desimone (approved):

PR #3755, opened 2022-11-17
proxy: add userinfo and webauthn endpoints

calebdoxsey:
## Summary
Add `/.pomerium` and `/.pomerium/webauthn` endpoints to the proxy service. Currently these are only handled by authenticate. With these changes they will also be available on any domain Pomerium handles.

I also moved all the common handlers to `internal/handlers`. (I wanted to avoid any circular references in `httputil`, which should be for more generic HTTP functionality)

The webauthn relying party is now derived from the request so that the origin used depends on the domain where the webauthn handlers live. As a consequence devices will now be domain-specific. I don't think a device registered on `a.example.com` could be used for `b.example.com`, but maybe this needs more testing.

## Related issues
- https://github.com/pomerium/internal/issues/1014
- https://github.com/pomerium/internal/issues/1013
 

## Checklist

- [x] updated unit tests 
- [x] ready for review


calebdoxsey:
After discussion, we should attempt to use the parent top-level-domain as the origin. For example: `console.example.com` should use `example.com` as the origin. This should make it possible to re-use devices across domain names for most use cases. For cases where routes are spread across different TLDs devices will only be re-usable for the same domain. In the future we can provide the ability to customize this behavior similar to `COOKIE_DOMAIN`.

pkg/webauthnutil/webauthnutil.go
```diff
@@ -2,14 +2,30 @@
...
 func GetRelyingParty(r *http.Request, client databroker.DataBrokerServiceClient) *webauthn.RelyingParty {
-	return webauthn.NewRelyingParty(urlutil.GetOrigin(r), NewCredentialStorage(client))
+	return webauthn.NewRelyingParty(
+		"https://"+GetEffectiveDomain(r),
+		NewCredentialStorage(client),
+	)
+}
+
+// GetEffectiveDomain returns the effective domain for an HTTP request.
+func GetEffectiveDomain(r *http.Request) string {
```
desimone, line 22: Maybe possible to use https://pkg.go.dev/golang.org/x/net/publicsuffix#EffectiveTLDPlusOne instead?

calebdoxsey:
It now uses the top-level domain name as the origin. Testing this with two routes it appears to register and authenticate existing devices properly.

calebdoxsey:
(worth mentioning, this is a breaking change, the next version of Pomerium will require devices to be re-registered because now they will be attached to a different name)

desimone:
Adding breaking tag so it's flagged.

internal/handlers/jwks.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// JWKSHandler returns the /.well-known/pomerium/jwks.json handler.
+func JWKSHandler(rawSigningKey string) http.Handler {
+	return cors.AllowAll().Handler(httputil.HandlerFunc(func(w http.ResponseWriter, r *http.Request) error {
+		var jwks jose.JSONWebKeySet
+		if rawSigningKey != "" {
+			decodedCert, err := base64.StdEncoding.DecodeString(rawSigningKey)
+			if err != nil {
+				return httputil.NewError(http.StatusInternalServerError, errors.New("bad signing key"))
```
desimone, line 22:
```suggestion
				return httputil.NewError(http.StatusInternalServerError, errors.New("bad base64 encoding for signing key"))
```

proxy/proxy.go
```diff
@@ -69,6 +71,7 @@ func New(cfg *config.Config) (*Proxy, error) {
 		currentOptions: config.NewAtomicOptions(),
 		currentRouter:  atomicutil.NewValue(httputil.NewRouter()),
 	}
+	p.webauthn = webauthn.New(p.getWebauthnState)
```
desimone, line 74: I think you can put the assignment in the above struct assignement? Maybe there's a reason I'm missing though.

internal/controlplane/http.go
```diff
@@ -37,7 +38,7 @@ func (srv *Server) addHTTPMiddleware(root *mux.Router, cfg *config.Config) {
 			Str("path", r.URL.String()).
 			Msg("http-request")
 	}))
-	root.Use(handlers.RecoveryHandler())
+	root.Use(gorillahandlers.RecoveryHandler())
```
desimone, line 41: Thought for FUP PR, we should probably just yank out the gorilla handlers depedency and carry the handler directly since we are using so few of them (I think this is the only one?). I think the project is deprecated too.

desimone (commented):
Code LGTM, with a few small suggestions, and comments.

proxy/proxy.go
calebdoxsey: It references itself (`p.getWebauthnState`), so it can't be done in the struct assignment

PR #3762, opened 2022-11-23
hpke: add HPKE key to JWKS endpoint

calebdoxsey:
## Summary
Add the HPKE public key to the JWKS endpoint and add code to fetch the HPKE public key from a JWKS endpoint. The format of an X25519 key is defined in [RFC8037](https://www.rfc-editor.org/rfc/rfc8037).

## Related issues
- https://github.com/pomerium/internal/issues/1015

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -30,6 +30,7 @@ import (
 	"github.com/pomerium/pomerium/internal/urlutil"
 	"github.com/pomerium/pomerium/pkg/cryptutil"
 	"github.com/pomerium/pomerium/pkg/grpc/config"
+	"github.com/pomerium/pomerium/pkg/hpke"
 )
 
 // DisableHeaderKey is the key used to check whether to disable setting header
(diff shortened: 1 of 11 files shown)


config/options.go
```diff
@@ -997,6 +998,16 @@ func (o *Options) GetSharedKey() ([]byte, error) {
 	return base64.StdEncoding.DecodeString(sharedKey)
 }
 
+// GetHPKEPrivateKey gets the hpke.PrivateKey dervived from the shared key.
+func (o *Options) GetHPKEPrivateKey() (hpke.PrivateKey, error) {
+	sharedKey, err := o.GetSharedKey()
+	if err != nil {
+		return hpke.PrivateKey{}, err
```
wasaga, line 1005: any specific reason for this not to be a pointer?

pkg/hpke/jwks.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return PublicKey{}, fmt.Errorf("hpke: error unmarshaling jwks endpoint: %w", err)
+	}
+
+	for _, key := range jwks.Keys {
+		if key.ID == jwkID && key.Type == jwkType && key.Curve == jwkCurve {
+			return PublicKeyFromString(key.X)
+		}
+	}
+	return PublicKey{}, fmt.Errorf("hpke key not found in JWKS endpoint")
```
wasaga, line 66: same question about pointers. everywhere else we use pointers

wasaga (approved):

config/options.go
calebdoxsey: I don't have a reason. I can change it.

PR #3774, opened 2022-11-29
autocert: support storing certificates in the databroker

calebdoxsey:
## Summary
Add support for using the databroker as an autocert backend.

## Related issues
Fixes https://github.com/pomerium/internal/issues/1058
 

## Checklist

- [x] updated unit tests 
- [ ] ready for review

### DIFF
--- a/internal/autocert/locker.go
+++ b/internal/autocert/locker.go
@@ -1,0 +1,137 @@
+package autocert
+
+import (
+	"context"
+	"time"
+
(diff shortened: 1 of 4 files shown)


calebdoxsey:
Currently we use a layered config that means we can't start the databroker without first starting the autocert manager. Because of this we can't use the databroker as the backend for autocert data. I'm not sure how to make this work.

PR #3777, opened 2022-11-30
identity: add identity profile

calebdoxsey:
## Summary
Add an `identity.Profile` type that contains information about an end-user after authentication.

## Related issues

## Checklist
- [x] updated unit tests
- [x] ready for review


pkg/grpc/identity/identity.proto
```diff
@@ -13,3 +15,10 @@ message Provider {
   string url = 7;
   map<string, string> request_params = 8;
 }
+
+message Profile {
+  string provider_id = 1;
+  bytes id_token = 2;
+  bytes oauth_token = 3;
+  google.protobuf.Struct claims = 4;
```
wasaga, line 23: what are claims in this context?

wasaga (approved):

pkg/grpc/identity/identity.proto
calebdoxsey: Claims are the claims from the token amended with data provided by the identity provider. Identity claim data is untyped and relies on a `.Claims` method on the token to be populated. Because of identity provider idiosyncrasies, we have code that hooks into these methods to add claims data as if it were in the token even though it's not.

For example github user info comes from API endpoints and then turned into claims:

```go
	var out struct {
		Email    string `json:"email"`
		Verified bool   `json:"email_verified"`
	}
	log.Debug(ctx).Interface("emails", response).Msg("github: user emails")
	for _, email := range response {
		if email.Primary && email.Verified {
			out.Email = email.Email
			out.Verified = true
			break
		}
	}
	b, err := json.Marshal(out)
	if err != nil {
		return err
	}
	return json.Unmarshal(b, v)
```

Unfortunately I don't know all the ways this is done or where its done. As another example, we actually modify the HTTP response from OIDC providers:

```go
	var userInfo map[string]interface{}
	if err := json.Unmarshal(bs, &userInfo); err == nil {
		// AWS Cognito returns email_verified as a string, so we'll make it a bool
		if ev, ok := userInfo["email_verified"]; ok {
			userInfo["email_verified"], _ = strconv.ParseBool(fmt.Sprint(ev))
		}

		// Some providers (ping) have a "mail" claim instead of "email"
		email, mail := userInfo["email"], userInfo["mail"]
		if email == nil && mail != nil && strings.Contains(fmt.Sprint(mail), "@") {
			userInfo["email"] = mail
		}

		bs, _ = json.Marshal(userInfo)
	}
```

This makes it look like they gave us back claims in the place we expect them, but they don't actually do that.

PR #3779, opened 2022-12-02
authenticate: implement hpke-based login flow

calebdoxsey:
## Summary
Rewrite the login flow so that it uses HPKE instead of symmetric encryption via the shared key. 

## Related issues

## Checklist
- [x] updated unit tests
- [x] ready for review


calebdoxsey:
This breaks programmatic login which is initiated by the proxy service and not the authorize service like normal login. I suspect this is already broken because it doesn't account for the identity provider id like it should.

calebdoxsey:
Fixed programmatic login

wasaga (approved):

Issue #3763, opened 2022-11-23
Unable to revoke Okta session


okgolove:
## What happened?
Okta signout does not work if I log out from Pomerium

## What did you expect to happen?
As soon as I logout on Pomerium page Okta session is stopped as well
## How'd it happen?

1. Authenticate in Pomerium (https://authenticate.example.com/.pomerium)
2. Logout on Pomerium page
3. Reopen https://authenticate.example.com/.pomerium
4. Okta page with login form should be opened

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`):
v0.20.0
- Server Operating System/Architecture/Cloud:
Google Kubernetes Engine

## What's your config.yaml?

```config.yaml
autocert: false
dns_lookup_family: V4_ONLY
address: :443
grpc_address: :443
certificate_authority_file: "/pomerium/ca/ca.crt"
certificates:
authenticate_service_url: https://authenticate.example.com/.pomerium.
authorize_service_url: https://pomerium-authorize.pomerium.svc.cluster.local
databroker_service_url: https://pomerium-databroker.pomerium.svc.cluster.local
idp_provider: okta
idp_scopes:
idp_provider_url: https://dev-1234567.okta.com
forward_auth_url: https://pomerium-proxy.pomerium
idp_client_id: <secret>
idp_client_secret: <secret>
databroker_storage_tls_skip_verify: false
routes:
  replaced
```

## What did you see in the logs?

```logs
{
    "level": "warn",
    "X-Forwarded-For": [
        "x.x.x.x"
    ],
    "X-Forwarded-Host": [
        "authenticate.example.com"
    ],
    "X-Forwarded-Port": [
        "443"
    ],
    "X-Forwarded-Proto": [
        "https"
    ],
    "X-Forwarded-Server": [
        "traefik-xxxxx"
    ],
    "X-Real-Ip": [
        "x.x.x.x"
    ],
    "ip": "127.0.0.1",
    "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/107.0.0.0 Safari/537.36",
    "referer": "https://authenticate.example.com/.pomerium/sign_out",
    "request-id": "7b3a12da-68bf-453e-80c7-6a13ce9983b0",
    "error": "identity/oidc: revoke not implemented",
    "time": "2022-11-23T10:40:54Z",
    "message": "authenticate: failed to revoke access token"
[...]

PR #3810, opened 2022-12-13
oidc: fix token revocation

calebdoxsey:
## Summary
Fix token revocation. We need to initialize the provider before checking the revocation URL.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/3763

## Checklist
- [x] updated unit tests
- [x] ready for review


internal/identity/oidc/oidc.go
```diff
@@ -176,18 +176,18 @@ func (p *Provider) UpdateUserInfo(ctx context.Context, t *oauth2.Token, v interf
 // Group membership is also refreshed.
```
wasaga, line 176: probably no longer true?

wasaga (approved):

Issue #3806, opened 2022-12-12
panic: runtime error: invalid memory address in getChallengeInfo


drewp:
## What happened?

I can get this:

```
2022/12/12 22:13:32 http: panic serving 10.42.1.0:31159: runtime error: invalid memory address or nil pointer dereference

goroutine 1855 [running]:

net/http.(*conn).serve.func1()

	/usr/local/go/src/net/http/server.go:1850 +0xbf

panic({0x24854c0, 0x82b6770})

	/usr/local/go/src/runtime/panic.go:890 +0x262

github.com/caddyserver/certmagic.(*Config).getChallengeInfo(0x0, {0x303b420, 0xc00148d5c0}, {0xc00165e1c1, 0x1c})

	/go/pkg/mod/github.com/caddyserver/certmagic@v0.17.2/config.go:970 +0x155

github.com/caddyserver/certmagic.(*ACMEIssuer).distributedHTTPChallengeSolver(0xc000de85a0, {0x3038900, 0xc000a7a460}, 0xc002617900)

	/go/pkg/mod/github.com/caddyserver/certmagic@v0.17.2/httphandler.go:74 +0x136

github.com/caddyserver/certmagic.(*ACMEIssuer).HandleHTTPChallenge(0xc000de85a0?, {0x3038900?, 0xc000a7a460?}, 0xc002617900?)

	/go/pkg/mod/github.com/caddyserver/certmagic@v0.17.2/httphandler.go:62 +0xb2

github.com/pomerium/pomerium/internal/autocert.(*Manager).handleHTTPChallenge(0xffffffffffffffff?, {0x3038900, 0xc000a7a460}, 0x0?)

	/go/pkg/mod/github.com/pomerium/pomerium@v0.20.1-0.20221205161217-df508f6bcc74/internal/autocert/manager.go:360 +0x4d

github.com/pomerium/pomerium/internal/autocert.(*Manager).updateServer.func1({0x3038900, 0xc000a7a460}, 0x7fe751d6ff38?)

	/go/pkg/mod/github.com/pomerium/pomerium@v0.20.1-0.20221205161217-df508f6bcc74/internal/autocert/manager.go:315 +0x4a

net/http.HandlerFunc.ServeHTTP(0x0?, {0x3038900?, 0xc000a7a460?}, 0x46c42e?)

	/usr/local/go/src/net/http/server.go:2109 +0x2f

[...]

PR #3816, opened 2022-12-16
autocert: use atomic pointer to allow nil

calebdoxsey:
## Summary
We currently use the `atomicutil.Value` to store the acme issuer with a zero-value acme issuer. This acme issuer will panic if accessed. Instead we should store nil, which exits early:

```
func (am *ACMEIssuer) HandleHTTPChallenge(w http.ResponseWriter, r *http.Request) bool {
	if am == nil {
		return false
	}
```

However `atomic.Value` doesn't support a nil value, so we can use `atomic.Pointer` which is a generic type that supports nil pointers.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/3806

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/autocert/manager.go
+++ b/internal/autocert/manager.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/caddyserver/certmagic"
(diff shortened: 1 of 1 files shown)


calebdoxsey:
The bug can be reproduced by enabling the http redirect server:

```
   "http_redirect_addr": ":80",
```

And curling the http server:

```
curl -v -k http://verify.example.com/.well-known/acme-challenge
*   Trying 127.0.0.1:80...
* Connected to verify.example.com (127.0.0.1) port 80 (#0)
> GET /.well-known/acme-challenge HTTP/1.1
> Host: verify.example.com
> User-Agent: curl/7.84.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 503 Service Unavailable
< content-length: 95
< content-type: text/plain
< date: Fri, 16 Dec 2022 18:35:42 GMT
< server: envoy
<
* Connection #0 to host verify.example.com left intact
upstream connect error or disconnect/reset before headers. reset reason: connection termination
```

wasaga (approved):

PR #3830, opened 2022-12-24
use tlsClientConfig instead of custom dialer

wasaga:
## Summary

When using custom CA, either supplied via `certificate_authority`/`certificate_authority_file`, or a custom certificate added to a system store, i.e. `mkcert --install`, `authz_check` and `proxy` were not able to communicate to the authenticate endpoint in order to fetch the signing public key. 

Apparently, installing a custom `TLSDialer` is not sufficient, and all requests to the authenticate endpoint would end up with 
EOF. 

This behaviour was observed for both Linux and MacOS using Go 1.19. 

This PR changes the way JWKSKeyFetcher operates, and adds an http Transport option to it, that is customized using current Config.

## Related issues

Related https://github.com/pomerium/internal/issues/1062

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,7 +7,8 @@
       "request": "launch",
       "mode": "debug",
       "program": "${workspaceRoot}/cmd/pomerium",
-      "args": ["-config", "${workspaceRoot}/.config.yaml"]
+      "args": ["-config", "${workspaceRoot}/.config.yaml"],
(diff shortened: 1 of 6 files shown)


calebdoxsey (approved):

Theghost256 (commented):
Thank you

PR #3858, opened 2023-01-06
hpke: fix hpke key fetcher when using self-signed certificates

calebdoxsey:
## Summary
When certificates are generated by pomerium the JWKS endpoint currently can't be fetched, resulting in 500s when trying to go through the authorization flow. 

## Related issues
Fixes https://github.com/pomerium/internal/issues/1062

## Checklist
- [ ] updated unit tests
- [x] ready for review


calebdoxsey:
We don't want to do this.

PR #3887, opened 2023-01-13
fix: concurrent map read and map writes

cfanbo:
## Summary
concurrent map read and map writes

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [X] add appropriate tag (`improvement` / `bug` / etc)
- [ ] ready for review

### DIFF
--- a/internal/sets/hash.go
+++ b/internal/sets/hash.go
@@ -1,8 +1,11 @@
 package sets
 
+import "sync"
+
 // A Hash is a set implemented via a map.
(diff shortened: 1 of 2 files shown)


calebdoxsey:
@cfanbo I believe with this container we had intended it to be locked higher up (treating it similar to a built in map, which is **not** safe for concurrent reads and writes). Is there a place where we are misusing this container?

PR #3952, opened 2023-02-09
envoy: optimize listener

wasaga:
## Summary

Previously, Envoy listeners were composed per loaded certificate, and contained all virtual hosts for all the routes. That 
was necessary in order to prevent issues with HTTP/2 connection coalescing, when browsers actively reused already opened connections, meaning 
a request for route `b.example.com` may arrive over the HTTP/2 connection originally opened for `a.example.com`. 

Since Envoy 1.25 it is now possible to supply multiple TLS certificates for a listener, and this PR uses to construct just 
one listeners with all known certificates, and register all virtual hosts representing Pomerium Routes in it. 

## Related issues

Fixes https://github.com/pomerium/internal/issues/1252

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


config/envoyconfig/listeners.go
```diff
@@ -518,65 +523,46 @@ func (b *Builder) buildRouteConfiguration(name string, virtualHosts []*envoy_con
...
-			AlpnProtocols:         alpnProtocols,
-			ValidationContextType: b.buildDownstreamValidationContext(ctx, cfg, serverName),
-		},
+		return []string{"h2", "http/1.1"}
 	}
 }
 
-func (b *Builder) buildDownstreamValidationContext(ctx context.Context,
+func (b *Builder) buildDownstreamValidationContext(
+	ctx context.Context,
```
calebdoxsey, line 563: do the changes to this method mean that if any route uses client certs, all routes will use client certs?

config/envoyconfig/listeners.go
```diff
@@ -349,7 +365,7 @@ func (b *Builder) buildMainHTTPConnectionManagerFilter(
 		UseRemoteAddress:  &wrappers.BoolValue{Value: true},
 		SkipXffAppend:     options.SkipXffAppend,
 		XffNumTrustedHops: options.XffNumTrustedHops,
-		LocalReplyConfig:  b.buildLocalReplyConfig(options, requireStrictTransportSecurity),
+		LocalReplyConfig:  b.buildLocalReplyConfig(options, true),
```
calebdoxsey, line 368: this appears to turn on HSTS headers for routes with self-signed certificates

calebdoxsey (approved):

config/envoyconfig/listeners.go
wasaga: right, we would not set the HSTS part of `localReply` then, as it's applied on the virtual host level.

```json
{
  "localReplyConfig": {
    "mappers": [
      {
        "filter": {
          "responseFlagFilter": {}
        },
        "headersToAdd": [
          {
            "header": {
              "key": "X-Frame-Options",
              "value": "SAMEORIGIN"
            },
            "appendAction": "OVERWRITE_IF_EXISTS_OR_ADD"
          },
          {
            "header": {
              "key": "X-XSS-Protection",
              "value": "1; mode=block"
            },
            "appendAction": "OVERWRITE_IF_EXISTS_OR_ADD"
          }
        ]
      }
    ]
  }
}
```

config/envoyconfig/listeners.go
wasaga: I think so yes, if there’s a client cert then it would be provided for any (reused) HTTP/2 connection. however the check of certs happens in authorize, so I suppose they’ll just be ignored for routes that do not require client cert?

calebdoxsey: I believe this would be a change in behavior but it would require testing.

Issue #3973, opened 2023-02-14
apple: tokens expire after 10 minutes


calebdoxsey:
## What happened?

After 10 minutes a user is logged out when using the Apple identity provider.

This is likely due to this error:

```json
{
  "level":"error",
  "config_file_source":"/home/caleb/src/github.com/calebdoxsey/pomerium-notes/2023/02/13/config.json",
  "bootstrap":true,
  "service":"identity_manager",
  "error":"identity/oidc: missing id_token",
  "user_id":"001468.7b378a7e37e54e2c885c62c77cb79ed8.0030",
  "session_id":"e549269c-01cb-493d-b07a-2d65d1cda4d3",
  "time":"2023-02-13T18:01:54-07:00",
  "message":"failed to update user info, deleting session"
}
```

## What did you expect to happen?

The user to remain logged in until their cookie expires.

## How'd it happen?

Use the Apple IdP

PR #3974, opened 2023-02-14
apple: fix userinfo

calebdoxsey:
## Summary
Fix the `UpdateUserInfo` call so that refreshing sessions doesn't kill the session anymore.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/3973

## Checklist
- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/identity/oauth/apple/apple.go
+++ b/internal/identity/oauth/apple/apple.go
@@ -5,8 +5,6 @@ package apple
 
 import (
 	"context"
-	"encoding/base64"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


wasaga (approved):

mnestor:
@calebdoxsey is this fixing the 10 minute timeout? I've been struggling all day to fix it and think I was just about to have it when I saw this. Darn!

calebdoxsey:
@mnestor Yes I believe it should fix that issue. I think Apple doesn't return extra info from the refresh token so there was no `id_token` to parse and use for user data.

mnestor:
Yeah, they also only give you 1 chance to get a name ever, the very first time they authenticate with your app. Which also means the who + to %20 can go away since we only need email there.

I wasn't sure if the id_token info was still needed so I had just about wrapped up pulling that info by doing a auth/token call again with the refresh_token. But given it should really only be used for the short time anyway I was way over thinking that.

Appreciate the response/acceptance/learning this have given me, and merging!

PR #3978, opened 2023-02-15
authenticate: save the session cookie with a different name

calebdoxsey:
## Summary
Save the session cookie with a different name so that it doesn't collide with the proxy session cookie.

## Related issues
- https://github.com/pomerium/internal/issues/1278
 

## Checklist
- [ ] updated unit tests
- [x] ready for review


desimone (approved):
What's old is new again. We used to do this way back in <v0.4 when we stored all tokens in the authN encrypted session cookie.

PR #3990, opened 2023-02-17
authenticate: jwks: use local bypass to fetch keys

wasaga:
## Summary

Both `authorize` and `proxy` services need access the `/.well-known/pomerium/jwks.json` endpoint of `authenticate`. 
When running all-in-one mode inside a non-root container with port mapping, it is not possible to reach that url. 

This PR changes to use a local control plane server to access it in that case. 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/state.go
+++ b/authorize/state.go
@@ -3,14 +3,12 @@ package authorize
 import (
 	"context"
 	"fmt"
-	"net/url"
 
 	googlegrpc "google.golang.org/grpc"
 
 	"github.com/pomerium/pomerium/authorize/evaluator"
 	"github.com/pomerium/pomerium/authorize/internal/store"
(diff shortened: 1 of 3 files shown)


desimone:
These changes assume that folks running in all in one mode will _always_ want to use their local instance of authenticate which would be incompatible with using the hosted authenticate service (AIO+hosted authN will result in a localhost call), right? .

calebdoxsey (changes requested):
Tests are failing.

As Bobby said, this solution prevents a shared authenticate service. This is a serious reduction in functionality, but I guess if we want to toss that capability, so be it 🤷 .

Issue #3976, opened 2023-02-15
Strange behavior with setting an identity provider client id per route 


cr1xu5:
## What happened?

It seems the `idp_client_id` per route is not used when set. Only the global `idp_client_id` is sent to the OIDC provider.

## What did you expect to happen?

I would expect the `idp_client_id` for a route would overwrite the global client id as stated on https://www.pomerium.com/docs/reference/routes/identity-provider-client-id-per-route

## How'd it happen?

See config.yaml

If I define the `idp_client_id` and `idp_client_secret` for this route globally, everything works.  However, I would like to configure multiple routes and a different client id for each route. The global client id is not needed in my use case.

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`):
- Server Operating System/Architecture/Cloud:

```
pomerium: 0.21.0-1675979974+3eaa6094
envoy: 1.24.0+05e1343657086268edaa16a1e8a1429d9c8f6e50a44f492cf629fdc71da2ecdc
```

Deployed by using the Debian repository.

## What's your config.yaml?

```config.yaml
# identity provider settings: https://www.pomerium.com/docs/identity-providers.html
idp_provider: oidc
idp_provider_url: https://zitadel.domain.com
idp_client_id: random@notinuse
idp_client_secret: notinuse

# https://pomerium.io/reference/#routes
routes:
  - from: https://acceptance.domain.com
    to: http://127.0.0.1:3009
    idp_client_id: ***
    idp_client_secret: ***
    pass_identity_headers: true
    allow_any_authenticated_user: true
```

## What did you see in the logs?

```logs
Feb 15 14:12:57 server1 pomerium[21471]: {"level":"error","ip":"127.0.0.1","user_agent":"Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/109.0","request-id":"***","error":"identity provider: server
_error","status":400,"status-text":"Bad Request","request-id":"***","time":"2023-02-15T14:12:57+01:00","message":"httputil: error"}
```
In the Zitadel access logs:
```
..."method":"GET","host":"zitadel.domain.com","uri":"/oa
uth/v2/authorize?client_id=random%40notinuse&redirect_uri=https%3A%2F%2Fpomerium.domain.com%2Foauth2%2Fcallback&response_type=code&scope=openid+profile+email+offline_access&state=SmFYW...
```

PR #4006, opened 2023-02-22
authenticate: fix identity provider id in encrypted query string

calebdoxsey:
## Summary
Per-route IDP credentials use a `pomerium_idp_id` query string parameter to indicate which credentials they need. The current code is not working because the `pomerium_idp_id` is inside an HPKE encrypted query string. This updates the code to decrypted the URL parameters to find it.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/3976

## Checklist
- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/authenticate/handlers.go
+++ b/authenticate/handlers.go
@@ -128,7 +128,7 @@ func (a *Authenticate) VerifySession(next http.Handler) http.Handler {
 		defer span.End()
 
 		state := a.state.Load()
-		idpID := r.FormValue(urlutil.QueryIdentityProviderID)
+		idpID := a.getIdentityProviderIDForRequest(r)
 
 		sessionState, err := a.getSessionFromCtx(ctx)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


authenticate/handlers.go
```diff
@@ -588,3 +588,24 @@ func (a *Authenticate) saveCallbackSession(w http.ResponseWriter, r *http.Reques
...
+	state := a.state.Load()
+	idpID := ""
+	if _, requestParams, err := hpke.DecryptURLValues(state.hpkePrivateKey, vs); err == nil {
+		if idpID == "" {
+			idpID = requestParams.Get(urlutil.QueryIdentityProviderID)
+		}
+	}
+	if idpID == "" {
+		idpID = vs.Get(urlutil.QueryIdentityProviderID)
+	}
```
wasaga, line 609: why this clause is still required if parameter is inside hpke query?

wasaga (approved):

authenticate/handlers.go
calebdoxsey: I think we pass the identity provider id from userinfo endpoints? I'm not totally sure. I kept it because I was afraid I'd break something by removing it.

Issue #4013, opened 2023-02-24
IC: x-frame-options header always returns the default value


ssveta7ak:
## What happened?

**Pomerium Ingress Controller**: 
It is impossible to override the `x-frame-options` header - it will still return the default value of SAMEORIGIN."
The issue is only related to `x-frame-options` header; other headers work properly.

## What's your config?

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    ingress.pomerium.io/set_response_headers: '{"x-frame-options":"DENY", "hello":"there"}'
``` 

## What did you see in the logs?

```logs
# Paste your logs here.
# Be sure to scrub any sensitive values
```

## Additional context

Related https://github.com/pomerium/pomerium/issues/4014

calebdoxsey:
to disable set response headers you have to use:

```
{
  "disable": true
}
```

https://www.pomerium.com/docs/reference/set-response-headers

wasaga:
> to disable set response headers you have to use:

this ticket is different - currently user cannot _override_ the default value to a different one.

desimone:
Unclear what is left here. @ssveta7ak  please reopen if I missed something. Looks complete.

PR #4026, opened 2023-02-27
config: fix set_response_headers

calebdoxsey:
## Summary
Fix the set response headers option so that its set with the correct precedence and `disable: 1` works for the route-specific option.

## Related issues
- https://github.com/pomerium/pomerium/issues/4013

 

## Checklist
- [x] updated unit tests
- [x] ready for review


calebdoxsey:
Per the docs:

> Specifies a set of headers that will be added to responses to requests matching this route. Headers specified at this level are applied before headers from the enclosing :ref:`envoy_v3_api_msg_config.route.v3.VirtualHost` and :ref:`envoy_v3_api_msg_config.route.v3.RouteConfiguration`.

So we can't use the virtual host `ResponseHeadersToAdd` option since we want the route-specific ones to be applied last.

wasaga (commented):
Is it expected that the following config results in no headers set in the response? 

```yaml
set_response_headers:
  "disable": "true"

routes:
    to: https://httpbin.org
    pass_identity_headers: true
    allowed_domains: ["gmail.com"]
    set_response_headers:
      "my-custom-header": "hello there"
```

calebdoxsey:
I don't know how it should behave.

desimone:
I want to find out why this functionality was added before making a decision.

calebdoxsey:
I don't know why this functionality was added.

desimone (approved):

wasaga (approved):

PR #4098, opened 2023-04-04
support loading route configuration via rds

calebdoxsey:
## Summary
Instead of defining the route configuration in the listener, define it as a separate resource and load it via RDS.

## Related issues
- https://github.com/pomerium/internal/issues/892

## Checklist
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
These changes are causing the ingress controller to fail with

```
type.googleapis.com/envoy.config.route.v3.RouteConfiguration: node 'id' and 'cluster' are required. Set it either in 'node' config or via --service-node and --service-cluster options.
```

calebdoxsey:
This requires changes to the ingress controller which manually validates configuration: https://github.com/pomerium/ingress-controller/blob/87d439530f778d485873c74e1e5d582d767277f8/pomerium/validate.go#L44

calebdoxsey:
The console has the same issue.

calebdoxsey:
There is no static resource for a route configuration. They must be included in a listener. These changes are incompatible with the validation logic in the enterprise console and the ingress controller.

calebdoxsey:
The code has been updated to support a fully static configuration option. We can update the code in the ingress controller and console to use this option.

PR #4131, opened 2023-04-24
config: add support for wildcard from addresses

calebdoxsey:
## Summary
Adds support for from addresses like `https://*.example.com`.

## Related issues
- https://github.com/pomerium/internal/issues/1329

## User Explanation
Adds support for `*` in from addresses that will match 0 or more characters. This can occur anywhere inside the host part of the URL.

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/authorize/evaluator/evaluator.go
+++ b/authorize/evaluator/evaluator.go
@@ -38,6 +38,7 @@ type Request struct {
 // RequestHTTP is the HTTP field in the request.
 type RequestHTTP struct {
 	Method            string            `json:"method"`
+	Hostname          string            `json:"hostname"`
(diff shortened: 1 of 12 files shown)


wasaga (approved):

wasaga:
the [`regex`](https://www.pomerium.com/docs/reference/routes/regex) path option won't work with wildcard `from`s, correct?

calebdoxsey:
The `regex` option will stork with wildcard `from`s. The route match requires all conditions to be met. So the `:authority` header will be matched against the wildcard `from` and the path will be matched against `regex`.

PR #4148, opened 2023-05-02
config: add cookie_same_site option

calebdoxsey:
## Summary
Add a new `cookie_same_site` option.

## Related issues
Fixes https://github.com/pomerium/internal/issues/1361
 
## Checklist
- [x] updated unit tests
- [x] ready for review


authenticate/state.go
```diff
@@ -101,6 +101,7 @@ func newAuthenticateStateFromConfig(cfg *config.Config) (*authenticateState, err
 			Secure:   cfg.Options.CookieSecure,
 			HTTPOnly: cfg.Options.CookieHTTPOnly,
 			Expire:   cfg.Options.CookieExpire,
+			SameSite: cfg.Options.GetCookieSameSite(),
```
wasaga, line 104: just wondering if there might be a situation when this setting may need be different for authenticate endpoint and proxy.

wasaga (approved):

authenticate/state.go
calebdoxsey: probably 🤷

PR #4186, opened 2023-05-23
improve certificate matching performance

calebdoxsey:
## Summary
Currently for every route we query for all the certificates, which results in a linear scan of all the certificates to see if anything matches `*` (which will never happen). This updates the code to only query for all the certificates once, and to remove the inner linear scan for matching `*` certificates to always generate a `*`.

## Related issues
- https://github.com/pomerium/internal/issues/1375

## Checklist
- [x] updated unit tests
- [x] ready for review


desimone (approved):

desimone:
Is this a backport candidate?

PR #4187, opened 2023-05-23
envoy: set re2 limits very high

calebdoxsey:
## Summary
Set the program size complexity for regular expressions in envoy to be very high.

## Related issues
Fixes https://github.com/pomerium/internal/issues/1373

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/envoyconfig/bootstrap_test.go
+++ b/config/envoyconfig/bootstrap_test.go
@@ -43,6 +43,10 @@ func TestBuilder_BuildBootstrapLayeredRuntime(t *testing.T) {
 			"staticLayer": {
 				"overload": {
 					"global_downstream_max_connections": 50000
+				},
+				"re2": {
+					"error_level": 1048576,
+					"warn_level": 1024
 				}
 			}
 		}] }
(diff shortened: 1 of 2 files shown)


config/envoyconfig/bootstrap_test.go
```diff
@@ -43,6 +43,10 @@ func TestBuilder_BuildBootstrapLayeredRuntime(t *testing.T) {
 			"staticLayer": {
 				"overload": {
 					"global_downstream_max_connections": 50000
+				},
+				"re2": {
+					"error_level": 1048576,
+					"warn_level": 1024
```
kenjenkins, line 49: These should be nested under the "max_program_size" key, right?

kenjenkins (approved):
I won't pretend to understand the `Runtime` configuration stuff yet, but this looks like the right idea to me.

config/envoyconfig/bootstrap_test.go
calebdoxsey: yes

PR #4193, opened 2023-05-23
config: separate wildcard hosts where possible

kenjenkins:
## Summary

Envoy supports virtual hosts with wildcard domains, as long as the wildcard token appears at the very beginning or the end of the domain (that is, Envoy supports suffix and prefix domain wildcards). Update the configuration builder logic to take advantage of this support whenever possible, using the current regex route matching as a fallback.

This may improve routing performance in some cases, by reducing the size of the routing table for the catch-all virtual host.

Add a utility method isSpecialWildcardHost() to determine whether a host has a wildcard token that is not at the very beginning or end. Update the existing route configuration test to add one of these special wildcard hosts.

## User Explanation

Improve routing performance when many routes have `*` at the beginning or end of the from address.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/listeners.go
+++ b/config/envoyconfig/listeners.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net"
 	"net/url"
-	"strings"
 	"time"
 
 	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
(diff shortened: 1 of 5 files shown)


calebdoxsey (approved):
This makes sense to me. I don't know if we use the routable hosts anywhere that wouldn't understand `*`, but if its just used for the virtual host I think it should work.

calebdoxsey (changes requested):
So re-reading the docs, there's a subtle difference here. For us `*x` would match just `x` whereas with virtual hosts `*x` means at-least-one character for `*`. `.*` vs `.+` in regex.

abl:
We do have this exact case - we want to match `anything.and.everything.yourmachine` as well as `yourmachine` but we don't want to match `notyourmachine`.

So what we really need is `/^(.+\.)?yourmachine$/`, which sort of makes sense, but does not neatly map in to how CNAMEs or wildcards in certs work.

PR #4196, opened 2023-05-25
config: simplify default set response headers

calebdoxsey:
## Summary
Currently we selectively remove the `Strict-Transport-Security` header for routes that use a generated certificate. This is done to help prevent the occurrence of an error page that requires users to manually type `thisisunsafe` to bypass.

We do this by searching the entire list of certificates for any that match the route. If there are none, we assume the certificate used will be a generated one and remove the `Strict-Transport-Security` header.

This PR updates the code to instead look for any certificates at all. If at least one certificate is supplied, we will assume the user would prefer to have this header added. If none are supplied, we will assume Pomerium is running in an entirely generated mode where the inclusion of this header is more annoying than helpful.

## Related issues
- https://github.com/pomerium/internal/issues/1375

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/envoyconfig/http_connection_manager.go
+++ b/config/envoyconfig/http_connection_manager.go
@@ -13,7 +13,6 @@ func (b *Builder) buildVirtualHost(
 	options *config.Options,
 	name string,
 	host string,
-	requireStrictTransportSecurity bool,
 ) (*envoy_config_route_v3.VirtualHost, error) {
 	vh := &envoy_config_route_v3.VirtualHost{
 		Name:    name,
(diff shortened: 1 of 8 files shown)


kenjenkins (approved):
What does "generated certificate" mean in this context? Is this the fallback self-signed certificate, or something else?

calebdoxsey:
We have a catch-all certificate: 


```
	return cryptutil.GenerateCertificate(sharedKey, "*")
```

desimone (approved):
Nice improvement to readability as well.

PR #4215, opened 2023-05-31
authorize: populate issuer even when policy is nil

kenjenkins:
Backport https://github.com/pomerium/pomerium/commit/6df4fba83236c8401fa487c0575453b0fd759d63 from https://github.com/pomerium/pomerium/pull/4211

This also backports the change from commit 18bc86d63284c64cd9c2d1e257b1c53198cb5d39 to add a `hostname` parameter to the `NewHeadersRequestFromPolicy()` method, in order to determine the issuer.


wasaga:
that's probably OK however i'm not sure if we would be cutting another v0.21 patch anytime soon?

wasaga (approved):

PR #4228, opened 2023-06-07
configure Envoy to validate client certificates

kenjenkins:
Currently, client certificate validation is performed within the authorize service, after user login. Instead, configure Envoy to perform certificate validation itself, at the time of the initial connection.

When a client certificate authority is configured, Envoy will reject any connection attempts that do not present a valid client certificate corresponding with a trust chain rooted at the configured certificate authority.

For end users without a client certificate configured in their browser, after this change they will see a browser default error page, rather than an HTML error page served by Pomerium.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/listeners.go
+++ b/config/envoyconfig/listeners.go
@@ -1,6 +1,7 @@
 package envoyconfig
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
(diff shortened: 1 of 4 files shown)


config/envoyconfig/listeners.go
```diff
@@ -668,42 +668,47 @@ func (b *Builder) buildDownstreamTLSContext(ctx context.Context,
...
-			if p.TLSDownstreamClientCA != "" {
-				needsClientCert = true
-				break
-			}
+// clientCAForDomain returns a bundle of all per-route client CAs configured
+// for the given domain, or else the globally configured client CA.
+func clientCAForDomain(cfg *config.Config, domain string) []byte {
+	var bundle []byte
+	for _, p := range getPoliciesForDomain(cfg.Options, domain) {
+		if p.TLSDownstreamClientCA == "" {
```
wasaga, line 690: there is also `TLSDownstreamClientCAFile` that need be accounted for.

config/envoyconfig/listeners.go
```diff
@@ -668,42 +668,47 @@ func (b *Builder) buildDownstreamTLSContext(ctx context.Context,
...
-		for _, p := range getPoliciesForDomain(cfg.Options, domain) {
-			if p.TLSDownstreamClientCA != "" {
-				needsClientCert = true
-				break
-			}
+// clientCAForDomain returns a bundle of all per-route client CAs configured
+// for the given domain, or else the globally configured client CA.
+func clientCAForDomain(cfg *config.Config, domain string) []byte {
+	var bundle []byte
+	for _, p := range getPoliciesForDomain(cfg.Options, domain) {
```
wasaga, line 689: if the user configured multiple routes per domain with different `tls_downstream_client_ca` options, this should return an invalid configuration error

config/envoyconfig/listeners.go
kenjenkins: At first I thought so as well, but then I noticed that `Policy.Validate()` will populate `TLSDownstreamClientCA` from `TLSDownstreamClientCAFile`: https://github.com/pomerium/pomerium/blob/0d34e3d00c16d0e951e6e18f1992531cc10c96c6/config/policy.go#L501

Is it safe to assume that `Validate()` will have been called on each `Policy` before we get to the main `BuildListeners()` method?

wasaga: indeed. in general, Validate() function shouldn't really have side effects, so I try to never rely on such magic behaviour. But we may do so here.

config/envoyconfig/listeners.go
wasaga: as discussed, making a bundle would work.

kenjenkins: So far in my manual testing this "bundle" approach appears to be working as intended:
 - I configure two routes for the same domain, each with a different CA
 - I get a browser prompt that lets me select certificates from either CA
 - selecting a certificate from the first route's CA lets me through to the login page
 - selecting a certificate from the second route's CA gives me a Pomerium 495 error page

kenjenkins: I've added an integration test case to exercise this behavior: [TestMultipleDownstreamClientCAs](https://github.com/pomerium/pomerium/pull/4228/files#diff-0eb2b3039ed5088bb153887ae325c6cf4c8ec35063b236524fd1247bc571c5a2R399).

desimone (approved):

integration/policy_test.go
```diff
@@ -320,3 +321,190 @@ func TestLoadBalancer(t *testing.T) {
...
+
+		// With cert1, we should get an HTTP error response for the /ca2 path.
+		req, err := http.NewRequestWithContext(ctx, "GET",
+			"https://client-cert-overlap.localhost.pomerium.io/ca2", nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		res, err = client.Do(req)
+		if !assert.NoError(t, err, "unexpected http error") {
```
wasaga, line 453: require.NoError

wasaga (approved):

integration/policy_test.go
kenjenkins: OK, I think I've switched over to `require.NoError` wherever appropriate.

PR #4229, opened 2023-06-08
ci: fix lint workflow

wasaga:
## Summary

Fixes failing `lint` workflow by pinning specific compatible versions.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -15,12 +15,14 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753
+      - uses: actions/setup-go@v4
         with:
           go-version: 1.20.x
-
+          cache: false
+  
       - run: make deps-build
 
-      - uses: golangci/golangci-lint-action@08e2f20817b15149a52b5b3ebe7de50aff2ba8c5
+      - uses: golangci/golangci-lint-action@v3.5.0
         with:
+          version: v1.52
           args: --timeout=10m
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -15,14 +15,14 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: actions/setup-go@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yaml
```diff
@@ -15,12 +15,14 @@ jobs:
...
-      - uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753
+      - uses: actions/setup-go@v4
         with:
           go-version: 1.20.x
-
+          cache: false
+  
       - run: make deps-build
 
-      - uses: golangci/golangci-lint-action@08e2f20817b15149a52b5b3ebe7de50aff2ba8c5
```
kenjenkins, line 24: I've seen a recommendation to always pin using commit hashes, because tags can be modified. (For example: https://www.nearform.com/blog/why-you-should-pin-your-github-actions-by-commit-hash/.)

Has there been any team discussion about whether it's better to pin to version tags or commit hashes? It would probably be good to standardize one way or the other.

kenjenkins (approved):

.github/workflows/lint.yaml
wasaga: I'm not exactly sure how dependabot bumps hashes. Is it just commits on mains of version is omitted? We probably would want a stable release instead, no?

Ie here the real dependency is golangci-lint version and not an action version, and dependabot won't recognize it I think. 

Also initially there was a bug with setup-go version which they seem to fix in the most recent release.

Let me try to revert back to hashes and see what happens.

wasaga:
Yes plus their release had an issue with setup-go v4

.github/workflows/lint.yaml
kenjenkins: I don't have a complete answer, but I think dependabot will wait for a new release. For example, here's an open dependabot PR from the datasource repository: https://github.com/pomerium/datasource/pull/193/files.

The PR name refers to the new version as 3.5.0 but the file update itself uses the commit hash.

But to your other point: yes, I don't think there's any way to get dependabot to keep the `version` parameter updated within the action configuration.

PR #4250, opened 2023-06-12
pin to a debian:latest image for casource base image

kenjenkins:
The debian 'stable' images configure apt to install from the rolling 'stable' repository, rather than a specific Debian release. Thus even though we pin to a specific Docker image digest, the packages installed by 'apt-get' can change when a new Debian release is promoted to stable.

Instead, pin to an image where apt is configured to install from repositories for a specific Debian release (in this case, bullseye).


desimone (approved):
Code & Changes LGTM but did not test.

PR #4251, opened 2023-06-13
fix: Remove default prompt

ma-ts:
## Summary

There's a difference in Azure if you remove the prompt parameter altogether (transparently redirects user in case user already has a valid session, but presents the user with a login screen in case the user doesn't). This is by default a good approach, also being used by default Azure apps, such as Azure AD or the Azure Portal. This change opens up the possibility for transparent login with Azure, but still allows for customisation of the prompt parameter.

## User Explanation

Implements the default Microsoft policy for when no prompt is explicitly given (so transparently redirects the user).

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


desimone:
Thanks for the pr @ma-ts ! There are some subtleties to how multi-account, and logout works with this code. We'll discuss and review shortly.

ma-ts:
> Thanks for the pr @ma-ts ! There are some subtleties to how multi-account, and logout works with this code. We'll discuss and review shortly.

@desimone: The thing is, it currently does not seem to be possible to not have a code (as an opt-in feature). Could I make a PR that we can overwrite the whole object (and that it then does not merge the _default_ options in there?) With that the default would not change, but the behaviour that we need is still there.

desimone:
@kenjenkins  -- makes sense to me.

Issue #4353, opened 2023-07-11
Feature request: stricter client certificate enforcement


kenjenkins:
Add a configuration option to enable stricter enforcement of downstream mTLS: when a connection is made without a valid client certificate, reject the connection attempt altogether, instead of serving a Pomerium error page.

Note that this would apply not just to routes defined in the Pomerium configuration but also to internal Pomerium routes (such as the HPKE public key endpoint and the JWKS endpoint).

~~Depends on #4352.~~

kenjenkins:
Note: the current plan is to add support for three new settings.
```yaml
downstream_mtls:
  enforcement: "reject_connection" | "policy_with_default_deny" | "policy"
  max_verify_depth: 1
  match_subject_alt_names:
    - dns: '.*\.corp\.example\.com'
```

Setting `enforcement: reject_connection` will trigger the new behavior of rejecting connection attempts during the initial TLS handshake. The default behavior will continue to be `enforcement: policy_with_default_deny`.

The `max_verify_depth` setting will behave just as the [Envoy setting](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto.html#envoy-v3-api-field-extensions-transport-sockets-tls-v3-certificatevalidationcontext-max-verify-depth) of the same name.

The `match_subject_alt_name` settings will allow for regex matching of certificate Subject Alternative Name entries.

kenjenkins:
This should be basically complete, so I think we can go ahead and close this issue.

Documentation for the new settings is available within the vNext documentation site, at https://main.docs.pomerium.com/docs/reference/downstream-mtls-settings.

PR #4365, opened 2023-07-15
envoy: separate gRPC listener configuration

### LINKED ISSUE #4363: `codec_type: http1` appears to be incompatible with split service mode
## What happened?

I edited [integration/clusters/multi/compose.yml](https://github.com/pomerium/pomerium/blob/main/integration/clusters/multi/compose.yml) to add `CODEC_TYPE: http1` to the `environment` section of each Pomerium service (pomerium-authenticate, pomerium-authorize, pomerium-databroker, pomerium-proxy) and then ran `docker-compose up -V`.

I expected Pomerium to run in split service mode even with `codec_type` set to `http1`.

## How'd it happen?

1. I edited [integration/clusters/multi/compose.yml](https://github.com/pomerium/pomerium/blob/main/integration/clusters/multi/compose.yml) to add `CODEC_TYPE: http1` to the `environment` section of each Pomerium service (pomerium-authenticate, pomerium-authorize, pomerium-databroker, pomerium-proxy).
2. I ran `docker-compose up -V`.
3. Running the integration tests appears to hang. Trying to access a configured route results in a browser default 500 error page.

 pomerium: v0.20.0-411-ga1388592+a1388592
envoy: 1.25.5+12c92a467a57b3d1d73da17128179abccb28f594834915eeb61abc62938622e2
- Server Operating System/Architecture/Cloud: Docker Desktop (macOS)

## What's your config.yaml?

n/a

Lots of errors like this (from multiple services):
```json
{
  "level": "error",
  "syncer_id": "databroker",
  "syncer_type": "type.googleapis.com/pomerium.config.Config",
  "error": "error during initial sync: error receiving record: rpc error: code = Unavailable desc = upstream connect error or disconnect/reset before headers. reset reason: protocol error",
  "time": "2023-07-13T21:31:32Z",
  "message": "sync"
}
```

## Additional context

This may be similar to #2844.

kenjenkins:
## Summary

Pomerium configures a gRPC listener in Envoy, for internal communication between the various Pomerium services. Currently this listener shares much of the same configuration as the main HTTP listener, based on the main Pomerium configuration options.

However, some configuration options don't make sense for the gRPC listener. Specifically, the `codec_type` option should not be applied to the gRPC listener, as gRPC requires HTTP/2. Also, any client certificate settings should not apply to the gRPC listener. (This is not currently a problem, but it will interfere with https://github.com/pomerium/pomerium/issues/4353.)

Separate the gRPC listener configuration from the main HTTP listener configuration, so we can avoid applying these configuration options. Instead set AlpnProtocols to just "h2" (HTTP/2), and do not set any ValidationContextType on the DownstreamTlsContext (no client certificate validation).

Specifically, inline the call to buildTLSSocket() within the body of buildGRPCListener(). Extract a new method envoyCertificates() from buildDownstreamTLSContextMulti(), to avoid repeating this logic.

## Related issues

Fixes #4363.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/listeners.go
+++ b/config/envoyconfig/listeners.go
@@ -381,35 +381,43 @@ func (b *Builder) buildGRPCListener(ctx context.Context, cfg *config.Config) (*e
 		return nil, err
 	}
 
+	filterChain := envoy_config_listener_v3.FilterChain{
+		Filters: []*envoy_config_listener_v3.Filter{filter},
+	}
+
 	li := newEnvoyListener("grpc-ingress")
+	li.FilterChains = []*envoy_config_listener_v3.FilterChain{&filterChain}
+
 	if cfg.Options.GetGRPCInsecure() {
 		li.Address = buildAddress(cfg.Options.GetGRPCAddr(), 80)
-		li.FilterChains = []*envoy_config_listener_v3.FilterChain{{
-			Filters: []*envoy_config_listener_v3.Filter{
-				filter,
-			},
-		}}
-	} else {
-		li.Address = buildAddress(cfg.Options.GetGRPCAddr(), 443)
-		li.ListenerFilters = []*envoy_config_listener_v3.ListenerFilter{
-			TLSInspectorFilter(),
-		}
+		return li, nil
+	}
 
-		allCertificates, err := getAllCertificates(cfg)
-		if err != nil {
-			return nil, err
-		}
+	li.Address = buildAddress(cfg.Options.GetGRPCAddr(), 443)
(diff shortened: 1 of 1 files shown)


calebdoxsey:
I think this makes sense but I'm not super clear if there was a reason why it was setup the way it was. (Possibly the listener could be on the same port as HTTP?)

calebdoxsey (approved):

PR #4367, opened 2023-07-17
config: validate log levels

calebdoxsey:
## Summary
Validate the `log_level` and `proxy_log_level` options to prevent envoy crashing on invalid values.

The supported log levels are the superset of all the know log levels from our documentation, envoy and zerolog, because this option was previously unconstrained, so this should not be a breaking change.

## Related issues
- https://github.com/pomerium/internal/issues/1452

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/log.go
+++ b/config/log.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"sync"
 
-	"github.com/rs/zerolog"
-
 	"github.com/pomerium/pomerium/internal/log"
 )
 
@@ -43,7 +41,6 @@ func (mgr *LogManager) OnConfigChange(_ context.Context, cfg *Config) {
 	}
 
 	if cfg.Options.LogLevel != "" {
-		log.SetLevel(cfg.Options.LogLevel.ToZerolog().String())
-		zerolog.SetGlobalLevel(cfg.Options.LogLevel.ToZerolog())
+		log.SetLevel(cfg.Options.LogLevel.ToZerolog())
 	}
 }
--- a/config/log.go
+++ b/config/log.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"sync"
 
+	"github.com/rs/zerolog"
+
 	"github.com/pomerium/pomerium/internal/log"
 )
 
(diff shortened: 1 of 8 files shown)


config/log.go
```diff
@@ -41,6 +43,7 @@ func (mgr *LogManager) OnConfigChange(_ context.Context, cfg *Config) {
 	}
 
 	if cfg.Options.LogLevel != "" {
-		log.SetLevel(cfg.Options.LogLevel)
+		log.SetLevel(cfg.Options.LogLevel.ToZerolog().String())
+		zerolog.SetGlobalLevel(cfg.Options.LogLevel.ToZerolog())
```
kenjenkins, line 47: I think this is the only place that calls internal/log `SetLevel()`? Should we go ahead and update `SetLevel()` to take a `zerolog.Level` directly, and then we can delete the mapping from string back to level there? https://github.com/pomerium/pomerium/blob/a1388592d8627368ef9421c98d43d500e20beb37/internal/log/log.go#L69-L84

(Please feel free to disagree, I'd value your opinion on this.)

config/log_level.go
```diff
@@ -0,0 +1,109 @@
...
+// ValidateLogLevel validates that a log level is one of the known log levels.
+func ValidateLogLevel(lvl LogLevel) error {
+	_, ok := logLevelLookup[lvl]
+	if !ok {
+		return fmt.Errorf("unknown log level: %s", lvl)
+	}
+	return nil
+}
+
+// ToZerolog converts the log level to a level zerolog expects
```
kenjenkins, line 63: It might be nice to document here that "unset" will default to "info".

(totally optional)

config/log_level.go
```diff
@@ -0,0 +1,109 @@
...
+	case LogLevelTrace:
+		return "trace"
+	case LogLevelDebug:
+		return "debug"
+	case LogLevelInfo, LogLevelUnset:
+		return "info"
+	case LogLevelWarn:
+		return "warn"
+	case LogLevelWarning:
+		return "warning"
```
kenjenkins, line 99: It looks like warn/warning are synonyms for Envoy (at least, they're listed together here: https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/run-envoy#debugging-envoy). Should we just pick one here, like how you have these two grouped together in `ToZerolog()` above?

(totally optional)

kenjenkins (approved):
Nice, I like the idea of introducing an explicit type for the log level.

PR #4371, opened 2023-07-19
autocert: suppress OCSP stapling errors

### LINKED ISSUE #4245: autocert / OCSP stapling errors are shown when autocert is not enabled
## Got

OCSP stapling errors are shown even when autocert is not enabled.  My understanding is our autocert and ocsp code is deeply coupled. This often creates user confusion when our users aren't actually using autocert

## Want

Ideally, we should de-couple OCSP functionality from our autocert stack.

Less ideally, we should not present confusing error messages to our users. 

### Related

- https://github.com/pomerium/pomerium/issues/193
- https://github.com/pomerium/pomerium/issues/4133
- https://github.com/pomerium/pomerium/issues/1946
_Originally posted by @wasaga in https://github.com/pomerium/ingress-controller/issues/658#issuecomment-1585319120_

calebdoxsey:
## Summary
Suppress "no OCSP server specified in certificate" errors. These errors are generated in the certmagic package and logged with a zap logger, so we need to implement a custom zap core to remove them.

## Related issues
Fixes #4245 
 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/autocert/certmagic_logger.go
+++ b/internal/autocert/certmagic_logger.go
@@ -1,0 +1,50 @@
+package autocert
+
+import (
+	"strings"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
+	"github.com/pomerium/pomerium/internal/log"
+)
+
+type certMagicLoggerCore struct {
+	core   zapcore.Core
+	fields []zapcore.Field
+}
+
+func (c certMagicLoggerCore) Enabled(lvl zapcore.Level) bool {
+	return c.core.Enabled(lvl)
+}
+
+func (c certMagicLoggerCore) With(fs []zapcore.Field) zapcore.Core {
+	return certMagicLoggerCore{core: c.core, fields: append(c.fields, fs...)}
+}
+
+func (c certMagicLoggerCore) Check(e zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
+	return ce.AddCore(e, c)
+}
+
+func (c certMagicLoggerCore) Write(e zapcore.Entry, fs []zapcore.Field) error {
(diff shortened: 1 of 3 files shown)


internal/autocert/certmagic_logger.go
```diff
@@ -0,0 +1,50 @@
...
+func (c certMagicLoggerCore) Enabled(lvl zapcore.Level) bool {
+	return c.core.Enabled(lvl)
+}
+
+func (c certMagicLoggerCore) With(fs []zapcore.Field) zapcore.Core {
+	return certMagicLoggerCore{core: c.core, fields: append(c.fields, fs...)}
+}
+
+func (c certMagicLoggerCore) Check(e zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
+	return ce.AddCore(e, c)
```
kenjenkins, line 26: Should this be conditional based on a call to `c.Enabled(e.Level)`, in order to respect the wrapped core's log level?

(I'm a little confused by the `Core` interface, so I might be misunderstanding the contract; please let me know if that's the case.)

internal/autocert/certmagic_logger.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (c certMagicLoggerCore) Write(e zapcore.Entry, fs []zapcore.Field) error {
```
kenjenkins, line 29: Would it be worth adding a unit test to exercise this behavior?

(optional)

kenjenkins (commented):
Very nice! I was a little intimidated at the thought of wrapping the `Core` interface, but this looks very reasonable. I just have one question about the log level.

kenjenkins (approved):

Issue #4352, opened 2023-07-11
Rely on Envoy to perform client certificate validation


kenjenkins:
**Is your feature request related to a problem? Please describe.**

We've had a customer request for stricter client certificate validation behavior: when downstream mTLS is enabled (via the [`client_ca`](https://www.pomerium.com/docs/reference/certificates#client-certificate-authority) setting) and a connection is made without a valid client certificate, reject the connection attempt altogether, instead of serving a Pomerium error page.

Currently client certificate validation is performed exclusively in the authorize service. If we instead configure Envoy to validate client certificates itself, then we can also provide a configuration option to enable this stricter client certificate validation behavior.

We will still need to perform additional validation in the authorize service, in order to support the per-route [`tls_downstream_client_ca`](https://www.pomerium.com/docs/reference/routes/tls-downstream-client-certificate-authority) setting.

**Describe the solution you'd like**

Configure the Envoy listener [`trusted_ca`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-field-extensions-transport-sockets-tls-v3-certificatevalidationcontext-trusted-ca) setting with a bundle of CA certificates formed from combining the Pomerium `client_ca` value with any `tls_downstream_client_ca` values. Configure Envoy to pass the validation status as an additional input to the Pomerium authorize service.

**Describe alternatives you've considered**

n/a

**Explain any additional use-cases**

This should also allow us to fix #4257.

**Additional context**

n/a

desimone:
Blocked until https://www.notion.so/pomerium/Client-mTLS-enforcement-c4507546f5c742cca709454665e2fec4#ba6186fff73845ff853f1b34fff41e9a

PR #4372, opened 2023-07-19
envoy: add a filter to store client cert info

kenjenkins:
## Summary

Add a new Lua filter that will store client certificate info as dynamic metadata. This will allow us to configure client certificate validation at the Envoy listener level, and then pass the results of that validation into our ExtAuthz service.

This also allows us to pass the entire client certificate chain (and not just the leaf certificate, which is how the 'include_peer_certificate' ExtAuthz setting behaves). This will allow us to add support for intermediate CA certificates supplied by the client.

However, if a client certificate does not validate successfully by Envoy, we will not store the certificate chain. (This should help guard against any possibility of making policy decisions based on unvalidated client certificate data.)

## Related issues

#4352 

## Checklist

- [x] ~~updated docs~~ [n/a]
- [x] ~~updated unit tests~~ [n/a]
- [x] ~~updated UPGRADING.md~~ [n/a]
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/luascripts/set-client-certificate-metadata.lua
+++ b/config/envoyconfig/luascripts/set-client-certificate-metadata.lua
@@ -1,0 +1,14 @@
+function envoy_on_request(request_handle)
+    local metadata = request_handle:streamInfo():dynamicMetadata()
+    local ssl = request_handle:streamInfo():downstreamSslConnection()
+    metadata:set("com.pomerium.client-certificate-info", "presented",
+                 ssl:peerCertificatePresented())
+    local validated = ssl:peerCertificateValidated()
+    metadata:set("com.pomerium.client-certificate-info", "validated", validated)
+    if validated then
+        metadata:set("com.pomerium.client-certificate-info", "chain", 
+                     ssl:urlEncodedPemEncodedPeerCertificateChain())
+    end
+end
+
+function envoy_on_response(response_handle) end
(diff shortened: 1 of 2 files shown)


config/envoyconfig/luascripts/set-client-certificate-metadata.lua
```diff
@@ -0,0 +1,14 @@
+function envoy_on_request(request_handle)
+    local metadata = request_handle:streamInfo():dynamicMetadata()
+    local ssl = request_handle:streamInfo():downstreamSslConnection()
+    metadata:set("com.pomerium.client-certificate-info", "presented",
```
kenjenkins, line 4: Let me know if you have any opinions on the metadata namespace to use for this data. The only guidance I've found from the Envoy documentation is this:

> Key is the reverse DNS filter name, e.g. com.acme.widget. The envoy.* namespace is reserved for Envoy’s built-in filters.

(from https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-field-config-core-v3-metadata-filter-metadata)

However, this metadata is for consumption by our ExtAuthz service, which can apparently ask for [any metadata namespaces it likes](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto.html#envoy-v3-api-field-extensions-filters-http-ext-authz-v3-extauthz-metadata-context-namespaces), so I don't think this needs to correspond exactly to a filter name.

Again, happy to change this if you have an opinion about what it should be.

calebdoxsey (approved):
LGTM

Issue #4257, opened 2023-06-14
Client CRL cannot be loaded to envoy listeners


gp666:
## What happened?

Hi! I tested the client side-mTLS feature by setting the `client_ca`  and `client_crl` in Pomerium config.
However, the Pomerium has errors occurred that Client CRL could not be loaded to envoy listener when I set the `client_crl` configuration in Pomerium config.

## What did you expect to happen?

Pomerium should be able to run with the `client_ca`  and `client_crl` settings , and Pomerium can deny users whose client certificates are revoked in the list of Client CRL.

## How'd it happen?

Pomerium is running in all-in-one mode :
1. Generated certificates and CRL (`client-ca.crt`, `client-ca.key` and `client-ca.crl`) by openssl
2. Set `client-ca-file` and `client-crl-file` in Pomerium config
3. Run the Pomerium

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`): pomerium: 0.22.2-1685134689+6efd1d6b
envoy: 1.25.5+b1095c058415dfb2261e695a0f144311a7dc346b6eb47ecbb0a01b7de2c7299f
- Server Operating System/Architecture/Cloud: Ubuntu 20.04

## What's your config.yaml?

```config.yaml
address: :443
grpc_address: :5443
grpc_insecure: false
authenticate_service_url: https://auth.domain.tld

certificate_authority_file: /certs/root/ca.crt
client_ca_file: /certs/client/client-ca.crt
client_crl_file: /certs/client/client-ca.crl

certificates:
  - cert: /certs/wildcard.crt
    key: /certs/wildcard.key
    
cookie_secret: <>
shared_secret: <>
signing_key: <>

idp_provider: google
idp_client_id: <>
idp_client_secret: <>

routes:
  - from: https://test.domain.tld
    to: https://127.0.0.1:9201
    allow_any_authenticated_user: true
    pass_identity_headers: true
```

## What did you see in the logs?

```logs
Jun 13 15:48:43 test-vm pomerium[1465383]: {"level":"info","service":"envoy","name":"upstream","time":"2023-06-13T15:48:43+08:00","message":"lds: add/update listener \\'metrics-ingress-11214784161331687928\\'"}
Jun 13 15:48:43 test-vm pomerium[1465383]: {"level":"info","service":"envoy","name":"upstream","time":"2023-06-13T15:48:43+08:00","message":"lds: add/update listener \\'outbound-ingress\\'"}
[...]

PR #4374, opened 2023-07-19
authorize: incorporate mTLS validation from Envoy

### LINKED ISSUE #4352: Rely on Envoy to perform client certificate validation

kenjenkins:
## Summary

Configure Envoy to validate client certificates, using the union of all relevant client CA bundles (that is, a bundle of the main client CA setting together with all per-route client CAs). Pass the validation status from Envoy through to the authorize service, by configuring Envoy to use the newly-added SetClientCertificateMetadata filter, and by also adding the relevant metadata namespace to the ExtAuthz configuration.

Remove the existing 'include_peer_certificate' setting from the ExtAuthz configuration, as the metadata from the Lua filter will include the full certificate chain (when it validates successfully by Envoy).

Update policy evaluation to consider the validation status from Envoy, in addition to its own certificate chain validation. (Policy evaluation cannot rely solely on the Envoy validation status while we still support the per-route client CA setting.)

## Related issues

Resolves #4352. Possibly #4257 as well (still need to add a test for that one).

## Checklist

- [x] ~~updated docs~~ [n/a]
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/authorize/grpc_test.go
+++ b/authorize/grpc_test.go
@@ -1,20 +1,25 @@
 package authorize
 
 import (
+	"bytes"
 	"context"
 	"net/http"
 	"net/url"
 	"testing"
 
+	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_service_auth_v3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
+	"github.com/rs/zerolog"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/pomerium/pomerium/authorize/evaluator"
 	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/internal/atomicutil"
 	"github.com/pomerium/pomerium/internal/sessions"
+	"github.com/pomerium/pomerium/internal/testutil"
 	"github.com/pomerium/pomerium/pkg/grpc/databroker"
 )
 
@@ -53,12 +58,9 @@ func Test_getEvaluatorRequest(t *testing.T) {
 		}},
 	})
 
-	actual, err := a.getEvaluatorRequestFromCheckRequest(
+	actual, err := a.getEvaluatorRequestFromCheckRequest(context.Background(),
 		&envoy_service_auth_v3.CheckRequest{
 			Attributes: &envoy_service_auth_v3.AttributeContext{
-				Source: &envoy_service_auth_v3.AttributeContext_Peer{
-					Certificate: url.QueryEscape(certPEM),
-				},
 				Request: &envoy_service_auth_v3.AttributeContext_Request{
 					Http: &envoy_service_auth_v3.AttributeContext_HttpRequest{
 						Id:     "id-1234",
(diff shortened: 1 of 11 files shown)


authorize/grpc_test.go
```diff
@@ -142,13 +158,144 @@ func Test_getEvaluatorRequestWithPortInHostHeader(t *testing.T) {
...
-			certPEM,
+			evaluator.ClientCertificateInfo{},
 			"",
 		),
 	}
 	assert.Equal(t, expect, actual)
 }
 
+func Test_getClientCertificateInfo(t *testing.T) {
+	const leafPEM = `-----BEGIN CERTIFICATE-----
```
calebdoxsey, line 169: optional: I think we can add `t.Parallel()`.

config/envoyconfig/listeners.go
```diff
@@ -574,6 +568,39 @@ func (b *Builder) buildDownstreamValidationContext(
...
+// clientCABundle returns a bundle of the globally configured client CA and any
+// per-route client CAs.
+func clientCABundle(ctx context.Context, cfg *config.Config) []byte {
+	var bundle bytes.Buffer
+	ca, _ := cfg.Options.GetClientCA()
+	addCAToBundle(&bundle, ca)
+	allPolicies := cfg.Options.GetAllPolicies()
+	for i := range allPolicies {
+		p := &allPolicies[i]
+		if p.TLSDownstreamClientCA == "" {
```
calebdoxsey, line 580: required: I think we need to account for `TLSDownstreamClientCAFile` too.

I *think* we handle file changes automatically in the config layer, so it's probably sufficient just to reproduce similar logic as `GetClientCA`

calebdoxsey (changes requested):

config/envoyconfig/listeners.go
kenjenkins: Denis had the same question on the previous v0.17 PR: https://github.com/pomerium/pomerium/pull/4228#discussion_r1222122189.

I agree it looks weird, but I think we don't need to check `TLSDownstreamClientCAFile` because the Policy `Validate()` method will populate `TLSDownstreamClientCA` when a file is configured. See config/policy.go here: https://github.com/pomerium/pomerium/blob/26bbcdfe07a81204b4357b15998e2fe152f7669f/config/policy.go#L506-L512

I've added a comment to help explain this. I will also test this locally.

authorize/grpc_test.go
kenjenkins: I wanted to also verify the log output from the `getClientCertificateInfo()` method, which is why I added a new `testutil.SetLogger()` method. If I'm understanding correctly, I think we need to make sure multiple parallel tests don't each try to set their own logger (as the logger is shared global state). Let me know if I'm misunderstanding how parallel tests work.

config/envoyconfig/listeners.go
kenjenkins: Setting `tls_downstream_client_ca_file` works for me locally when building with these changes.

calebdoxsey (approved):

PR #4380, opened 2023-07-20
envoy: configure upstream IP SAN match as needed

### LINKED ISSUE #4379: Investigate SAN matching configuration for https routes using IP addresses
## What happened?

For routes with a `to` URL using https we include a SAN matching condition in the configured Envoy cluster: https://github.com/pomerium/pomerium/blob/26bbcdfe07a81204b4357b15998e2fe152f7669f/config/envoyconfig/clusters.go#L345-L347

This condition will always match against DNS type entries: https://github.com/pomerium/pomerium/blob/26bbcdfe07a81204b4357b15998e2fe152f7669f/config/envoyconfig/tls.go#L45-L52

However, the route domain might be an IP address rather than a domain.

If Pomerium configures a SAN match condition in this case, I'd expect it to be of type "IP_ADDRESS" rather than "DNS".

## How'd it happen?

1. Configure Pomerium with an https route using an IP address, and enable the Envoy admin interface by setting `envoy_admin_address`.
2. Navigate to the configured Envoy admin interface and select 'config_dump'.
3. Find the UpstreamTlsContext corresponding to this route and observe the configured validation context, e.g.:
```json
         "validation_context": {
          "trusted_ca": {
           "filename": "<redacted>.pem"
          },
          "match_typed_subject_alt_names": [
           {
            "san_type": "DNS",
            "matcher": {
             "exact": "127.0.0.1"
            }
           }
          ]
         },
```

pomerium: v0.20.0-419-g7f533ba1-dirty+7f533ba1
envoy: 1.25.5+ecf50a958e5c053e5016b994943d8e77710b8c7ddeef5bc6ca32b8ca09e7bcbc

## What's your config.yaml?

```config.yaml
envoy_admin_address: '127.0.0.1:29091'

routes:
  - from: https://example.localhost.pomerium.io
    to: https://127.0.0.1:4443
    allow_any_authenticated_user: true
    tls_custom_ca_file: /path/to/custom/CA.pem
```

n/a

## Additional context

Noticed while investigating https://github.com/pomerium/pomerium/issues/4366#issuecomment-1644521456.

kenjenkins:
## Summary

When building an upstream validation context for a particular URL, check whether the hostname is an IP address. If so, configure the SAN match to use type IP_ADDRESS rather than DNS.

## Related issues

Resolves #4379.

## User Explanation

Fix TLS certificate validation logic for routes where the `to` URL uses https with an IP address.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/tls.go
+++ b/config/envoyconfig/tls.go
@@ -5,7 +5,7 @@ import (
 	"crypto/x509"
 	"encoding/asn1"
 	"fmt"
-	"net"
+	"net/netip"
 	"net/url"
 	"regexp"
 	"strings"
@@ -25,12 +25,16 @@ func (b *Builder) buildSubjectAltNameMatcher(
 		sni = overrideName
 	}
 
-	if net.ParseIP(sni) != nil {
+	if ip, err := netip.ParseAddr(sni); err == nil {
+		// Strip off any IPv6 zone.
+		if ip.Zone() != "" {
+			ip = ip.WithZone("")
+		}
 		return &envoy_extensions_transport_sockets_tls_v3.SubjectAltNameMatcher{
 			SanType: envoy_extensions_transport_sockets_tls_v3.SubjectAltNameMatcher_IP_ADDRESS,
 			Matcher: &envoy_type_matcher_v3.StringMatcher{
 				MatchPattern: &envoy_type_matcher_v3.StringMatcher_Exact{
-					Exact: sni,
+					Exact: ip.String(),
 				},
 			},
 		}
--- a/config/envoyconfig/tls.go
+++ b/config/envoyconfig/tls.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"encoding/asn1"
 	"fmt"
+	"net"
 	"net/url"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


config/envoyconfig/tls.go
```diff
@@ -24,6 +25,17 @@ func (b *Builder) buildSubjectAltNameMatcher(
 		sni = overrideName
 	}
 
+	if net.ParseIP(sni) != nil {
```
calebdoxsey, line 28: I vaguely recall something about the netip package being more efficient? https://pkg.go.dev/net/netip#ParseAddr

calebdoxsey (approved):

config/envoyconfig/tls.go
calebdoxsey: (I think maybe that was only for memory if you actually use IPs in code, so probably makes no difference.

kenjenkins: Sure, it probably makes sense to use the newer package. I see the netip package also supports [scoped IPv6 addresses](https://en.wikipedia.org/wiki/IPv6_address#Scoped_literal_IPv6_addresses_.28with_zone_index.29). I won't pretend to fully understand this concept, but I figure if someone uses a scoped IPv6 address in their route's `to` URL, it probably makes sense to recognize that as an IP address.

If a zone _is_ present I believe we need to strip that from the SAN matcher (IPv6 SAN's cannot contain a zone, if I'm reading https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6 correctly). I've added another unit test case for this.

Of course I really hope no one is creating certificates with link-local IPv6 addresses in the first place.

PR #4383, opened 2023-07-21
config: add customization options for logging

calebdoxsey:
## Summary
Add two new config options to customize logging: `access_log_fields` and `authorize_log_fields`. These options accept a list of strings that define components of a log message. 

An access log is produced by the proxy service when an HTTP request is made in envoy. An authorize log is produced when the authorize service runs an authorization check for access.

The defined fields (and defaults) mirror the existing log messages.

~~For the authorization log we also support `debug.` which indicates we should only log that field if the log level is set to `debug`. (this is to replicate the existing behavior of only logging request headers on `debug`)~~

I attempted to set these options as defaults in the `defaultOptions` struct, but unfortunately Viper fills values by appending to slices rather than replacing them, so instead we have getters that look for nil. (I also attempted to workaround this using a decode func but it didn't work) There's likely some way of dealing with this deep in the innards of Viper but I didn't want to spend any more time on it.

A `StringList` type is used in the protobuf so we can distinguish between `nil` and an empty list.

## Related issues
- https://github.com/pomerium/internal/issues/1470

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/log/authorize.go
```diff
@@ -0,0 +1,62 @@
...
+	AuthorizeLogFieldQuery,
+	AuthorizeLogFieldIP,
+	AuthorizeLogFieldSessionID,
+	AuthorizeLogFieldImpersonateSessionID,
+	AuthorizeLogFieldImpersonateUserID,
+	AuthorizeLogFieldImpersonateEmail,
+	AuthorizeLogFieldServiceAccountID,
+	AuthorizeLogFieldUser,
+	AuthorizeLogFieldEmail,
+	AuthorizeLogFieldHeaders.ForDebugOnly(),
```
kenjenkins, line 43: My first impression is that `ForDebugOnly()` / `IsForDebugOnly()` mechanism feels a little confusing.

What would you think about making `DefaultAuthorizeLogFields` into a function instead? Something like this:

```
func DefaultAuthorizeLogFields(logLevel zerolog.Level) []AuthorizeLogField {
    fields := []AuthorizeLogField{
        AuthorizeLogFieldRequestID,
        AuthorizeLogFieldCheckRequestID,
        AuthorizeLogFieldMethod,
        AuthorizeLogFieldPath,
        AuthorizeLogFieldHost,
        AuthorizeLogFieldQuery,
        AuthorizeLogFieldIP,
        AuthorizeLogFieldSessionID,
        AuthorizeLogFieldImpersonateSessionID,
        AuthorizeLogFieldImpersonateUserID,
        AuthorizeLogFieldImpersonateEmail,
        AuthorizeLogFieldServiceAccountID,
        AuthorizeLogFieldUser,
        AuthorizeLogFieldEmail,
    }
    if logLevel <= zerolog.DebugLevel {
        fields = append(fields, AuthorizeLogFieldHeaders)
    }
    return fields
} 
```

(definitely optional, please let me know if it feels simpler to keep it the way you have it here)

authorize/log.go
```diff
@@ -92,56 +73,140 @@ func (a *Authorize) logAuthorizeCheck(
...
+		return evt
+	case log.AuthorizeLogFieldSessionID:
+		if s, ok := s.(*session.Session); ok {
+			evt = evt.Str(string(field), s.GetId())
+		}
+		return evt
+	case log.AuthorizeLogFieldUser:
+		return evt.Str(string(field), u.GetId())
+	default:
+		return evt.Str(string(field), "<<UNKNOWN AUTHORIZE LOG FIELD>>")
```
kenjenkins, line 208: This feels a little awkward to me. Should we add logic to Options.Validate() to verify that there aren't any unknown log fields? Would it make sense to silently skip unknown fields here if we logged a warning for any unknown fields from Options.Validate()?

kenjenkins (commented):
Generally looks good to me, I just have a couple questions.

authorize/log.go
calebdoxsey: Yes I can add code to validate the options, in which case silently ignoring is ok.

internal/log/authorize.go
calebdoxsey: but then you'd have no way of defining this functionality yourself... maybe that's ok because if you want to customize it you probably don't care about log levels?

internal/log/access.go
```diff
@@ -19,18 +24,48 @@ const (
...
+		AccessLogFieldSize,
+		AccessLogFieldResponseCode,
+		AccessLogFieldResponseCodeDetails,
+	}
+}
+
+// ErrUnknownAccessLogField indicates that an access log field is unknown.
+var ErrUnknownAccessLogField = errors.New("unknown access log field")
+
+var accessLogFieldLookup = map[AccessLogField]struct{}{
```
kenjenkins, line 48: It would be nice if there were some way to generate this map automatically, so we didn't have to worry about keeping it in sync with the definitions above, but I can't think of anything easy at the moment.

(no action required)

config/options.go
```diff
@@ -1292,15 +1304,15 @@ func (o *Options) GetSigningKey() ([]byte, error) {
 // GetAccessLogFields returns the access log fields. If none are set, the default fields are returned.
 func (o *Options) GetAccessLogFields() []log.AccessLogField {
 	if o.AccessLogFields == nil {
-		return log.DefaultAccessLogFields
+		return log.DefaultAccessLogFields()
```
kenjenkins, line 1307: Do you have a sense of how much we should care about the memory allocation for the slice of AccessLogField here? I assume this will be called frequently under production load, but I assume we already make lots of other allocations while in the process of streaming access logs from Envoy.

(no action required)

kenjenkins (approved):
Looks great!

config/options.go
calebdoxsey: yes this would probably not be great. I can change it to use two pre-allocated arrays.

internal/log/access.go
calebdoxsey: We could use the list we made for defaults, but when we add new fields they won't be part of the defaults, so we wouldn't be able to use that slice anymore.

Issue #3444, opened 2022-06-22
[Feature Request] Allow envoy to log additional HTTP request headers to the log statement


kking124:
We would like the ability to designate additional http request headers to appear in the envoy access logs (e.g. `Geolocation` or `X-Amzn-Trace-Id`). 

**Describe the solution you'd like**

An additional configuration parameter/environment variable where a list of headers can be supplied that, if they exist, would be included in the envoy access log.

**Describe alternatives you've considered**

Alternatives include: 
- placing an additional proxy after Pomerium for the sole purpose of logging the headers
- extending the application behind Pomerium to log the additional headers

**Explain any additional use-cases**

The main value is to reduce the complexity of logging http request information. Additionally, in contexts where an additional request ID exists (such as X-Amzn-Trace-Id) it provides an additional capability to tie the request into a wider request context.

**Additional context**

example `yaml` config:

```yaml

envoy_log_http_headers:
  - X-Amz-Trace-Id
  - Geolocation
  - X-Application-Version

```

example `.env` config:
```env
ENVOY_LOG_HTTP_HEADERS=X-Amz-Trace-Id,Geolocation,X-Application-Version
```

expected output:
```json
{"level":"info","service":"envoy","upstream-cluster":"route-a0acefa27f89480a","method":"GET","authority":"verify.localhost.pomerium.io","path":"/api/verify-info","user-agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.5005.124 Safari/537.36 Edg/102.0.1245.44","referer":"https://verify.localhost.pomerium.io/","forwarded-for":"192.168.X.X","request-id":"40e9136e-0ddf-4889-b3d4-8051b4813d1e","duration":5.065997,"size":959,"response-code":200,"response-code-details":"via_upstream","time":"2022-06-22T14:40:52Z","message":"http-request","X-Amz-Trace-Id":"-","Geolocation":"geo:37.786971,-122.399677","X-Application-Version":"0.17.3-android"}
```

kking124:
Is there any update on this request?

desimone:
@kking124 -- perhaps the most direct way to do this is to expose envoy's access log formatter directly.

https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#

So the default

```
[...]

PR #4388, opened 2023-07-24
proxy: add support for logging http request headers

calebdoxsey:
## Summary
Add support for logging http headers in access and authorization logs by setting a field name of `headers.{HeaderName}` in the config. For example:

```yaml
access_log_fields: ["headers.accept"]
```

The header name is normalized using `http.CanonicalHeaderKey` so case should not matter.

Headers are displayed as an object in the logs:

```json
{
  "level": "info",
  "service": "authorize",
  "headers": {
    ":authority": "example.com",
    ":method": "GET",
    ":path": "/get",
    ":scheme": "https",
    "Accept": "application/json",
    "Accept-Encoding": "gzip, deflate, br"
  }
}
```

The authorization log can log all headers using the `"headers"` field. The access log does **not** support this because the request headers to send to the access logs have to be explicitly defined in Envoy. (The access log does support explicit header names)

## Related issues
- https://github.com/pomerium/pomerium/issues/3444
 
## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/log/headers.go
```diff
@@ -0,0 +1,56 @@
...
+	src map[string]string,
+) *zerolog.Event {
+	all := false
+	include := sets.NewHash[string]()
+	for _, field := range fields {
+		if field == headersFieldName {
+			all = true
+			break
+		} else if strings.HasPrefix(string(field), headersFieldPrefix) {
+			include.Add(http.CanonicalHeaderKey(string(field[len(headersFieldPrefix):])))
```
kenjenkins, line 39: Can you help me understand: is this `include` set needed for the access logs, or the authorize logs, or both?

(I'm a little confused about what Envoy's [`request_headers`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/data/accesslog/v3/accesslog.proto#envoy-v3-api-field-data-accesslog-v3-httprequestproperties-request-headers) map contains in the absence of the `AdditionalRequestHeadersToLog` setting.)

config/envoyconfig/envoyconfig.go
```diff
@@ -92,6 +93,13 @@ func buildAccessLogs(options *config.Options) []*envoy_config_accesslog_v3.Acces
 		return nil
 	}
 
+	var additionalRequestHeaders []string
+	for _, field := range options.AccessLogFields {
+		if headerName, ok := log.GetHeaderField(field); ok {
+			additionalRequestHeaders = append(additionalRequestHeaders, http.CanonicalHeaderKey(headerName))
```
kenjenkins, line 99: Do we intend to support HTTP/2 pseudo-headers (`:authority`, `:method`, etc.)? I think CanonicalHeaderKey won't canonicalize those (cf. https://go.dev/play/p/MRbeUn11-fp).

(Does that matter here? I'm not sure.)

internal/log/headers_test.go
```diff
@@ -0,0 +1,63 @@
...
+		evt = HTTPHeaders(evt, A{"a", "b", "c"}, M{
+			"a": "1",
+			"b": "2",
+			"c": "3",
+		})
+		evt.Send()
+
+		assert.Equal(t, `{"level":"info"}`, strings.TrimSpace(buf.String()))
+	})
+	t.Run("one", func(t *testing.T) {
```
kenjenkins, line 48: nit: "two" instead of "one"?

internal/log/headers.go
calebdoxsey: It is for access logs when called from the envoy control plane code and for authorize logs when called from the authorize service.

Envoy's `request_headers` contains nothing when `AdditionalRequestHeadersToLog` is empty.

config/envoyconfig/envoyconfig.go
calebdoxsey: Yes they are supported. `headers.:authority` worked for me. Should we write custom code to lower case `:` header names?

kenjenkins: I could go either way. It's probably fine not to convert something like `headers.:Authority`; that feels like an odd thing to type.

internal/log/headers.go
kenjenkins: Maybe I'm still missing something... if we already configure the set of request headers from Envoy, then we wouldn't need to filter them out here, right? Does that mean this is only needed for the authorize logs, where we always start with the full set of request headers?

kenjenkins (approved):

internal/log/headers.go
calebdoxsey: Ok yes I can write different logic for the access logs.

kenjenkins (approved):

Issue #4376, opened 2023-07-20
logging: add option to add a user's `id_token` to the authorization log


desimone:
### What

While our current authorization logs contain [basic user information](https://github.com/pomerium/pomerium/blob/26bbcdfe07a81204b4357b15998e2fe152f7669f/authorize/log.go#L64-L66
) like user email and groups, being able to quickly see a user's request detail's (including extra custom claims) can be helpful in a security audit, or to troubleshoot access issues quickly.

### Solution

We should pass the user's most recent **encoded** AND base64 **decoded** `id_token` as part of the request. 

### Alternatives

We provide a user configurable way to select the claims. But this would likely add latency to each request for parsing, and open up the possibility for information leakage. In theory, id_token's should be relatively safe to put in logs. 

### Considerations

- Every identity provider does slightly different things with `id_tokens` which could make this result inconsistent. For example, Azure provides you the complete user contexts only _once_ during initial sign in. The following id_token / user info calls are assumed to be updates only.  

### Related issues

- https://github.com/pomerium/pomerium/issues/4209

kenjenkins:
From chatting with @desimone I believe it's fine if we want to add two separate options for logging the raw ID token and the decoded ID token claims. (I have a slight preference for doing it that way, as to me it feels more consistent with the other log fields, but it's probably fine to leave it the way #4392 has it for now. We can always split these up later if we need to.)

PR #4392, opened 2023-07-25
authorize: add support for logging id token

calebdoxsey:
## Summary
Add support for logging the id token. When this field is included in the logs the raw id token will be logged to `id-token` and the decoded claims to `id-token-claims`.

## Related issues
- https://github.com/pomerium/pomerium/issues/4376
 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authorize/log.go
+++ b/authorize/log.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	envoy_service_auth_v3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
+	"github.com/go-jose/go-jose/v3/jwt"
 	"github.com/rs/zerolog"
 
 	"github.com/pomerium/pomerium/authorize/evaluator"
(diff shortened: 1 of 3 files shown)


authorize/log.go
```diff
@@ -161,6 +162,17 @@ func populateLogEvent(
...
 	case log.AuthorizeLogFieldHost:
 		return evt.Str(string(field), in.GetAttributes().GetRequest().GetHttp().GetHost())
+	case log.AuthorizeLogFieldIDToken:
+		if s, ok := s.(*session.Session); ok {
+			evt = evt.Str("id-token", s.GetIdToken().GetRaw())
+
+			if t, err := jwt.ParseSigned(s.GetIdToken().GetRaw()); err == nil {
+				var m map[string]any
+				_ = t.UnsafeClaimsWithoutVerification(&m)
+				evt = evt.Interface("id-token-claims", m)
```
kenjenkins, line 172: (no changes required)

Do we already verify the ID token signature somewhere else, and that's why it's fine to fine to call UnsafeClaimsWithoutVerification() here?

kenjenkins (approved):

authorize/log.go
calebdoxsey: I don't know about the security ramifications of this. Verifying the signature of the ID token against the configured identity provider would be difficult to implement here (and also likely very slow)

kenjenkins: It looks like there's some verification related to ID tokens here: https://github.com/pomerium/pomerium/blob/baf8918676aeca3f6f0b017252c1c9560a0d7381/internal/identity/oidc/oidc.go#L225

Do you know if all ID tokens come from that `getIDToken()` method?

calebdoxsey: The ID token comes from the identity provider and is saved to the session.

PR #4442, opened 2023-08-07
config: move mTLS settings to new struct

kenjenkins:
## Summary

Move downstream mTLS settings to a nested config file object, under the key 'downstream_mtls', and add a new DownstreamMTLSSettings struct for these settings.

Deprecate the existing ClientCA and ClientCAFile fields in the Options struct, but continue to honor them for now (log a warning if either is populated).

Delete the ClientCRL and ClientCRLFile fields entirely (in current releases these cannot be set without causing an Envoy error, so this should not be a breaking change).

Update the Settings proto to mirror this nested structure.

## Related issues

Preparation for adding new mTLS settings for #4353.

## User Explanation

Group downstream mTLS settings under a new `downstream_mtls` config file key. Deprecate the existing `client_ca` and `client_ca_file` settings in favor of `downstream_mtls.ca` and `downstream_mtls.ca_file`.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


config/mtls.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// CRLFile is the path to a file containing the certificate revocation
+	// list (or bundle of CRLs) to use when validating client certificates.
+	CRLFile string `mapstructure:"crl_file" yaml:"crl_file,omitempty"`
+}
+
+// GetCA returns the certificate authority (or nil if unset).
+func (s *DownstreamMTLSSettings) GetCA() ([]byte, error) {
+	if s.CA != "" {
+		return base64.StdEncoding.DecodeString(s.CA)
```
wasaga, line 37: could you please wrap the error with explanation of which specific field it relates to?

config/mtls.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	if s.CAFile != "" {
+		return os.ReadFile(s.CAFile)
```
wasaga, line 40: same

config/mtls.go
```diff
@@ -0,0 +1,92 @@
...
+	// CRLs) to use when validating client certificates.
+	CRL string `mapstructure:"crl" yaml:"crl,omitempty"`
+
+	// CRLFile is the path to a file containing the certificate revocation
+	// list (or bundle of CRLs) to use when validating client certificates.
+	CRLFile string `mapstructure:"crl_file" yaml:"crl_file,omitempty"`
+}
+
+// GetCA returns the certificate authority (or nil if unset).
+func (s *DownstreamMTLSSettings) GetCA() ([]byte, error) {
```
wasaga, line 35: please add a check that `ca` and `ca_file` cannot be set simultaneously

config/mtls.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// GetCRL returns the certificate revocation list bundle (or nil if unset).
+func (s *DownstreamMTLSSettings) GetCRL() ([]byte, error) {
+	if s.CRL != "" {
+		return base64.StdEncoding.DecodeString(s.CRL)
+	}
+	if s.CRLFile != "" {
+		return os.ReadFile(s.CRLFile)
+	}
```
wasaga, line 52: same as above

config/mtls.go
```diff
@@ -0,0 +1,92 @@
...
+		if _, err := base64.StdEncoding.DecodeString(s.CA); err != nil {
+			return fmt.Errorf("CA: %w", err)
+		}
+	}
+
+	if s.CAFile != "" {
+		if _, err := os.ReadFile(s.CAFile); err != nil {
+			return fmt.Errorf("CA file: %w", err)
+		}
+	}
```
wasaga, line 67: seems like a duplicate of `GetCA()`

config/mtls.go
wasaga, line 40: I don't remember whether it was during audit or not, but mostly we wrap reading from sources into LimitReader just in case we're accidentally pointed to some large / infinite file. i.e. https://github.com/pomerium/pomerium/blob/bb2fb9730ad4f714d025578305236081fba32db9/pkg/cryptutil/certificates.go#L235

config/mtls.go
```diff
@@ -0,0 +1,92 @@
...
+		return base64.StdEncoding.DecodeString(s.CA)
+	}
+	if s.CAFile != "" {
+		return os.ReadFile(s.CAFile)
+	}
+	return nil, nil
+}
+
+// GetCRL returns the certificate revocation list bundle (or nil if unset).
+func (s *DownstreamMTLSSettings) GetCRL() ([]byte, error) {
```
wasaga, line 46: should this function parse CRL and return `x509.RevocationList` instead of []byte ?

pkg/grpc/config/config.proto
```diff
@@ -239,3 +241,8 @@ message Settings {
   optional string favicon_url = 90;
   optional string error_message_first_paragraph = 91;
 }
+
+message DownstreamMtlsSettings {
+  optional string ca = 1;
+  optional string crl = 2;
```
wasaga, line 247: there was some historic b64 strings / bytes confusion in the protobuf, I believe we wanted to use use actual `bytes` in protobuf? i.e. https://github.com/pomerium/pomerium/blob/bb2fb9730ad4f714d025578305236081fba32db9/pkg/grpc/config/config.proto#L136-L140

config/mtls.go
kenjenkins: Done

config/mtls.go
kenjenkins: Done

PR #4445, opened 2023-08-08
zero: resource bundle reconciler

wasaga:
## Summary

Bundle Reconciler watches a group of resource bundles, downloads them and reconciles their contents with the databroker. 

Bundles are not expected to have any intersection across record types. 

For caching purposes, an entry with etag/last-modified is kept in the databroker, and the upload is conditional. That way, there's no need to keep a cache of downloaded bundles. 

## Related issues

Implements https://github.com/pomerium/internal/issues/1423

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


internal/zero/reconciler/records.go
```diff
@@ -0,0 +1,132 @@
+package reconciler
+
+// RecordSetBundle is an index of databroker records by type
+type RecordSetBundle[T Record[T]] map[string]RecordSet[T]
```
calebdoxsey, line 4: nit: I'm not sure why this is a generic when it's only ever used for `DatabrokerRecord`?

internal/zero/reconciler/bundles_queue.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+	}
+
+	newBundle := bundle{id: id, synced: false, priority: b.counter}
+	heap.Push(&b.bundles, newBundle)
+	b.counter++
+}
+
+// MarkForSyncLater marks the bundle with the given ID for syncing later (after all other bundles).
+func (b *BundleQueue) MarkForSyncLater(id string) {
```
calebdoxsey, line 84: nit: this method is `O(n)`. If the goal of using the heap package was to achieve `O(log n)` performance, shouldn't all the methods be that or better? Otherwise why not just use a loop?

calebdoxsey (approved):

internal/zero/reconciler/records.go
wasaga: I found it convenient to use generics for a higher level functionality like this one, as it allows writing more concise and easy to read tests compared to using the full data structure as it might be pretty large. See `record_tests.go`

PR #4456, opened 2023-08-11
authorize: rework token substitution in headers

kenjenkins:
## Summary

Currently Pomerium replaces dynamic `set_request_headers` tokens sequentially. As a result, if a replacement value itself contained a supported "$pomerium" token, Pomerium may treat that as another replacement, resulting in incorrect output.

This is unlikely to be a problem given the current set of dynamic tokens, but if we continue to add additional tokens, this will likely become more of a concern.

To forestall any issues, let's perform all replacements in one pass, using the `os.Expand()` method. This does require a slight change to the syntax, as tokens containing a '.' (i.e. all of them) will need to be wrapped in curly braces, like `${pomerium.id_token}`.

## Related issues

Resolves https://github.com/pomerium/internal/issues/1501.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/authorize/evaluator/headers_evaluator.go
+++ b/authorize/evaluator/headers_evaluator.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
 
 	envoy_config_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/types"
 
 	"github.com/pomerium/pomerium/authorize/evaluator/opa"
 	"github.com/pomerium/pomerium/authorize/internal/store"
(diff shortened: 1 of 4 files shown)


authorize/evaluator/headers_evaluator.go
```diff
@@ -54,6 +57,45 @@ type HeadersResponse struct {
...
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("input_string", types.S),
+			types.Named("replacements",
+				types.NewObject(nil, types.NewDynamicProperty(types.S, types.S))),
+		),
+		types.Named("output", types.S),
+	),
+}, func(bctx rego.BuiltinContext, op1 *ast.Term, op2 *ast.Term) (*ast.Term, error) {
+	inputString, ok := op1.Value.(ast.String)
```
kenjenkins, line 71: I've tried to follow the example of [getGoogleCloudServerlessHeadersRegoOption](https://github.com/pomerium/pomerium/blob/5568606f03023c1f936ed70b5a41fe3679ba6426/authorize/evaluator/google_cloud_serverless.go#L37C1-L59) here, but I don't completely understand whether all these type checks are needed (given the `Decl` type above).

calebdoxsey (approved):

authorize/evaluator/headers_evaluator.go
```diff
@@ -54,6 +57,48 @@ type HeadersResponse struct {
...
+	replacements, ok := op2.Value.(ast.Object)
+	if !ok {
+		return nil, fmt.Errorf("invalid replacements type: %T", op2.Value)
+	}
+
+	var err error
+	output := os.Expand(string(inputString), func(key string) string {
+		if key == "$" {
+			return "$" // allow a dollar sign to be escaped using $$
+		}
```
kenjenkins, line 85: I realized that someone might conceivably want to include the $ character in a header value (directly followed by some text), so I added a special case for this. This allows setting the literal value "$foo" using the syntax "$$foo" (just like how dollar signs can be escaped in Makefiles).

calebdoxsey (approved):

PR #4545, opened 2023-09-15
core/authenticate: validate the identity profile

calebdoxsey:
## Summary
Validate the identity profile by checking that the oauth token is still valid and calling the user info endpoint. If the access token has been revoked that call should fail and the identity profile will be treated as if it doesn't exist. This will then trigger the re-authentication flow.

## Related issues
- https://github.com/pomerium/internal/issues/1527

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/authenticate/identity_profile.go
+++ b/authenticate/identity_profile.go
@@ -99,3 +99,28 @@ func (a *Authenticate) storeIdentityProfile(w http.ResponseWriter, aead cipher.A
 	cookie.Path = "/"
 	return cookieChunker.SetCookie(w, cookie)
 }
+
+func (a *Authenticate) validateIdentityProfile(ctx context.Context, profile *identitypb.Profile) error {
+	authenticator, err := a.cfg.getIdentityProvider(a.options.Load(), profile.GetProviderId())
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


authenticate/identity_profile.go
```diff
@@ -99,3 +99,28 @@ func (a *Authenticate) storeIdentityProfile(w http.ResponseWriter, aead cipher.A
...
+	}
+
+	oauthToken := new(oauth2.Token)
+	err = json.Unmarshal(profile.GetOauthToken(), oauthToken)
+	if err != nil {
+		return fmt.Errorf("invalid oauth token in profile: %w", err)
+	}
+
+	if !oauthToken.Valid() {
+		return fmt.Errorf("invalid oauth token in profile")
```
wasaga, line 116: can AccessToken be missing or it's just expiration condition?

wasaga (approved):

authenticate/identity_profile.go
calebdoxsey: I don't think it can be missing, but I'm not sure.

PR #4622, opened 2023-10-05
Allow to match on other SAN fields for mTLS policies

### LINKED ISSUE #4615: Add more capabilities to certificate-matcher for mTLS routes
Currently the Certificate Matcher that can be configured in a route when using mTLS only offers the following capabilities:
  - fingerprint (SHA256)
  - spki_hash

This is insufficient and not scalable when configuring policies. Instead, one would expect the certificate matcher to allow a lot more granulatity, such as filtering on the SANs and using complex expression to build better mTLS route-logic.
In addition, when a route matches on mTLS, there is nothing in the JWT assertion to indicate that this is what allowed the connection to get through, instead one has to add custom HTTP header that can only include the fingerprint of the certificate.
It would be interesting to add claims to the JWT token to indicate reference the "method" that was used to authenticate (user, bearer, mtls certificate) and carry more metadata.

The ability to build better mTLS route ACL using more complex matching, for instance:

```
- client_certificate:
    match_subject_alt_names:
      - email: '.*@blah\.com'
```

I have tried using the global SAN matching but it seems a little buggy.
I have reported https://github.com/pomerium/pomerium/issues/4614 on the side of this feature request.

calderonth:
## Summary

This introduces more capabilities for the mTLS matching as opposed to only supporting `fingerprint` and `spki_hash`.
Indeed, this allows to match on valuable fields of the mTLS certificate such as SubjectAltName fields such as email address, dns names, ip addresses and URIs.

## Related issues

Fixes #4615 

## User Explanation

Self-explanatory in the above section.

## Checklist

- [ ] updated docs
- [X] updated unit tests
- [ ] updated UPGRADING.md
- [X] ready for review

### DIFF
--- a/pkg/policy/criteria/client_certificate.go
+++ b/pkg/policy/criteria/client_certificate.go
@@ -4,6 +4,9 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"net/mail"
+	"net/url"
+	"net"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


calderonth:
Sure thing, feel free to discuss, it was essentially a first stab at it.

calderonth:
Hey there,

I'm.in conversations with Bobby on agreeing on a corporate CLA instead as I
can't sign it as an individual having work on this during my employment.

IP SAN matching seems less useful, I just think it might be nice to have
feature parity. Otherwise, DNS, email and URI field are likely the most
useful ones.

PR #4632, opened 2023-10-19
identity: override TokenSource expiry behavior

kenjenkins:
## Summary

The current session refresh loop attempts to refresh access tokens when they are due to expire in less than one minute. However, the code to perform the refresh relies on a `TokenSource` from the x/oauth2 package, which has its own internal [`expiryDelta`](https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.12.0:oauth2.go;l=299) threshold, with a default of 10 seconds. As a result, the first four or five attempts to refresh a particular access token will not actually refresh the token. The refresh will happen only when the access token is within 10 seconds of expiring.

Instead, before we obtain a new `TokenSource`, first clear any existing access token. This causes the `TokenSource` to consider the token invalid, triggering a refresh. This should give the refresh loop more control over when refreshes happen.

Consolidate this logic in a new `Refresh()` method in the oidc package. Add unit tests for this new method.

## User Explanation

Improve reliability of session refresh.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/internal/identity/oauth/apple/apple.go
+++ b/internal/identity/oauth/apple/apple.go
@@ -130,16 +130,9 @@ func (p *Provider) Authenticate(ctx context.Context, code string, v identity.Sta
 
 // Refresh renews a user's session.
 func (p *Provider) Refresh(ctx context.Context, t *oauth2.Token, v identity.State) (*oauth2.Token, error) {
-	if t == nil {
-		return nil, oidc.ErrMissingAccessToken
-	}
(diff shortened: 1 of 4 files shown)


calebdoxsey (approved):
Makes sense to me. It was surprising to me that the oauth2 package had this caching behavior and we clearly didn't anticipate it.

PR #4641, opened 2023-10-25
core/envoy: fix remove cookie lua script

calebdoxsey:
## Summary
Rewrite the `remove_pomerium_cookie` lua function to handle `=` inside of cookie values.

## Related issues
Fixes https://github.com/pomerium/internal/issues/1530

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


config/envoyconfig/luascripts/clean-upstream.lua
```diff
@@ -1,15 +1,22 @@
...
-
 function has_prefix(str, prefix)
     return str ~= nil and str:sub(1, #prefix) == prefix
 end
 
+function remove_pomerium_cookie(cookie_name, cookie)
+    local result = ""
+    for c in cookie:gmatch("([^;]+)") do
+        c = c:gsub("[ ]+", "")
+        if not has_prefix(c, cookie_name) then
```
kenjenkins, line 9: I think this may remove other cookies too (e.g. `_pomerium_something_else` in addition to `_pomerium`).

Should we also split on the first `=`? Would something like this work?

```suggestion
        local name = c:match("^ *([^=]+)")
        if name ~= cookie_name then
```

kenjenkins (approved):

PR #4650, opened 2023-10-30
core/hpke: reduce memory usage from zstd

calebdoxsey:
## Summary
Based on https://pkg.go.dev/tailscale.com/smallzstd we can use a smaller window size and some options to reduce the memory usage for zstd.

Re-using the encoder and decoder appears to be supported based on the packages documentation. The defaults are not a memory leak, but they do use a lot of memory since they are optimized for large payloads. 

## Related issues
- https://github.com/pomerium/pomerium/issues/4652
- #4648 
 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/hpke/url.go
+++ b/pkg/hpke/url.go
@@ -145,21 +145,30 @@ func withoutHPKEParams(values url.Values) url.Values {
 	return filtered
 }
 
-var zstdEncoder, _ = zstd.NewWriter(nil, zstd.WithEncoderLevel(zstd.SpeedBestCompression))
-
 func encodeQueryStringV1(values url.Values) []byte {
 	return []byte(values.Encode())
 }
+
+const zstdWindowSize = 8 << 10 // 8kiB
+
+var zstdEncoder, _ = zstd.NewWriter(nil,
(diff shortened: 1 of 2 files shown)


pkg/hpke/url.go
```diff
@@ -145,22 +145,31 @@ func withoutHPKEParams(values url.Values) url.Values {
...
-
 func encodeQueryStringV1(values url.Values) []byte {
 	return []byte(values.Encode())
 }
 
+const zstdWindowSize = 8 << 10 // 8kiB
+
+var zstdEncoder, _ = zstd.NewWriter(nil,
+	zstd.WithEncoderLevel(zstd.SpeedBestCompression),
+	zstd.WithEncoderConcurrency(1),
```
kenjenkins, line 156: It looks like with this option only one call to `EncodeAll()` can happen at a time (it may block while acquiring the one available encoder instance, see [here](https://github.com/klauspost/compress/blob/8eb6542f08f086fb92796f6d8a4531d754947309/zstd/encoder.go#L495)).

Do we feel comfortable about that?

PR #4655, opened 2023-10-31
databroker: build config concurrently, option to bypass validation

wasaga:
## Summary

1. refactors databroker.ConfigSource to spread config build across multiple goroutines
2. Adds (code-only) option to bypass certain expensive validations, which is useful in case config originates from source that already validates config (i.e. console).

## Related issues

- https://github.com/pomerium/internal/issues/1568
- https://github.com/pomerium/internal/issues/1579

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/cmd/pomerium/pomerium.go
```diff
@@ -40,7 +40,7 @@ func Run(ctx context.Context, src config.Source) error {
 	if err != nil {
 		return err
 	}
-	src = databroker.NewConfigSource(ctx, src)
+	src = databroker.NewConfigSource(ctx, src, databroker.EnableConfigValidation(true))
```
wasaga, line 43: this one is to keep current behaviour for now

internal/databroker/config_source.go
```diff
@@ -74,108 +88,131 @@ func (src *ConfigSource) rebuild(ctx context.Context, firstTime firstTime) {
...
+
+	return policy, nil
+}
+
+func (src *ConfigSource) addPolicies(ctx context.Context, cfg *config.Config, policies []*config.Policy) {
+	seen := make(map[uint64]struct{})
+	for _, policy := range cfg.Options.GetAllPolicies() {
+		id, err := policy.RouteID()
 		if err != nil {
-			metrics.SetDBConfigRejected(ctx, cfg.Options.Services, id, cfgpb.version, err)
```
wasaga, line 118: those metrics were originally added for console, however they are not compatible with chunked config anyway currently, thus removing them for now as they require redesign.

internal/databroker/config_source.go
```diff
@@ -74,108 +88,131 @@ func (src *ConfigSource) rebuild(ctx context.Context, firstTime firstTime) {
...
-		cfgpb := src.dbConfigs[id]
-
+	for i := 0; i < len(ids) && ctx.Err() == nil; i++ {
+		cfgpb := src.dbConfigs[ids[i]]
 		cfg.Options.ApplySettings(ctx, certsIndex, cfgpb.Settings)
-		var errCount uint64
+	}
+}
 
-		err := cfg.Options.Validate()
```
calebdoxsey, line 116: I believe `.Validate` has side effects and is necessary for some config options to be usable. I'm not sure what those are though.

internal/databroker/config_source.go
```diff
@@ -74,108 +88,131 @@ func (src *ConfigSource) rebuild(ctx context.Context, firstTime firstTime) {
...
+			return fmt.Errorf("validating settings: %w", err)
 		}
-		seen[id] = ""
+		return nil
+	})
+	policies := slices.NewSafeSlice[*config.Policy]()
+	for _, cfgpb := range src.dbConfigs {
+		for _, routepb := range cfgpb.GetRoutes() {
+			routepb := routepb
+			eg.Go(func() error {
```
calebdoxsey, line 132: I think this will change the order of routes, which is a breaking change.

internal/databroker/config_source.go
calebdoxsey: Oh I see we call it above.

internal/databroker/config_source.go
wasaga: added sorting

PR #4657, opened 2023-11-01
core/config: refactor change dispatcher

calebdoxsey:
## Summary
Refactor the config Change Dispatcher to use an `events.Target`.

`events.Target` is a new events implementation that triggers events in separate goroutines so that config source subscribers can be called concurrently.

For example if (1) generating envoyconfig takes 30 seconds, (2) rebuilding authorization takes 45 seconds and (3) updating XDS takes 60 seconds, previously we would've done 1+2+3 sequentially resulting in 135 seconds total time. We will do the authorization and XDS update concurrently 1+max(2,3) resulting in 90 seconds total time. (assuming cores are available)

In addition since updates are handled each in their goroutine, the next call to 1 can happen while the previous call to 2,3 is still completing, further reducing the amount of total time taken. (1b-1a)+max(2,3)

## Related issues
- https://github.com/pomerium/internal/issues/1575

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/events/target.go
```diff
@@ -0,0 +1,160 @@
...
+//
+// Target is safe to use in its zero state.
+//
+// The first time any method of Target is called a background goroutine is started that handles
+// any requests and maintains the state of the listeners. Each listener also starts a
+// separate goroutine so that all listeners can be invoked concurrently.
+//
+// The channels to the main goroutine and to the listener goroutines have a size of 1 so typically
+// methods and dispatches will return immediately. However a slow listener will cause the next event
+// dispatch to block. This is the opposite behavior from Manager.
```
wasaga, line 43: for follow-up: that might be a great place to inject telemetry instrumentation to, in order to automatically calculate delays, especially if we give human readable names to subscribers

internal/events/target.go
```diff
@@ -0,0 +1,160 @@
...
+		go t.run()
+	})
+}
+
+func (t *Target[T]) run() {
+	// listen for add/remove/dispatch events and call functions
+	for {
+		select {
+		case <-t.ctx.Done():
+			return
```
wasaga, line 116: should we evict all listeners on context cancellation, so that all goroutines we spawned in addListener would quit?

wasaga (approved):

internal/events/target.go
calebdoxsey: they will now exit at line 137

internal/events/target.go
```diff
@@ -0,0 +1,165 @@
...
+
+// Close closes the event target. This can be called multiple times safely.
+// Once closed the target cannot be used.
+func (t *Target[T]) Close() {
+	t.init()
+
+	t.cancel(errors.New("target closed"))
+}
+
+// Dispatch dispatches an event to any listeners.
```
wasaga, line 81:
```suggestion
// Dispatch dispatches an event to all listeners.
```

internal/events/target.go
```diff
@@ -0,0 +1,165 @@
...
+	"context"
+	"errors"
+	"sync"
+
+	"github.com/google/uuid"
+)
+
+type (
+	// A Listener is a function that listens for events of type T.
+	Listener[T any] func(T)
```
wasaga, line 13: the contract with this function is that we allow listener to run for some time, we do not expect it to return almost immediatelly, correct? 

if that's the case, do we maybe want to add context here for the sake of telemetry context propagation? 

And also add the context to Dispatch(), so that if you're sending an event that's part of some operation in telemetry terms, you could collect spans from listeners as well?

calebdoxsey: We probably don't want a trace span that contains the entire execution stack of Pomerium, but sure I can add context to more places.

calebdoxsey:
Personally I don't understand how context should be used with this package. I suspect naive usage will result in unexpected behavior. Like canceling the context in A leading to process B stopping unexpectedly. The purpose of events was to de-couple systems. We've now re-coupled them.

internal/events/target.go
```diff
@@ -0,0 +1,166 @@
...
+// these functions are not thread-safe. They are intended to be called only by "run".
+
+func (t *Target[T]) addListener(listener Listener[T], handle Handle) {
+	ch := make(chan dispatchEvent[T], 1)
+	t.listeners[handle] = ch
+	// start a goroutine to send events to the listener
+	go func() {
+		for {
+			select {
+			case <-t.ctx.Done():
```
kenjenkins, line 138: I think this case is missing a `return`, to stop the goroutine from looping.

internal/events/target.go
```diff
@@ -0,0 +1,166 @@
...
+			case evt := <-ch:
+				listener(evt.ctx, evt.event)
```
kenjenkins, line 140: Do we also need some way to break out of the loop when `ch` is closed? It looks like calling RemoveListener() will result in the listener being called with a zero event value.

PR #4697, opened 2023-11-01
core/filemgr: use xxhash instead of sha512 for filenames

calebdoxsey:
## Summary
For envoy we use files as a DataSource whenever possible. When we have raw data we save it to a cache directory with a filename like `tls-crt-xxx.pem`. The hash of the filename is computed using the SHA512_256 of the bytes.

For this use case its not necessary to have a cryptographic hash. We can instead use `xxhash` which is much faster. Roughly 10x improvement:

```
// BenchmarkGetFileNameWithBytesHash-10    	   12331	     95284 ns/op	    1657 B/op	      63 allocs/op
// BenchmarkGetFileNameWithBytesHash-10    	  112304	      9472 ns/op	     112 B/op	       5 allocs/op

```

## Related issues
- https://github.com/pomerium/internal/issues/1573
 

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/envoyconfig/clusters_test.go
+++ b/config/envoyconfig/clusters_test.go
@@ -23,7 +23,7 @@ import (
 func Test_buildPolicyTransportSocket(t *testing.T) {
 	ctx := context.Background()
 	cacheDir, _ := os.UserCacheDir()
-	customCA := filepath.Join(cacheDir, "pomerium", "envoy", "files", "custom-ca-32484c314b584447463735303142374c31414145374650305a525539554938594d524855353757313942494d473847535231.pem")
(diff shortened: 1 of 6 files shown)


wasaga:
does xxhash require cgo? 

> The package is written with optimized pure Go and also contains even faster assembly implementations for amd64 and arm64. If desired, the purego build tag opts into using the Go code even on those architectures.

calebdoxsey:
Not as far as I know. I didn't add xxhash. We already use it. If we don't trust it I can replace it with something else everywhere we use it.

wasaga (approved):
I didn't notice it's an existing dependency

PR #4700, opened 2023-11-01
core/events: fix flaky test

calebdoxsey:
## Summary
<s>
Currently the event target pushes an operation onto a channel for each of its methods. This is then handled by a goroutine. Since each of the operations goes through a separate channel, and the channels were defined with a size of 1, it was possible for call order to behave unpredictably. For example:

```go
AddListener(li)
Dispatch(ctx, evt)
```

Could end up with the `Dispatch` occurring before `AddListener`. 

By making these channels synchronous we can avoid this behavior. Now `AddListener` will block until the goroutine picks it up, and `Dispatch` will always happen after `AddListener`.
</s>

Refactor the `events.Target` to use mutexes instead of a background goroutine.

### DIFF
--- a/internal/events/target.go
+++ b/internal/events/target.go
@@ -40,9 +40,9 @@ type (
 // any requests and maintains the state of the listeners. Each listener also starts a
 // separate goroutine so that all listeners can be invoked concurrently.
 //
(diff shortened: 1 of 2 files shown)


wasaga (approved):

calebdoxsey:
I will refactor this to use mutexes.

PR #4701, opened 2023-11-02
databroker: add utility recordset and changeset

wasaga:
## Summary

Adds utility functions to calculate the difference between observed and derived state of databroker. Used by the console and zero managed mode. 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


pkg/grpc/databroker/recordset.go
```diff
@@ -0,0 +1,132 @@
+package databroker
+
+// RecordSetBundle is an index of databroker records by type
+type RecordSetBundle[T IRecord[T]] map[string]RecordSet[T]
+
+// RecordSet is an index of databroker records by their id.
+type RecordSet[T IRecord[T]] map[string]T
+
+// IRecord is an abstract record
+type IRecord[T any] interface {
```
calebdoxsey, line 10: There's already a `Record` type in this package. It's not an interface, it's a struct with an `Any` field.

Since RecordSets and RecordSetBundles don't actually contain Records (rather they contain a new type called IRecord that's not compatible with Record) could we use a different name for these things? 

I'm not sure I understand what we're trying to do with this.

wasaga: removed generics, hopefully the intent is clearer now.

PR #4702, opened 2023-11-02
core/config: refactor file watcher

calebdoxsey:
## Summary
Refactor the file watcher, which had several problems:

1. By using `Add` we were constantly creating new file notification watches on existing files. We will now track what we've watched and use a single method to add/remove watches accordingly.
2. The `FileWatcherSource` always triggers updates even if files don't actually change. We were building a SHA hash of the file contents, but we never actually used them. This has been replaced with building an xx hash and only triggering the update on changes.
3. The `FileWatcherSource` had a single check method that was called whenever the underlying config changed. Previously there was a bug where changes didn't bubble up unless one of the watched files was changed (hence why that logic disappeared). We now handle these two cases separately. We always bubble up underlying config changes.
4. We didn't handle empty strings properly, so we were checking the `""` file frequently.

## Related issues
https://github.com/pomerium/internal/issues/1574

## Checklist

- [x] updated unit tests
- [x] ready for review


internal/fileutil/watcher.go
```diff
@@ -31,50 +30,64 @@ func NewWatcher() *Watcher {
...
-	})
 	watcher.initLocked(ctx)
 
-	if watcher.eventWatcher != nil {
-		if err := watcher.eventWatcher.Add(filePath); err != nil {
-			log.Error(ctx).Msg("fileutil/watcher: failed to watch file with event-based file watcher")
+	var add []string
+	seen := map[string]struct{}{}
+	for _, filePath := range filePaths {
+		if _, ok := watcher.watching[filePath]; !ok {
```
kenjenkins, line 43: Do we know if there are any corner cases where we might still need to add a watch even if the same path already exists in the `watching` set?

If I add a file path to be watched (e.g. `foo`) and then do something like `rm foo && mv bar foo` or `mv foo bar && mv bar foo` will we lose the file watch? (Do we need to handle operations like that?)

calebdoxsey: I'm not sure how the underlying library we use handles this. I will look at the source code.

calebdoxsey: > A path will remain watched if it gets renamed to somewhere else on the same filesystem, but the monitor will get removed if the path gets deleted and re-created, or if it's moved to a different filesystem.

So the fsnotify watcher will miss these events.

It appears that he polling watcher will handle this though.

kenjenkins: I assume we can't just call Remove() and Add() for each path in `filePaths`, as that would leave a brief window where we might miss an event in between?

Would it make sense to add some logic to the "handle events" goroutine below to automatically remove a path from the `watching` set whenever it observes an event where `evt.Op.Has(fsnotify.Remove)`?

calebdoxsey: Should we only using the polling watcher then?

calebdoxsey: Ok I added a test for this and only use the polling watcher. It seems to detect events on newly created files.

kenjenkins: Sorry, I totally misunderstood that it always uses both the event-based watcher and the polling watcher (I thought it was either/or). I think either way should be fine.

kenjenkins (approved):

PR #4703, opened 2023-11-02
core/authorize: cache prepared query building

calebdoxsey:
## Summary
Add a new `RegoCompiler` that can cache prepared queries. 

## Related issues
- https://github.com/pomerium/internal/issues/1572

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


authorize/evaluator/rego_compiler.go
```diff
@@ -0,0 +1,101 @@
...
+// A RegoCompiler compiles rego scripts.
+type RegoCompiler struct {
+	Store        *store.Store
+	policyCache  *lru.Cache[string, rego.PreparedEvalQuery]
+	headersCache *lru.Cache[string, rego.PreparedEvalQuery]
+}
+
+// NewRegoCompiler creates a new RegoCompiler using the given store.
+func NewRegoCompiler(store *store.Store) *RegoCompiler {
+	policyCache, err := lru.New[string, rego.PreparedEvalQuery](10_000)
```
kenjenkins, line 24: Do we have a sense of how much memory a PreparedEvalQuery will typically require?

kenjenkins (approved):

authorize/evaluator/rego_compiler.go
calebdoxsey: I ran a benchmark with a memory profile. I think its about 50 bytes per PreparedEvalQuery.

calebdoxsey: Actually I don't think that's right. My memory profile used 900MB of RAM, but I'm not sure how many queries that worked out to. It was about 50 bytes per object allocated, but likely many objects were allocated per query.

kenjenkins: What do you think about adding a TTL in addition to an entry size limit, using https://pkg.go.dev/github.com/hashicorp/golang-lru/v2@v2.0.7/expirable#NewLRU? Maybe something like 24 hours? I'm thinking of small deployments where the total number of policies is low but there's some steady modification over time — it would be nice to keep this from growing and growing over the longer term.

PR #4716, opened 2023-11-07
core/go: upgrade go to 1.20.10

calebdoxsey:
## Summary
Upgrade go to 1.20.10.

## Related issues
Fixes https://github.com/pomerium/internal/issues/1550

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


kenjenkins (approved):
Just to make sure I understand: we don't need to update the release.yaml workflow on this branch, because it uses `go-version: 1.20.x`, so it will pick the latest Go 1.20 release every time it runs?

calebdoxsey:
Yes I believe it should always use the latest 1.20 release.

calebdoxsey:
This should not be backported. AFAIK v0.24 uses go 1.21

Issue #4197, opened 2023-05-26
pass_identity_headers should default to true


calderonth:
**`pass_identity_headers` defaults to `false`**

Downstream services integrated with Pomerium are supposed to validate the JWT assertion to be really secure against unauthenticated/unauthorized access.
However, the current way to do this means setting the value of `pass_identity_headers` to `true` for every single route/policies.

This feels like a wrong decision that goes against secure by default principle and the behavior should be changed the other way around.

**Describe the solution you'd like**
`pass_identity_headers` default to `true`

**Describe alternatives you've considered**
N/A

**Explain any additional use-cases**
N/A

**Additional context**
N/A

calderonth:
Yes that would be great, would naming the variable `global_pass_identity_headers` be less confusing that re-using an existing variable name?

desimone:
In retrospect, we probably should have named all global variables `global_*`, but to be consistent with how other global/route settings operate we use the same key name.

calebdoxsey:
This option has been added and will be available in the next release.

PR #4720, opened 2023-11-07
core/config: add pass_identity_headers option

calebdoxsey:
## Summary
Add a global config option for `pass_identity_headers`.

## Related issues
- https://github.com/pomerium/pomerium/issues/4197

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/grpc/config/config.proto
```diff
@@ -94,7 +94,8 @@ message Route {
   // AuthorizationHeaderMode set_authorization_header = 54;
 
   bool preserve_host_header = 24;
-  bool pass_identity_headers = 25;
+  bool deprecated_pass_identity_headers = 25;
```
kenjenkins, line 97: What's the history behind the existing pass_identity_headers field?

Is it backwards incompatible to change a proto field from required to optional?

calebdoxsey: I wasn't sure if it was backwards compatible. Looking at how optional works, I think it should be ok to modify the existing field, so I will remove this one.

pkg/grpc/config/config.proto
```diff
@@ -235,6 +235,7 @@ message Settings {
   optional string logo_url = 89;
   optional string favicon_url = 90;
   optional string error_message_first_paragraph = 91;
+  optional bool pass_identity_headers = 117;
```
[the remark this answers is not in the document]
calebdoxsey: I updated it.

kenjenkins (approved):
Thanks! Looks good to me.

PR #4722, opened 2023-11-08
authorize: build evaluators cache in parallel

wasaga:
## Summary

Builds Rego evaluators in parallel, which makes sense for large route counts. 
Also moves the parallel building code into separate utility package.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/databroker/config_source.go
```diff
@@ -125,30 +124,31 @@ func (src *ConfigSource) buildNewConfigLocked(ctx context.Context, cfg *config.C
...
-			return nil
-		}
-	}
-
+	var policyBuilders []errgrouputil.BuilderFunc[config.Policy]
 	for _, cfgpb := range src.dbConfigs {
 		for _, routepb := range cfgpb.GetRoutes() {
-			builders = append(builders, buildPolicy(len(builders), routepb))
+			policyBuilders = append(policyBuilders, func(ctx context.Context) (*config.Policy, error) {
+				p, err := src.buildPolicyFromProto(ctx, routepb)
```
calebdoxsey, line 131: `routepb` will always end up as the last `route`. I believe you need to alias `routepb`:

```
for _, routepb := range cfgpb.GetRoutes() {
    routepb := routepb
```

calebdoxsey (changes requested):

authorize/evaluator/evaluator.go
```diff
@@ -157,15 +167,35 @@ func getOrCreatePolicyEvaluators(
...
+				id:        id,
+				evaluator: evaluator,
+			}, nil
+		})
+	}
+
+	evals, errs := errgrouputil.Build(ctx, builders...)
+	if len(errs) > 0 {
+		for _, err := range errs {
+			log.Error(ctx).Msg(err.Error())
```
kenjenkins, line 185: If a single error will fail this operation, I wonder if we should take advantage of the error handling behavior built in to errgroup.Group, and have Build() return the first error encountered?

internal/errgrouputil/builder.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// BuilderFunc is a function that builds a value of type T
+type BuilderFunc[T any] func(ctx context.Context) (*T, error)
+
+// Build builds a slice of values of type T using the provided builders concurrently
+// and returns the results and any errors.
+func Build[T any](
+	ctx context.Context,
+	builders ...BuilderFunc[T],
```
kenjenkins, line 20: I'm not entirely sure about this API. I'm wondering if there's an alternative where we wouldn't need to allocate this entire slice of functions up front.

Have you considered something more like this?
```go
func Build[I, O any](ctx context.Context, f func(I) (O, error))
```

internal/errgrouputil/builder.go
```diff
@@ -0,0 +1,54 @@
...
+) ([]*T, []error) {
+	eg, ctx := errgroup.WithContext(ctx)
+	eg.SetLimit(runtime.NumCPU()/2 + 1)
+
+	results := make([]*T, len(builders))
+	errors := make([]error, len(builders))
+
+	fn := func(i int) func() error {
+		return func() error {
+			result, err := builders[i](ctx)
```
kenjenkins, line 30: I'm not sure about this context propagation either. Would it be better to check for cancellation here (or around line 41?), and return early if cancelled, without the builder function needing to know about the errgroup context?

authorize/evaluator/evaluator.go
wasaga: that would be behaviour in authorize, but not in databroker package, which is why we return results and errors separately

calebdoxsey (approved):

internal/errgrouputil/builder.go
wasaga: agree its not the most elegant interface. i primarily wanted to concentrate the logic of concurrent building (and the magic formula of SetLimit(runtime.NumCPU()/2+1)) somewhere in one place for the config building speedup in v0.24

we can certainly iterate on a design that looks cleaner, as we still seem to have a few places where we can build heavy configs in parallel.

PR #4763, opened 2023-11-24
zero/k8s: deployment manifests

wasaga:
## Summary

Creates manifests to deploy Pomerium Zero into Kubernetes cluster (see README.md)

## Related issues

Implements https://github.com/pomerium/pomerium-zero/issues/1479

## Checklist

- [x] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/k8s/zero/deployment/base.yaml
+++ b/k8s/zero/deployment/base.yaml
@@ -1,9 +1,13 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
 metadata:
   name: pomerium
 spec:
+  serviceName: "pomerium-proxy"
   replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: pomerium-zero
   template:
     spec:
       containers:
--- a/k8s/zero/deployment/base.yaml
+++ b/k8s/zero/deployment/base.yaml
@@ -1,0 +1,11 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: pomerium
+spec:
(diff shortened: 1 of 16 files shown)


k8s/zero/deployment/base.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: apps/v1
+kind: StatefulSet
```
calebdoxsey, line 2: Why `StatefulSet` as opposed to `Deployment`?

wasaga: Mainly because we need PVC where we store piece of configuration that tells us how to connect to the databroker. 

I can try change to Deployment.

calebdoxsey (approved):

k8s/zero/deployment/base.yaml
wasaga: after some more consideration I believe we should keep it a StatefulSet for Zero: 

1. Zero requires storage that is persistent across restarts, that keeps:
 - zero bootstrap config, so that it can know how to connect to Postgres
 - autocert data
2. The only recommended way to have a persistent storage is `StatefulSet`, i.e. see explanation [here](https://cloud.google.com/kubernetes-engine/docs/concepts/persistent-volumes#deployments_vs_statefulsets). 
3. Zero mode operation does not really require multiple replicas. we know Pomerium and Envoy can scale to very high load numbers per single instance by bumping CPU & RAM.

calebdoxsey (approved):

PR #4765, opened 2023-11-27
support both stateful and stateless authenticate

kenjenkins:
## Summary

Update the initialization logic for the authenticate, authorize, and proxy services to automatically select between the stateful authentication flow and the stateless authentication flow, depending on whether Pomerium is configured to use the hosted authenticate service.

## Related issues

- https://github.com/pomerium/pomerium/issues/4819

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/authenticate/handlers.go
+++ b/authenticate/handlers.go
@@ -84,6 +84,7 @@ func (a *Authenticate) mountDashboard(r *mux.Router) {
 		AllowedHeaders:   []string{"*"},
 	})
 	sr.Use(c.Handler)
+	sr.Use(a.RetrieveSession)
 
 	// routes that don't need a session:
 	sr.Path("/sign_out").Handler(httputil.HandlerFunc(a.SignOut))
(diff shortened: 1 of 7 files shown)


authenticate/handlers.go
```diff
@@ -84,14 +84,14 @@ func (a *Authenticate) mountDashboard(r *mux.Router) {
 		AllowedHeaders:   []string{"*"},
 	})
 	sr.Use(c.Handler)
+	sr.Use(a.RetrieveSession)
```
calebdoxsey, line 87: I think `RetrieveSession` triggers authentication? In which case the `signed_out` page wouldn't work properly?

calebdoxsey: Ok maybe it's just the `VerifySession` that does that.

calebdoxsey (approved):

authenticate/handlers.go
kenjenkins: Yes, I think it's `VerifySession` that will trigger authentication. I think `RetrieveSession` will only populate some values into the request context: https://github.com/pomerium/pomerium/blob/a771b82a7246ccab3d40dc0fe9d49ce683367ca1/internal/sessions/middleware.go#L25-L28

Let me see if I can add a test case to verify this behavior.

kenjenkins: OK, I've added a new test case in [authenticate/handlers_test.go](https://github.com/pomerium/pomerium/pull/4765/files#diff-05242e92555eb106ddb7a54833cc89eacb25138dc4f36d52d983b48f95168683).

PR #4766, opened 2023-11-28
core/go: use max procs

calebdoxsey:
## Summary
Use `automaxprocs` to set `GOMAXPROCS` to the cgroup limits for containers. Also use `GOMAXPROCS` instead of `runtime.NumCPU`.

## Related issues
- https://github.com/pomerium/internal/issues/1602
- https://github.com/pomerium/internal/issues/1603

## Checklist

- [x] updated unit tests
- [x] ready for review


internal/errgrouputil/builder.go
```diff
@@ -20,7 +20,7 @@ func Build[T any](
 	builders ...BuilderFunc[T],
 ) ([]*T, []error) {
 	eg, ctx := errgroup.WithContext(ctx)
-	eg.SetLimit(runtime.NumCPU()/2 + 1)
+	eg.SetLimit(runtime.GOMAXPROCS(0)/2 + 1)
```
kenjenkins, line 23: Should this consider both GOMAXPROCS and NumCPU, and base the limit on whichever is lower?

kenjenkins (approved):

internal/errgrouputil/builder.go
calebdoxsey: I'm not sure what the purpose of this was. It sounds like it was put in place to avoid CPU starvation, but Go has had a pre-emptive scheduler for a while now, so starvation shouldn't be an issue.

@wasaga what do you think it should do?

PR #4776, opened 2023-12-01
metrics: add linear probabilistic counter

wasaga:
## Summary

Adds an utility package with linear probabilistic counter, to be used to estimate session statistics 
while only occupying a limited storage space.

## Related issues

Related: https://github.com/pomerium/pomerium-zero/issues/821

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -108,6 +108,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sts v1.22.0 // indirect
 	github.com/aws/smithy-go v1.16.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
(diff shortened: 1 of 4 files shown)


calebdoxsey:
This is just `go.mod` changes?

calebdoxsey (approved):
Code LGTM but I'm not sure what it's for. For counting sessions can't we just do a DB query?

wasaga:
in order to measure _monthly active users_ you need keep track of all the unique user IDs within a specific period of time (a month, for MAU).

if you just count sessions, say, every hour, that would be a different kind of metric, also useful, but you cannot convert it into a MAU.

PR #4810, opened 2023-12-04
zero/telemetry: calculate DAU and MAU

wasaga:
## Summary

When running with Zero, in managed mode, periodically calculate DAU and MAU.
Note this PR does not include code to push those stats anywhere. 

## Related issues

Related: https://github.com/pomerium/pomerium-zero/issues/821

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/zero/controller/controller.go
```diff
@@ -97,3 +99,23 @@ func (c *controller) runPomeriumCore(ctx context.Context) error {
 func (c *controller) runConnect(ctx context.Context) error {
 	return c.api.Connect(ctx)
 }
+
+func (c *controller) runAnalytics() func(ctx context.Context) error {
+	disable := false
+	return func(ctx context.Context) error {
+		if disable {
```
kenjenkins, line 106: When would this function be invoked more than once? I see only one call to the `runFn` passed to `run()` (on line 82), but I don't see any surrounding loop. Is there some other invocation?

internal/zero/analytics/storage.go
```diff
@@ -0,0 +1,116 @@
...
+const (
+	dataKey      = "data"
+	lastResetKey = "last_reset"
+)
+
+// ToAny marshals a MetricState into an anypb.Any
+func (r *MetricState) ToAny() *anypb.Any {
+	return protoutil.NewAny(&structpb.Struct{
+		Fields: map[string]*structpb.Value{
+			dataKey:      structpb.NewStringValue(base64.StdEncoding.EncodeToString(r.Data)),
```
kenjenkins, line 77: Do protobuf strings not support arbitrary data?

More generally, is there a reason to prefer an ad-hoc Struct proto to a defined message type? Do we anticipate needing to evolve this data type in the future?

internal/zero/analytics/sessions.go
```diff
@@ -0,0 +1,52 @@
...
+
+var (
+	sessionTypeURL = protoutil.GetTypeURL(new(session.Session))
+)
+
+// CurrentUsers returns a list of users active within the current UTC day
+func CurrentUsers(
+	ctx context.Context,
+	client databroker.DataBrokerServiceClient,
+) ([]string, error) {
```
kenjenkins, line 21: I'm wondering if it would be worthwhile to avoid keeping all user IDs in memory while updating the counters? You could imagine an alternative design where this method accepts some callback that is invoked per user ID. Although I don't have a clear sense of what user/session volume this needs to support.

Actually, doesn't the identity manager keep all current user sessions in memory already? Do you think it would be any simpler to integrate analytics reporting with the identity manager code rather than adding a totally separate syncer here?

internal/zero/analytics/sessions.go
```diff
@@ -0,0 +1,52 @@
...
+		if s.UserId == "" { // session creation is in progress
+			continue
+		}
+		if s.AccessedAt == nil {
+			continue
+		}
+		if s.AccessedAt.AsTime().Before(threshold) {
+			continue
+		}
+		users = append(users, s.UserId)
```
kenjenkins, line 48: Each user may have many active sessions; would it make sense to de-duplicate user IDs here?

internal/zero/analytics/collector.go
```diff
@@ -0,0 +1,100 @@
...
+	users, err := CurrentUsers(ctx, c.client)
+	if err != nil {
+		return fmt.Errorf("failed to get current users: %w", err)
+	}
+
+	now := time.Now()
+	for key, counter := range c.counters {
+		updated := counter.Update(users, now)
+		if !updated {
+			log.Ctx(ctx).Debug().Msgf("metric %s not changed: %d", key, counter.Count())
```
kenjenkins, line 83: Should "not changed" really be "not reset" in this log message?

Should this still call SaveMetricState() when the counter is updated but not reset?

internal/zero/analytics/storage.go
wasaga: > Do protobuf strings not support arbitrary data?
doesn't seem like it https://protobuf.dev/reference/protobuf/google.protobuf/#value

> is there a reason to prefer an ad-hoc Struct proto to a defined message type? 

I can try that, but then it'll be not in `internal/zero`

calebdoxsey: In general, for the databroker, I kinda wish all the existing types were structs. The reason why is that to unmarshal an `Any` type, you need that type's protobuf definition to be imported. This makes it much harder for consumers of Databroker data to work with arbitrary data.

internal/zero/analytics/sessions.go
wasaga: > Do you think it would be any simpler to integrate analytics reporting with the identity manager code

there might be benefits for doing that. if we move it there, it would be running for any kind of deployment and not just zero. let me look into that. 

the zero control loop can then run counter syncer and report the changes to the Zero. the cloud interaction would continue being severed from rest of the core.

wasaga: if we move it to identity_manager, this is where it can be hooked up to: 
https://github.com/pomerium/pomerium/blob/3b2bdd059a02b20654d19e2e9b04a3359d0e28cc/internal/identity/manager/manager.go#L389

the counter logic would have to maintain an additional control loop to periodically store the counters.

wasaga: after some prototyping I would prefer to leave it here for now.

internal/zero/analytics/activeusers.go
```diff
@@ -0,0 +1,78 @@
+package analytics
+
+import (
+	"time"
+
+	"github.com/pomerium/pomerium/pkg/counter"
+)
+
+const (
+	activeUsersCap = 10_000
```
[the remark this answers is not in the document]
wasaga: added

kenjenkins (approved):

PR #4853, opened 2023-12-12
config: no longer stub out HPKE public key fetch

kenjenkins:
## Summary

This partially reverts commit a1388592d8627368ef9421c98d43d500e20beb37.

Fetching the authenticate service HPKE public key is required only for the stateless authentication flow. Now that Pomerium will instead use the older (stateful) authentication flow when configured for a self-hosted authenticate service, this logic shouldn't be needed at all.

Removing this logic should also make it easier to test against a local instance of the hosted authenticate service.

## Related issues

- https://github.com/pomerium/internal/issues/1628

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


calebdoxsey (approved):
Because the typical flow will be the stateful one, which doesn't use HPKE, this makes sense.

Merely creating the HPKE key fetcher won't initiate an HTTP request right?

PR #4858, opened 2023-12-15
core/authenticate: refactor identity authenticators to initiate redirect

calebdoxsey:
## Summary
Cognito uses a `logout_uri` parameter on their `/logout` endpoint as a return URL after logging a user out. Unfortunately this has two limitations:

1. The `logout_uri` must be whitelisted ahead of time.
2. The `logout_uri` does not support arbitrary query string parameters.

As a consequence if the proxy service initiates a logout with a redirect uri to return to, there's not currently a way to support this in Cognito.

These changes refactor the `Authenticator` so that instead of having `GetSignInURL` and `GetSignOutURL` they have `SignIn` and `SignOut` and initiate the redirect themselves. In this way we can also write a cookie for the redirect URL and after logging the user out, Cognito will redirect to the `/signed-out` page where we can read this cookie and initiate the redirect to the final redirect URL.

All providers will now receive the `authenticateSignedOutURL` separately from the `redirectToURL`. The previous logic would default `redirectToURL` to `authenticateSignedOutURL` when not specified by the incoming request or by the configuration `signout_redirect_url` option. This logic is preserved for all providers, except for Cognito, where it will now always redirect to the `authenticateSignedOutURL`, but support was added to the `/signed-out` page to do another redirect. This makes it easier to whitelist the URL in Cognito. We should update the Cognito documentation to mention that:

* `https://{AUTHENTICATE_DOMAIN}/.pomerium/signed-out` should be added to the sign out URL whitelist

The only user-facing change here is for Cognito. Other providers should work exactly as they did before.

## Related issues
- https://github.com/pomerium/pomerium-zero/issues/1385
 

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/httputil/signedout.go
```diff
@@ -0,0 +1,26 @@
...
+	http.SetCookie(w, cookie)
+	return cookie.Value, true
+}
+
+// SetSignedOutRedirectURICookie sets the redirect uri cookie for the signed-out page.
+func SetSignedOutRedirectURICookie(w http.ResponseWriter, redirectURI string) {
+	http.SetCookie(w, &http.Cookie{
+		Name:   signedOutRedirectURICookieName,
+		Value:  redirectURI,
+		MaxAge: 5 * 60,
```
kenjenkins, line 24: Should we set `HttpOnly: true` and/or `Secure: true` here too?

(optional)

internal/identity/oidc/cognito/cognito_test.go
```diff
@@ -53,9 +53,11 @@ func TestProvider(t *testing.T) {
...
 
-	t.Run("GetSignOutURL", func(t *testing.T) {
-		signOutURL, err := p.GetSignOutURL("", "https://www.example.com?a=b")
+	t.Run("SignOut", func(t *testing.T) {
+		w := httptest.NewRecorder()
+		r := httptest.NewRequest(http.MethodGet, "https://authenticate.example.com/.pomerium/sign_out", nil)
+		err := p.SignOut(w, r, "", "https://authenticate.example.com/.pomerium/signed_out", "https://www.example.com?a=b")
 		assert.NoError(t, err)
-		assert.Equal(t, srv.URL+"/logout?client_id=CLIENT_ID&logout_uri=https%3A%2F%2Fwww.example.com%3Fa%3Db", signOutURL)
+		assert.Equal(t, srv.URL+"/logout?client_id=CLIENT_ID&logout_uri=https%3A%2F%2Fauthenticate.example.com%2F.pomerium%2Fsigned_out", w.Header().Get("Location"))
```
kenjenkins, line 61: What do you think about adding an additional assertion here that the redirect cookie is set?

kenjenkins (approved):

Issue #4892, opened 2024-01-02
Identity manager expects to be able to read user ID from deleted session records


kenjenkins:
## What happened?

While working on debugging another issue, I noticed an unexpected behavior related to the (not-yet-released) stateful authentication flow changes (https://github.com/pomerium/pomerium/issues/4819): some time after signing out of a Pomerium session, Pomerium would attempt to use the (now revoked) OAuth2 token from that session to try to refresh the user info.

## What did you expect to happen?

After signing out of Pomerium and ending an underlying OAuth2 session with the configured identity provider, Pomerium should no longer attempt to use the access token from that OAuth2 session.

## How'd it happen?

1. Run a build of Pomerium from the main branch, configured with a self-hosted authenticate service.
2. Sign in at a Pomerium route and then sign back out.
3. Wait 10 minutes.

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`):
pomerium: v0.23.0-247-g18717a96+18717a96
envoy: 1.28.0+cf3c0b96f4f3258bda685a245620b8d4d5c34eb680716d2bb82cffb80d4085cc
- Server Operating System/Architecture/Cloud: macOS

## What's your config.yaml?

```config.yaml
authenticate_service_url: https://authenticate.localhost.pomerium.io

idp_provider: google
idp_client_id: [redacted]
idp_client_secret: [redacted]

routes:
  - from: https://verify.localhost.pomerium.io
    to: http://localhost:8000
    allow_any_authenticated_user: true
    pass_identity_headers: true

...
```

## What did you see in the logs?

```logs
{"level":"info","config_file_source":".config.basic.yaml","bootstrap":true,"service":"identity_manager","user_id":"111124447713896378131","time":"2024-01-02T12:38:08-08:00","message":"refreshing user"}
{"level":"error","config_file_source":".config.basic.yaml","bootstrap":true,"service":"identity_manager","error":"identity/oidc: user info endpoint: 401 Unauthorized: {\"error\":\"invalid_request\",\"error_description\":\"Invalid Credentials\"}","user_id":"111124447713896378131","session_id":"c943e916-3ebc-4592-b41d-a24421249939","time":"2024-01-02T12:38:09-08:00","message":"failed to update user info, deleting session"}
{"level":"info","address":"127.0.0.1:57958","time":"2024-01-02T12:38:09-08:00","message":"grpc: dialing"}
[...]

PR #4893, opened 2024-01-02
authenticateflow: change how sessions are deleted

kenjenkins:
## Summary

The identity manager expects to be able to read session ID and user ID from any deleted databroker session records. The `session.Delete()` wrapper method is not compatible with this expectation, as it calls `Put()` with a record containing an empty session. The stateful authentication flow currently calls `session.Delete()` from its `RevokeSession()` method.

The result is that the identity manager will not correctly track sessions deleted by the the stateful authentication flow, and will still try to use them during session refresh and user info refresh.

Instead, let's change the stateful authentication flow `RevokeSession()` method to perform deletions in a way that is compatible with the current identity manager code. That is, include the existing session data in the `Put()` call to delete the revoked session.

## Related issues

https://github.com/pomerium/pomerium/issues/4892

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/internal/authenticateflow/stateful.go
+++ b/internal/authenticateflow/stateful.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pomerium/pomerium/pkg/grpc/databroker"
 	"github.com/pomerium/pomerium/pkg/grpc/session"
 	"github.com/pomerium/pomerium/pkg/grpc/user"
+	"github.com/pomerium/pomerium/pkg/grpcutil"
 )
 
 // Stateful implements the stateful authentication flow. In this flow, the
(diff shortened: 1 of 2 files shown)


internal/authenticateflow/stateful.go
```diff
@@ -261,15 +262,44 @@ func (s *Stateful) RevokeSession(
 		return ""
 	}
 
+	// Note: session.Delete() cannot be used safely, because the identity
+	// manager expects to be able to read both session ID and user ID from
+	// deleted session records. Instead, we match the behavior used in the
+	// identity manager itself: fetch the existing databroker session record,
+	// explicitly set the DeletedAt timestamp, and Put() that record back.
+
+	res, err := s.dataBrokerClient.Get(ctx, &databroker.GetRequest{
```
calebdoxsey, line 271: Can you use `Patch` instead?

kenjenkins: That's a good idea, but I think we can't, because the `deleted_at` field is part of the databroker `Record` message itself, rather than the data it contains (e.g. the `Session` message).

It might be worth a bigger conversation at some point about whether we should modify the `Patch` method to make it be able to delete a record.

calebdoxsey (approved):

PR #4895, opened 2024-01-04
core/opa: update for rego 1.0

calebdoxsey:
## Summary
Update `rego` code to be compatible with v1.0. The major change is the requirement that a rule:

```rego
x := y {
  some condition
}
```

Now requires an `if`:

```rego
x := y if {
  some condition
}
```

## Related issues
- https://github.com/pomerium/internal/issues/1619
 

## Checklist
- [x] updated unit tests
- [x] ready for review


calebdoxsey:
(this should not be merged until after v0.25)

PR #4897, opened 2024-01-05
core/envoy: upgrade to v1.25.10

calebdoxsey:
## Summary
Upgrade envoy to v1.25.10. We don't have darwin binaries for this version so only upgrade for linux.

## Related issues
Fixes https://github.com/pomerium/internal/issues/1551

## Checklist
- [x] ready for review


wasaga (changes requested):
holding it off until we cut v0.25 version branch

Issue #4757, opened 2023-11-17
Flaky test in internal/autocert package


kenjenkins:
It looks like TestConfig has failed in two recent test runs:

https://github.com/pomerium/pomerium/actions/runs/6880759340/job/18715639777
```
--- FAIL: TestConfig (1.38s)
    testing.go:1225: TempDir RemoveAll cleanup: unlinkat /var/folders/3s/vfzpb5r51gs6y328rmlgzm7c0000gn/T/TestConfig1921913482/001: directory not empty
{"level":"info","time":"2023-11-15T17:37:03Z","logger":"obtain","msg":"acquiring lock","service":"autocert","identifier":"from.example.com"}
{"level":"info","service":"autocert-manager","addr":"127.0.0.1:49339","time":"2023-11-15T17:37:03Z","message":"starting http redirect server"}
{"level":"info","time":"2023-11-15T17:37:03Z","logger":"maintenance","msg":"started background certificate maintenance","service":"autocert","cache":"0xc000264580"}
{"level":"info","time":"2023-11-15T17:37:03Z","logger":"obtain","msg":"lock acquired","service":"autocert","identifier":"from.example.com"}
{"level":"info","time":"2023-11-15T17:37:03Z","logger":"obtain","msg":"obtaining certificate","service":"autocert","identifier":"from.example.com"}
{"level":"warn","time":"2023-11-15T17:37:03Z","logger":"acme_client","msg":"HTTP request failed; retrying","service":"autocert","url":"http://127.0.0.1:49313/acme/new-nonce","error":"performing request: Head \"[http://127.0.0.1:49313/acme/new-nonce\](http://127.0.0.1:49313/acme/new-nonce/)": context canceled"}
{"level":"error","time":"2023-11-15T17:37:03Z","logger":"obtain","msg":"could not get certificate from issuer","service":"autocert","identifier":"from.example.com","issuer":"127.0.0.1:49313-acme-directory","error":"registering account [mailto:pomerium-test@example.com] with server: fetching new nonce from server: context canceled"}
{"level":"info","time":"2023-11-15T17:37:03Z","logger":"obtain","msg":"releasing lock","service":"autocert","identifier":"from.example.com"}
[...]

PR #4899, opened 2024-01-05
core/autocert: fix flaky test

calebdoxsey:
## Summary
Revert back to manual temporary directory creating and removal in the autocert test. Apparently sometimes a background process will continue writing files even after the test completes. `t.TempDir` causes the test to fail when this happens. Doing it manually if `RemoveAll` fails we can ignore the error.

I was not able to reproduce the flakiness despite running the test with `-count 100`, but it happened on github actions once.

## Related issues
Fixes https://github.com/pomerium/pomerium/issues/4757

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/autocert/manager_test.go
+++ b/internal/autocert/manager_test.go
@@ -18,6 +18,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/autocert/manager_test.go
```diff
@@ -215,12 +217,12 @@ func TestConfig(t *testing.T) {
 
 	mockACME = newMockACME(ca, srv)
 
-	tmpdir := t.TempDir()
+	tmpdir := filepath.Join(os.TempDir(), uuid.New().String())
```
kenjenkins, line 220: Would you mind adding a comment here to mention the issue with t.TempDir(), maybe linking to https://github.com/pomerium/pomerium/issues/4757?

kenjenkins (approved):
Thanks for investigating!

PR #4916, opened 2024-01-29
core/ppl: add groups criterion

calebdoxsey:
## Summary
Add the `groups` criterion for PPL. It references data that is only populated from the Enterprise Console.

## Related issues
- https://github.com/pomerium/internal/issues/1673

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/policy/criteria/groups_test.go
+++ b/pkg/policy/criteria/groups_test.go
@@ -5,36 +5,44 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/pomerium/datasource/pkg/directory"
 	"github.com/pomerium/pomerium/pkg/grpc/databroker"
 	"github.com/pomerium/pomerium/pkg/grpc/session"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/policy/criteria/groups_test.go
```diff
@@ -0,0 +1,48 @@
...
+					Id:     "session1",
+					UserId: "user1",
+				}),
+				makeStructRecord(directory.UserRecordType, "user1", map[string]any{
+					"group_ids": []any{"group1"},
+				}),
+			},
+			Input{Session: InputSession{ID: "session1"}})
+		require.NoError(t, err)
+		require.Equal(t, A{true, A{ReasonGroupsOK}, M{}}, res["allow"])
```
kenjenkins, line 45: Would it make sense to another test case where the user is authenticated but the groups criterion results in a Deny decision because the user is not a member of either allowed group?

kenjenkins (approved):

Issue #4951, opened 2024-02-01
core/authorize: denied response is not an error page anymore


calebdoxsey:
## What happened?
Using the `main` branch, the authorize service is no longer returning an error page in the browser for access denied. Instead there's a JSON object returned:

```json
{
  "error": "Forbidden",
  "request_id": "8736001c-c6db-4386-855c-441d2ae9b4b7"
}
```

## What did you expect to happen?
To see an HTML page.

## Additional context
This broke with #4904. The issue is that browsers send `*/*`. So `application/json` matches.

ssveta7ak:
Fixed. Checked on 
Core v0.25.0-53-gd405a53b + d405a53b
Enterprise v0.25.0 + 13e803f7 + 2024-02-29T19:12:42Z

Issue #4912, opened 2024-01-22
robots.txt Disallow / for public routes


stuba:
**robots.txt is always responding with content that disallows crawling.**

I was blaming framework for this, but now it seems that Pomerium replies to all robots.txt requests automatically. I did read that during 0.11 there should have been fix for this https://github.com/pomerium/pomerium/pull/1361. 

**When public domain robots.txt should get passed to backend service.**

**Setup rules in pomerium config**

**No other use cases**

If there are any use-cases that would help us understand the use/need/value please share them as they can help us decide on acceptance and prioritization.

**Here is my config for this route**
```
  - name: Default route
    from: https://my_service.url
    to: http://nginx
    preserve_host_header: true
    policy:
      allow:
        and:
          - accept: true

```

kenjenkins:
Hi @stuba,

It looks like currently the logic from https://github.com/pomerium/pomerium/pull/1361 is based on the [allow_public_unauthenticated_access](https://www.pomerium.com/docs/reference/routes/public-access) option. Please try updating your route configuration to set `allow_public_unauthenticated_access` instead of `policy`, something like this:
```yaml
  - name: Default route
    from: https://my_service.url
    to: http://nginx
    preserve_host_header: true
    allow_public_unauthenticated_access: true
```

stuba:
Surely this works. I do not know why I had understood this allow_public_unauthenticated_access is all ready deprecated in flavour of new policy. Thank you for clarifying this!

calebdoxsey:
I agree we should update the logic to also work with PPL.

kenjenkins:
Per Caleb: we could refactor this logic to move it to the authorizer service-- if we get a request for /robots.txt and the request is not authorized, we could directly return the "disallow" response. (This should avoid some of the complication of recognizing equivalent PPL policies.)

calebdoxsey:
The logic around robots.txt handling has been updated to happen in the authorize service so that any policy that allows access without requiring a login should behave the same.

The example route:

```
 - name: Default route
    from: https://my_service.url
    to: http://nginx
    preserve_host_header: true
    policy:
      allow:
        and:
          - accept: true
```

Should forward requests to the backend. Whereas a route with an authorization policy should return 

```
User-agent: *
Disallow: /
```

stuba:
Perfect, thanks!

PR #4952, opened 2024-02-01
core/authorize: result denied improvements

calebdoxsey:
## Summary
Fix the access denied error page. Handle robots.txt in the authorize service.

## Related issues
- #4951 
- #4912 

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/envoyconfig/routes.go
+++ b/config/envoyconfig/routes.go
@@ -68,10 +68,6 @@ func (b *Builder) buildPomeriumHTTPRoutes(
 			b.buildControlPlanePathRoute(options, "/.well-known/pomerium"),
 			b.buildControlPlanePrefixRoute(options, "/.well-known/pomerium/"),
 		)
-		// per #837, only add robots.txt if there are no unauthenticated routes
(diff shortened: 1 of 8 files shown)


config/envoyconfig/routes.go
```diff
@@ -102,6 +98,7 @@ func (b *Builder) buildPomeriumAuthenticateHTTPRoutes(
 			return []*envoy_config_route_v3.Route{
 				b.buildControlPlanePathRoute(options, options.AuthenticateCallbackPath),
 				b.buildControlPlanePathRoute(options, "/"),
+				b.buildControlPlanePathRoute(options, "/robots.txt"),
```
kenjenkins, line 101: Why does the authenticate service need a robots.txt route?

kenjenkins (approved):
Looks good to me.

config/envoyconfig/routes.go
calebdoxsey: Because routes to authenticate bypass the authorize service. Without this exception the robots.txt route will return a 404.

PR #4957, opened 2024-02-08
core/ui: fix page title

calebdoxsey:
## Summary
Set the page title for UI pages. For pages like the dashboard they are set to a title that makes sense for that page. For errors we set the title to something like `403 Forbidden`.

The `index.html` file was converted to a Go template.

## Related issues
- https://github.com/pomerium/internal/issues/1698
 

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/ui/embed.go
+++ b/ui/embed.go
@@ -3,7 +3,7 @@ package ui
 
 import (
 	"bytes"
-	"encoding/json"
+	"html/template"
 	"io"
 	"net/http"
 	"path/filepath"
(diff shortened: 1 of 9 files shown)


nhayfield (approved):

ui/embed.go
```diff
@@ -26,35 +26,44 @@ func ServeFile(w http.ResponseWriter, r *http.Request, filePath string) error {
...
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	bs = bytes.Replace(bs,
-		[]byte("window.POMERIUM_DATA = {}"),
-		append([]byte("window.POMERIUM_DATA = "), jsonData...),
-		1)
+	tpl, err := template.New("").Parse(string(bs))
```
kenjenkins, line 61: (optional)

Would it make sense to parse the template once (during startup), rather than for each call to `ServePage()`?

kenjenkins (approved):
Thanks, looks good!

PR #4961, opened 2024-02-13
connect: add gRPC keep-alive

wasaga:
## Summary

When running in Zero managed mode, Pomerium must maintain a streaming gRPC to the Connect signalling service in the cloud, in order to receive configuration updates from the cloud as they appear. 

It was noticed that sometimes the gRPC connection may be in the `READY` state while the end to end connection to the Connect gRPC server, is in fact, long time gone. 

This PR introduces a client-side keepalive with infrequent 1-minute interval for the Connect service, to ensure we are still maintain gRPC connection.

## Related issues

Fixes https://github.com/pomerium/pomerium-zero/issues/1711

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/zero/api/api.go
+++ b/internal/zero/api/api.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"time"
 
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/keepalive"
+
 	"github.com/pomerium/pomerium/internal/zero/apierror"
 	connect_mux "github.com/pomerium/pomerium/internal/zero/connect-mux"
 	"github.com/pomerium/pomerium/internal/zero/grpcconn"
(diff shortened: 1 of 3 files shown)


internal/zero/api/api.go
```diff
@@ -34,6 +37,13 @@ const (
 	minTelemetryTokenTTL = time.Minute * 5
 )
 
+// see https://github.com/pomerium/pomerium-zero/issues/1711
+var connectClientKeepaliveParams = keepalive.ClientParameters{
+	Time:                time.Minute, // send pings every minute
+	Timeout:             time.Minute, // wait 1 minute for ping ack
+	PermitWithoutStream: false,
```
calebdoxsey, line 44: Shouldn't this be `true`?

> // If true, client sends keepalive pings even with no active RPCs. If false,
> // when there are no active RPCs, Time and Timeout will be ignored and no
> // keepalive pings will be sent.

wasaga: couple reasons: 

1. we are talking to GLB rather then to our server app, and it may not like it (I've seen it shut down the connection if keep-alive is too aggressive, so it definitely has some _unknown_ keep-alive enforcement policy in place); thus the choice of very conservative 1 min interval. 
2. there is a control loop to restart `Subscribe()` if it fails,

calebdoxsey (approved):
> we are talking to GLB rather then to our server app, and it may not like it (I've seen it shut down the connection if keep-alive is too aggressive, so it definitely has some unknown keep-alive enforcement policy in place); thus the choice of very conservative 1 min interval.

Out of scope, but it sounds like gRPC is very unreliable for this service.

PR #4969, opened 2024-02-29
core/zero: fix ticker usage

calebdoxsey:
## Summary
In several places we use a `time.Ticker` based on a `backoff.Backoff`. We create a new `time.Ticker` that fires immediately by using a short duration. Unfortunately this causes excessive CPU usage. There are several ways to fix this:

1. We can use the `backoff.NewTicker` function, which returns a "Ticker" that behaves as expected by using a `time.Timer` instead. This approach does not work if we need to call `Reset` on the `Backoff`.
2. We can use the `backoff.Retry` function with a `backoff.WithContext`. We then don't need a timer or ticker.
3. Instead of a `time.Ticker` we can use a `time.Timer` set to the backoff interval. Since a `time.Timer` is allowed to be set to 0 with no downside, this can handle the case when we want to fire immediately. The benefit over a `time.After` is we can call `.Stop` to collect resources if another case in the select fires.

Option (3) was used for the `retry.Retry` function because the logic around resetting is incompatible with (1) and (2).

## Related issues
Fixes https://github.com/pomerium/internal/issues/1727

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
I should also mention that a call to `Reset` on `Timer` can **only** be done if the `Timer` has already fired, which prevents some ways of solving this problem.

internal/zero/leaser/restart.go
```diff
@@ -68,13 +66,7 @@ func restartContexts(
 			return
 		}
 
-		if time.Since(start) > bo.MaxInterval {
```
wasaga, line 71: I believe we need keep this condition; otherwise `bo.Reset` would never be called.

either the `restarterFn` should take the bo.Reset as a parameter to call on success (like in connect-mux/service.go) or we need keep a current empirical condition that restart was not requested for over `bo.MaxInterval` (indicating a normal operation) 

in practical terms, based on how this function is used up the stack, if we won't do that the following would happen: 
- you set databroker connection string that's incorrect, the `restartFn` would fail continuously and quickly get the back-off interval to max value
- now you set a correct connection string, but because we never reset this back-off, the restart won't happen until after bo.MaxInterval (~1 minute)

internal/zero/leaser/restart.go
```diff
@@ -49,15 +48,14 @@ func restartContexts(
 	contexts chan<- context.Context,
 	restartFn func(context.Context) error,
 ) {
-	bo := backoff.NewExponentialBackOff()
-	bo.MaxElapsedTime = 0 // never stop
+	b := backoff.NewExponentialBackOff()
+	b.MaxElapsedTime = 0 // never stop
 
-	ticker := time.NewTicker(bo.InitialInterval)
+	ticker := backoff.NewTicker(b)
```
wasaga, line 54: maybe let's just change it to time.Timer() and do manual reset on each iteration?

wasaga (changes requested):

internal/zero/leaser/restart.go
calebdoxsey: This doesn't make any sense. You should never compare to the MaxInterval. I don't know why you would want to do that. You end up with a "backoff" that backs off exponentially and then goes back to 0 again in a stair step pattern. Why would you want to do this? Why would you name this "backoff" when it bares no resemblance to what an exponential backoff is supposed to be?

If all of what you say above is true then I don't think I understand the intention of any of this code and don't feel qualified to work on it. (Which is exactly what I've said in code reviews on the zero code and now it's broken and I can't fix it because its too complex)

calebdoxsey:
Sorry forgot I wasn't supposed to close PRs. Re-opening.

wasaga (approved):

PR #5027, opened 2024-03-12
ci: bump Go to 1.21.8 in docker

wasaga:
## Summary

Bump Go in Dockerfiles to 1.21.8 
The GitHub actions are supposed to pick up latest from 1.21 

https://groups.google.com/g/golang-announce/c/5pwGVUPoMbg

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN make yarn
 COPY ./ui/ ./ui/
 RUN make build-ui
 
-FROM golang:1.21.8-bookworm@sha256:ac14cc827536ef1a124cd2f7a03178c3335c1db8ad3807e7fdd57f74096abfa0 as build
+FROM golang:1.21.8-bookworm@sha256:05a9064db595ba2a6aa7c2d48d16ba5872c42583606741c750b0d895e9d0a09d as build
 WORKDIR /go/src/github.com/pomerium/pomerium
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN make yarn
 COPY ./ui/ ./ui/
 RUN make build-ui
 
-FROM golang:1.21.5-bookworm@sha256:1415bb0b25d3bffc0a44dcf9851c20a9f8bbe558095221d931f2e4a4cc3596eb as build
+FROM golang:1.21.8-bookworm@sha256:ac14cc827536ef1a124cd2f7a03178c3335c1db8ad3807e7fdd57f74096abfa0 as build
```
kenjenkins, line 16: I think this is the amd64 image, rather than the multi-platform manifest.

Although I think it shouldn't matter, as these dockerfiles aren't used for the release builds? (Is that right?)

wasaga: right; I backported the script to check https://github.com/pomerium/pomerium/pull/5028 and updated images

kenjenkins: OK thanks!

kenjenkins (approved):

PR #5033, opened 2024-03-20
grpc: add some debug options

wasaga:
## Summary

Adds some gRPC related debug options configurable via setting environment variables:

- `POMERIUM_GRPC_LOG_CONNECTION_STATE` - enables logging gRPC connection state transition. this is significantly less verbose compared to enabling standard gRPC debug logging.
- `POMERIUM_GRPC_CONNECT_KEEPALIVE` - enables gRPC keepalive (on by default).

Also updates some deprecated grpc configuration options.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/internal/featureflags/env.go
+++ b/internal/featureflags/env.go
@@ -20,6 +20,7 @@ var flags = make(map[string]struct{})
 func init() {
 	for _, env := range []string{
 		GRPCLogConnectionState,
+		GRPCConnectDisableKeepalive,
 	} {
 		if _, ok := os.LookupEnv(fmt.Sprintf("POMERIUM_%s", env)); ok {
 			flags[env] = struct{}{}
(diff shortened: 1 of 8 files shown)


calebdoxsey (approved):
I don't understand the reasoning behind the way these changes were made.

1. We already have a configuration mechanism in the `config` package. We should not have multiple ways of doing the same thing.
2. This violates the requirements for Pomerium Zero: "With Pomerium Zero, the one and only source of configuration should be Pomerium Zero cloud."

I assume there was a good reason for this that I'm not aware of.

wasaga:
closing in favour of https://github.com/pomerium/pomerium/pull/5050

PR #5034, opened 2024-03-21
grpc: add databroker keepalive option

wasaga:
## Summary

Various components maintain a streaming `databroker.Sync` call to a databroker, that may potentially be located behind an external load balancer, that are outside of our control. 

The load balancer may quietly drop the gRPC connection to the upstream while keeping a TCP connection live with the downstream (databroker gRPC client). 

This PR adds an experimental option that forces HTTP2 PING frames to be sent to the (remote) databroker gRPC cluster. We cannot use native Go gRPC keepalives as that part of the connection is to the localhost Envoy.

## Related issues

Related: https://github.com/pomerium/internal/issues/1730

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/bootstrap.go
+++ b/config/envoyconfig/bootstrap.go
@@ -219,7 +219,7 @@ func (b *Builder) BuildBootstrapStaticResources(
 				},
 			},
 		},
-		TypedExtensionProtocolOptions: buildTypedExtensionProtocolOptions(nil, upstreamProtocolHTTP2),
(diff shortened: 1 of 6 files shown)


calebdoxsey (approved):

wasaga:
superceded by https://github.com/pomerium/pomerium/pull/5050

PR #5050, opened 2024-04-01
config: add runtime flags

wasaga:
## Summary

Adds `runtime_flags` map[string]bool option that may be used to enable/disable certain default runtime 
options. 

Adds first option to enable gRPC keepalive for the databroker connection (off by default).

## Related issues

Related: https://github.com/pomerium/internal/issues/1730

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


config/custom.go
```diff
@@ -543,6 +543,52 @@ func decodeStringToMapHookFunc() mapstructure.DecodeHookFunc {
...
+func decodeRuntimeFlagsHookFunc() mapstructure.DecodeHookFunc {
+	return mapstructure.DecodeHookFuncType(func(from, to reflect.Type, data any) (any, error) {
+		if to != reflect.TypeOf(RuntimeFlags{}) {
+			return data, nil
+		}
+		if from.Kind() != reflect.Map {
+			return nil, fmt.Errorf("%w: got %s", ErrRuntimeFlagsInvalidValue, from.Kind())
+		}
+
+		src, ok := data.(map[string]interface{})
```
calebdoxsey, line 562:
```suggestion
		src, ok := data.(map[string]any)
```

config/custom.go
```diff
@@ -543,6 +543,52 @@ func decodeStringToMapHookFunc() mapstructure.DecodeHookFunc {
...
+		if !ok {
+			return nil, fmt.Errorf("decoding runtime flags: expected map[string]interface{}, got %T", from.String())
```
calebdoxsey, line 564:
```suggestion
			return nil, fmt.Errorf("decoding runtime flags: expected map[string]any, got %T", from.String())
```

config/runtime_flags.go
```diff
@@ -0,0 +1,26 @@
...
+func runtimeFlag(txt string, def bool) RuntimeFlag {
+	key := RuntimeFlag(txt)
+	defaultRuntimeFlags[key] = def
+	return key
+}
+
+var defaultRuntimeFlags = map[RuntimeFlag]bool{}
+
+func DefaultRuntimeFlags() RuntimeFlags {
+	out := make(RuntimeFlags, len(defaultRuntimeFlags))
```
calebdoxsey, line 21: You can use `golang.org/x/exp/maps.Clone` or `golang.org/x/exp/maps.Copy`.

config/custom.go
```diff
@@ -543,6 +543,52 @@ func decodeStringToMapHookFunc() mapstructure.DecodeHookFunc {
 	})
 }
 
+var (
+	ErrRuntimeFlagUnknown         = fmt.Errorf("decoding runtime flags: unknown flag")
```
calebdoxsey, line 547: Is there a good reason to enforce this? Doesn't this just limit the flexibility of runtime flags? (e.g. we remove one in the future, and now all config that had it fails, we add one an older version of pomerium fails because it doesn't exist there, etc...)

config/envoyconfig/protocols.go
```diff
@@ -41,34 +44,56 @@ var http1ProtocolOptions = &envoy_config_core_v3.Http1ProtocolOptions{
...
+type Keepalive bool
+
 var http2ProtocolOptions = &envoy_config_core_v3.Http2ProtocolOptions{
 	AllowConnect:                true,
 	MaxConcurrentStreams:        wrapperspb.UInt32(maxConcurrentStreams),
 	InitialStreamWindowSize:     wrapperspb.UInt32(initialStreamWindowSizeLimit),
 	InitialConnectionWindowSize: wrapperspb.UInt32(initialConnectionWindowSizeLimit),
 }
 
+var http2ProtocolOptionsWithKeepalive = &envoy_config_core_v3.Http2ProtocolOptions{
```
calebdoxsey, line 57: Can we derive this in a different way so that all the other default settings are only defined in one place?

config/envoyconfig/protocols.go
```diff
@@ -41,34 +44,56 @@ var http1ProtocolOptions = &envoy_config_core_v3.Http1ProtocolOptions{
 	},
 }
 
+// Keepalive is a type to enable or disable keepalive
+type Keepalive bool
```
calebdoxsey, line 48: Why a typed boolean and not just a normal boolean?

config/custom.go
kenjenkins, line 1: What benefit do we gain from a custom decode hook, compared with adding a map directly to the Options struct, like this?
```go
	RuntimeFlags map[RuntimeFlag]bool `mapstructure:"runtime_flags" yaml: "runtime_flags,omitempty"`
```

config/envoyconfig/protocols.go
wasaga: I always struggle to remember what random booleans mean in various functions. Isn't Keepalive(true) more readable ?

config/custom.go
wasaga: I believe runtime flags are not covered by any version compatibility contract. 

I don't expect users to flip those flags unless explicitly instructed. 

It's like a command line flags - if you provide a flag to a program version that doesn't support it the customary behaviour is to quit with error, not to ignore. 

Note this only applies to flags supplied via config file and not via the protobuf.

config/custom.go
wasaga: We don't overwrite defaults

config/envoyconfig/protocols.go
calebdoxsey: Sorry I don't understand this. I feel like we've had this conversation before. I'll probably keep leaving this comment because it will never make sense to me. Sorry.

I guess we do this all over the place so 🤷 . Sorry for not offering constructive feedback. I cannot defend my perspective.

config/custom.go
calebdoxsey: My understanding is that currently our default is to warn on unknown config fields. I can add a follow up issue to error out.

This will bite us, but ok.

wasaga: we can keep current warning behaviour for consistency.

config/custom.go
wasaga: it tried to do few things:

- error on unknown keys: we can turn it into warning, and it'll have to happen outside of the hook then anyway
- erroring if something that's not boolean
- ensure default values from `defaultRuntimeFlags` are always present. 
 
we parse config via Viper 
https://github.com/pomerium/pomerium/blob/be9c14fc046ee83dd33c1de8c9b28cd58c5b5c5c/config/options.go#L389

that in turn calls mapstructure. 
https://github.com/spf13/viper/blob/947eb59667db361a0c00a5f0b809b41e5944a8b1/viper.go#L1154C1-L1169C1

I had a feeling viper previously configured mapstructure to zero out maps but it doesn't seem to be the case, maps are merged. 

--- 

in short, given that we don't want to default on unknown keys, and by default viper configures WeaklyTypedInput: true, and doesn't seem to override maps, we can get rid of this custom hook and just add a warning about inexistent keys.

config/options.go
```diff
@@ -328,6 +330,13 @@ var defaultOptions = Options{
 	EnvoyAdminAccessLogPath:             os.DevNull,
 	EnvoyAdminProfilePath:               os.DevNull,
 	ProgrammaticRedirectDomainWhitelist: []string{"localhost"},
+
+	RuntimeFlags: DefaultRuntimeFlags(),
```
kenjenkins, line 334: I think you were right that there is some subtlety around default flags. When I tried this out locally, I did not observe the default flag value in the RuntimeFlags map after config parsing.

I think maybe the initialization of RuntimeFlags should be moved into the NewDefaultOptions() function, so we get a distinct map instance each time NewDefultOptions() is called?

kenjenkins (approved):
Mostly looks good to me, just a couple questions.

kenjenkins (approved):
Thanks for adding a test!

PR #5058, opened 2024-04-09
core/config: disable gRPC ingress when address is the empty string

calebdoxsey:
## Summary
Currently if the `grpc_address` is set to the empty string:

```yaml
grpc_address: ""
```

We will start the `grpc-ingress` on port `80`. With these changes we will instead disable the `grpc-ingress`. The databroker and authorize service will only listen on the gRPC port, which is randomly chosen on start, and only bound to the local ip.

## Related issues
- https://github.com/pomerium/pomerium-zero/issues/1987

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/databroker/config_source.go
+++ b/internal/databroker/config_source.go
@@ -272,15 +272,8 @@ func (src *ConfigSource) runUpdater(cfg *config.Config) {
 	}, databroker.WithTypeURL(grpcutil.GetTypeURL(new(configpb.Config))),
 		databroker.WithFastForward())
 	go func() {
-		var databrokerURLs []string
-		urls, _ := cfg.Options.GetDataBrokerURLs()
-		for _, url := range urls {
-			databrokerURLs = append(databrokerURLs, url.String())
-		}
-
 		log.Debug(ctx).
 			Str("outbound_port", cfg.OutboundPort).
-			Strs("databroker_urls", databrokerURLs).
 			Msg("config: starting databroker config source syncer")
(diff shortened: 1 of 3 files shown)


internal/databroker/config_source.go
```diff
@@ -272,15 +272,8 @@ func (src *ConfigSource) runUpdater(cfg *config.Config) {
 	}, databroker.WithTypeURL(grpcutil.GetTypeURL(new(configpb.Config))),
 		databroker.WithFastForward())
 	go func() {
-		var databrokerURLs []string
-		urls, _ := cfg.Options.GetDataBrokerURLs()
-		for _, url := range urls {
-			databrokerURLs = append(databrokerURLs, url.String())
-		}
```
kenjenkins, line 279: I'm not sure I understand this part. Is this logging no longer correct if the gRPC listener is disabled?

kenjenkins (approved):
Looks good to me, just one question.

internal/databroker/config_source.go
calebdoxsey: It will show:

```json
{
  "level":"debug",
  "outbound_port":"51313",
  "databroker_urls":["http://127.0.0.1:5443"],
  "time":"2024-04-10T11:31:30-06:00",
  "message":"config: starting databroker config source syncer"
}
```

But that's not actually correct. I don't think this log was particularly useful so I thought it made more sense to just remove it rather than fixing it.

PR #5059, opened 2024-04-10
healthcheck: add common package, zero reporter and first xds check

wasaga:
## Summary

1. Adds package `pkg/health` where health checks are registered and reported.
2. This package is made global so that its easier to call from various places. 
3. By default, the reporter is no-op. 
4. In Zero managed mode, the reporter is added that sends health check results back to 
telemetry endpoint. 

## Related issues

- https://github.com/pomerium/pomerium-zero/issues/2102
- https://github.com/pomerium/pomerium-zero/issues/2107

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


pkg/health/provider.go
```diff
@@ -0,0 +1,64 @@
...
+	Value string
+}
+
+// StrAttr creates a new string attribute
+func StrAttr(key, value string) Attr {
+	return Attr{Key: key, Value: value}
+}
+
+// ReportOK reports that a check was successful
+func ReportOK(check Check, attributes ...Attr) {
```
calebdoxsey, line 19: Is the reason we didn't use https://pkg.go.dev/go.opentelemetry.io/otel@v1.25.0/attribute because we're afraid of future incompatible changes?

wasaga: just not to entangle the interface and implementation together. 
also, theoretically, there might be some other kind of health check reporter for OSS or Enterprise in the future.

PR #5068, opened 2024-04-12
healthcheck: only report transitions

wasaga:
## Summary

For healthchecks we are only interested in state transitions. 
This PR adds a wrapper around default Provider that filters out repetitive calls.

## Related issues

Implements: https://github.com/pomerium/pomerium-zero/issues/2174

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


kenjenkins (commented):
What impact will this have on the health check timestamps? Will this cause some health checks to appear staler and staler over time if there is no change in their status?

wasaga:
> Will this cause some health checks to appear staler and staler over time if there is no change in their status?

yes. note that a lot of health checks are only triggered on configuration update. however there might be some control loops that are restarted because of some external condition; this would prevent from health checks that are triggered within them to be reported multiple time.

pkg/health/deduplicate.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (r *record) equalError(other *record) bool {
+	if r.err == nil || other.err == nil {
+		return r.err == other.err
+	}
+	return *r.err == *other.err
+}
+
+func (r *record) equalAttrs(other *record) bool {
```
calebdoxsey, line 50: I think this can be done with https://pkg.go.dev/golang.org/x/exp/maps#Equal

calebdoxsey (approved):

PR #5084, opened 2024-04-23
core/identity: add enabler

calebdoxsey:
## Summary

Add a new `Enabler` to simplify enabling and disabling components dynamically.
 
 

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/disabler/disabler.go
```diff
@@ -0,0 +1,111 @@
...
+	Disable()
+}
+
+type disabler struct {
+	name            string
+	handler         Handler
+	onChangeEnabled chan bool
+}
+
+// New creates a new Disabler. When the Disabler is enabled, the Handler's
```
kenjenkins, line 29: I'd put in a vote for naming this Enabler rather than Disabler. The phrase "When the Disabler is enabled" makes my head hurt 🙂.

internal/disabler/disabler.go
```diff
@@ -0,0 +1,111 @@
...
+		return err
+	}
+}
+
+func (d disabler) runEnabledOnce(ctx context.Context) error {
+	ctx, cancel := context.WithCancelCause(ctx)
+	defer cancel(context.Canceled) // ensure we cancel the context if we exit early
+
+	// start a background goroutine that will listen for a transition to disabled
+	go func() {
```
kenjenkins, line 70: I wonder if we do need a background goroutine for this? I haven't completely thought it through yet but I wonder if we could simplify this a little by storing the `cancel` function in a `sync.Pointer` field on the `disabler` struct.

I'm thinking the Enable operation would create a new `context.WithCancelCause(ctx)`, then `CompareAndSwap(nil, cancel)`, and then call `RunEnabled()` if the swap succeeded. The Disable operation could retrieve the cancel function using `Swap(nil)` and then invoke it if it isn't nil.

Does that make any sense? Any thoughts / feelings?

kenjenkins (commented):
Thanks, I think something like this could make sense. Added a couple comments.

internal/disabler/disabler.go
calebdoxsey: Run happens in a separate goroutine from Enable/Disable. I'm not sure how to communicate the state change apart from using a channel.

We could wait for `enableCtx.Done()` and cancel it in Enable and then wait for `disableCtx.Done()` and cancel it in Disable... but I think that wouldn't work and fixing it feels racy.

I suppose if we track the current state and wrap that with a mutex we could avoid the race. I can try to implement this.

calebdoxsey: Why do we want to avoid using a goroutine?

calebdoxsey: Using a mutex did simplify the code.

calebdoxsey:
I renamed it to enabler and re-worked the logic to use a `context.CancelCauseFunc` for signaling from `Enable`/`Disable` to `Run`.

internal/enabler/enabler.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Run calls RunEnabled if enabled, otherwise it waits until enabled.
+func (d *enabler) Run(ctx context.Context) error {
+	for {
+		err := d.runOnce(ctx)
+		if errors.Is(err, errCauseEnabler) {
+			continue
+		}
+		return err
+	}
```
kenjenkins, line 66: I find this loop a little confusing, but I think I'm slowly understanding. Let me see if I can summarize the behavior:

 - This loop will keep going until the underlying context is canceled for some reason unrelated to the enabled/disabled state.
 - If currently enabled, runOnce() will synchronously call RunEnabled(), blocking until that function returns.
 - If currently disabled, runOnce() will still block, waiting for a transition to the enabled state.

I think the name "runOnce" confused me a little (because it might not run anything), but I don't have any naming alternatives to offer.

kenjenkins (approved):
@wasaga, would you mind taking a look at this as well? I don't really trust myself when reasoning about concurrency.

internal/disabler/disabler.go
kenjenkins: OK great, like I said I probably didn't think it through all the way. I think I didn't really understand the contract for the `Run()` method. I'll defer to your judgement on whatever looks simplest.

internal/enabler/enabler.go
calebdoxsey: I changed the loop to avoid the continue.

When a `ctx` is involved like this I try to have the loop separate from the implementation so that the scope of the new `ctx` doesn't override the original one (a subtle bug that has bit me several times before), and also so we can use normal returns to exit early instead of continues.

I renamed the method to make it more clear and added some comments.

Basically we get the current `enabled` state, create a new cancelable context and call the cancel function in `Enable` and `Disable`. 

In the case when we're enabled, we call `RunEnabled` with the new context which will be canceled by a call to `Disable`. We then check the underlying cause. If it's due to us, we return that error, which is checked in the loop. But any other error (or nil) we want to exit to preserve the original behavior of `Run`.

In the case when we're disabled, we just wait for the context to be canceled, which can only happen if the parent context is canceled or we cancel it with a call to `Enable`. `context.Cause` should surface that error, and the loop will continue if the cause was us, or return if the parent context was canceled.

wasaga (approved):

internal/enabler/enabler.go
kenjenkins: Thanks, I like the name `runOrWaitForEnabled`.

Issue #4460, opened 2023-08-14
Allow wildcard port in route host/from


caesay:
This is similar to https://github.com/pomerium/pomerium/issues/1677 and the merged https://github.com/pomerium/pomerium/pull/4131 which allows you to use wildcards in the 'from' route property. Also related to: https://github.com/pomerium/pomerium/issues/659

**Is your feature request related to a problem? Please describe.**
I have devices accessing the proxy via various local networks, and also remotely from behind other proxies. Pomerium runs/listens on port 8443. If accessing the proxy externally, generally you access via port 443 - which is then forwarded to 8443 on the local network by another proxy. If a device needs to access the proxy on the same local network, then they will directly use 8443. In some occasional cases, ports need to be re-mapped between local networks, so the client may access on even other ports.

Currently, to support this you would need to create a route for each possible port to the same service, and copy over any additional route configuration (such as access policies, headers etc):

```yaml
  - from: https://one.example.com
    to: local.service
  - from: https://one.example.com:443
    to: local.service
  - from: https://one.example.com:8443
    to: local.service
  - from: https://one.example.com:18443
    to: local.service
```

This duplication needs to be done for each route/service, and can get particularly erroneous if the routes have complicated access policies and request/response rewriting.

**Describe the solution you'd like**
It would be good if you could define multiple "from" addresses, or use wildcards and/or regexes that also cover the port number in the host address. 

For example, a solution using multiple "from" addresses, requiring you to spell out each port but share access policies and request rewriting properties -
```yaml
  - from: 
      - https://one.example.com
      - https://one.example.com:8443
      - https://one.example.com:18443
    to: local.service
```

Or a solution using wildcard (not covered by current implementation in #4131)
```yaml
  - from: https://one.example.com:*
    to: local.service
```

Or a solution using regexes (not covered by specification of #659)
```yaml
  - from: https://one.example.com:(\d*)
    to: local.service
```

kenjenkins:
I wonder if we can ignore the port number entirely (or perhaps ignore it unless specified explicitly). I believe Pomerium doesn't currently listen on multiple ports, so I'm not sure when you'd want it to make a distinction between different port numbers.

wasaga:
@kenjenkins if the server is configured to listen at port `8443` (`address: ":8443"`) while running behind a TCP load balancer that forwards requests arriving to `one.example.com:443` to Pomerium host:8443, your `from` should look like `https://one.example.com`. 

[...]

PR #5085, opened 2024-04-24
core/config: add support for stripping the port for matching routes

calebdoxsey:
## Summary
Change the way route matching works so that if a route is defined without an explicit port that route will match a request on any port.

The current behavior is that a route defined without an explicit port will only match a request with no port, or port `:443`.

For example, with the current behavior a route like:

```yaml
from: https://www.example.com
```

Would match incoming requests for `https://www.example.com` or `https://www.example.com:443`, but not `https://www.example.com:8443`.

With the new behavior all 3 routes would match.

This new behavior can be disabled via a runtime flag (`strip_from_port`), but the default is enabled since we believe its unlikely to break existing configurations.

TCP routes should behave as they do currently. The port is always required for them and must match explicitly.

## Related issues
- https://github.com/pomerium/pomerium/issues/4460
- https://github.com/pomerium/ingress-controller/issues/926

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/from.go
+++ b/config/from.go
@@ -9,8 +9,8 @@ import (
 )
 
 // FromURLMatchesRequestURL returns true if the from URL matches the request URL.
-func FromURLMatchesRequestURL(fromURL, requestURL *url.URL) bool {
-	for _, domain := range urlutil.GetDomainsForURL(fromURL) {
+func FromURLMatchesRequestURL(fromURL, requestURL *url.URL, stripPort bool) bool {
+	for _, domain := range urlutil.GetDomainsForURL(fromURL, true) {
 		if domain == requestURL.Host {
 			return true
 		}
@@ -19,7 +19,7 @@ func FromURLMatchesRequestURL(fromURL, requestURL *url.URL) bool {
 			continue
 		}
 
-		reStr := WildcardToRegex(domain)
+		reStr := WildcardToRegex(domain, stripPort)
 		re := regexp.MustCompile(reStr)
 		if re.MatchString(requestURL.Host) {
 			return true
@@ -29,7 +29,7 @@ func FromURLMatchesRequestURL(fromURL, requestURL *url.URL) bool {
 }
 
 // WildcardToRegex converts a wildcard string to a regular expression.
-func WildcardToRegex(wildcard string) string {
+func WildcardToRegex(wildcard string, stripPort bool) string {
 	var b strings.Builder
 	b.WriteByte('^')
 	for {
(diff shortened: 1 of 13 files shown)


config/runtime_flags.go
```diff
@@ -2,8 +2,13 @@ package config
...
 import "golang.org/x/exp/maps"
 
-// RuntimeFlagGRPCDatabrokerKeepalive enables gRPC keepalive to the databroker service
-var RuntimeFlagGRPCDatabrokerKeepalive = runtimeFlag("grpc_databroker_keepalive", false)
+var (
+	// RuntimeFlagGRPCDatabrokerKeepalive enables gRPC keepalive to the databroker service
+	RuntimeFlagGRPCDatabrokerKeepalive = runtimeFlag("grpc_databroker_keepalive", false)
+
+	// RuntimeFlagStripFromPort enables stripping the port from incoming requests for matching from addresses
+	RuntimeFlagStripFromPort = runtimeFlag("strip_from_port", true)
```
kenjenkins, line 10: One comment on the name: with this implementation it doesn't really "strip" the port, does it? I think maybe something along the lines of "match_any_incoming_port" might be more accurate.

config/from.go
```diff
@@ -42,6 +42,9 @@ func WildcardToRegex(wildcard string) string {
 		wildcard = wildcard[idx+1:]
 	}
 	b.WriteString(regexp.QuoteMeta(wildcard))
+	if stripPort && !strings.Contains(wildcard, ":") {
```
kenjenkins, line 45: I'm trying to think if there are any corner cases here... do we allow IP addresses in `from` URLs? I'm wondering if we should do anything about IPv6 addresses, which could contain `:` characters without specifying a port number.

(Not sure if it's something we want to support though.)

config/from_test.go
```diff
@@ -21,18 +21,19 @@ func TestFromURLMatchesRequestURL(t *testing.T) {
...
+		assert.Equal(t, tc.matches, FromURLMatchesRequestURL(&fromURL, &requestURL, true),
 			"from-url: %s\nrequest-url: %s", tc.pattern, tc.input)
 	}
 }
 
 func TestWildcardToRegex(t *testing.T) {
 	t.Parallel()
 
-	re, err := regexp.Compile(WildcardToRegex("*.internal.*.example.com"))
+	re, err := regexp.Compile(WildcardToRegex("*.internal.*.example.com", true))
```
kenjenkins, line 36: Would it be worth adding a unit test exercise the `WildcardToRegex(..., false)` code path too?

kenjenkins (approved):
Thanks! I think this makes sense.

config/from.go
calebdoxsey: good catch

PR #5091, opened 2024-04-26
core/identity: refactor identity manager

calebdoxsey:
## Summary
Implement a new version of the identity manager. The legacy version can be enabled via a runtime flag, but the new version is the default.

The identity manager is responsible for two things: refreshing user sessions and updating user information. A typical Pomerium session lasts for 14 hours, whereas a typical IdP OIDC token is valid for 1 hour. So we refresh the user's IdP token so that they don't have to login again. We also periodically (every 10 minutes) update the user's info from the IdP so that policy evaluation is always against relatively fresh data.

The identity manager receives session and user data from a Syncer to the Databroker. Every new session that gets created is pushed to the identity manager.

The legacy implementation maintained a sorted list of scheduled session refreshes and user updates. Each loop iteration we would retrieve the next ready session, refresh it, and then move on to users once all soon-to-expire sessions have been refreshed. The problem with this design is that if refreshing sessions or updating user information takes a long time, it blocks the identity manager from doing anything else. In particular if a session has an hour to be refreshed and all the sessions roughly expire at the same time, and calls to refresh a session take several seconds, we could very easily fail to schedule in time, leading to the session token expiring during Authorization and users being logged out.

Rather than attempting to refresh multiple sessions or update multiple users concurrently, the new identity manager uses an entirely different approach.

For each new session we start an independent `refreshSessionScheduler` and for each user we start an independent `updateUserInfoScheduler`. These schedulers run in their own goroutines and when ready invoke a method on the `Manager`. This removes the bottleneck as all sessions can now refresh concurrently. It also makes the code easier to understand.

The downside of this approach is that we will be starting lots of goroutines. A goroutine for each session and user object. A goroutine uses roughly 2KB of RAM, so for 10,000 sessions and 1,000 users this would mean roughly 21MB of RAM. In addition to memory usage there's the added work put on the Go scheduler, which ought to be negligible. Based on these back-of-the-napkin calculations I don't think this needs further optimization, but we could pursue that in a subsequent PR if that's desired.

## Related issues
- https://github.com/pomerium/ingress-controller/issues/938

## User Explanation
The new identity manager should change no functionality.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


internal/identity/manager/datastore_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	u, ss := ds.getUserAndSessions("U1")
+	assert.Nil(t, u, "should return a nil user when none exists")
+	assert.Empty(t, ss, "should return an empty list of sessions when no user exists")
+
+	s = &session.Session{Id: "S1", UserId: "U1"}
+	ds.putSession(s)
+	s1, u1 := ds.getSessionAndUser("S1")
+	assert.NotNil(t, s1, "should return a non-nil session")
+	assert.False(t, s == s1, "should return different pointers")
```
kenjenkins, line 30: totally optional: you might like to use https://pkg.go.dev/github.com/stretchr/testify/assert#NotSame for these pointer assertions

internal/identity/manager/schedulers.go
```diff
@@ -0,0 +1,161 @@
...
+	ctx, uuis.cancel = context.WithCancel(context.WithoutCancel(uuis.baseCtx))
+	go uuis.run(ctx)
+	return uuis
+}
+
+func (uuis *updateUserInfoScheduler) Reset() {
+	select {
+	case uuis.reset <- struct{}{}:
+	default:
+	}
```
kenjenkins, line 43: This effectively de-duplicates Reset() calls, is that right? Because `reset` is a buffered channel, this should never "miss" a reset, because sending to the reset channel should never block unless there's already a pending item in the channel buffer?

(not requesting any changes, just want to make sure I understand how this works)

internal/identity/manager/schedulers.go
```diff
@@ -0,0 +1,161 @@
...
+func (uuis *updateUserInfoScheduler) run(ctx context.Context) {
+	ticker := time.NewTicker(uuis.updateUserInfoInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-uuis.reset:
+			ticker.Reset(uuis.updateUserInfoInterval)
```
kenjenkins, line 59: I think I'm still missing something— when do we need to reset the ticker?

kenjenkins (commented):
Thanks for taking this on, I think this is a good approach.

Just a few questions / suggestions from me... I think I _almost_ understand the new scheduler logic but I'm not quite there yet.

internal/identity/manager/data.go
```diff
@@ -88,18 +40,41 @@ func (s Session) NextRefresh() time.Time {
...
 	}
 
 	return tm
 }
 
-// UnmarshalJSON unmarshals json data into the session object.
-func (s *Session) UnmarshalJSON(data []byte) error {
-	if s.Session == nil {
-		s.Session = new(session.Session)
+type multiUnmarshaler []any
```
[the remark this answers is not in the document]
calebdoxsey: yes. I added a comment.

internal/identity/manager/schedulers.go
calebdoxsey: We reset after the user is updated. This fixes an unnecessary update in the original scheduler. When a session is refreshed it also updates the user data, in which case it would be better to reset the timer and wait another 10 minutes before doing it again. The original scheduler would keep to its 10 minute interval regardless.

internal/identity/manager/schedulers.go
calebdoxsey: Added a comment. Yes we're basically using the channel as a signaling mechanism. There are other ways to accomplish this, but using channels allows us to use the `select` statement.

internal/identity/manager/schedulers.go
kenjenkins: Thanks, got it.

internal/identity/manager/manager.go
```diff
@@ -450,21 +357,70 @@ func (mgr *Manager) deleteSession(ctx context.Context, userID, sessionID string)
...
+
+	mgr.mu.Lock()
+	mgr.dataStore.putSession(s)
+	if rss, ok := mgr.refreshSessionSchedulers[s.GetId()]; ok {
+		rss.Update(s)
+	}
+	mgr.mu.Unlock()
+}
+
+// updateUser updates the user in the databroker, the local data store, and resets the scheduler
```
kenjenkins, line 403: I think it could be helpful to capture the reset explanation here (feel free to edit if this isn't quite right).
```suggestion
// updateUser updates the user in the databroker, the local data store, and resets the scheduler.
// (Whenever we refresh a session, we also refresh the user info. By resetting the user info 
// scheduler here we can avoid refreshing user info more often than necessary.)
```

kenjenkins (approved):

PR #5096, opened 2024-05-01
log/grpc: use standard logger

wasaga:
## Summary

Previously, gRPC used default global logger, that caused mismatch with levels and format options configured across Pomerium.

This PR: 

1. sets default gRPC logger to use Pomerium global logger
2. only log gRPC related messages if debug level is enabled. 
 - By default, gRPC logs are rather verbose and have numerous transient errors as various Pomerium modules are initializing, that may be confusing to the user. 
 - if user enables pomerium `log_level: debug`, they would also observe all the gRPC transport related errors, without need to specifically adjust gRPC verbosity via [special env variables](https://pkg.go.dev/google.golang.org/grpc/examples/features/debugging#section-readme) which should be handy when troubleshooting Pomerium installation.
3. set default level to `info` in main.go, to avoid interfering with numerous tests that expect zerolog global default level to be debug

## Related issues

Fixes https://github.com/pomerium/internal/issues/1774

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/log/grpc.go
+++ b/internal/log/grpc.go
@@ -14,55 +14,55 @@ func init() {
 type grpcLogger struct{}
 
 func (c *grpcLogger) Info(args ...interface{}) {
-	if GetLevel() == zerolog.DebugLevel {
+	if GetLevel() <= zerolog.DebugLevel {
 		Logger().Info().Msg(fmt.Sprint(args...))
 	}
 }
 
 func (c *grpcLogger) Infoln(args ...interface{}) {
-	if GetLevel() == zerolog.DebugLevel {
+	if GetLevel() <= zerolog.DebugLevel {
 		Logger().Info().Msg(fmt.Sprintln(args...))
 	}
 }
 
 func (c *grpcLogger) Infof(format string, args ...interface{}) {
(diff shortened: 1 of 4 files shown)


internal/log/grpc.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+func init() {
+	grpclog.SetLoggerV2(&grpcLogger{})
+}
+
+type grpcLogger struct{}
+
+func (c *grpcLogger) Info(args ...interface{}) {
+	if GetLevel() == zerolog.DebugLevel {
```
calebdoxsey, line 17: does this need to be something like `<=`? I think zerolog has `trace` too

calebdoxsey (approved):

internal/log/grpc.go
wasaga: Debug is zero in zero log but sure to be future proof we can do <=

wasaga: // TraceLevel defines trace log level.
	TraceLevel Level = -1

it seems like they added it indeed as -1

PR #5099, opened 2024-05-02
core/lint: upgrade golangci-lint, replace interface{} with any

calebdoxsey:
## Summary
Upgrade golangci-lint to `1.57.2`. This version requires a variable to be used or named `_`. Also update everywhere we use `interface{}` to be `any`.
 
## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


config/autocert_test.go
```diff
@@ -64,15 +64,15 @@ func TestAutocertOptions_Validate(t *testing.T) {
 		cleanup func()
 	}
 	tests := map[string]func(t *testing.T) test{
-		"ok/custom-ca": func(t *testing.T) test {
+		"ok/custom-ca": func(_ *testing.T) test {
```
kenjenkins, line 67: Note to self: it looks wrong to have a `func(*testing.T)` where `T` is unused, but it looks like this is set up so that some specific test cases can perform extra initialization steps (and the test should fail if these initialization steps fail). I think this could be refactored to make the intent clearer; filed https://github.com/pomerium/pomerium/issues/5100 to track separately.

pkg/grpc/audit/audit.pb.go
```diff
@@ -111,7 +111,7 @@ func file_audit_proto_rawDescGZIP() []byte {
 }
 
 var file_audit_proto_msgTypes = make([]protoimpl.MessageInfo, 1)
-var file_audit_proto_goTypes = []interface{}{
+var file_audit_proto_goTypes = []any{
```
kenjenkins, line 114: I'm not sure about modifying the generated protobuf code directly. Won't all these changes get reverted the next time someone runs `make proto`?

kenjenkins (approved):
Looks good to me, just one concern about the generated protobuf code.

pkg/grpc/audit/audit.pb.go
kenjenkins: Thanks, looks good to me.

PR #5106, opened 2024-05-06
Envoy resource monitoring & overload manager configuration

kralicky:
## Summary

This adds a new "resource manager" feature that continuously monitors cgroup memory saturation and feeds it to envoy's overload manager. It supports cgroup v1/v2/hybrid and container-style layouts.

It works by using envoy's [injected resource](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/resource_monitor/injected_resource/v2alpha/injected_resource.proto) resource monitor, effectively allowing us manual control over the overload action states. Envoy can only monitor its own memory usage, so we can give it a more accurate measurement of memory pressure since we know more about the runtime environment.

The most common use case will be when both pomerium and envoy are running in the same container, and the container has memory restrictions configured. Pomerium will monitor the memory usage of the cgroup containing both processes (since that's what matters from the oom-killer's perspective) and envoy will use that value to control its overload actions. 

A runtime flag `envoy_resource_manager_enabled` (default true) can be used to soft-disable the resource manager. Since the envoy overload manager settings are part of the bootstrap config, they can't be changed without a restart. If this flag is set to false, pomerium will always treat the memory saturation value as `0`, ensuring all overload actions stay disabled.

If the current cgroup has an unbounded memory limit ("max") then it is also treated as `0`. The resource manager will watch for changes to the memory limit to allow updating it at runtime (such as via k8s 1.30 [dynamic resource allocation](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/))

TODOs:
- [x] Adjust and document thresholds for various overload actions
- [x] Add new metrics for memory usage and overload manager actions
- [x] Write tests
- [ ] Document how this interacts with pod resource restrictions

## Related issues

* https://github.com/pomerium/internal/issues/1744

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/envoy/resource_monitor_linux.go
```diff
@@ -0,0 +1,767 @@
...
+					if err != nil {
+						log.Error(ctx).Err(err).Msg("failed to get memory saturation")
+						continue
+					}
+					saturation = float64(usage) / float64(limit)
+				}
+			}
+
+			saturationStr := fmt.Sprintf("%.6f", saturation)
+			nextInterval := (monitorMaxTickInterval - (time.Duration(float64(monitorMaxTickInterval-monitorMinTickInterval) * saturation))).
```
calebdoxsey, line 334: maybe we could pull this into a separate function?

Also do we need to handle `saturation > 1`?

pkg/envoy/resource_monitor_linux.go
```diff
@@ -0,0 +1,767 @@
...
+	}
+	max := strings.TrimSpace(string(data))
+	if max == "max" {
+		return 0, nil
+	}
+	return strconv.ParseUint(max, 10, 64)
+}
+
+func (w *memoryLimitWatcher) Watch(ctx context.Context) error {
+	fd, err := syscall.InotifyInit1(syscall.IN_CLOEXEC)
```
calebdoxsey, line 691: We have a file watcher in `fileutil`. But I guess its made for more of the cross-platform case and would also start a polling watcher, which is not ideal. Would using the [underlying package](namespacelabs.dev/go-filenotify) simplify things?  Or is it really no different?

calebdoxsey (approved):
LGTM

I did see these notices in the docs:

> This extension is functional but has not had substantial production burn time, use only with this caveat.
> The injected resource monitor allows injecting a synthetic resource pressure into Envoy via a text file, which must contain a floating-point number in the range [0..1] representing the resource pressure and be updated atomically by a symbolic link swap. This is intended primarily for integration tests to force Envoy into an overloaded state.

If we're ok with those caveats I think this is a good solution!

pkg/envoy/resource_monitor_linux.go
kralicky: Sure, I did this manually since the usage here is simple enough maybe not to warrant using a library for, but I can rewrite to use github.com/fsnotify/fsnotify if you'd like. It might make the code a bit more readable.

pkg/envoy/resource_monitor_linux.go
kralicky: Yeah I think we should handle that, good catch

pkg/envoy/resource_monitor_linux.go
kralicky: Sure, I did this manually since the usage here is simple enough maybe not to warrant using a library for, but I can rewrite to use github.com/fsnotify/fsnotify if you'd like. It might make the code a bit more readable.

calebdoxsey: No preference from me. Whichever way seems best.

wasaga:
Looks great!

I have a couple follow ups that I'll open separate tickets for.

wasaga (approved):
Looks great!

I have a couple follow ups that I'll open separate tickets for.

pkg/envoy/resource_monitor_linux.go
kralicky: Looks like there are some wrappers around the low level syscall functions in golang.org/x/sys/unix that I can use here, that might be good for now.

calebdoxsey (approved):

PR #5108, opened 2024-05-08
core/envoy: exclude unauthorized access from local replies

calebdoxsey:
## Summary
Exclude `ext_authz_denied` from the local reply handling so it retains the data returned by the authorize service.

I couldn't find a way of doing `NOT` with the filter, so I just took all the possible values and removed `UAEX`.

## Related issues
For https://github.com/pomerium/internal/issues/1783

## Checklist

- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/config/envoyconfig/http_connection_manager.go
+++ b/config/envoyconfig/http_connection_manager.go
@@ -56,9 +56,10 @@ func (b *Builder) buildLocalReplyConfig(
 	}
 
 	data := map[string]any{
-		"status":     "%RESPONSE_CODE%",
-		"statusText": "%RESPONSE_CODE_DETAILS%",
-		"requestId":  "%STREAM_ID%",
+		"status":        "%RESPONSE_CODE%",
+		"statusText":    "%RESPONSE_CODE_DETAILS%",
+		"requestId":     "%STREAM_ID%",
+		"responseFlags": "%RESPONSE_FLAGS%",
 	}
 	httputil.AddBrandingOptionsToMap(data, options.BrandingOptions)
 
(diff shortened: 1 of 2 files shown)


config/envoyconfig/http_connection_manager.go
```diff
@@ -71,7 +72,38 @@ func (b *Builder) buildLocalReplyConfig(
 		Mappers: []*envoy_http_connection_manager.ResponseMapper{{
 			Filter: &envoy_config_accesslog_v3.AccessLogFilter{
 				FilterSpecifier: &envoy_config_accesslog_v3.AccessLogFilter_ResponseFlagFilter{
-					ResponseFlagFilter: &envoy_config_accesslog_v3.ResponseFlagFilter{},
+					ResponseFlagFilter: &envoy_config_accesslog_v3.ResponseFlagFilter{
+						Flags: []string{
```
kenjenkins, line 76: The fact that Envoy doesn't provide any sort of `NOT` operator here is definitely annoying.

Would we be better off using a [Metadata Filter](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/accesslog/v3/accesslog.proto#envoy-v3-api-msg-config-accesslog-v3-metadatafilter)? Could we set a bit of metadata in the ext_authz response that tells this filter not to override the response body?

calebdoxsey: For reasons I don't understand, using the metadata filter ends up always matching regardless of what I set.

```go
	filter := &envoy_config_accesslog_v3.AccessLogFilter{
		FilterSpecifier: &envoy_config_accesslog_v3.AccessLogFilter_MetadataFilter{
			MetadataFilter: &envoy_config_accesslog_v3.MetadataFilter{
				Matcher: &envoy_type_matcher_v3.MetadataMatcher{
					Filter: "envoy.filters.http.ext_authz",
					Path: []*envoy_type_matcher_v3.MetadataMatcher_PathSegment{
						{
							Segment: &envoy_type_matcher_v3.MetadataMatcher_PathSegment_Key{
								Key: DynamicMetadataPomeriumServiceKey,
							},
						},
					},
					Value: &envoy_type_matcher_v3.ValueMatcher{
						MatchPattern: &envoy_type_matcher_v3.ValueMatcher_StringMatch{
							StringMatch: &envoy_type_matcher_v3.StringMatcher{
								MatchPattern: &envoy_type_matcher_v3.StringMatcher_Exact{
									Exact: "authorizexxx",
								},
							},
						},
					},
				},
			},
		},
	}
```

I expected that filter to match nothing, but it was still true.

calebdoxsey: Possibly https://github.com/envoyproxy/envoy/issues/17802

calebdoxsey:
Ok looking into this, although it would simplify the code to remove the default error case where we generate an HTML response, there are other code paths to handle various types of requests (`robots.txt`, json web, grpc web, ...). Given the limitations on how local reply configuration is constructed, I don't think we'd be able to reproduce that logic effectively, so we'd still need to carve out the exception for authorize denied responses.

PR #5109, opened 2024-05-08
core/ui: improve frontend build size

calebdoxsey:
## Summary
Improve the UI build size from nearly 4MB to 400KB.

- upgrade `esbuild` 
- use external, linked sourcemaps
- remove `luxon`, which we apparently only used to convert unix timestamps to ISO strings, which JS can do natively
- switch to `import { } from '@mui/material'` everywhere, because it didn't seem to matter which we did it for tree-shaking
- use imports from `lodash` like `lodash/isArray` to tree shake better

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


desimone:
wow that's quite the improvement!

nhayfield (approved):

ui/scripts/esbuild.ts
```diff
@@ -1,16 +1,26 @@
...
-    ".woff": "dataurl",
-    ".woff2": "dataurl",
-  },
-});
+async function run() {
+  const cfg: BuildOptions = {
+    entryPoints: ["src/index.tsx"],
+    bundle: true,
+    outdir: "dist",
+    sourcemap: "linked",
```
kenjenkins, line 9: If the UI in Core isn't super complicated, I wonder if we even need to generate source maps at all?

(I'd be fine either way, just curious what you think.)

.gitignore
kenjenkins, line 1: It looks like index.js and index.css also appear in the [.dockerignore file](https://github.com/pomerium/pomerium/blob/d225288ab3b9224861a7a750aab03e52468017d5/.dockerignore#L3-L4). Do you know if we should update that list to include index.js.map and index.css.map as well?

kenjenkins (approved):
Nice work!

ui/scripts/esbuild.ts
calebdoxsey: Sure. I changed it to only generate the sourcemap when using watch mode (for local development, where it's probably more useful)

PR #5111, opened 2024-05-10
health-checks: zero route availability improvements

wasaga:
## Summary

This PR does some improvements to the Zero Route Availability health check: 

- ignores certificate errors, as the point of that health check is not to check the validity of the cert, but rather whether there is an expected pomerium instance available at a given hostname
- if check fails, re-check it with exponential back-off
- previously it was only ran on schedule (every 30 minutes), now it would also be triggered on configuration change
- also adds an "internal error" reason for health checks when health check cannot complete because of some external condition unrelated to the health check itself

## Related issues

Fixes: https://github.com/pomerium/pomerium-zero/issues/2360

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


internal/zero/healthcheck/syncer.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"context"
+
+	configpb "github.com/pomerium/pomerium/pkg/grpc/config"
+	"github.com/pomerium/pomerium/pkg/grpc/databroker"
+	"github.com/pomerium/pomerium/pkg/health"
+	"github.com/pomerium/pomerium/pkg/protoutil"
+)
+
+func (c *checker) ConfigSyncer(ctx context.Context) error {
```
kenjenkins, line 12: Why should the health checks sync config separately from the rest of the app?

What are the difference between this approach compared with registering an OnConfigChange() listener on the main config source?

wasaga: that's a fair point, however here we only have access to the lowest layer of config. 
[by design](https://github.com/pomerium/pomerium/blob/7d8af09036fe0e6de0122c3c490393d1da68f4c5/config/config_source.go#L49-L53) higher layers are notified when lower layer configs are changed but not the other way around. 

we would have to move this health check to some entirely different place to be able to attach to a regular config.Source; there are benefits doing that; I can make a separate PR for that.

kenjenkins (approved):

PR #2405, opened 2021-07-29
docs: concepts (suggestions on #2390) 

desimone:
See commit messages for breakdown of changes. Should be git-cherry-pickable.

PR #5114, opened 2024-05-22
Add new configurable bootstrap writers (#2405)

kralicky:
## Summary

This PR adds the ability to configure different backends to use for storing modifications to the zero bootstrap config. The two currently implemented backends allow writing changes to a file or to a Kubernetes secret. Backend selection is determined by the scheme in a URI passed to the flag '--config-writeback-uri'.

In a Kubernetes environment, where the bootstrap config is mounted into the pod from a secret, this option allows Pomerium to write changes back to the secret, as writes to the mounted secret file on disk are not persisted.

## Related issues

Part of #2405 

## User Explanation

This would generally not be used by users, but configured automatically as part of a deployment manifest or controller.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


internal/zero/bootstrap/writers/writers.go
```diff
@@ -0,0 +1,43 @@
...
+	"crypto/cipher"
+	"fmt"
+	"net/url"
+	"sync"
+
+	cluster_api "github.com/pomerium/pomerium/pkg/zero/cluster"
+)
+
+type ConfigWriter interface {
+	WriteConfig(ctx context.Context, src *cluster_api.BootstrapConfig, cipher cipher.AEAD) error
```
wasaga, line 14: I'm not sure about passing cipher in this function:
- if we keep encryption for both writers, lets make encrypt/decrypt responsibility of the caller. 
- if we only keep encryption for the file writer, the cipher need be provided to it during initialization

internal/zero/bootstrap/writers/k8s/secret.go
```diff
@@ -0,0 +1,176 @@
...
+	switch resp.StatusCode {
+	case http.StatusOK, http.StatusCreated:
+		return nil
+	case http.StatusForbidden:
+		if resp.Header.Get("Content-Type") == "application/json" {
+			// log the detailed status message if available
+			status, err := io.ReadAll(resp.Body)
+			if err != nil && len(status) > 0 {
+				log.Ctx(ctx).Error().
+					RawJSON("response", status).
```
wasaga, line 106: RawJSON requires no carriage returns in the output; something we may not be able to provide a guarantee for

internal/zero/bootstrap/writers/k8s/secret.go
```diff
@@ -0,0 +1,176 @@
...
+					Msg("forbidden")
```
wasaga, line 107: let's add the `PATCH /api/v1/namespaces/....` so that it'll be evident where we tried to write
and a little more explanation in the message (what did we tried to write)

internal/zero/cmd/command.go
```diff
@@ -41,6 +40,11 @@ func Run(ctx context.Context, configFile string) error {
 	} else {
 		log.Ctx(ctx).Info().Str("file", bootstrapConfigFileName).Msg("cluster bootstrap config path")
 		opts = append(opts, controller.WithBootstrapConfigFileName(bootstrapConfigFileName))
+
+		if uri := getBootstrapConfigWritebackURI(); uri != "" {
+			log.Ctx(ctx).Info().Str("uri", uri).Msg("cluster bootstrap config writeback URI")
```
wasaga, line 45:
```suggestion
			log.Ctx(ctx).Debug().Str("uri", uri).Msg("cluster bootstrap config writeback URI")
```

internal/zero/cmd/env.go
```diff
@@ -10,6 +12,12 @@ const (
 	// PomeriumZeroTokenEnv is the environment variable name for the API token.
 	//nolint: gosec
 	PomeriumZeroTokenEnv = "POMERIUM_ZERO_TOKEN"
+
+	// BootstrapConfigFileName can be set to override the default location of the bootstrap config file.
+	BootstrapConfigFileName = "BOOTSTRAP_CONFIG_FILE"
```
wasaga, line 17: there's a general preference to avoid introduction of additional environment variables; pomerium already utilizes cache dir

internal/zero/cmd/env.go
```diff
@@ -50,3 +58,26 @@ func getOTELAPIEndpoint() string {
 	}
 	return "https://telemetry.pomerium.app"
 }
+
+func getBootstrapConfigFileName() (string, error) {
+	if filename := os.Getenv(BootstrapConfigFileName); filename != "" {
+		return filename, nil
+	}
```
wasaga, line 65:
```suggestion
```

internal/zero/cmd/env.go
```diff
@@ -50,3 +58,26 @@ func getOTELAPIEndpoint() string {
...
+
+	return filepath.Join(dir, "bootstrap.dat"), nil
+}
+
+func getBootstrapConfigWritebackURI() string {
+	if uri := os.Getenv(BootstrapConfigWritebackURI); uri != "" {
+		return uri
+	}
+	return ""
```
wasaga, line 82:
```suggestion
	return os.Getenv(BootstrapConfigWritebackURI)
```

internal/zero/cmd/env.go
```diff
@@ -10,6 +12,12 @@ const (
 	// PomeriumZeroTokenEnv is the environment variable name for the API token.
 	//nolint: gosec
 	PomeriumZeroTokenEnv = "POMERIUM_ZERO_TOKEN"
+
+	// BootstrapConfigFileName can be set to override the default location of the bootstrap config file.
+	BootstrapConfigFileName = "BOOTSTRAP_CONFIG_FILE"
+	// BootstrapConfigWritebackURI controls how changes to the bootstrap config are persisted.
+	// See controller.WithBootstrapConfigWritebackURI for details.
+	BootstrapConfigWritebackURI = "BOOTSTRAP_CONFIG_WRITEBACK_URI"
```
wasaga, line 20: I wonder if we should start calling such env variables that are required purely for deployment purposes something like `INTERNAL_` hinting they are not part of the public API / config surface and may be changed in the future.

internal/zero/bootstrap/writers/writers.go
kralicky: Yeah I agree, it makes more sense to keep encryption as an implementation detail.

internal/zero/cmd/env.go
kralicky: Good idea. For context, this was originally a command line flag but that was clunky to wire up all the way to here, but let me know if you think there is a better way to configure these. Adding an internal prefix seems good to me otherwise.

internal/zero/bootstrap/writers/k8s/secret.go
```diff
@@ -0,0 +1,176 @@
...
+		RawQuery: url.Values{
+			"fieldManager": {"pomerium-zero"},
+			"force":        {"true"},
+		}.Encode(),
+	})
+	plaintext, err := json.Marshal(src)
+	if err != nil {
+		return err
+	}
+	ciphertext := cryptutil.Encrypt(cipher, plaintext, nil)
```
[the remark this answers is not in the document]
kralicky: Could go either way I think. It may be easier to encrypt it so that it works the same regardless of writer backend. Depends if we want this to be user-serviceable or not. If not, having the data be raw bytes without a visible encoding might help signal that this is not meant to be hand-edited.

internal/zero/cmd/env.go
kralicky: On second thought, taking a look at some of the other internal variables we use ("CLUSTER_API_ENDPOINT", "CONNECT_SERVER_ENDPOINT" etc) vs other ones that could reasonably be user-serviceable ("POMERIUM_OTEL_ENDPOINT") the lack of prefix seems to be a consistent way to denote internal variables. What do you think?

internal/zero/cmd/env.go
kralicky: This is an intentional pattern, but not a big deal, I will change it. :+1:

wasaga (approved):

internal/zero/cmd/env.go
wasaga: I see. yes those are confusing as well. the only reason to call `POMERIUM_OTEL_ENDPOINT` with `POMERIUM_` prefix was to avoid a potential conflict with OTEL SDK env vars.

calebdoxsey (approved):

PR #5118, opened 2024-05-29
telemetry: add prometheus to otel types converter

wasaga:
## Summary

This PR adds a converter that takes a stream of Prometheus metrics and outputs OTEL metric objects.
This is just a first pass, so not all types are implemented.
It additionally processes incoming metrics one by one from a stream, so that we could skip them without first reading and loading all the dump (that may be extremely large). 

## Related issues

Related: https://github.com/pomerium/pomerium-zero/issues/2454

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


wasaga:
that will have to be split differently, will reopen in a new

PR #5126, opened 2024-06-04
core/zero: use exit code 0 for bad requests

calebdoxsey:
## Summary
Instead of exiting with 1 on a bad request error exit with 0.

## Related issues
- https://github.com/pomerium/pomerium-zero/issues/1748
 

## Checklist

- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/cmd/pomerium/main.go
+++ b/cmd/pomerium/main.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/rs/zerolog"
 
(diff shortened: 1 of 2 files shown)


cmd/pomerium/main.go
```diff
@@ -40,6 +42,12 @@ func main() {
 	}
 
 	if err := runFn(ctx); err != nil && !errors.Is(err, context.Canceled) {
+		// if the error was due to a bad request, return 0 as the exit code
+		if strings.Contains(err.Error(), "bad request") {
```
wasaga, line 46: terminal errors are supposed to implement `IsTerminal()` interface we should be able to check that vs error text, please let me double-check if its propagated up to this point

PR #5131, opened 2024-06-08
zero/connect: add telemetry request command

wasaga:
## Summary

Adds a new command to Zero/Connect service to collect telemetry. 

## Related issues

Related: https://github.com/pomerium/pomerium-zero/issues/2453

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


pkg/zero/connect/connect.proto
```diff
@@ -32,6 +33,30 @@ message ConfigUpdated {
...
+
+// SessionAnalyticsRequest is used to request current MAU/DAU data
+message SessionAnalyticsRequest {}
+
+// EnvoyMetricsRequest is used to request current envoy metrics
+message EnvoyMetricsRequest {
+  // only include metrics that match the provided labels, and relabel them with the provided labels
+  map<string, string> metrics = 1;
+  // only include labels that match the provided labels
+  map<string, string> labels = 2;
```
kenjenkins, line 54: Could you expand on these comments? I'm not sure I understand what they're saying.

In particular, what do the values in the `labels` map represent?

kenjenkins (approved):

pkg/zero/connect/connect.proto
wasaga: the second value is a name of a metric to be assigned. however, relabeling may be a responsibility of the telemetry receiver so it may be just `repeated string`.

kenjenkins: Sure, `repeated string` sounds fine to me.

PR #5137, opened 2024-06-14
core/userinfo: remove excess userinfo data

calebdoxsey:
## Summary
We have a lot of unnecessary data on the userinfo dashboard. This PR removes it.

## Related issues
For https://github.com/pomerium/internal/issues/1832
 
## Checklist

- [ ] updated unit tests 
- [x] ready for review


internal/handlers/userinfo.go
```diff
@@ -62,6 +53,62 @@ func (data UserInfoData) ToJSON() map[string]any {
...
+	if data.Session == nil {
+		return nil
+	}
+
+	m := map[string]any{}
+	claims := make(map[string]any)
+	for k, vs := range data.Session.GetClaims() {
+		claims[k] = vs.AsSlice()
+	}
+	m["claims"] = claims
```
kralicky, line 80: are there any claims we might want to omit here? the id token `aud` claim would have pomerium's client id, though it's not a secret and might be necessary for validation purposes

calebdoxsey: I don't know what claims are considered sensitive.

kralicky: Me neither, but I am not 100% confident that none of them are. @wasaga do you know?

kenjenkins (approved):

nhayfield (approved):

PR #5142, opened 2024-06-17
zero/telemetry: collect limited core metrics

wasaga:
## Summary

Adds ability to collect certain limited core metrics (i.e. related to resource pressure) by pulling them via the opencensus bridge and filtering.

## Related issues

Related: https://github.com/pomerium/pomerium-zero/issues/2457

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


pkg/zero/connect/connect.proto
```diff
@@ -55,7 +55,10 @@ message EnvoyMetricsRequest {
 }
 
 // PomeriumMetricsRequest is used to request current pomerium metrics
-message PomeriumMetricsRequest {}
+message PomeriumMetricsRequest {
+  // only include metrics that match the provided labels, and relabel them with the provided labels
```
kralicky, line 59: this comment doesn't really make sense here, is it outdated?

wasaga: yes we moved relabeling to the cloud side indeed

internal/zero/telemetry/opencensus/opencensus.go
```diff
@@ -0,0 +1,63 @@
...
+	"sync/atomic"
+
+	"go.opentelemetry.io/otel/bridge/opencensus"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+)
+
+type Producer struct {
+	producer metric.Producer
+	filter   atomic.Value
```
calebdoxsey, line 15: I think it might be better to use `sets.Hash` with an `atomic.Pointer`. 

`SetFilter` becomes something like:

```go
func (p *Producer) SetFilter(names []string) {
  p.filter.Store(sets.NewHash(names...))
}
```

`loadFilter` becomes something like

```go
func (p *Producer) loadFilter() *sets.Hash[string] {
	p.filter.Load()
}
```

And you can then just call `filter.Has(metric.Name)`.

calebdoxsey (approved):

kralicky (approved):

PR #5143, opened 2024-06-19
authorize: hot path identity provider lookup optimizations

kralicky:
## Summary

GetIdentityProviderForRequestURL is very slow when many routes are configured. This makes it faster by querying against a cache instead of checking every policy.

Some benchmarks:
```
Old/5 policies                                               16034      75547 ns/op   110410 B/op      479 allocs/op
Old/50 policies                                               1405     858198 ns/op  1329344 B/op     5612 allocs/op
Old/500 policies                                               136    8800093 ns/op 13640888 B/op    57342 allocs/op
Old/5000 policies                                               15   79413080 ns/op 138233461 B/op   578966 allocs/op

New/5 policies (domain matching only)                       386138       3119 ns/op     1050 B/op       22 allocs/op
New/50 policies (domain matching only)                      382624       3008 ns/op     1049 B/op       22 allocs/op
New/500 policies (domain matching only)                     391855       3089 ns/op     1076 B/op       22 allocs/op
New/5000 policies (domain matching only)                    360991       2953 ns/op     1087 B/op       23 allocs/op
New/5 policies (path matching only)                         383300       2985 ns/op      953 B/op       22 allocs/op
New/50 policies (path matching only)                        383926       3049 ns/op      953 B/op       22 allocs/op
New/500 policies (path matching only)                       366984       3083 ns/op      980 B/op       22 allocs/op
New/5000 policies (path matching only)                      357968       3024 ns/op      997 B/op       23 allocs/op
New/25 policies (5 domains, 5 paths per domain)             358186       3194 ns/op     1062 B/op       22 allocs/op
New/500 policies (50 domains, 10 paths per domain)          359605       3263 ns/op     1091 B/op       23 allocs/op
New/500 policies (10 domains, 50 paths per domain)          362532       3300 ns/op     1092 B/op       23 allocs/op
New/2500 policies (50 domains, 50 paths per domain)         374217       3206 ns/op     1101 B/op       24 allocs/op
New/5000 policies (100 domains, 50 paths per domain)        357696       3168 ns/op     1102 B/op       24 allocs/op
New/5000 policies (50 domains, 100 paths per domain)        355652       3185 ns/op     1103 B/op       24 allocs/op
New/10000 policies (100 domains, 100 paths per domain)      396895       3162 ns/op     1103 B/op       24 allocs/op
New/25 policies (5 domains, 5 prefixes per domain)          365000       3196 ns/op     1059 B/op       23 allocs/op
New/500 policies (50 domains, 10 prefixes per domain)       360031       3335 ns/op     1091 B/op       23 allocs/op
New/500 policies (10 domains, 50 prefixes per domain)       327415       3831 ns/op     1150 B/op       23 allocs/op
New/2500 policies (50 domains, 50 prefixes per domain)      287326       3736 ns/op     1159 B/op       24 allocs/op
New/5000 policies (100 domains, 50 prefixes per domain)     290416       3693 ns/op     1160 B/op       24 allocs/op
New/5000 policies (50 domains, 100 prefixes per domain)     246459       4345 ns/op     1237 B/op       24 allocs/op
New/10000 policies (100 domains, 100 prefixes per domain)   276199       4416 ns/op     1237 B/op       24 allocs/op
```

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


internal/urlutil/url.go
```diff
@@ -149,6 +149,59 @@ func GetDomainsForURL(u *url.URL, includeDefaultPort bool) []string {
 	return []string{u.Hostname(), net.JoinHostPort(u.Hostname(), defaultPort)}
 }
 
+func AllDomainsForURL(u *url.URL, includeDefaultPort bool) func(yield func(string, string) bool) {
```
calebdoxsey, line 152: The use of closure functions like this is to avoid allocating a slice?

calebdoxsey: Oh I see this is compatible with v1.23's upcoming `Seq2` type?

pkg/cryptutil/kek.go
```diff
@@ -4,10 +4,9 @@ import (
 	"crypto/rand"
 	"fmt"
 
+	"github.com/akamensky/base58"
```
calebdoxsey, line 7: So for some context, we removed references to the the original base58 library (and directly embedded it) because people didn't like our dependency on a bitcoin library. Assuming this library produces the same results I suppose that's better.

kralicky: I probably should have mentioned it in the description - I swapped this library out because profiling revealed the base58 encoding (primarily bigint math) as a significant portion of the runtime for this, so I found this library which has some optimizations that result in big performance gains

internal/urlutil/url.go
kralicky: Yep, this was to avoid duplicating some slices - when go 1.23 is released it can be rewritten as `for host, port := range AllDomainsForUrl(...)` but for now we call it "manually" like `AllDomainsForUrl(func(host, port){ ... })` which has the same effect

pkg/cryptutil/kek.go
kralicky: That said, it's still super slow, so perhaps it might be a good idea to switch to a different encoding scheme in the future

calebdoxsey (approved):

pkg/cryptutil/kek.go
calebdoxsey: `base58` was used as a dense encoding that's file safe. Certainly open to alternatives.

calebdoxsey: I think [base62](https://pkg.go.dev/github.com/jxskiss/base62) might be a good option. We can't swap out all usage of base58 but I think we could this one if we want to.

kralicky: Might be worth benchmarking :eyes:

kralicky:
Closing in favor of https://github.com/pomerium/pomerium/pull/5167

Issue #5138, opened 2024-06-16
Live config change causing autocert to fail, OK on restart


bluecmd:
## What happened?

It seems that when I add a new domain to a running Pomerium instance the LE ACME certification process fails. If I restart Pomerium, it succeeds without issue.

## What did you expect to happen?

Live-adding a new domain into the configuration causes the ACME integration to successfully obtain a certificate.

## How'd it happen?

1. Added the following stanza:
```yaml
  - from: https://test1.local.sonix.network
    to: http://127.0.0.1
    host_rewrite_header: true
    allowed_users: *sonix_admins
```

2. Certificate is not successfully obtained
3. Restart Pomerium
4. Certificate is successfully obtained

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`):
```
pomerium: 0.26.0-1715969560+adb5f781
envoy: 1.30.1+7571ddfa650801c56d470f3507c3e7a5fd1f71591416dc0c2a42b07627aefadd
```

- Server Operating System/Architecture/Cloud: Rocky Linux 9.4

Nginx on port :80 doing 301 redirects for http:// requests to https://. See additional section.

## What's your config.yaml?

```config.yaml
address: '[::]:443'
authenticate_service_url: https://authenticate.sonix.network
autocert: true
autocert_dir: /etc/pomerium/certs
cookie_secret: xyz
idp_provider: "github"
idp_client_id: x
idp_client_secret: y
jwt_claims_headers: "email,user,sub"
signing_key_file: "/etc/pomerium/keys/ec_private.pem"
metrics_address: :9090
policy:
[..]
  - from: https://test1.local.sonix.network
    to: http://127.0.0.1
    host_rewrite_header: true
    allowed_users: *sonix_admins
[..]
```

## What did you see in the logs?

```logs
Jun 16 14:29:03 pomerium pomerium[270191]: {"level":"info","config_file_source":"/etc/pomerium/config.yaml","config_change_id":"fb72993c-5f49-4e43-975e-420632b0834a","domain":"test1.local.sonix.network","time":"2024-06-16T14:29:03+02:00","message":"obtaining certificate"}
[...]

PR #5150, opened 2024-06-25
core/autocert: fix filter chain, handshake

calebdoxsey:
## Summary
Fix several issues with autocert:

1. Disable HTTP challenges when the redirect server is not enabled.
2. At some point the TLS ALPN filter was removed from envoy, so the autocert code was never being called. How it works at all is a mystery to me, but because of the missing envoy filter it was impossible for certificates to be provisioned via TLS ALPN challenges at runtime.
3. We weren't calling `Handshake` after accepting the TLS connection. This is necessary to complete the ACME protocol.

## Related issues
- https://github.com/pomerium/pomerium/issues/5138

## Checklist
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/internal/autocert/manager.go
+++ b/internal/autocert/manager.go
@@ -52,9 +52,10 @@ type Manager struct {
 	acmeMgr   atomic.Pointer[certmagic.ACMEIssuer]
 	srv       *http.Server
 
-	acmeTLSALPNLock   sync.Mutex
-	acmeTLSALPNPort   string
-	acmeTLSALPNConfig *tls.Config
+	acmeTLSALPNLock     sync.Mutex
+	acmeTLSALPNPort     string
+	acmeTLSALPNListener net.Listener
+	acmeTLSALPNConfig   *tls.Config
 
 	*ocspCache
 
@@ -158,7 +159,7 @@ func (mgr *Manager) getCertMagicConfig(ctx context.Context, cfg *config.Config) 
 		}
 	}
 	acmeMgr := certmagic.NewACMEIssuer(mgr.certmagic, mgr.acmeTemplate)
-	acmeMgr.DisableHTTPChallenge = cfg.Options.HTTPRedirectAddr == ""
+	acmeMgr.DisableHTTPChallenge = !shouldEnableHTTPChallenge(cfg)
 	err = configureCertificateAuthority(acmeMgr, cfg.Options.AutocertOptions)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


internal/autocert/manager.go
```diff
@@ -364,6 +376,20 @@ func (mgr *Manager) updateACMETLSALPNServer(ctx context.Context, cfg *config.Con
...
 			}
+
+			// initiate the TLS handshake
+			mgr.acmeTLSALPNLock.Lock()
+			tlsConfig := mgr.acmeTLSALPNConfig.Clone()
+			mgr.acmeTLSALPNLock.Unlock()
+
+			orig := tlsConfig.GetCertificate
+			tlsConfig.GetCertificate = func(chi *tls.ClientHelloInfo) (*tls.Certificate, error) {
+				log.Info(ctx).Str("server-name", chi.ServerName).
```
kenjenkins, line 387: Nice, adding some logging here seems useful.

internal/autocert/manager.go
```diff
@@ -155,6 +158,7 @@ func (mgr *Manager) getCertMagicConfig(ctx context.Context, cfg *config.Config)
 		}
 	}
 	acmeMgr := certmagic.NewACMEIssuer(mgr.certmagic, mgr.acmeTemplate)
+	acmeMgr.DisableHTTPChallenge = cfg.Options.HTTPRedirectAddr == ""
```
kenjenkins, line 161: I think we might even want to strengthen this condition, and require that the redirect address uses port 80, as the http-01 challenge can only be done on port 80.

(see the [Let's Encrypt docs](https://letsencrypt.org/docs/challenge-types/#:~:text=The%20HTTP%2D01%20challenge%20can%20only%20be%20done%20on%20port%2080.%20Allowing%20clients%20to%20specify%20arbitrary%20ports%20would%20make%20the%20challenge%20less%20secure%2C%20and%20so%20it%20is%20not%20allowed%20by%20the%20ACME%20standard.))

internal/autocert/manager.go
```diff
@@ -342,20 +346,28 @@ func (mgr *Manager) updateServer(ctx context.Context, cfg *config.Config) {
...
 	}
 
-	tlsConfig := mgr.certmagic.TLSConfig()
-	ln, err := tls.Listen("tcp", addr, tlsConfig)
+	// store the updated port
+	mgr.acmeTLSALPNPort = cfg.ACMETLSALPNPort
+
+	// start the listener
+	addr := net.JoinHostPort("127.0.0.1", cfg.ACMETLSALPNPort)
+	ln, err := net.Listen("tcp", addr)
```
kenjenkins, line 364: Is there anything that will signal a previous listener to stop listening (if the `ACMETLSALPNPort` changes)?

Should we be storing the `ln` reference on `Manager` struct and call `Close()` here if there's an existing listener? (It looks like `updateServer()` does something similar to this.)

kenjenkins (approved):
Nice debugging work!

PR #5165, opened 2024-07-05
core/authorize: require new login when authenticate url changes

calebdoxsey:
## Summary
Currently if the authenticate URL changes sessions are not invalidated. When switching between local and hosted authenticate we would expect users to have to login again, so this PR updates the code to include the authenticate URL as part of the identity provider options and hash. This causes the authorize service to drop the session when it's loaded and triggers the login flow.

Also switch from `base58` to `base62` for these IDs, a performance issue @kralicky flagged earlier.

`base58`:

```
goos: linux
goarch: amd64
pkg: github.com/pomerium/pomerium/pkg/grpc/identity
cpu: AMD Ryzen 7 5800H with Radeon Graphics         
BenchmarkHash-16    	  458904	      6188 ns/op	     936 B/op	      50 allocs/op
```

`base62`:

```
goos: linux
goarch: amd64
pkg: github.com/pomerium/pomerium/pkg/grpc/identity
cpu: AMD Ryzen 7 5800H with Radeon Graphics         
BenchmarkHash-16    	 1809844	       587.7 ns/op	     400 B/op	       4 allocs/op
```

## Related issues
- https://github.com/pomerium/internal/issues/1844

## Checklist
- [x] updated unit tests
- [x] ready for review


kenjenkins (approved):
Looks good to me!

PR #5167, opened 2024-07-09
Optimize IDP lookup using envoy route metadata

kralicky:
Supersedes https://github.com/pomerium/pomerium/pull/5143

## Summary

GetIdentityProviderForRequestURL is very slow when many routes are configured, because it will match routes by checking each one individually. This change makes use of the envoy route metadata passed to our ext_authz handler so we can avoid having to match urls to routes ourselves.

For programmatic login (and device auth), this lookup is accomplished by making a new request to envoy with the same URL, and with the internal header `X-Pomerium-Check-Route` set. When this header is present, our ext_authz handler will deny the request but return the envoy context extensions in the response body. The current implementation uses HTTP `TRACE` to make this request to envoy, which seemed appropriate, but I might change it to just use the same verb as the incoming request, since I don't think it makes any practical difference.

## Related issues

Fixes https://github.com/pomerium/internal/issues/1833

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/luascripts/drop-check-route-requests.lua
+++ b/config/envoyconfig/luascripts/drop-check-route-requests.lua
@@ -1,0 +1,7 @@
+function envoy_on_request(request_handle)
+  local headers = request_handle:headers()
+  if headers:get("x-pomerium-check-route") ~= nil or headers:get(":method") == "TRACE" then
+    request_handle:logErr("check-route request caught by filter")
+    request_handle:respond({[":status"] = "500"}, "Internal Server Error")
+  end
+end
--- a/proxy/match.go
+++ b/proxy/match.go
@@ -1,0 +1,71 @@
+package proxy
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/pomerium/pomerium/internal/atomicutil"
+	"github.com/pomerium/pomerium/internal/urlutil"
+)
+
+type RouteMatcher interface {
+	GetIdentityProviderForRequest(r *http.Request) (string, error)
+}
+
+type RouteMatcherFunc func(r *http.Request) (string, error)
+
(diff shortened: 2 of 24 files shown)


config/envoyconfig/luascripts/drop-check-route-requests.lua
```diff
@@ -0,0 +1,7 @@
+function envoy_on_request(request_handle)
+  local headers = request_handle:headers()
+  if headers:get("x-pomerium-check-route") ~= nil or headers:get(":method") == "TRACE" then
```
calebdoxsey, line 3: Does `TRACE` ever occur in the wild?

kralicky: unlikely; it's usually disabled on web servers, and entirely unsupported by browsers. I originally used it here as an extra precaution to basically guarantee that even if somehow both our ext_authz filter and this filter let the request through, it would still be a no-op. However, it is probably not necessary, if that situation happens then we have much bigger problems...

proxy/match.go
```diff
@@ -0,0 +1,71 @@
...
+			Path: r.FormValue(urlutil.QueryRequestPath),
+		}),
+	)
+
+	ctx, ca := context.WithTimeout(r.Context(), 10*time.Second)
+	defer ca()
+	req, _ := http.NewRequestWithContext(ctx, r.Method, requestURL.String(), nil)
+	req.Header.Set("X-Pomerium-Check-Route", "1")
+	req.Header.Set("Accept", "application/json")
+	resp, err := http.DefaultTransport.RoundTrip(req)
```
wasaga, line 48: what if there's a client certificate requirement for a route?

kralicky: That wouldn't affect the route matching, would it?

wasaga: wouldn't request fail on TLS handshake, before route matching can happen?

kralicky: oh, good point. I'll look into this

Issue #5176, opened 2024-07-18
Support OtherName(UPN) SAN for mTLS Verification


arulthileeban:
Title: Support OtherName(UPN) SAN for mTLS Verification

**Is your feature request related to a problem? Please describe.**
Currently, Envoy supports SAN value matching against DNS, URI, Email, and IP Address types, which are utilized by Pomerium. In the context of Pomerium mTLS with downstream clients being user devices, client certificates are typically user certificates. MDM solutions like Jamf and Intune integrating with Cert Vendors (Eg: Digicert) only allow Email SAN in user certificates. For any SAN that requires a customized value, they resort to the OtherName option, primarily using UserPrincipalName (UPN). Due to the lack of support for OtherName verification in Envoy, this option looks to not be supported in Pomerium. Thus, we are not able to enforce mTLS Verification based on the SAN OtherName type.

**Describe the solution you'd like**
To enable verification against OtherName SAN value, we added a [change](https://github.com/envoyproxy/envoy/pull/34471) to Envoy to support this, which will be released in the latest version 1.31.0 (Pending release with expected date [07/16](https://github.com/envoyproxy/envoy/blob/main/RELEASES.md#major-release-schedule)). Could this change be bubbled up to Pomerium?

Envoy uses OID as a parameter to describe the OtherName type to match against. Either the change can just support only UPN, which is widely used (or) accept OID through config.

PR #5177, opened 2024-07-19
add mTLS UserPrincipalName SAN match

kenjenkins:
## Summary

Add a new `user_principal_name` type to the downstream mTLS `match_subject_alt_names` option.

UserPrincipalName is a SubjectAltName [OtherName](https://datatracker.ietf.org/doc/html/rfc3280#:~:text=OtherName%20%3A%3A%3D%20SEQUENCE%20%7B%0A%20%20%20%20%20%20%20%20type%2Did%20%20%20%20OBJECT%20IDENTIFIER%2C%0A%20%20%20%20%20%20%20%20value%20%20%20%20%20%20%5B0%5D%20EXPLICIT%20ANY%20DEFINED%20BY%20type%2Did%20%7D) with type-id 1.3.6.1.4.1.311.20.2.3 and a UTF8String as the value type.

This change has two parts:
1. For the `reject_connection` enforcement mode, we need to populate the corresponding Envoy setting when building the Envoy listener configuration. (The OTHER_NAME option is newly added in Envoy 1.31, and requires an update to the Envoy API proto dependency for the new `oid` field.)
2. For the other enforcement modes, we need to add new logic to the authorize/evaluator package. The crypto/x509 standard library package does not have first-class support for UserPrincipalName, so we need to manually parse the raw SubjectAltName extension data.

Add unit tests. Unfortunately, adding a new test cert in the authorize/evaluator package requires regenerating the existing ones as well, because the CA key isn't saved.

Based on initial work by @arulthileeban. (thanks!)

## Related issues

- https://github.com/pomerium/pomerium/issues/5176

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


authorize/evaluator/functions_test.go
```diff
@@ -452,3 +486,64 @@ func TestClientCertConstraintsFromConfig(t *testing.T) {
...
+func TestGetUserPrincipalNamesFromSAN(t *testing.T) {
+	b64 := func(raw string) []byte {
+		b, err := base64.StdEncoding.DecodeString(raw)
+		require.NoError(t, err)
+		return b
+	}
+
+	// One UserPrincipalName.
+	names, err := getUserPrincipalNamesFromSAN(b64(`
+MBegFQYKKwYBBAGCNxQCA6AHDAVoZWxsbw==`))
```
kralicky, line 499: might be better to build these from asn1.Marshal (like in gen-test-certs.go)

authorize/evaluator/functions.go
```diff
@@ -256,3 +270,58 @@ func parseCertificate(pemStr string) (*x509.Certificate, error) {
 	}
 	return x509.ParseCertificate(block.Bytes)
 }
+
+var (
+	oidSubjectAltName    = asn1.ObjectIdentifier{2, 5, 29, 17}
+	oidUserPrincipalName = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 311, 20, 2, 3}
+	otherNameTag         = cb_asn1.Tag(0).Constructed().ContextSpecific()
+	upnValueTag          = cb_asn1.Tag(0).Constructed().ContextSpecific()
```
kralicky, line 278: suggest renaming to something like `otherNameValueTag` to make it more obvious where this tag id comes from

authorize/evaluator/functions.go
```diff
@@ -256,3 +270,58 @@ func parseCertificate(pemStr string) (*x509.Certificate, error) {
...
+}
+
+func getUserPrincipalNamesFromCert(cert *x509.Certificate) ([]string, error) {
+	for _, ext := range cert.Extensions {
+		if ext.Id.Equal(oidSubjectAltName) {
+			return getUserPrincipalNamesFromSAN(ext.Value)
+		}
+	}
+	return nil, nil
+}
```
kralicky, line 327: To handle the case where e.g. there are _only_ UPN/OtherName SANs in the certificate, I believe we need to check for the ID in `cert.UnhandledCriticalExtensions`, and remove it if we successfully read a UPN from the extensions list

kenjenkins: I think we can unconditionally remove the SAN extension ID from `UnhandledCriticalExtensions` if it exists there. The rationale being that we support certain SAN matching options, and we enforce no particular SAN requirements beyond what is explicitly configured by those SAN matching options.

kralicky (approved):

PR #5184, opened 2024-07-22
Optimize policy iterators

kralicky:
## Summary

This modifies (*Options).GetAllPolicies() to use a go 1.23 iterator instead of copying all policies on every call, which can be extremely expensive. All existing usages of this function were updated as necessary.

Additionally, a new (*Options).NumPolicies() method was added which quickly computes the number of policies that would be given by GetAllPolicies(), since there were several usages where only the number of policies was needed.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


calebdoxsey:
LGTM but I would vote for waiting for the full release of go 1.23. Core Pomerium is used as a library by several projects.

calebdoxsey (approved):

PR #5185, opened 2024-07-22
Optimize Policy RouteID and Checksum

kralicky:
## Summary

This significantly optimizes the (*Policy).RouteID() and (*Policy).Checksum() methods for both speed and memory usage.

A new method (*Policy).ChecksumWithID(uint64) can be used to skip a call to RouteID() if the ID is already known. Checksum() is implemented in terms of this new method, and will always recompute the route ID on each call.

RouteID() does not allocate heap memory. Checksum() may allocate heap memory, depending on which fields are set. If all of the following are true, Checksum() makes zero allocations:
1. The policy uses redirect or direct-response mode
2. The policy has no sub-policies
3. The policy has no response header rewrite config

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/config/policy.go
+++ b/config/policy.go
@@ -4,15 +4,22 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/base64"
+	"encoding/binary"
 	"fmt"
 	"net/url"
 	"os"
 	"regexp"
+	"slices"
 	"sort"
+	"strconv"
 	"strings"
+	"sync"
 	"time"
+	"unsafe"
 
+	"github.com/cespare/xxhash/v2"
 	envoy_config_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	"github.com/valyala/bytebufferpool"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/durationpb"
 
@@ -113,7 +120,7 @@ type Policy struct {
 	TLSClientKey      string           `mapstructure:"tls_client_key" yaml:"tls_client_key,omitempty"`
 	TLSClientCertFile string           `mapstructure:"tls_client_cert_file" yaml:"tls_client_cert_file,omitempty"`
 	TLSClientKeyFile  string           `mapstructure:"tls_client_key_file" yaml:"tls_client_key_file,omitempty"`
-	ClientCertificate *tls.Certificate `yaml:",omitempty" hash:"ignore"`
+	ClientCertificate *tls.Certificate `yaml:",omitempty" hash:"-"`
 
 	// TLSDownstreamClientCA defines the root certificate to use with a given route to verify
 	// downstream client certificates (e.g. from a user's browser).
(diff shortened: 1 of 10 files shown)


config/policy.go
```diff
@@ -568,35 +637,171 @@ func (p *Policy) Validate() error {
...
+	pr.reset()
+	policyPool.Put(pr)
+	return sum
+}
 
+// RouteID returns a unique identifier for a route.
+// Only the fields From, To, Redirect, and Response are used to compute the ID.
+func (p *Policy) RouteID() (uint64, error) {
+	// this function is in the hot path, try not to allocate too much memory here
+	hash := xxhash.New()
```
calebdoxsey, line 750: I had an idea. What about creating a new `Digest` type in `hashutil` to wrap `xxhash.Digest` so we can add these additional methods:

```go
package hashutil

import (
	"encoding/binary"

	"github.com/cespare/xxhash/v2"
)

// Digest is a hash digest computed via xxhash.Digest.
type Digest struct {
	xxhash.Digest
}

// NewDigest creates a new Digest.
func NewDigest() *Digest {
	return &Digest{
		Digest: *xxhash.New(),
	}
}

// WriteBool writes a bool.
func (hash *Digest) WriteBool(b bool) {
	if b {
		hash.Write([]byte{1})
	} else {
		hash.Write([]byte{0})
	}
}

// WriteUint32 writes a uint32.
func (hash *Digest) WriteUint32(t uint32) {
	var buf [4]byte
	binary.LittleEndian.PutUint32(buf[:], t)
	hash.Write(buf[:])
}

// WriteInt32 writes an int32.
func (hash *Digest) WriteInt32(t int32) {
	var buf [4]byte
	binary.LittleEndian.PutUint32(buf[:], uint32(t))
	hash.Write(buf[:])
}
```

Maybe could add a `WriteNullableString` (or similarly named) method to cut down on all the ifs too.

calebdoxsey: Incidentally, for those nullable fields, we might need to write a `0` byte or something... otherwise how to do we tell the difference between:

```
{ "httpsRedirect": "VALUE" }
```

and

```
{ "schemeRedirect": "VALUE" }
```

wouldn't they hash to the same thing?

config/policy.go
```diff
@@ -607,6 +812,26 @@ func (p *Policy) MustRouteID() uint64 {
...
+
+func writeUint32(hash *xxhash.Digest, t uint32) {
+	var buf [4]byte
+	binary.LittleEndian.PutUint32(buf[:], t)
+	hash.Write(buf[:])
+}
+
+func writeInt32(hash *xxhash.Digest, t int32) {
+	var buf [4]byte
+	binary.LittleEndian.PutUint32(buf[:], *(*uint32)(unsafe.Pointer(&t)))
```
calebdoxsey, line 831: I ran a benchmark and didn't see much difference between `*(*uint32)(unsafe.Pointer(&t))` and `uint32(t)`. Any other reason to do it this way?

kralicky: Oops, yeah this can just be uint32(t)

kralicky: This method is probably redundant anyway

config/policy.go
kralicky: Good idea; and also yes you are correct, this is missing some null bytes (also unit tests)

internal/hashutil/hashutil.go
```diff
@@ -27,3 +29,167 @@ func Hash(v any) (uint64, error) {
...
 }
+
+// MapHash efficiently computes a non-cryptographic hash of a map of strings.
+func MapHash(iv uint64, m map[string]string) uint64 {
+	accum := iv
+	var lenBuf [4]byte
+	binary.BigEndian.PutUint32(lenBuf[:], uint32(len(m)))
+	accum ^= xxhash.Sum64(lenBuf[:])
+	var kvBuf [16]byte
+	for k, v := range m {
```
calebdoxsey, line 40: Is this using the same trick as hashstructure? XORing all of them together so iteration order doesn't matter?

calebdoxsey (approved):
I guess it is what it is, but this is going to bite us so hard with memory re-use bugs.

calebdoxsey:
To clarify my comment:

This pull request improves performance by (1) implementing a custom hashing algorithm for routes and (2) using a sync pool to reuse memory and avoid garbage collector pressure.

I agree that both approaches can improve performance. I attempted to review the code associated with them and identify any issues. I approved this PR because I found no further issues.

That said I do not know the motivation behind the changes. I have not been part of those discussions. I assume the changes are necessary. Though, in general, philosophically, I don't agree with them. 

I think we should try to use a library like hashstructure to generate this hash, because I believe there are subtle edge cases around generating hashes that those libraries already account for, and getting those edge cases wrong will cause bugs which are very hard to diagnose. 

I also believe that garbage collection is a good thing and that Go made the right decision to have a garbage collector. Yes there is a performance cost, but it also prevents many bugs. If we forget to erase one slice, or update one pointer in the struct incorrectly, our code will share memory between objects inadvertently. A route could end up with the wrong headers or the wrong security policy.

I am not confident in my ability to avoid these bugs so I try to look for other ways to improve performance. If we're at the point where this is all we can do then I guess we have no choice, though if that's true, an honest evaluation would say we should not be using Go at all to write this product.

internal/hashutil/hashutil.go
kralicky: Yes, it is the same except this allows for domain separation using a custom initialization value for the hash. this is done because we remove the map fields before performing proto serialization (since they get copied/sorted internally when using deterministic mode) then add the hashes of those maps at the end. this would mess up the field ordering so the hash of the field name is used as the starting value to distinguish between multiple map fields in a single message.

kralicky:
Appreciate the feedback. I don't want to merge these changes unless we are satisfied with them though, so I will take another pass at some areas where we can improve safety and try to catch future bugs.

These changes were made in conjunction with other PRs (I just split them up for ease of review), specifically to optimize this function https://github.com/pomerium/pomerium/pull/5186/files#diff-3a0c95710d25065101c5c3463d0ccaf0f58d08617910ebc35997e45467846c1fR348 as much as possible. Profiling and tracing revealed that as the number of routes scales up, the majority of the time it takes to update the config is spent computing policy hashes. With the current config system, we don't know what specifically was changed during a config update, so we have to recompute the hashes for every policy to check them. 

The focus on reducing memory allocations here is because the amount of memory we were allocating was putting tremendous pressure on the GC. This can be seen clearly when comparing the gc trace and memory stats after loading a large config file with/without GOGC=off:


The config I am using has 5000 routes that look like this:
```yaml
routes:
  - &route
    from: https://route1.localhost.pomerium.io:8443
    response:
      status: 200
      body: OK
    policy:
      - allow:
          and:
            - email:
                is: test@example.com
  - <<: *route
    from: https://route2.localhost.pomerium.io:8443
  - <<: *route
    from: https://route3.localhost.pomerium.io:8443
    ...
  - <<: *route
    from: https://route5000.localhost.pomerium.io:8443
```

This is about where memory usage settles after startup, so in this test we exit here.
```patch
--- a/pkg/grpc/databroker/syncer.go
+++ b/pkg/grpc/databroker/syncer.go
@@ -158,6 +158,9 @@ func (syncer *Syncer) sync(ctx context.Context) error {
 	}
 
 	log.Debug(ctx).Msg("listening for updates")
+	if syncer.cfg.typeURL == "type.googleapis.com/pomerium.config.Config" {
+		os.Exit(1)
+	}
 
 	for {
 		res, err := stream.Recv()
```

### main branch:

1. `GODEBUG=gctrace=1 /usr/bin/time -v bin/pomerium -config dev/5000-routes.yaml`
[5000-routes-gc.log](https://github.com/user-attachments/files/16431614/5000-routes-gc.log)
summary: 785 GC cycles, maximum RSS 5.02GB
	```
	User time (seconds): 243.23
	System time (seconds): 4.56
	Percent of CPU this job got: 804%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:30.79
	Maximum resident set size (kbytes): 5029816
	```

2. `GOGC=off GODEBUG=gctrace=1 /usr/bin/time -v bin/pomerium -config dev/5000-routes.yaml`
[5000-routes-nogc.log](https://github.com/user-attachments/files/16431629/5000-routes-nogc.log)
summary: no GC, maximum RSS **101.62GB**
	```
	User time (seconds): 46.88
	System time (seconds): 11.06
	Percent of CPU this job got: 193%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:29.90
	Maximum resident set size (kbytes): 101622888
	```

### this branch (no other optimization changes):
1. `GODEBUG=gctrace=1 /usr/bin/time -v bin/pomerium -config dev/5000-routes.yaml`
[5000-routes-gc-new.log](https://github.com/user-attachments/files/16431711/5000-routes-gc-new.log)
summary: 132 GC cycles, maximum RSS 5.06GB
	```
	User time (seconds): 97.86
	System time (seconds): 2.48
	Percent of CPU this job got: 462%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:21.71
	Maximum resident set size (kbytes): 5066996
	```

2. `GOGC=off GODEBUG=gctrace=1 /usr/bin/time -v bin/pomerium -config dev/5000-routes.yaml`
[5000-routes-nogc-new.log](https://github.com/user-attachments/files/16431717/5000-routes-nogc-new.log)
summary: no GC, maximum RSS **22.82GB**
	```
	User time (seconds): 40.86
	System time (seconds): 6.18
	Percent of CPU this job got: 284%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:16.52
	Maximum resident set size (kbytes): 22823620
	```

This change saves 80GB total memory allocation and thus triggers 653 fewer GC cycles when loading the config file.

I absolutely agree that the garbage collector is a good thing and Go's GC especially is fantastic. In 99.9% of cases we should never need to worry about it or try to go out of our way to optimize for memory usage to avoid it, but this may be a special case where it is worth the tradeoffs. I don't want it to cause issues later on though, so I will try to make some things safer where possible and add some more tests that can check for memory safety.

config/policy_test.go
```diff
@@ -344,28 +373,947 @@ func TestPolicy_SortOrder(t *testing.T) {
...
+		// to use later is stored in a uintptr
+		debug.SetGCPercent(-1)
+		t.Cleanup(func() {
+			debug.SetGCPercent(100)
+		})
+
+		{
+			p := newStaticPolicy()
+			checksum := p.Checksum()
+			lp := uintptr(unsafe.Pointer(p)) // launder a copy of p so the GC can't see it
```
kenjenkins, line 921: Can you help me understand what behavior this test is exercising?

calebdoxsey (approved):

config/policy.go
```diff
@@ -333,115 +350,190 @@ func NewPolicyFromProto(pb *configpb.Route) (*Policy, error) {
...
 			Header: rwh.Header,
 			Matcher: &configpb.RouteRewriteHeader_Prefix{
 				Prefix: rwh.Prefix,
 			},
 			Value: rwh.Value,
 		})
 	}
+}
+
+func copySrcToOptionalDest[T comparable](dst **T, src *T) {
```
kenjenkins, line 489: Can you help me understand this method? Why does it sometimes copy a pointer and sometimes copy a value?

kralicky: `src` is a pointer to a (non-pointer/non-optional) struct field, and `dst` is a pointer to a (pointer/optional) struct field. This does the following:
- If `*src` is a zero-valued `T`, `*dst` is set to nil. This is a best-effort preservation of optional semantics. fields for which a value of zero is valid will be unset anyway, but there is no way around this.
- Otherwise (`*src` has a non-zero value)
	- If `*dst` is a non-nil pointer, copy the value `*src` into `**dst`, to reuse the heap memory already pointed to by `*dst` (if possible; strings etc. may or may not end up needing to be resized)
	- If `*dst` is a nil pointer, set `*dst` to `src` directly. the pointer field in the destination struct now holds an address interior to the source struct.

kenjenkins: > * If `*dst` is a non-nil pointer, copy the value `*src` into `**dst`, to reuse the heap memory already pointed to by `*dst` (if possible; strings etc. may or may not end up needing to be resized)

This is the case I find confusing. In what circumstances is it better to copy the value rather than the pointer?

Or to put the question another way, if it's OK to alias parts of the source struct, why not always copy the `src` pointer?

kralicky: To reduce GC pressure. if we overwrite the original pointer, that memory needs to be garbage collected. If we copy the value, it doesn't. This allows using a sync.Pool to store commonly used fields that need to be heap allocated.

kenjenkins: OK, I think I understand — this trades additional memory usage for reduced GC time?

kralicky: Both, this ends up drastically reducing memory usage, which helps achieve the actual goal of reducing GC time.

PR #5186, opened 2024-07-22
Optimize evaluator

kralicky:
## Summary

This optimizes the Evaluator in the Authorize service to scale to very large numbers of routes. Additional caching was also added when building rego policy query evaluators in parallel to allow sharing work and to avoid building evaluators for scripts with the same contents.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/authorize/evaluator/evaluator.go
+++ b/authorize/evaluator/evaluator.go
@@ -2,20 +2,27 @@
 package evaluator
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
+	"maps"
+	"math/bits"
 	"net/http"
 	"net/url"
+	"runtime"
+	rttrace "runtime/trace"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-jose/go-jose/v3"
 	"github.com/open-policy-agent/opa/rego"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/singleflight"
 
 	"github.com/pomerium/pomerium/authorize/internal/store"
 	"github.com/pomerium/pomerium/config"
-	"github.com/pomerium/pomerium/internal/errgrouputil"
 	"github.com/pomerium/pomerium/internal/httputil"
 	"github.com/pomerium/pomerium/internal/log"
 	"github.com/pomerium/pomerium/internal/telemetry/trace"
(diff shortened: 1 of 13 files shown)


authorize/evaluator/policy_evaluator.go
```diff
@@ -145,41 +156,49 @@ func NewPolicyEvaluator(
...
-		if err != nil {
-			return nil, err
+	// for each script, create a rego object and prepare a query.
+	rttrace.WithRegion(ctx, "Compile Rego", func() {
+		numCached := 0
+		for i := range e.queries {
+			var cached bool
+			e.queries[i].PreparedEvalQuery, cached, err = cache.LookupOrBuild(&e.queries[i], func() (rego.PreparedEvalQuery, error) {
+				r := rego.New(
+					rego.Store(store),
```
calebdoxsey, line 166: Looking at the code there's a potential issue here involving the `store` variable. If it were changed I don't think we'd detect the change and end up using a prepared query with the wrong store.

That said, it looks like we never actually change the `store` so this bug won't happen. But it seems like a footgun.

kralicky: Good catch, I will add cache invalidation here if the store changes.

authorize/evaluator/evaluator.go
```diff
@@ -142,61 +288,224 @@ func New(
...
+				chunkStatus |= 1 << i
+			}
+		}
+		wctx.StatusBits[chunkIdx] = chunkStatus
+		popcnt := bits.OnesCount64(chunkStatus)
+		rttrace.Logf(wctx, "", "chunk %d: %d/%d changed", chunkIdx, popcnt, limit)
+		wctx.StatusCounts[chunkIdx] = uint8(popcnt)
+		wctx.TotalModified.Add(int32(popcnt))
+	}
+}
```
calebdoxsey, line 361: I'm sorry but I'm not really following what this code is doing. I didn't really understand the original code either.

calebdoxsey: (I mean all this building, computing checksums, worker chunking, bit manipulation, etc)

I think there are too many things going on here put together and I can't pull them apart to understand it.

kralicky: Yeah sorry this definitely needs a big comment explaining what's going on :smile:

calebdoxsey (approved):

PR #5187, opened 2024-07-22
Optimize envoy config building

kralicky:
## Summary

This optimizes the envoy config builder to scale to a very large number of routes, by aggresively caching and reusing policy, url, and route objects. This significantly reduces GC overhead and minimizes the amount of time we spend duplicating work while building the envoy route list from the configuration.

There is still more room for optimization here; in particular, use of the new go 1.23 `unique` package might provide some performance improvements around URL comparisons.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/route_configurations.go
+++ b/config/envoyconfig/route_configurations.go
@@ -2,64 +2,93 @@ package envoyconfig
 
 import (
 	"context"
+	"fmt"
+	"strings"
+
+	rttrace "runtime/trace"
 
 	envoy_config_route_v3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 
 	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/internal/telemetry/trace"
 )
 
 // BuildRouteConfigurations builds the route configurations for the RDS service.
-func (b *Builder) BuildRouteConfigurations(
+func (b *ScopedBuilder) BuildRouteConfiguration(
 	ctx context.Context,
-	cfg *config.Config,
-) ([]*envoy_config_route_v3.RouteConfiguration, error) {
+) (*envoy_config_route_v3.RouteConfiguration, error) {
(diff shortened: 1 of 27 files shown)


internal/urlutil/url.go
```diff
@@ -52,6 +57,55 @@ func ParseAndValidateURL(rawurl string) (*url.URL, error) {
 	return u, nil
 }
 
+type SharedURL struct {
+	*url.URL
+	initDone uint32
+	hostname func() string
```
calebdoxsey, line 63: Would it be simpler to pre-compute the hostname?

```go
type SharedURL struct {
	*url.URL
	hostname string
}

func (s *SharedURL) Hostname() string {
	return s.hostname
}

var urlCache sync.Map // map[string]url.SharedURL
```

Does using singleflight with a sync.Map exhibit better performance than just a mutex and normal map? I've only ever used singleflight with IO-bound tasks.

config/envoyconfig/route_configurations.go
```diff
@@ -30,36 +32,63 @@ func (b *Builder) BuildRouteConfigurations(
...
+func (b *Builder) buildRouteConfiguration(name string, virtualHosts []*envoy_config_route_v3.VirtualHost) (*envoy_config_route_v3.RouteConfiguration, error) {
+	return &envoy_config_route_v3.RouteConfiguration{
+		Name:         name,
+		VirtualHosts: virtualHosts,
+		// disable cluster validation since the order of LDS/CDS updates isn't guaranteed
+		ValidateClusters: &wrapperspb.BoolValue{Value: false},
+	}, nil
+}
+
+func (b *Builder) buildMainRouteConfiguration() (*envoy_config_route_v3.RouteConfiguration, error) {
```
calebdoxsey, line 44: Is there some way to break this method up? I'm having a hard time following the catch-all loop

calebdoxsey (approved):

internal/urlutil/url.go
kralicky: This can probably be improved, yeah

PR #5188, opened 2024-07-23
cert-index: add delete, fix usage bitmask

wasaga:
## Summary

Adds `Delete` method to `CertificatesIndex` so that certificate data may be deleted from the index.
Also fixes an issue that overlapping certs with non-conflicting usages would not persist existing bits.

## Related issues

Related: https://github.com/pomerium/pomerium-console/issues/4158

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/pkg/cryptutil/certificates_index.go
+++ b/pkg/cryptutil/certificates_index.go
@@ -32,6 +32,14 @@ func (c *CertificatesIndex) Add(cert *x509.Certificate) {
 	}
 }
 
+// Delete deletes a certificate from the index.
+func (c *CertificatesIndex) Delete(cert *x509.Certificate) {
+	usage := getCertUsage(cert)
+	for _, name := range cert.DNSNames {
+		c.delete(name, usage)
+	}
+}
+
 // OverlapsWithExistingCertificate returns true if the certificate overlaps with an existing certificate.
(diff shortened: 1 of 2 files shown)


pkg/cryptutil/certificates_index.go
```diff
@@ -32,6 +32,14 @@ func (c *CertificatesIndex) Add(cert *x509.Certificate) {
...
 }
 
+// Delete deletes a certificate from the index.
+func (c *CertificatesIndex) Delete(cert *x509.Certificate) {
+	usage := getCertUsage(cert)
+	for _, name := range cert.DNSNames {
+		c.delete(name, usage)
+	}
+}
+
```
kralicky, line 42: Would the CertificatesIndex ever need mutex locking? hard to tell from usages if it could be concurrently accessed/modified.

wasaga: that is out of scope for this particular component

kralicky: I only ask because before it was add-only, which can race but won't blow up your program; on the other hand, concurrent add/delete is a runtime panic. this may or may not actually be possible, but it was hard to tell

kralicky (approved):

Issue #5171, opened 2024-07-10
Feature request: always pass a Pomerium JWT to upstream services


kenjenkins:
**Is your feature request related to a problem? Please describe.**

If I understand correctly, the Pomerium JWT serves two purposes:
1. Verifying the JWT lets the upstream service know that the request was indeed authorized by Pomerium, if the network connection is not otherwise secured (e.g. if the upstream service is exposed over the public internet).
2. The upstream service can obtain user account information from the JWT, so it knows which user is currently signed in.

Currently the Pomerium JWT is passed to upstream services when the [Pass Identity Headers](https://www.pomerium.com/docs/reference/pass-identity-headers) option is enabled.

This makes sense when considering purpose (2), but not necessarily for purpose (1). Even if an upstream service does not need information about the user's identity, it may still be desirable to verify that an incoming request was in fact authorized by Pomerium.

**Describe the solution you'd like**

Always pass a Pomerium JWT to upstream services. If the Pass Identity Headers option is off, include a minimal set of claims (e.g. `iss`, `aud`, `iat`, and `exp`). If the Pass Identity Headers option is on, include user info claims (the current Pomerium behavior).

**Describe alternatives you've considered**

n/a

**Explain any additional use-cases**

n/a

**Additional context**

If there are cases where the `X-Pomerium-Jwt-Assertion` header causes a problem for a particular upstream service, it can still be removed by using the [Remove Request Headers](https://www.pomerium.com/docs/reference/routes/headers#remove-request-headers) option.

kenjenkins:
Potential downside: we've occasionally seen problems with header size due to large number of groups or with user avatar image data inline. If we start passing a JWT unconditionally we'll want to make sure it contains only a limited amount of data.

calebdoxsey:
This will invalidate https://github.com/pomerium/pomerium/issues/3805.

kenjenkins:
> This will invalidate #3805.

Oh, that's a good point. If the Pomerium JWT signifies "this request was authorized by Pomerium" I'm not entirely sure what to do with the /.pomerium/jwt endpoint. 

Would it make sense to serve the /.pomerium/jwt endpoint only for routes with the Pass Identity Headers option enabled, and preserve the current 401 behavior if there is no signed-in user?

calebdoxsey:
This is not possible. Routes include paths as part of their definition so its impossible to know which route `/.pomerium/jwt` corresponds to.

PR #5190, opened 2024-07-26
core/authorize: always pass jwt

calebdoxsey:
## Summary
Always pass the JWT assertion header regardless of the pass identity header option. The option will now only remove the user-specific claims. 

If the header is not desired it can be removed using the remove request headers option: https://www.pomerium.com/docs/reference/routes/headers#remove-request-headers.

## Related issues
For https://github.com/pomerium/pomerium/issues/5171
 

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


kenjenkins (commented):
Sorry for the slow response. I think I need to write up some thoughts on what to do with the /.pomerium/jwt endpoint in general going forward. There are a few complications and I'd like to make sure we have answers for some of these questions before moving forward with this change.

Issue #5130, opened 2024-06-06
Enable downstream MTLS failures logging


ViswanathB:
**Is your feature request related to a problem? Please describe.**
When client connection request fails in TLS negotiation with MTLS enforced on the proxy, [access log](https://www.pomerium.com/docs/reference/access-log-fields) doesn't capture this. 

We would like to log 
1. client IP
2. client cert information(subject to start with, if present)
3. connection failure reason

for all such cases.

**Describe the solution you'd like**
Be able to extract information from listener logs.

**Describe alternatives you've considered**
N/A

**Explain any additional use-cases**
N/A

We want to track the number of failures happening for different reasons. Metrics on this would serve in understanding source of such requests, detect an active attack in progress to name a few..

**Additional context**
Slack discussion with @kenjenkins

ssveta7ak:
Implemented. Checked on 
pomerium: 0.27.0+dad954ae
envoy: 1.31.0+59c34f2ead46e56f647993324faed008a6910f1606ba34d78e77acf5d3ab89c3

PR #5196, opened 2024-07-31
envoy: enable access logging for main http/grpc listeners

kralicky:
## Summary

This configures envoy access logs for the main http-ingress listener and the grpc-ingress listener using the same configuration as the http connection manager. This enables additional envoy logs for connections that pass through these listeners and/or do not reach the http connection manager filter.

Additionally, more builtin access log fields have been made available, including mtls certificate info and transport error messages.

## Related issues

https://github.com/pomerium/pomerium/issues/5130

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


config/envoyconfig/listeners.go
```diff
@@ -133,6 +133,8 @@ func (b *Builder) buildMainListener(
 		li.ListenerFilters = append(li.ListenerFilters, ProxyProtocolFilter())
 	}
 
+	li.AccessLog = buildAccessLogs(cfg.Options)
```
kenjenkins, line 136: I think we need a separate handler for listener access logs, as they appear to be scoped to the connection rather than the request, so most of the existing access log fields will be empty/zero. For example:

> {"level":"info","service":"envoy","upstream-cluster":"","method":"METHOD_UNSPECIFIED","authority":"","path":"","user-agent":"","referer":"","forwarded-for":"","request-id":"","duration":0,"size":1947,"response-code":0,"response-code-details":"","time":"2024-08-01T16:12:12-07:00","message":"http-request"}

kralicky:
Reworked this a bit, still need to update tests though

PR #5197, opened 2024-07-31
envoy: dynamic access log fields

kralicky:
This adds the ability to include any available field in the envoy access logs, in addition to the built-in field names we have predefined.

Entries in the access_log_fields list in the config that have the form 'name=path', where path is a protopath string with an implied root of type envoy.data.accesslog.v3.HTTPAccessLogEntry, will add fields to all access log events. These fields will use the given name as the key, and use the path to look up the value in the log entry sent by envoy.

Paths can be used to look up fields deeply nested in the message, as well as access dynamic metadata (with google.protobuf.Any values), specific entries in arrays or maps, or select entire message fields to log as a json object.

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


internal/controlplane/grpc_accesslog.go
```diff
@@ -64,6 +69,16 @@ func populateLogEvent(
 	referer, _, _ := strings.Cut(entry.GetRequest().GetReferer(), "?")
 	path, query, _ := strings.Cut(entry.GetRequest().GetPath(), "?")
 
+	if !field.IsWellKnownField() && field.IsDynamicField() {
+		name, pathStr, _ := strings.Cut(string(field), "=")
+		name = strings.ToValidUTF8(strings.TrimSpace(name), "")
+		path, err := paths.Parse(entry.ProtoReflect().Descriptor(), pathStr)
```
calebdoxsey, line 75: Is there a way to parse this path once and cache it? Or does the path depend on not only the descriptor but the contents of the log entry?

calebdoxsey: (if the parsing is inexpensive then this may be premature optimization anyway)

pkg/protoutil/paths/paths.go
```diff
@@ -0,0 +1,306 @@
...
+	"strings"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protopath"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+func Parse(root protoreflect.MessageDescriptor, pathStr string) (protopath.Path, error) {
```
calebdoxsey, line 16: I think I understand some of this... I was surprised that there's a `protopath` package but this parsing/derefencing functionality doesn't exist? Weird.

There might be some gaps in this code, but the unit tests are good.

calebdoxsey (approved):

pkg/protoutil/paths/paths.go
kralicky: Yeah, the builtin protopath is rather incomplete. I've written a number of variations of this function for so many projects, maybe I should just try to contribute it to upstream instead, or collect all this code into a library

internal/controlplane/grpc_accesslog.go
kralicky: Yes we can cache these, the parsing step only operates on the descriptor, the dereference step operates on the actual message contents. The parsing is reasonably cheap but it would still make sense to cache it I think

kralicky:
Added an automatic parser cache, should be quite a bit faster now to parse the same descriptor/string combination multiple times.
```
BenchmarkParser/Without_cache-32       	  424480	     2512 ns/op	   2544 B/op	     27 allocs/op
BenchmarkParser/With_cache-32          	33914374	    35.90 ns/op	      0 B/op	      0 allocs/op
```

calebdoxsey (approved):

PR #5218, opened 2024-08-14
core/proto: update protoc dependencies

calebdoxsey:
## Summary
We were using `github.com/golang/protobuf/protoc-gen-go` which has been deprecated for a while. The newer `google.golang.org/protobuf/cmd/protoc-gen-go` doesn't support plugins in the same way, so the `protoc.bash` script had to be rewritten.

Also update the `buf.gen.yaml` to use remote plugins instead of local binaries. (which wasn't working for me because I don't have `protoc-gen-go` or `protoc-gen-go-grpc` on my path locally)

There shouldn't be any real change in behavior from these changes.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


pkg/storage/postgres/registry.go
```diff
@@ -12,12 +12,13 @@ import (
 )
 
 type registryServer struct {
+	registry.UnimplementedRegistryServer
```
kralicky, line 15: Forgot to remove this one?

kenjenkins (approved):

PR #5227, opened 2024-08-19
core/grpc: add IterateAll method

calebdoxsey:
## Summary
Add an `IterateAll` method to the databroker that's like `InitialSync` but returns an iterator:

```go
	for record, err := range databroker.IterateAll[session.Session](ctx, client) {
		// record.Object is a session.Session
	}
```

This avoids having to create a full slice of records and also handles unmarshaling the data automatically.

## Related issues
- https://github.com/pomerium/pomerium-zero/issues/2764

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review


wasaga (approved):
the new iterator interface looks quite elegant

PR #5229, opened 2024-08-20
zero/health-checks: fix early checks sometimes missing

wasaga:
## Summary

Health checks are reported and should be picked up by the reporter. During start up, some checks were sometimes reported before the reporting provider was set, and were discarded. 
This PR fixes that behaviour and makes sure the latest known state of the health checks is replayed when provider is set, allowing it to catch up.

It also fixes the telemetry shutdown context to allow up to 10s for spooling of the pending events.

## Related issues

Fixes https://github.com/pomerium/pomerium-zero/issues/2893

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/pkg/health/deduplicate.go
+++ b/pkg/health/deduplicate.go
@@ -5,18 +5,19 @@ import (
 	"sync"
 )
 
-var _ Provider = (*deduplicator)(nil)
+var _ Provider = (*Deduplicator)(nil)
 
-// deduplicator is a health check provider that deduplicates health check reports
+// Deduplicator is a health check provider that deduplicates health check reports
 // i.e. it only reports a health check if the status or attributes have changed
-type deduplicator struct {
-	seen     sync.Map
+type Deduplicator struct {
+	lock     sync.Mutex
+	records  map[Check]*record
 	provider Provider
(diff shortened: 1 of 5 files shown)


pkg/health/deduplicate.go
```diff
@@ -5,75 +5,121 @@ import (
...
-func (d *deduplicator) ReportError(check Check, err error, attrs ...Attr) {
-	cur := newErrorRecord(err, attrs)
-	prev := d.swap(check, cur)
-	if prev == nil || !cur.Equals(prev) {
-		d.provider.ReportError(check, err, attrs...)
+func (d *Deduplicator) SetProvider(p Provider) {
+	d.setProvider(p)()
+}
+
+func (d *Deduplicator) setProvider(p Provider) func() {
```
calebdoxsey, line 78: What's the purpose of the closure function? It seems to always be invoked immediately.

wasaga: To be called after releasing the the lock. It's not very idiomatic, I updated implementation.

PR #5231, opened 2024-08-20
core/zero: periodically report usage

calebdoxsey:
## Summary
Periodically report usage to the zero API by iterating over all session and user records.

## Related issues
- https://github.com/pomerium/pomerium-zero/issues/2764
 

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


wasaga:
I was under impression from yesterday's discussion that we would only send this request once when user session is created, and rename that "last seen" column in the UI into "last signed in"

calebdoxsey:
The usage reporter will now use sync to get new sessions. Any updated records are reported every minute.

wasaga:
I believe during an RFC review meeting this week we agreed on changing the "Last seen" (represented by last session access time) into "Last signed in" (represented by session created time).

calebdoxsey:
If this is not correct what do you want instead?

```
nr.accessedAt = latest(nr.accessedAt, s.GetIssuedAt().AsTime())
```

wasaga:
can we please rename the `accessedAt` property to match what it represents?

calebdoxsey:
I interpreted that meeting differently. But sure whatever you want.

internal/zero/controller/usage_reporter.go
```diff
@@ -0,0 +1,278 @@
+package controller
```
wasaga, line 1: (nit) `controller` package currently only has top-level logic, while other packages reside in their dedicated packages - can we move the usage reporter into some other package for consistency?

internal/zero/controller/usage_reporter.go
```diff
@@ -0,0 +1,278 @@
...
+		records := make([]usageReporterRecord, 0, len(ur.updates))
+		for userID := range ur.updates {
+			records = append(records, ur.byUserID[userID])
+		}
+		clear(ur.updates)
+		ur.mu.Unlock()
+
+		// report the records with a backoff in case the API is temporarily unavailable
+		if len(records) > 0 {
+			log.Info(ctx).Int("updated-users", len(records)).Msg("reporting usage")
```
wasaga, line 116:
```suggestion
			log.Debug(ctx).Int("updated-users", len(records)).Msg("reporting usage")
```

internal/zero/controller/usage_reporter.go
```diff
@@ -0,0 +1,278 @@
...
+			err := backoff.Retry(func() error {
+				err := ur.report(ctx, records)
+				if err != nil {
+					log.Error(ctx).Err(err).Msg("error reporting usage")
```
wasaga, line 120:
```suggestion
					log.Warn(ctx).Err(err).Msg("error reporting usage")
```

as it does not affect pomerium itself

internal/zero/controller/usage_reporter.go
```diff
@@ -0,0 +1,278 @@
...
+		for userID := range ur.updates {
+			records = append(records, ur.byUserID[userID])
+		}
+		clear(ur.updates)
+		ur.mu.Unlock()
+
+		// report the records with a backoff in case the API is temporarily unavailable
+		if len(records) > 0 {
+			log.Info(ctx).Int("updated-users", len(records)).Msg("reporting usage")
+			err := backoff.Retry(func() error {
```
wasaga, line 117: would it make sense to move the backoff loop down to `ur.report` function to save on continuous record conversion in case of retries?

internal/zero/controller/usage_reporter.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func latest(t1, t2 time.Time) time.Time {
+	if t2.After(t1) {
+		return t2
+	}
+	return t1
+}
+
+func syncRecords[T any, TMessage interface {
```
wasaga, line 185: does it make sense to move `syncRecords` and `syncRecordsLatest` to `pkg/grpc/databroker` ?

wasaga (approved):

internal/zero/controller/usage_reporter.go
calebdoxsey: I doubt they will never be used by anything else. They use callbacks and don't handle errors. The generic implementation I built using iterators is now completely unused because we changed requirements. I will remove it in a separate PR.

internal/zero/controller/usage_reporter.go
calebdoxsey: I don't understand the objection, but ok.

PR #5248, opened 2024-08-29
zero/k8s: use deployments

wasaga:
## Summary

Previously, Pomerium in Zero mode was deployed using `StatefulSet` and required a PVC to store the bootstrap piece.

Support to store the bootstrap piece in a kubernetes secret was added in #5114. 

This PR updates manifests accordingly to use stateless `Deployment` and configures the bootstrap writeback to the `pomerium-zero/pomerium` `Secret` that is already required to store `POMERIUM_ZERO_TOKEN`. 

## Related issues

Related: https://github.com/pomerium/pomerium-zero/issues/2405

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/k8s/zero/deployment/env.yaml
+++ b/k8s/zero/deployment/env.yaml
@@ -22,7 +22,7 @@ spec:
             - name: BOOTSTRAP_CONFIG_FILE
               value: "/var/run/secrets/pomerium/bootstrap.dat"
             - name: BOOTSTRAP_CONFIG_WRITEBACK_URI
-              value: "secret://pomerium-zero/pomerium/bootstrap"
+              value: "secret://$(POMERIUM_NAMESPACE)/pomerium/bootstrap"
             - name: POD_IP
               valueFrom:
                 fieldRef:
--- a/k8s/zero/deployment/env.yaml
+++ b/k8s/zero/deployment/env.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: StatefulSet
+kind: Deployment
 metadata:
   name: pomerium
 spec:
@@ -19,6 +19,10 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: metadata.namespace
+            - name: BOOTSTRAP_CONFIG_FILE
+              value: "/var/run/secrets/pomerium/bootstrap.dat"
+            - name: BOOTSTRAP_CONFIG_WRITEBACK_URI
+              value: "secret://pomerium-zero/pomerium/bootstrap"
             - name: POD_IP
               valueFrom:
                 fieldRef:
--- a/k8s/zero/deployment/volumes.yaml
+++ b/k8s/zero/deployment/volumes.yaml
@@ -21,6 +21,7 @@ spec:
               name: tmp
             - mountPath: "/var/run/secrets/pomerium"
               name: bootstrap
+              readOnly: true
       volumes:
         - name: tmp
           emptyDir: {}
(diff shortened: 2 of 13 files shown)


calebdoxsey (approved):

k8s/zero/deployment/env.yaml
```diff
@@ -19,6 +19,10 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: metadata.namespace
+            - name: BOOTSTRAP_CONFIG_FILE
+              value: "/var/run/secrets/pomerium/bootstrap.dat"
+            - name: BOOTSTRAP_CONFIG_WRITEBACK_URI
+              value: "secret://pomerium-zero/pomerium/bootstrap"
```
kralicky, line 25:
```suggestion
              value: "secret://$(POMERIUM_NAMESPACE)/pomerium/bootstrap.dat"
```

kralicky: Do you think we should use the same `pomerium` secret for this, or a different one?

k8s/zero/deployment/volumes.yaml
```diff
@@ -13,22 +13,20 @@ spec:
...
             - name: XDG_DATA_HOME
-              value: "/var/cache"
+              value: "/tmp/pomerium/cache"
           volumeMounts:
             - mountPath: "/tmp/pomerium"
               name: tmp
-            - mountPath: "/var/cache"
-              name: pomerium-cache
+            - mountPath: "/var/run/secrets/pomerium"
+              name: bootstrap
```
kralicky, line 23:
```suggestion
              name: bootstrap
              readOnly: true
```

k8s/zero/deployment/env.yaml
wasaga: I wasn't sure if envs would expand here, let me double-check

wasaga: > Do you think we should use the same `pomerium` secret for this, or a different one?

it seemed to make sense: the user has to create that secret with their PZ token, until its created PZ is not starting, and it is kept outside of installation manifests so that when you upgrade it doesn't get overwritten.

k8s/zero/deployment/volumes.yaml
```diff
@@ -13,22 +13,21 @@ spec:
...
-      accessModes: [ "ReadWriteOnce" ]
-      resources:
-        requests:
-          storage: 100Mi
+        - name: bootstrap
+          secret:
+            secretName: pomerium
+            items:
+              - key: bootstrap
+                path: bootstrap.dat
```
kralicky, line 33: This also needs `optional: true`, if the secret exists but doesn't have this specific key (zero token only but no bootstrap), the pod won't start

k8s/zero/deployment/env.yaml
kralicky: That should work, but a possible complication is that we have to mark the whole secret with `optional: true` since the `bootstrap.dat` key won't exist on the first run. However, we still want the secret to exist to make sure the zero token will be present.

wasaga: the volume mount is now optional, so the secret doesn't need to have this property anymore; we reference the `POMERIUM_ZERO_TOKEN` env , so if the secret does not exist it would still wait until the user creates this secret.

Theghost256 (commented):
Thank you for letting me help you 😌🙏

PR #5252, opened 2024-08-29
zero: generate error methods for response types

kralicky:
## Summary

This replaces the hand-written functions in pkg/zero/cluster/client_errors.go with identical auto-generated versions, removing the need to copy-paste-modify these when adding new apis. This also adds support for generating methods for response codes other than 400/500, preventing error messages from being lost after calling `apierror.CheckResponse`

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review


pkg/zero/cluster/client-with-responses.tmpl
```diff
@@ -0,0 +1,207 @@
...
+
+    return response, nil
+}
+{{end}}{{/* range . $opid := .OperationId */}}
+
+{{/* End upstream code /*}}
+
+
+{{/* Custom code below */}}
+
```
kralicky, line 149: new code starts here, everything above is copied from https://github.com/oapi-codegen/oapi-codegen/blob/main/pkg/codegen/templates/client-with-responses.tmpl

calebdoxsey (approved):

PR #5275, opened 2024-09-10
core/zero: add organization id and cluster id to bootstrap config

calebdoxsey:
## Summary
Add the organization id and the cluster id to the zero config.

## Related issues
- https://github.com/pomerium/pomerium-zero/issues/2964

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -49,6 +49,11 @@ type Config struct {
 
 	// MetricsScrapeEndpoints additional metrics endpoints to scrape and provide part of metrics
 	MetricsScrapeEndpoints []MetricsScrapeEndpoint
+
+	// ZeroClusterID is the zero cluster id, only set when in zero mode.
+	ZeroClusterID string
+	// ZeroOrganizationID is the zero organization id, only set when in zero mode.
+	ZeroOrganizationID string
 }
 
 // Clone creates a clone of the config.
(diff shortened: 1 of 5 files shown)


wasaga:
did you consider reusing existing `InstallationID` ?

calebdoxsey:
I did not. What was discussed was the organization id. If we would like change the requirements I can do that and use the installation id instead.

calebdoxsey:
Trying to understand the ask here,

Are you saying we should put something like `{OBFUSCATED_ORGANIZATION_ID}|{OBFUSCATED_CLUSTER_ID}` in the installation id?

Are you also saying that we should not add this to the bootstrap but instead ship it with the existing installation id property in the config.proto options?

I see two issues with this:

1. This stores data unstructured in a field that's not obvious. We can't just use `organization id` because doesn't "installation id" imply a cluster on the zero side? Or we fine with multiple clusters all having the same installation id? If the data's stringly-typed its prone to confusion and parsing errors. I generally prefer separate fields.
2. Having the data in the config proto makes it slightly harder to work with when starting the usage reporter since we have to wait for it to show up. Not insurmountable but I think using the bootstrap config is easier.

That was my perspective but I'm fine changing it.

wasaga (approved):
Thanks for clarification I understand now: we need specifically an identifier that's stable across organization. The installationID won't work in that case as it'll turn out cluster-specific.

Theghost256 (commented):
Thank you for letting me help ☺️

PR #5281, opened 2024-09-11
core/zero: add usage reporter

calebdoxsey:
## Summary
Add a usage reporter when in Zero mode. This will record pseudonymous user data based on active sessions.

## Related issues
- https://github.com/pomerium/pomerium-zero/issues/2964

## Checklist

- [x] updated unit tests
- [x] ready for review


pkg/grpc/databroker/sync.go
```diff
@@ -0,0 +1,110 @@
...
+	"io"
+
+	"google.golang.org/protobuf/proto"
+
+	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/pkg/protoutil"
+)
+
+// SyncRecords calls fn for every record using Sync.
+func SyncRecords[T any, TMessage interface {
```
kralicky, line 16: was this code from https://github.com/pomerium/pomerium/pull/5234/files? should we un-revert it instead?

calebdoxsey: That was an iterator based version. I had originally intended it for this work, but then the code changed and I couldn't use it anymore. In particular the SyncLatest returns the server and latest record versions, so the stream has two sets of data within it. `iter.Seq2` is already kinda clumsy with errors, but it started getting really gnarly with those versions too.

It could maybe work with just the normal `Sync`, but the consistency of these two methods was nice.

internal/zero/controller/usagereporter/usagereporter.go
```diff
@@ -0,0 +1,217 @@
...
+		Users: convertUsageReporterRecords(ur.organizationID, records),
+	}
+	return backoff.Retry(func() error {
+		log.Debug(ctx).Int("updated-users", len(req.Users)).Msg("reporting usage")
+		err := ur.api.ReportUsage(ctx, req)
+		if err != nil {
+			log.Warn(ctx).Err(err).Msg("error reporting usage")
+		}
+		return err
+	}, backoff.WithContext(backoff.NewExponentialBackOff(), ctx))
```
wasaga, line 88: my reading of this loop is that we capture the updates and try to deliver them for the maximum of default exponential update (15 minutes). 
after that period, the undelivered records are dropped and do not go back into the update set. 

would it make sense to consider moving this retry up to the `runReporter` loop so that undelivered updates would be merged with pending updates or dropping records is intentional?

internal/zero/api/api.go
```diff
@@ -119,3 +119,16 @@ func (api *API) GetClusterResourceBundles(ctx context.Context) (*cluster_api.Get
...
 	return api.telemetryConn
 }
+
+func (api *API) ReportUsage(ctx context.Context, req cluster_api.ReportUsageRequest) error {
+	res, err := api.cluster.ReportUsageWithResponse(ctx, req)
+	if err != nil {
+		return err
+	}
+
+	if res.StatusCode()/100 != 2 {
```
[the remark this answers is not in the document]
calebdoxsey: Doesn't compile. `in call to apierror.CheckResponse, type *cluster.ReportUsageResp of api.cluster.ReportUsageWithResponse(ctx, req) does not match apierror.APIResponse[T]`.

internal/zero/controller/usagereporter/usagereporter.go
calebdoxsey: My understanding is that `backoff.Retry` returns an error once stop is reached. This is returned by `report`, which is returned by `runReporter`, which is returned by `runSync`, which is returned by `Run`, which will be retried by `WithBackoff`.

The retry was done here to avoid excessive calls to sync in the databroker. But I can change it to just always return the error if that's preferable.

internal/zero/api/api.go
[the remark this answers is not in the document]
calebdoxsey: I added additional statuses to the openapi and now it works.

internal/zero/controller/usagereporter/usagereporter.go
wasaga: what I meant is maybe combining this retry loop with the `runReporter` main loop, so that unsent records would merge back with `ur.updates` until they can be spooled to the cloud

Theghost256 (commented):
Thank you for letting me help ☺️

PR #5287, opened 2024-09-13
core/authorize: remove default header logging in debug

calebdoxsey:
## Summary
Currently we log headers during authorization if the log level is set to debug. This PR changes it to never log headers by default. If a user wants headers logged they need to set the authorization header field to include headers explicitly.

## Related issues
- https://github.com/pomerium/internal/issues/1892

## Checklist

- [ ] updated unit tests
- [x] ready for review


kenjenkins (approved):
Thanks!

PR #5289, opened 2024-09-16
ppl: fix empty/no-op allow block added in some cases to converted PPL policies

kralicky:
When converting a core route to a PPL policy, if an embedded policy object is present but none of the shorthand options that translate to PPL snippets (`allowed_users`, etc.) are present, omit the empty (no-op) allow block from the resulting policy.

This doesn't change any semantics of the policy, but it may be confusing for users if we display the converted yaml.

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/config/policy_ppl.go
+++ b/config/policy_ppl.go
@@ -77,10 +77,14 @@ func (p *Policy) ToPPL() *parser.Policy {
 				},
 			})
 	}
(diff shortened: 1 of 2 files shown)


kenjenkins (approved):

Theghost256 (commented):
Thank you 👍

Issue #5246, opened 2024-08-27
Missing Pomerium claims from User Info Dashboard


goldie-bloomberg:
## What happened?

- We have a programmatic sanity check tool built using pomerium's authclient and certstore packages. The tool attempts retrieve a user's claims from the dashboard endpoint and cross check 403 response codes against the user's claims. 

- However the users claims (and other details) are only returned by web browser access. 
- Via programmatic access, all fields in POMERIUM_DATA's jwt, except for webAuthnUrl, are empty. 

## What did you expect to happen?
I expected all the fields POMERIUM_DATA's jwt to be identical whether accessed programmatically or via a browser. 

## How'd it happen?

1. Ran an internal tool programmatic_access.exe
2. It retrieved the pomerium_jwt
3. And then it set the Authorization header to the pomerium_jwt and accessed the dashboard url
4. the returned jwt did not have any claims set. 

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`): 
  - Unfortunately we're missing the  pomerium version due to our messed up build that we will fix. 
  - The most recent commit that's pulled in is #e3e7de7
- Server Operating System/Architecture/Cloud: Server operating system : 
  - RHEL8

## What's your config.yaml?

Not particularly relevant

```config.yaml
# Paste your configs here
# Be sure to scrub any sensitive values
```

## What did you see in the logs?
```logs
 {
  "level": "info",
  "service": "envoy",
  "authority": "scrubbed",
  "path": "/.pomerium/",
  "query": "",
  "ip": "scrubbed",
  "forwarded-for": "scrubbed",
  "method": "GET",
  "duration": 21.085387,
  "referer": "",
  "request-id": "7dcf38d7-cc92-439f-91d3-de741c28b152",
  "response-code": 200,
  "response-code-details": "via_upstream",
  "size": 706,
  "upstream-cluster": "pomerium-control-plane-http",
  "user-agent": "Go-http-client/1.1",
  "client-certificate": {
    "issuer": "scrubbed",
    "subject": "scrubbed"
  },
  "time": "2024-08-27T14:22:26-04:00",
  "message": "http-request"
}
```

calebdoxsey:
This should be fixed after https://github.com/pomerium/pomerium/commit/52d4899d4c883360815cf58abf4a74f56fcfac9f on main or in the next tagged release of 0.27.

PR #5291, opened 2024-09-17
core/proxy: support loading sessions from headers and query string

calebdoxsey:
## Summary
Currently the proxy service will only load sessions from cookies. We also support sessions stored in headers and query string parameters. This PR updates the proxy code to support loading sessions from these places as well.

## Related issues
- https://github.com/pomerium/pomerium/issues/5246

## Checklist

- [ ] updated docs
- [x] updated unit tests
- [ ] updated UPGRADING.md
- [x] ready for review

### DIFF
--- a/proxy/handlers_test.go
+++ b/proxy/handlers_test.go
@@ -15,7 +15,9 @@ import (
 
 	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/internal/atomicutil"
+	"github.com/pomerium/pomerium/internal/encoding/jws"
 	"github.com/pomerium/pomerium/internal/httputil"
+	"github.com/pomerium/pomerium/internal/sessions"
 	"github.com/pomerium/pomerium/internal/urlutil"
 )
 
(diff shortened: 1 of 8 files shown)


proxy/handlers_test.go
```diff
@@ -260,3 +262,79 @@ func TestProxy_registerDashboardHandlers_jwtEndpoint(t *testing.T) {
...
+		sharedKey, err := opts.GetSharedKey()
+		require.NoError(t, err)
+
+		encoder, err := jws.NewHS256Signer(sharedKey)
+		require.NoError(t, err)
+
+		sessionBS, err := encoder.Marshal(&sessions.State{
+			ID: "___SESSION_ID___",
+		})
+		require.NoError(t, err)
```
kenjenkins, line 330: (totally optional)

Might be nice to extract this session token setup code out of the individual "cookie session" / "header session" tests to make it super clear that the same exact token can appear either in a cookie or in the Authorization header.

kenjenkins (approved):
Thanks! Looks like some good code clean-up too.

Issue #5299, opened 2024-09-19
build-dev-docker.sh script error when run on macOS


kenjenkins:
I wanted to run the integration tests locally, so I followed the steps under https://www.pomerium.com/docs/community/contributing#development. However running the build-dev-docker.bash script did not succeed:

```
==> get-envoy
fork/exec /var/folders/f9/91j5hnyd333bctgnqvx44gsw0000gn/T/go-build3058201948/b001/exe/get-envoy: exec format error
make: *** [get-envoy] Error 1
```

Since we switched the `get-envoy` target to use a `go run` command, now it is affected by the `GOOS` setting. So the `env GOOS=linux make build` line in build-dev-docker.bash will no longer succeed when run on macOS.

Probably the simplest fix would be to _unset_ GOOS before running the `get-envoy` command, so that it defaults to the host OS.

PR #5300, opened 2024-09-19
Makefile: unset GOOS for get-envoy target

kenjenkins:
## Summary

The build-dev-docker.sh script will run

    env GOOS=linux make build

to build Pomerium for Linux (before copying this binary into a minimal Docker image).

However, now that the `get-envoy` target is implemented using a Go command, it too will be affected by `GOOS`. As a result running this step on macOS will result in an error when trying to execute a Linux `get-envoy` binary.

Instead, let's unset `GOOS` when running the `get-envoy` command, so that it defaults to building for the host OS.

## Related issues

- https://github.com/pomerium/pomerium/issues/5299

## Checklist

- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] ready for review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ all: clean build-deps test lint build ## Runs a clean, build, fmt, lint, test, a
 .PHONY: get-envoy
 get-envoy: ## Fetch envoy binaries
 	@echo "==> $@"
-	@cd pkg/envoy/files && go run ../get-envoy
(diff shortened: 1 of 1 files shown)


wasaga (approved):

Theghost256 (commented):
Thank you for letting me help 🐱

Issue #5316, opened 2024-10-07
authorize: databroker connection error results in redirect loop


calebdoxsey:
## What happened?
If the connection between the databroker and the authorize service is lost, all calls to the authorize service's `Check` method will result in an `unauthenticated` result. This then results in a redirect to the authenticate service, and eventually back to the authorize service, etc.

The loop is the result of this code:

```go
	if sessionState != nil {
		s, err = a.getDataBrokerSessionOrServiceAccount(ctx, sessionState.ID, sessionState.DatabrokerRecordVersion)
		if err != nil {
			log.Info(ctx).Err(err).Str("request-id", requestID).Msg("clearing session due to missing or invalid session or service account")
			sessionState = nil
		}
	}
```

This code works properly when the error code is `NotFound`. But when the error code is `Unavailable` the correct result should be to return the error (and log it).

## What did you expect to happen?

The user should get a 500 error when this happens. Not be put in a redirect loop.

## How'd it happen?
This error was seen while writing integration tests. It could probably be fairly easily reproduce in a separate service deploy.

calebdoxsey:
This is done.

PR #5317, opened 2024-10-07
authorize: fix transient errors from databroker causing redirect loop

kralicky:
## Summary

When the databroker is temporarily unavailable, do not clear sessions during authorization, instead return the 500 error so it can be detected and retried.

## Related issues

https://github.com/pomerium/pomerium/issues/5316

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/authorize/grpc.go
+++ b/authorize/grpc.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 
 	envoy_service_auth_v3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/structpb"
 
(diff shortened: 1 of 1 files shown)


authorize/grpc.go
```diff
@@ -56,6 +58,10 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v3.CheckRe
 	if sessionState != nil {
 		s, err = a.getDataBrokerSessionOrServiceAccount(ctx, sessionState.ID, sessionState.DatabrokerRecordVersion)
 		if err != nil {
+			if status.Code(err) == codes.Unavailable {
```
calebdoxsey, line 61: could we move this one level up and do the `if err != nil` as an `else if err != nil`

calebdoxsey (approved):

kenjenkins (approved):
Good work!

PR #5323, opened 2024-10-09
config/k8s: allow configuring claims used to detect user/group

kralicky:
## Summary

This adds `kubernetes_impersonate_user_claim` (default 'email') and `kubernetes_impersonate_group_claim` (default 'groups') route settings, which can be used to control the claims that are used to obtain user and group names for kubernetes impersonation. 

Some IDPs (auth0) place restrictions on custom claims. In particular, they disallow using the name 'groups' which makes it impossible to use group impersonation. This will allow users to override the default claims to support using these IDPs (or if they want to use a different claim than 'email' for user impersonation)

## Related issues

https://github.com/pomerium/pomerium/issues/5324

## Checklist

- [x] updated unit tests
- [x] ready for review


calebdoxsey (commented):
Groups are an enterprise feature. They are populated via directory sync. I think these changes make it so that kubernetes impersonation no longer works with that.

calebdoxsey:
I think if we want to support using a different claim for groups, it should be a setting not associated with kubernetes, since I would expect groups to be used consistently. For:

```rego
jwt_payload_groups := v if {
	v = array.concat(group_ids, get_databroker_group_names(group_ids))
	v != []
} else := v if {
	v = session.claims.groups
	v != null
} else := []
```

We could do something like: `session.claims[input.groups_claim]`.

I'm not sure how we would make the user dynamic.

Issue #5328, opened 2024-10-15
when in k8s api server proxy mode, return typed response


wasaga:
**Is your feature request related to a problem? Please describe.**

currently, if proxying a request from `kubectl`, pomerium returns a standard json response, that is not understood by `kubectl` and the error message is generic "unknown", that complicates troubleshooting.  

<img width="1508" alt="image" src="https://github.com/user-attachments/assets/504a4eb1-9bcc-448f-be28-3ad18646cc88">

**Describe the solution you'd like**

1. return a typed response that is understood by kubectl
2. make request-id part of the error displayed by kubectl
3. if Show Error Details is enabled on the route in the Enterprise, show an extended policy decision results

wasaga:
https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Status

PR #5331, opened 2024-10-21
authorize: serialize errors in k8s Status format when gvk parameters are present in request media type

kralicky:
## Summary

When responding to a Kubernetes api request with a (pomerium) auth error, serialize the error in the format expected by the Kubernetes rest client. 

The presence of GVK parameters in the Accept media type are used to identify these requests.

TODO: 
- [ ] update unit tests

## Related issues

https://github.com/pomerium/pomerium/issues/5328

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/authorize/check_response.go
+++ b/authorize/check_response.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"mime"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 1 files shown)


authorize/check_response.go
```diff
@@ -134,21 +135,48 @@ func (a *Authorize) deniedResponse(
 ) (*envoy_service_auth_v3.CheckResponse, error) {
 	respHeader := []*envoy_config_core_v3.HeaderValueOption{}
 
+	kind := wellKnownRequestKind(in)
 	var respBody []byte
 	switch {
 	case getCheckRequestURL(in).Path == "/robots.txt":
```
calebdoxsey, line 141: could we make this a `kind` too? Then just switch on `switch kind {`?

kralicky:
The presence of these headers is not as consistent as I initially thought, so I'll probably go with user-agent based detection instead.

PR #5334, opened 2024-10-22
authorize: serialize errors in metav1.Status format when kubernetes user-agent is detected

kralicky:
## Summary

When responding to a Kubernetes api request with a (pomerium) auth error, serialize the error in the format expected by the Kubernetes rest client. 

Kubernetes client requests are detected by the presence of the string "kubernetes/" in the user-agent, as set in client-go: https://github.com/kubernetes/kubernetes/blob/9b56413d56a9805ce37207b940b0e566f30db785/staging/src/k8s.io/client-go/rest/config.go#L495-L498. This covers kubectl, standard client-go sdk, and the controller-runtime client.

## Related issues

#5328 

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/authorize/check_response.go
+++ b/authorize/check_response.go
@@ -135,7 +135,36 @@ func (a *Authorize) deniedResponse(
 	respHeader := []*envoy_config_core_v3.HeaderValueOption{}
 
 	var respBody []byte
+
+	hdrs := in.GetAttributes().GetRequest().GetHttp().GetHeaders()
+	userAgent := getHeader(hdrs, "User-Agent")
 	switch {
+	case strings.Contains(userAgent, "kubernetes/"):
+		message := reason
+		var statusReason string
+		switch code {
+		case http.StatusUnauthorized:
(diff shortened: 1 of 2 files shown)


calebdoxsey (approved):

kralicky:
@wasaga re: your comment on the previous PR, can you give an example of what the details would look like? We can put info in the (user-facing) message field, for example:
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "pods is forbidden: User \"example\" cannot list resource \"pods\" in API group \"\" at the cluster scope",
  "reason": "Forbidden",
  "details": {
    "kind": "pods"
  },
  "code": 403
}
```

wasaga:
I think we can deal with `details` field in a follow-up. I was thinking of it containing a JSON form of route errors:

image

however its of a little use if those details are not user-facing.

Issue #5335, opened 2024-10-22
ppl: more flexible matchers


calebdoxsey:
**Is your feature request related to a problem? Please describe.**

Currently some PPL criteria match a value directly, like `claim`:

```yaml
- allow:
    and:
      - claim/family_name: Smith
```

Whereas other PPL criteria use a `string matcher` or a `string list matcher`:

```yaml
- allow:
    or:
      - domain:
          is: domain1.com
```

```yaml
allow:
  and:
    - groups:
        has: '00gv40ki4gmtCyl5d4x6'
```

Sometimes users are confused by these differences and end up using a string matcher where they should be using a value directly, or vice-versa.

**Describe the solution you'd like**

With some small changes we can ease this:

1. For a `string matcher`, in addition to the existing options, we can support using a string instead of an object, and if this is done, it will be the same as `is`:

  ```yaml
  domain: domain1.com
  # same as
  domain:
    is: domain1.com  
  ```

2. For a `string list matcher`, in addition to the existing `has` operator, we can support an `is` operator which does the same thing as `has`, and like with the `string matcher`, support using a string, which would be the equivalent of has:

  ```yaml
  groups: '00gv40ki4gmtCyl5d4x6'
  # same as 
  groups:
    is: '00gv40ki4gmtCyl5d4x6'
  # same as 
  groups:
    has: '00gv40ki4gmtCyl5d4x6'
  ```

3. Convert the `claim` criterion to use a `string list matcher`.

kenjenkins:
I like this idea. I'm all for making PPL simpler to write wherever we can.

I think you're right that "is" vs. "has" is confusing, and I think the shorthand for "is" makes sense too.

calebdoxsey:
This was implemented with a slight change: `is` for a list matcher is now `len(x) == 1 && x[0] == y`.

kenjenkins:
@calebdoxsey will this need any follow-up changes to the policy builder UI in Enterprise / Zero?

calebdoxsey:
That was not my expectation but I can make whatever changes we want to make. This was intended merely as a fallback to prevent footguns. User's shouldn't use it.

kenjenkins:
Ah OK, thanks for the clarification.

PR #5336, opened 2024-10-22
ppl: more flexible matchers

calebdoxsey:
## Summary
Add support for more flexible PPL matchers.

## Related issues
For #5335 

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/policy/criteria/claim_test.go
+++ b/pkg/policy/criteria/claim_test.go
@@ -11,8 +11,12 @@ import (
 	"github.com/pomerium/pomerium/pkg/grpc/user"
 )
 
-func TestClaims(t *testing.T) {
+func TestClaim(t *testing.T) {
+	t.Parallel()
+
 	t.Run("no session", func(t *testing.T) {
+		t.Parallel()
+
 		res, err := evaluate(t, `
 allow:
   and:
@@ -40,6 +44,8 @@ allow:
 		require.Equal(t, A{false, A{}}, res["deny"])
 	})
 	t.Run("by session claim", func(t *testing.T) {
+		t.Parallel()
+
 		res, err := evaluate(t, `
 allow:
   and:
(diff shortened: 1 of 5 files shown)


pkg/policy/criteria/matchers_test.go
```diff
@@ -66,4 +84,12 @@ func TestStringListMatcher(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, `count([true | some v; v = example[_]; v == "test"]) > 0`, str(body))
 	})
+	t.Run("string", func(t *testing.T) {
```
kenjenkins, line 87: If `is` is now a synonym for `has`, should we also duplicate the `has` test above, and swap in `is` instead of `has`?

kenjenkins (approved):

pkg/policy/criteria/claim_test.go
```diff
@@ -40,10 +44,39 @@ allow:
...
+					Id:    "USER_ID",
+					Email: "test@example.com",
+				}),
+			},
+			Input{Session: InputSession{ID: "SESSION_ID"}})
+		require.NoError(t, err)
+		require.Equal(t, A{true, A{ReasonClaimOK}, M{}}, res["allow"])
+		require.Equal(t, A{false, A{}}, res["deny"])
+	})
+	t.Run("by session claim via has", func(t *testing.T) {
```
kralicky, line 72: We should also probably test the opposite case, e.g. if the value doesn't contain "Smith", it is not accepted

pkg/policy/criteria/matchers.go
```diff
@@ -56,13 +63,21 @@ func matchStringStartsWith(dst *ast.Body, left *ast.Term, right parser.Value) er
...
+	}
+
 	obj, ok := right.(parser.Object)
 	if !ok {
 		return fmt.Errorf("expected object for string list matcher, got: %T", right)
 	}
 
 	lookup := map[string]matcher{
 		"has": matchStringListHas,
+		"is":  matchStringListHas,
```
kralicky, line 80: is `is` the same as `has`? I might expect `is` to perform an exact match, but `has` to be more of a "contains" operation.

calebdoxsey: This is matching a list. Do you mean: `is: Smith` should match `["Smith"]` but not `["Smith","Jones"]`.

kralicky: Yes, that is what I would expect.

calebdoxsey: Ok that seems reasonable.

PR #5338, opened 2024-10-23
Add new jwt issuer format route option

kralicky:
## Summary

This is a prerequisite for supporting k8s structured authentication config. Adds a new `jwt_issuer_format` route option that can be used to change the format of the `iss` header in pomerium JWTs for that route. 

The default value `"hostOnly"` is the same as the current behavior, where the issuer will be set to the route host. The other available value `"uri"` will set the issuer to a complete URI including the https scheme, and with a trailing slash.

## Related issues

https://github.com/pomerium/pomerium/issues/5343

## Checklist

- [x] updated unit tests
- [x] ready for review


calebdoxsey (approved):
LGTM

PR #5233, opened 2024-08-22
New integration test fixtures

kralicky:
## Summary

This adds a new integration test fixture in the style of controller-runtime's test environment which can be used for lightweight in-process integration tests. This is not intended to replace the existing integration test system, only to cover a different set of use cases.

Only a basic feature set is implemented, additional features will be added as needed. 

The first example test using this is in protocols_int_test.go. I left the previous implementation there for now (will delete it once this is moved out of draft) to compare against.

Possibly obsoletes https://github.com/pomerium/pomerium/pull/5195

## Checklist

- [ ] reference any related issues
- [ ] updated docs
- [ ] updated unit tests
- [ ] updated UPGRADING.md
- [ ] add appropriate tag (`improvement` / `bug` / etc)
- [ ] ready for review

PR #5340, opened 2024-10-23
Fix many instances of contexts and loggers not being propagated

kralicky:
## Summary

Fixes many instances of contexts (and loggers, by extension) not being propagated throughout the codebase, resulting in inadvertent usage of the global logger which may not have the correct inherited fields.

This also replaces instances where we manually write "return ctx.Err()" with "return context.Cause(ctx)" which is functionally identical, but will also correctly propagate cause errors if present.

Additionally, a couple new global debug variables were added which can be enabled to suppress log messages that occur in places where it is impossible to propagate context. These are not used yet, but are important for https://github.com/pomerium/pomerium/pull/5233

## Related issues

https://github.com/pomerium/pomerium/pull/5233

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/authorize/evaluator/policy_evaluator.go
+++ b/authorize/evaluator/policy_evaluator.go
@@ -147,7 +147,8 @@ func NewPolicyEvaluator(
 
 	// for each script, create a rego and prepare a query.
 	for i := range e.queries {
-		log.Ctx(ctx).Debug().
+		log.Ctx(ctx).
+			Trace().
 			Str("script", e.queries[i].script).
 			Str("from", configPolicy.From).
 			Interface("to", configPolicy.To).
--- a/internal/databroker/server.go
+++ b/internal/databroker/server.go
@@ -28,24 +28,25 @@ import (
 type Server struct {
 	cfg *serverConfig
 
-	mu       sync.RWMutex
-	backend  storage.Backend
-	registry registry.Interface
+	mu         sync.RWMutex
+	backend    storage.Backend
+	backendCtx context.Context
+	registry   registry.Interface
 }
 
 // New creates a new server.
-func New(options ...ServerOption) *Server {
-	srv := &Server{}
-	srv.UpdateConfig(options...)
+func New(ctx context.Context, options ...ServerOption) *Server {
+	srv := &Server{
+		backendCtx: ctx,
+	}
+	srv.UpdateConfig(ctx, options...)
 	return srv
 }
 
 // UpdateConfig updates the server with the new options.
-func (srv *Server) UpdateConfig(options ...ServerOption) {
+func (srv *Server) UpdateConfig(ctx context.Context, options ...ServerOption) {
 	srv.mu.Lock()
 	defer srv.mu.Unlock()
-
-	ctx := context.TODO()
 
 	cfg := newServerConfig(options...)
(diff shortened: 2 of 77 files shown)


calebdoxsey:
The integration tests are failing.

calebdoxsey (approved):

authorize/evaluator/policy_evaluator.go
```diff
@@ -147,7 +147,8 @@ func NewPolicyEvaluator(
 
 	// for each script, create a rego and prepare a query.
 	for i := range e.queries {
-		log.Ctx(ctx).Debug().
+		log.Ctx(ctx).
+			Trace().
```
kenjenkins, line 151: This is an intentional change from debug to trace? Is the `script` field a bunch of Rego, and generally just extra noise in the logs?

internal/databroker/server.go
```diff
@@ -452,18 +454,15 @@ func (srv *Server) getBackend() (backend storage.Backend, err error) {
...
 	switch srv.cfg.storageType {
 	case config.StorageInMemoryName:
-		log.Ctx(ctx).Info().Msg("using in-memory store")
+		log.Ctx(ctx).Info().Msg("initializing new in-memory store")
 		return inmemory.New(), nil
 	case config.StoragePostgresName:
-		log.Ctx(ctx).Info().Msg("using postgres store")
-		backend = postgres.New(srv.cfg.storageConnectionString)
+		log.Ctx(ctx).Info().Msg("initializing new postgres store")
+		return postgres.New(srv.backendCtx, srv.cfg.storageConnectionString), nil
```
kenjenkins, line 464: Can you help me understand `backendCtx`? Why is this different from the new `ctx` parameter on `getBackend()`?

pkg/cmd/pomerium/pomerium.go
```diff
@@ -30,9 +31,30 @@ import (
...
+
+type RunOption func(*RunOptions)
+
+func (o *RunOptions) apply(opts ...RunOption) {
+	for _, op := range opts {
+		op(o)
+	}
+}
+
+func WithOverrideFileManager(fileMgr *filemgr.Manager) RunOption {
```
kenjenkins, line 46: The `fileMgr` changes aren't directly related to contexts and loggers, are they? Would you mind updating the PR description to include the rationale for these changes (or maybe split these off into a separate PR)?

kenjenkins (approved):
Looks good to me, just a few questions.

Looking forward to more accurate log output after this change!

authorize/evaluator/policy_evaluator.go
kralicky: This happened to be cherry-picked along with changes to the log.Ctx line, but yes this writes out massive rego strings and we should probably change it.

internal/databroker/server.go
kenjenkins: OK thanks, makes sense. Would you mind adding that explanation in a comment in the code?

authorize/evaluator/policy_evaluator.go
kenjenkins: Sounds good, just wanted to double check.

kenjenkins (approved):
Thanks!

PR #5349, opened 2024-10-30
HTTP/3 Support

calebdoxsey:
## Summary
Add support for HTTP/3 via a new `codec_type` of `http3`. When chosen two listeners will be configured: a TLS listener the same as `auto` today, except an `alt-svc` header is added to responses indicating `h3` is available, and a new UDP listener on the same port handling Quic traffic.

## Related issues
- https://github.com/pomerium/internal/issues/1999

## Checklist

- [ ] updated unit tests
- [ ] ready for review


calebdoxsey:
Needs some cleanup, unit tests and an integration test.

Testing this is surprisingly difficult. Chrome won't let you use `h3` with self-signed certificates and most pre-packaged versions of curl don't have http3 support. However using a version of curl that does have http3 support requests seemed to be handled properly with this configuration.

calebdoxsey:
I think this is working as it should. I setup an instance of Pomerium, provisioned a valid certificate, and was able to access it using curl, my browser and a public HTTP/3 online check tool. (make sure to open UDP firewall ports though) HTTP/2 also still works as expected.

kenjenkins (approved):

Issue #5357, opened 2024-11-05
authorize: reimplement header evaluation in Go


calebdoxsey:
**Is your feature request related to a problem? Please describe.**

Currently header evaluation is done in `headers.rego`. This has a couple issues:

1. Rego evaluation performance is a bottleneck in benchmarks
2. Rego is difficult for developers to understand and maintain

**Describe the solution you'd like**

We should reimplement headers evaluation in Go.

calebdoxsey:
This is done.

PR #5362, opened 2024-11-06
authorize: rewrite header evaluator to use go instead of rego

calebdoxsey:
## Summary
Rewrite the header evaluator so that its written in Go instead of using the policy defined in `headers.rego`.

## Related issues
- https://github.com/pomerium/pomerium/issues/5357
 

## Checklist

- [ ] updated unit tests
- [ ] ready for review


calebdoxsey:
I will break out the unit test changes and merge those in a PR first to confirm behavior hasn't changed and to get benchmark numbers.

calebdoxsey:
On my machine this is the benchmark:

```
goos: linux
goarch: amd64
pkg: github.com/pomerium/pomerium/authorize/evaluator
cpu: AMD Ryzen 7 5800H with Radeon Graphics         
BenchmarkHeadersEvaluator-16    	    9436	    112199 ns/op	   59549 B/op	    1094 allocs/op
PASS
ok  	github.com/pomerium/pomerium/authorize/evaluator	1.087s
```

Approximately 10x faster.

wasaga (approved):

authorize/evaluator/headers_evaluator_evaluation.go
```diff
@@ -0,0 +1,476 @@
...
+	}
+
+	lv, ok := claims[field]
+	if !ok {
+		return nil, false
+	}
+
+	strs := make([]string, len(lv.Values))
+	for i, v := range lv.Values {
+		strs[i] = fmt.Sprint(v.AsInterface())
```
kralicky, line 413: Do we want to support object/list claim values? This will serialize objects into a format that looks like `map[key:value key2:value2]`. The existing behavior is that if you have any object or list in the claim value, it skips the entire claim, but that doesn't seem correct either.

calebdoxsey: I updated the code to just skip them.

kralicky (approved):

PR #5400, opened 2024-12-18
authorize: handle gRPC requests

calebdoxsey:
## Summary
Add support to the authorize service for gRPC requests. We already supported gRPC-Web, which uses the same headers `grpc-status` and `grpc-message`. 

Updated the code to use `http.Header` consistently. There was also a bug in the `okResponse` with sorting, and now it will just use the same `toEnvoyHeaders` function we were already using.

## Related issues
- https://linear.app/pomerium/issue/ENG-1750/core-respond-with-json-to-a-grpc
 

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/authorize/check_response.go
+++ b/authorize/check_response.go
@@ -5,11 +5,11 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"maps"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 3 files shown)


kralicky (approved):
LGTM

PR #5406, opened 2024-12-20
envoyconfig: unix socket for outbound listener

kenjenkins:
WIP -- move the `outbound-ingress` listener to a Unix domain socket.

At startup Pomerium allocates a number of ports randomly. However, it does so in a way that is liable to a potential race condition.

The OutboundPort is the only one of these ports that is used for an Envoy listener. If we use a Unix domain socket instead, we could pick a fixed path (like we do for the Envoy admin socket) and potentially avoid the race condition.


pkg/grpc/client.go
```diff
@@ -15,6 +16,8 @@ import (
 	"github.com/pomerium/pomerium/pkg/telemetry/requestid"
 )
 
+var OutboundAddress = filepath.Join(os.TempDir(), "pomerium-outbound.sock")
```
calebdoxsey, line 19: Possibly we should use `$XDG_RUNTIME_DIR` if its available?

calebdoxsey (approved):
LGTM

pkg/grpc/client.go
kralicky: The test environment uses/modifies $TMPDIR (picked up by os.TempDir()) to separate extracted files for individual envoy instances, so if we put the socket somewhere else by default we'd also need a way to override that

PR #5411, opened 2024-12-30
evaluator: halt on databroker errors during policy and header evaluation

kralicky:
This updates the evaluator to handle all databroker client errors (previously they were ignored). If such an error is encountered, rego evaluation is immediately stopped and "internal-server-error" allow/deny reasons are returned. This allows us to treat the "databroker not ready" condition differently from the "user not found" condition - the latter of which will continue the auth process and start redirecting the user to sign in, which is not desirable if the databroker is not ready. This causes some confusion especially in cases where the databroker [client] comes online between the time that the user is redirected to sign in (not because the user was not found, but because the databroker was not ready) and the time that the user is redirected back after the oauth flow.

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/authorize/evaluator/evaluator.go
+++ b/authorize/evaluator/evaluator.go
@@ -13,6 +13,8 @@ import (
 	"github.com/hashicorp/go-set/v3"
 	"github.com/open-policy-agent/opa/rego"
 	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/pomerium/pomerium/authorize/internal/store"
 	"github.com/pomerium/pomerium/config"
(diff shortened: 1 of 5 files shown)


calebdoxsey:
I don't agree with this change, but if you are confident its correct we can merge it.

calebdoxsey (approved):

calebdoxsey:
To clarify I am dubious these changes can be done safely. We rely on the existing behavior in subtle ways. I would expect this to break customers in ways we can't predict.

Furthermore returning every error in this way may result in a much harsher experience than was intended originally.

For example, suppose we are syncing the directory data and a user record is updated to reference a new group, but that new group doesn't existing the databroker yet. This change would mean that the evaluation errors out and the request isn't proxied. This was the behavior before. And at least in this case, the previous behavior seems objectively better.

kralicky:
That makes sense, we wouldn't want to change any existing [lack of] short-circuiting logic for errors unrelated to the databroker being unavailable. I'll make some adjustments to this

kralicky:
Closing this, found a simpler alternative solution

PR #5420, opened 2025-01-08
[tracing] refactor to use custom extension for trace id editing

kralicky:
## Summary

By extending Envoy's built-in opentelemetry tracer, we can overwrite trace IDs as they are generated, based on the presence of the pomerium_traceparent query parameter. This means we won't need to batch and post-process trace data, since the trace IDs will always be correct from the start. The implementation is in this PR: https://github.com/pomerium/envoy-custom/pull/1

The only special case is with the /oauth2/callback handler, since the query parameters `code` and `state` are standardized and managed by oauth2 clients. However, we control the contents of the state parameter, so we can add the trace ID into the unencrypted section of the state string to be read by Envoy. The state string technically already contains this data as part of the redirect URL, but the URL is encrypted, so we can't read it from within Envoy.

(note: tests will not work here until the PR in envoy-custom is merged and new binaries are built for v1.32.3)

## Checklist

- [ ] updated unit tests
- [ ] ready for review


calebdoxsey (approved):
bunch of lint issues, but lgtm

PR #5422, opened 2025-01-09
config: set default tracing sample rate to 1.0

kralicky:
The previous default sample rate of 0.0001 is very low, so traces are unlikely to be visible after enabling them until many thousands of requests have been sent. This could be confusing to users.

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/testdata/main_http_connection_manager_filter.json
+++ b/config/envoyconfig/testdata/main_http_connection_manager_filter.json
@@ -52,7 +52,9 @@
             },
             "timeout": "10s"
           },
-          "metadataContextNamespaces": ["com.pomerium.client-certificate-info"],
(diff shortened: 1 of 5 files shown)


config/envoyconfig/testdata/main_http_connection_manager_filter.json
```diff
@@ -180,10 +182,10 @@
...
       "randomSampling": {
-        "value": 0.01
+        "value": 100
       }
     },
     "useRemoteAddress": true,
     "xffNumTrustedHops": 1
   }
-}
+}
```
calebdoxsey, line 191: needs an EOL

calebdoxsey (approved):

PR #5425, opened 2025-01-14
envoy: add internal_address_config to address deprecation warning

kralicky:
## Summary

See docs: https://www.envoyproxy.io/docs/envoy/v1.32.3/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-internal-address-config
## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/filters.go
+++ b/config/envoyconfig/filters.go
@@ -47,6 +47,7 @@ func ExtAuthzFilter(grpcClientTimeout *durationpb.Duration) *envoy_extensions_fi
 func HTTPConnectionManagerFilter(
 	httpConnectionManager *envoy_extensions_filters_network_http_connection_manager.HttpConnectionManager,
 ) *envoy_config_listener_v3.Filter {
+	applyGlobalHTTPConnectionManagerOptions(httpConnectionManager)
 	return &envoy_config_listener_v3.Filter{
 		Name: "envoy.filters.network.http_connection_manager",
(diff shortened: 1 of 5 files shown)


calebdoxsey:
We create `HttpConnectionManager`s in several places. Is this the only one that needs to be updated? Or should we update the others as well? If so perhaps we need a `newHTTPConnectionManager` function that sets it and then we can use it everywhere?

config/envoyconfig/listeners_main.go
```diff
@@ -212,6 +212,21 @@ func (b *Builder) buildMainHTTPConnectionManagerFilter(
...
+
+				// RFC1918
+				{AddressPrefix: "10.0.0.0", PrefixLen: wrapperspb.UInt32(8)},
+				{AddressPrefix: "192.168.0.0", PrefixLen: wrapperspb.UInt32(16)},
+				{AddressPrefix: "172.16.0.0", PrefixLen: wrapperspb.UInt32(12)},
+
+				// RFC4193
+				{AddressPrefix: "fd00::", PrefixLen: wrapperspb.UInt32(8)},
+			},
+		},
```
calebdoxsey, line 229:
```suggestion
```

Shouldn't be needed anymore right?

kralicky: good catch thanks

calebdoxsey (approved):

PR #5439, opened 2025-01-21
authorize: cache warming

calebdoxsey:
## Summary
Add a cache warmer that will sync records from the databroker and add them to the cache. These records are set with a TTL a year in the future so they won't expire. 

## Related issues
- https://linear.app/pomerium/issue/ENG-1915/authorize-improve-group-query-performance

## Checklist

- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/authorize/authorize.go
+++ b/authorize/authorize.go
@@ -10,7 +10,10 @@ import (
 	"time"
 
 	"github.com/rs/zerolog"
+	oteltrace "go.opentelemetry.io/otel/trace"
+	"golang.org/x/sync/errgroup"
 
+	"github.com/pomerium/datasource/pkg/directory"
 	"github.com/pomerium/pomerium/authorize/evaluator"
 	"github.com/pomerium/pomerium/authorize/internal/store"
 	"github.com/pomerium/pomerium/config"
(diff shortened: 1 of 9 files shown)


authorize/grpc.go
```diff
@@ -33,11 +33,8 @@ func (a *Authorize) Check(ctx context.Context, in *envoy_service_auth_v3.CheckRe
 
 	querier := storage.NewTracingQuerier(
```
wasaga, line 34: we don't seem to use .Traces() of this querier outside of tests, do we need it here?

authorize/authorize.go
```diff
@@ -154,5 +166,6 @@ func (a *Authorize) OnConfigChange(ctx context.Context, cfg *config.Config) {
 		log.Ctx(ctx).Error().Err(err).Msg("authorize: error updating state")
 	} else {
 		a.state.Store(state)
+		a.groupsCacheWarmer.UpdateConn(state.dataBrokerClientConnection)
```
wasaga, line 169: it seems like this would be called on _any_ configuration change - i.e. route/etc, so in a high churn environment where a lot of routes get updated all the time, that will be constantly invalidating the cache?

calebdoxsey: I attempted to account for this. What are you suggesting I do instead?

authorize/grpc.go
calebdoxsey: Tracing was probably removed with audit logging. It was a product feature. I guess I can also remove it.

authorize/authorize.go
calebdoxsey: I will update this to only call this function if the connection changes and remove the code I wrote to account for this in the cache warmer.

PR #5448, opened 2025-01-24
proxy: add logo discovery

calebdoxsey:
## Summary
Depends on https://github.com/pomerium/pomerium/pull/5443

Implement favicon discovery. This works by fetching the `to` address, looking for any icons on it, retrieving their contents and embedding them as data-uris.

## Checklist

- [x] updated unit tests
- [x] ready for review


proxy/handlers_portal.go
```diff
@@ -45,7 +48,25 @@ func (p *Proxy) getPortalRoutes(u handlers.UserInfoData) []portal.Route {
...
 		}
 	}
-	return portal.RoutesFromConfigRoutes(routes)
+	portalRoutes := portal.RoutesFromConfigRoutes(routes)
+	for i, pr := range portalRoutes {
+		r := routes[i]
+		for _, to := range r.To {
+			if pr.LogoURL == "" {
+				var err error
+				pr.LogoURL, err = p.logoProvider.GetLogoURL(ctx, pr.From, to.URL.String())
```
kenjenkins, line 57: I'm trying to think what will happen if there are many routes. Are the logos fetched sequentially? Will this end up blocking the page from being served until they are all fetched?

calebdoxsey: Yes and Yes. I could fetch them in parallel.

kenjenkins: I think this is good for now, and we can consider any improvements separately from this PR.

One other idea: maybe we could add a separate handler for serving an individual logo image? (Something like a `/routes/:name/logo` endpoint that we could reference by URL in the generated HTML, without needing to fetch the logo itself in advance.) I think that way the browser will naturally request these in parallel, and if there happens to be a slow upstream service, it shouldn't affect any of the other logos?

(Just a suggestion, like I said I think what you have here is good for now.)

PR #5450, opened 2025-01-27
proto: fix dependencies

calebdoxsey:
## Summary
The script to generate protobuf downloaded several 3rd party repositories. It would skip this if the local directory already existed. Because of the way this was done it would not detect if the URL was updated. This PR updates the code to check the url and if it changes the zip is downloaded. To make this work the URLs also need to be pinned to a specific commit, so I updated the URLs accordingly.

### DIFF
--- a/scripts/protoc
+++ b/scripts/protoc
@@ -12,25 +12,32 @@ fi
 
 # fetch_zip dir name url
 download() {
-  local dst="$1/$2"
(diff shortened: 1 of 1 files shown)


kenjenkins (approved):
Thanks, pinning the proto dependencies makes sense to me.

PR #5460, opened 2025-01-31
config: escape % signs in local reply format string

kenjenkins:
## Summary

Since v0.26, Pomerium configures Envoy to use a custom HTML error page format string for most errors served by Envoy itself. This format string uses %COMMAND% directives to include details about the error.

The HTML error page template also includes any branding options set via the corresponding Enterprise settings. We need to ensure that any % signs in the branding options strings are escaped to %% so that Envoy will not interpret them as the start of a %COMMAND% directive, which could lead to Envoy rejecting the format string as invalid.

## Related issues

https://linear.app/pomerium/issue/ENG-1958/some-logo-url-values-may-result-in-bad-envoy-config

## Checklist

- [x] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/http_connection_manager.go
+++ b/config/envoyconfig/http_connection_manager.go
@@ -56,13 +56,19 @@ func (b *Builder) buildLocalReplyConfig(
 		headers = toEnvoyHeaders(options.GetSetResponseHeaders())
 	}
 
-	data := map[string]any{
-		"status":        "%RESPONSE_CODE%",
-		"statusText":    "%RESPONSE_CODE_DETAILS%",
-		"requestId":     "%STREAM_ID%",
-		"responseFlags": "%RESPONSE_FLAGS%",
-	}
+	data := make(map[string]any)
 	httputil.AddBrandingOptionsToMap(data, options.BrandingOptions)
+	for k, v := range data {
+		// Escape any % signs in the branding options data, as Envoy will
+		// interpret the page output as a substitution format string.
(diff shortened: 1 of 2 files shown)


wasaga (approved):

config/envoyconfig/http_connection_manager.go
```diff
@@ -56,13 +56,19 @@ func (b *Builder) buildLocalReplyConfig(
...
 
-	data := map[string]any{
-		"status":        "%RESPONSE_CODE%",
-		"statusText":    "%RESPONSE_CODE_DETAILS%",
-		"requestId":     "%STREAM_ID%",
-		"responseFlags": "%RESPONSE_FLAGS%",
-	}
+	data := make(map[string]any)
 	httputil.AddBrandingOptionsToMap(data, options.BrandingOptions)
+	for k, v := range data {
```
calebdoxsey, line 61: Should we do this recursively somehow? Maybe JSON marshal, replace `%`, and then JSON unmarshal?

kenjenkins: Good question — I think for now we don't need to, as it looks like all of the branding options are plain strings (based on the [BrandingOptions](https://github.com/pomerium/pomerium/blob/1e9a09269b6ebd8464cbc9f6a91fe58ff212ae76/internal/httputil/branding.go#L4) interface), so we shouldn't need to do anything recursive.

But if we start passing more complicated data into the error page then we'll probably need to change this.

PR #5466, opened 2025-02-03
config: add support for http3 advertise port

### LINKED ISSUE #5465: config: add http3 advertised port option

When HTTP3 is enabled and a request is made to HTTP1 or HTTP2, we add an `Alt-Svc` header indicating `h3` is available:

```go
func newQUICAltSvcHeaderFilter(cfg *config.Config) *envoy_extensions_filters_network_http_connection_manager.HttpFilter {
	listenAddr := buildUDPAddress(cfg.Options.Addr, 443)
	listenPort := listenAddr.GetSocketAddress().GetPortValue()
	return HTTPHeaderMutationsFilter(&envoy_extensions_filters_http_header_mutation_v3.HeaderMutation{
		Mutations: &envoy_extensions_filters_http_header_mutation_v3.Mutations{
			ResponseMutations: []*envoy_config_common_mutation_rules_v3.HeaderMutation{{
				Action: &envoy_config_common_mutation_rules_v3.HeaderMutation_Append{
					Append: &envoy_config_core_v3.HeaderValueOption{
						Header: &envoy_config_core_v3.HeaderValue{
							Key:   "alt-svc",
							Value: fmt.Sprintf(`h3=":%d"; ma=86400`, listenPort),
						},
					},
				},
			}},
		},
	})
}
```

This deduces the port based on the listening address of Pomerium. However sometimes Pomerium listens on a different port from the external one users access because Pomerium is sitting behind a Load Balancer. When this happens we end up advertising a port like `:8443` which won't work because the user needs to hit `:443`.

We should add a configuration option `http3_advertised_port` that will be used for the `Alt-Svc` header. If unset we will keep the current behavior.

Alternatively we could just always assume `:443`, but I'm not sure that's universally the right thing to do.

calebdoxsey:
## Summary
Add a new option to change the http3 advertise port used in the `Alt-Svc` header. This option is only available from the protobuf.

## Related issues
- Fixes #5465 

## Checklist

- [x] updated unit tests
- [x] ready for review


kenjenkins (approved):
Sounds good to me. I think adding this only to the proto API makes sense for now.

PR #5484, opened 2025-02-17
authorize: support authenticating with idp tokens

calebdoxsey:
## Summary
Add support for authenticating with IdP access and identity tokens. Only Azure AD access tokens are implemented.

## Related issues
- [ENG-2001](https://linear.app/pomerium/issue/ENG-2001/core-add-verifyaccesstoken-and-verifyidentitytoken-authenticator)

 

## Checklist

- [ ] updated unit tests
- [x] ready for review


config/bearer_token_format.go
```diff
@@ -0,0 +1,98 @@
...
+	case BearerTokenFormatIDPIdentityToken:
+		return configpb.BearerTokenFormat_BEARER_TOKEN_FORMAT_IDP_IDENTITY_TOKEN.Enum()
+	default:
+		panic(fmt.Sprintf("unknown bearer token format: %v", bearerTokenFormat))
+	}
+}
+
+func decodeBearerTokenFormatHookFunc() mapstructure.DecodeHookFunc {
+	return func(_, t reflect.Type, data any) (any, error) {
+		if t != reflect.TypeOf(BearerTokenFormat("")) {
```
kralicky, line 83:
```suggestion
		if t != reflect.TypeFor[BearerTokenFormat]() {
```

config/session.go
```diff
@@ -112,3 +128,292 @@ func (store *SessionStore) LoadSessionStateAndCheckIDP(r *http.Request) (*sessio
...
+}
+
+func (c *incomingIDPTokenSessionCreator) createSessionAccessToken(
+	ctx context.Context,
+	cfg *Config,
+	policy *Policy,
+	rawAccessToken string,
+) (*session.Session, error) {
+	sessionID := uuid.NewSHA1(accessTokenUUIDNamespace, []byte(rawAccessToken)).String()
+	s, err := c.getSession(ctx, sessionID)
```
kralicky, line 180: should the session id be tied the route as well? what happens if you use the same access token for different routes?

config/session.go
kralicky, line 1: Should probably add some tests here, this logic seems important

config/session.go
calebdoxsey: Sessions are typically shared across routes, but I could make it route-specific.

kralicky: If you use a token that authenticates you to one route, then try to use it for a different route (with a separate policy) it would still check that route's policy, right? if multiple routes have identical policy though then i suppose it makes sense to share the session

calebdoxsey: These sessions will now be route-specific.

If we don't think sessions should be shared across routes in Pomerium we should create a separate follow up issue. I thought session sharing was a feature of the product.

kralicky: If it makes more sense to have the sessions shared, then they should be shared. I'm not sure what would be best. If the session persists, is it just that the user's identity is cached? or does it affect whether the user is authorized for different routes?

calebdoxsey: I don't know what would be best either. I guess I can update it so the session is only shared based on the identity provider config since that can vary by routes (but doesn't usually).

A session is used to track authentication. It tells us who the user is. It does not grant access to routes unless the "any authenticated user" policy is used.

kralicky (commented):
LGTM just a couple of nits

internal/jwtutil/jwtutil.go
```diff
@@ -0,0 +1,180 @@
...
+	if claims == nil {
+		return tm, false
+	}
+
+	raw, ok := claims[name]
+	if !ok {
+		return tm, false
+	}
+
+	switch v := raw.(type) {
```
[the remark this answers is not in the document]
calebdoxsey: Yes because the map may've been built directly without going through JSON unmarshal.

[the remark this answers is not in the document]
kralicky: Ok no problem

kralicky (approved):

PR #5485, opened 2025-02-18
netutil: improve port allocation

calebdoxsey:
## Summary
Try to improve the way we allocate ports so we don't end up re-using a recently allocated port. Hopefully this will improve some test flakiness.

## Checklist

- [ ] updated unit tests
- [ ] ready for review


kralicky (approved):
unsure if this will work, but it's worth a try

Issue #5444, opened 2025-01-22
file updates not picked up


calebdoxsey:
## What happened?

Sometimes when a certificate is updated Pomerium does not pick up the new certificate but continues serving the old certificate. This also seems to happen when updating the `config.yaml`.

There is a log message:

```json
{"level":"info","config_file_source":"/some_path/config.yaml","name":"/some_path/config.yaml","op":"WRITE","time":"2025-01-22T14:43:31-05:00","message":"fileutil/watcher: file notification event"}
```

There is also this log message:

```
"message":"config/filewatchersource: change detected, triggering update"
```

## What did you expect to happen?

After the change is detected for the new config.yaml or certificate to be used.

## Additional context

Just speculating here, but I think the issue may be that we're detecting the change, loading the config, but the new value hasn't been entirely written yet. If multiple updates happen to a file in quick succession its likely the modification time will not change, since the modification time for a file often lacks sufficient precision.

calebdoxsey:
We use https://github.com/namespacelabs/go-filenotify, which is forked from a package in Moby. The original package in Moby has since been removed. I suspect it was not very reliable.

kenjenkins:
I realize it wouldn't directly solve this problem, but I sometimes wonder if Pomerium should install a SIGHUP handler that would trigger a config reload. (I think this is a fairly common pattern; I know nginx does something like this.) That way, at least users could force a reload if they needed to, after they make some config change.

calebdoxsey:
Possibly the issue could be that the files are using timestamps from an archive. If that were the case then two different files could be missed if they have the same size and the same timestamp (because the timestamp is from an archive and not an actual modification time).

Perhaps if we additionally used the logic from [os.SameFile](https://pkg.go.dev/os#SameFile), where the device and inode id are used, we could detect this.

calebdoxsey:
Although `os.SameFile` likely would catch this issue sometimes, it won't always work. If the same file is overwritten it ends up with the same inode and device id, so we still won't detect the change.

I think what we need to do is in addition to checking files via polling, we need to use filesystem notification events on the directories containing those files. The combination of both of these should detect files with the same size and timestamp when using filesystems and operating systems that support notifications. There'd still be a gap for filesystems and operating systems that don't support notifications, but we'd be no worse off than our current behavior.

PR #5498, opened 2025-02-25
fileutil: reimplement file watcher

calebdoxsey:
## Summary
Reimplement the `fileutil.Watcher`.

The current version of the watcher works by periodically polling file metadata (size and modification time) every 200ms. If this information about a file changes, a signal is broadcast triggering configuration to reload.

This design has two issues:

1. A file could be updated but preserve the modification time and file size. Such a change would go undetected by the file watcher.
2. The watcher broadcasts changes unnecessarily if a file is updated but the contents stay the same.

The new design also polls file metadata. If a change to metadata is detected the file contents are hashed and only if that hash is different will a change be signaled (solving 2).

In addition an fsnotify watcher - which uses filesystem notification events to detect changes - is used on all of the directories of all of the files being watched. If a change event is detected via this mechanism we flag the file as requiring a full hash check (solving 1).

The polling interval is also used as a debounce mechanism since files often require many writes before being completed.

It is still possible to read a partially written file, but I don't think this is something we should solve. Rather, applications should write files by writing to a separate temporary file and then renaming them after the file is fully written. 

## Related issues
- #5444 

## Checklist

- [x] updated unit tests
- [x] ready for review


config/config_source.go
```diff
@@ -136,12 +136,13 @@ func NewFileOrEnvironmentSource(
 		watcher:    fileutil.NewWatcher(),
 		config:     cfg,
 	}
+	context.AfterFunc(ctx, func() { src.watcher.Close() })
```
calebdoxsey, line 139: This appears to be what we were doing in the file watcher originally. I changed the watcher to have an explicit Close and tried to make sure we were calling it as before.

kenjenkins (approved):

PR #5508, opened 2025-02-28
add global jwt_issuer_format option

kenjenkins:
## Summary

Add a corresponding global setting for the existing route-level `jwt_issuer_format` option. The route-level option will take precedence when set to a non-empty string.

## Related issues

https://linear.app/pomerium/issue/ENG-2072/config-jwt-issuer-format-should-also-be-support-at-the-global-level

## Checklist

- [x] updated unit tests
- [ ] ready for review


authorize/evaluator/headers_evaluator_evaluation.go
```diff
@@ -248,17 +248,15 @@ func (e *headersEvaluatorEvaluation) getGroupIDs(ctx context.Context) []string {
...
 	}
 	switch issuerFormat {
-	case "uri":
+	case config.JWTIssuerFormatURI:
 		return fmt.Sprintf("https://%s/", e.request.HTTP.Hostname), nil
-	case "", "hostOnly":
-		return e.request.HTTP.Hostname, nil
 	default:
-		return "", fmt.Errorf("unsupported JWT issuer format: %s", issuerFormat)
+		return e.request.HTTP.Hostname, nil
```
calebdoxsey, line 259: this function now never returns an error

kenjenkins: I thought it was odd to return an error here, but I can restore this logic if you prefer.

I've added a new check in Options.Validate() and Policy.Validate() to make sure the option is valid.

calebdoxsey: Sorry I meant it was not necessary to return `(string, error)` and we could just return `(string)` because it now always returns nil as the error.

calebdoxsey (approved):

kralicky (approved):
lgtm

authorize/evaluator/headers_evaluator_evaluation.go
kenjenkins: Oh thanks, I understand now. Should be fixed.

Issue #5378, opened 2024-11-24
Accessing Pomerium Metrics Endpoint Causing Pod Memory Spike


doctor500:
## What happened?
Currently, I have a use case where I'm using Pomerium as an Ingress Controller setup in a GKE Cluster. This ingress controller has more than 500+ Ingress manifests configured. I want to set Cloud Monitoring can consume Pomerium Custom Metrics using Google Managed Prometheus, so we decided to set up our managed collection with the following manifests:

```
apiVersion: monitoring.googleapis.com/v1
kind: PodMonitoring
metadata:
  name: pomerium
  namespace: pomerium
spec:
  endpoints:
    - port: metrics
      interval: 30s
      metricRelabeling:
      - action: keep
        regex: "http_client_requests_total"
        sourceLabels: [__name__]
      - action: keep
        regex: "http_server_requests_total"
        sourceLabels: [__name__]
      - action: keep
        regex: "grpc_client_requests_total"
        sourceLabels: [__name__]
      - action: keep
        regex: "grpc_server_requests_total"
        sourceLabels: [__name__]
  selector:
    matchLabels:
      app.kubernetes.io/name: pomerium
``` 

But turns out it causing all of the Pomerium Controller pod memory to increase almost touching pod limit specs (4GB) in our case. I'm observing by sampling the metrics payload size and found that the payload size is having a size for more than 100MB for each call. Even if we increase the interval collection time, the memory spike still happened.

```
root@xxxxx-6zk5478998-wrr5q:/# curl http://pomerium-metrics.pomerium.svc.cluster.local:9090/metrics > sample.txt
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  122M    0  122M    0     0  16.0M      0 --:--:--  0:00:07 --:--:-- 36.1M
```
```
root@xxxxx-6zk5478998-wrr5q:/# ls -lh --block-size=MB sample.txt
-rw-r--r-- 1 root root 128MB Nov 24 07:48 sample.txt
```
 
This is the current pod condition after the curl process is done (monitored using k9s)
<img width="1149" alt="image" src="https://github.com/user-attachments/assets/2170eb7e-c467-49a2-b0c3-ff8ef53a0fa1">

## What did you expect to happen?
[...]

PR #5530, opened 2025-03-18
metrics: reduce gc pressure

wasaga:
## Summary

We add extra labels (hostname and installation id) to the Prometheus metrics output.
Previously we were parsing the incoming streams (from Envoy and internal ones) with `expfmt` prometheus library, updated each entry with extra labels and encoded back.

Unfortunately in a large streams that creates a significant GC pressure as a lot of small objects are created and destroyed.

This PR tries to do metric label augmentation with minimal GC allocations. It also adds a runtime flag `add_extra_metrics_labels` that would disable adding extra labels entirely, if they are not explicitly used during queries. 

## Related issues

Ref: https://linear.app/pomerium/issue/ENG-2117
Ref: https://github.com/pomerium/pomerium/issues/5378

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/telemetry/metrics/bench_test.go
+++ b/internal/telemetry/metrics/bench_test.go
@@ -1,0 +1,115 @@
+package metrics_test
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/pomerium/pomerium/config"
+	"github.com/pomerium/pomerium/internal/testenv"
+	"github.com/pomerium/pomerium/internal/testenv/snippets"
+	"github.com/pomerium/pomerium/internal/testenv/upstreams"
+	"github.com/zeebo/assert"
+)
+
+func TestScrapeMetricsEndpoint(t *testing.T) {
+	t.Skip("this test is for profiling purposes only")
+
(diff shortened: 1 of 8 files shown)


calebdoxsey (approved):

internal/telemetry/metrics/bench_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"time"
+
+	"github.com/pomerium/pomerium/config"
+	"github.com/pomerium/pomerium/internal/testenv"
+	"github.com/pomerium/pomerium/internal/testenv/snippets"
+	"github.com/pomerium/pomerium/internal/testenv/upstreams"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestScrapeMetricsEndpoint(t *testing.T) {
```
kenjenkins, line 18: Would it be feasible to write this as a Go benchmark test, so we can measure memory allocations with the `-benchmem` option?

wasaga: added a separate benchmark old vs new 
```
BenchmarkRelabelTestStream/RelabelTextStream-10         	 6379164	       166.3 ns/op	       6 B/op	       6 allocs/op
BenchmarkRelabelTestStream/Previous-10                  	  211172	      5620 ns/op	    2805 B/op	      97 allocs/op
```

PR #5540, opened 2025-03-24
databroker: preserve data type when deleting changeset

calebdoxsey:
## Summary
When the `Reconciler` deletes a record it currently creates an object like this:

```
&Record{
  Type:      typ,
  Id:        id,
  DeletedAt: cs.now,
  Data:      &anypb.Any{TypeUrl: typ},
}
```

Unfortunately this fails to account for objects where the type stored in the databroker does not match the type of the data. For example directory groups and users have the type `pomerium.io/DirectoryGroup` and `pomerium.io/DirectoryUser`, but their data is stored as a struct.

This PR updates the code to preserve the existing data and just update the deleted at timestamp instead.

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/grpc/databroker/changeset.go
+++ b/pkg/grpc/databroker/changeset.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 
-	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/grpc/databroker/changeset.go
```diff
@@ -33,13 +33,10 @@ type changeSet struct {
...
-func (cs *changeSet) Remove(typ string, id string) {
-	cs.updates = append(cs.updates, &Record{
-		Type:      typ,
-		Id:        id,
-		DeletedAt: cs.now,
-		Data:      &anypb.Any{TypeUrl: typ},
-	})
+func (cs *changeSet) Remove(record *Record) {
+	record = proto.Clone(record).(*Record)
+	record.DeletedAt = timestamppb.Now()
```
kralicky, line 38: should this stay `record.DeletedAt = cs.now`?

kenjenkins (approved):

Issue #5549, opened 2025-03-27
New Route Portal leads to infinite redirects


aponert:
## What happened?
I updated several pomerium instances from 0.28.0 to 0.29.0. I wanted to try the new routes portal. on clicking on the menu Item I get errors and infinite redirects

## What did you expect to happen?
I expected to see the new feature

## How'd it happen?

1. Went to my pomerium url
2. Logged in via my IDP
3. Clicked on 'Routes'

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`): 0.29.0
- Server Operating System/Architecture/Cloud: ubuntu linux server

## What's your config.yaml?

```config.yaml
runtime_flags:
  config_hot_reload: false
authenticate_service_url: https://auth.<domain>
address: :443
signing_key: <redacted>
metrics_address: :9099
autocert: true
http_redirect_addr: :80
downstream_mtls:
  ca_file: /etc/pomerium/mtlsCA.pem
  enforcement: reject_connection

idp_provider: oidc
idp_client_id: <redacted>
idp_client_secret: <redacted>
idp_provider_url: <redacted>

routes:

  - from: tcp+https://ssh-ap.<domain>:22
    name: ap_ssh

    to: tcp://10.0.0.10:22
    pass_identity_headers: false
    preserve_host_header:  false

    tls_skip_verify: false
    allow_websockets: false
    allow_spdy: false


    policy:
    - allow:
        or:
        - claim/groups: pomerium:user



  - from: https://argocd.<domain>:443
    name: argocd

    to: https://argocd.home:443
    pass_identity_headers: false
    preserve_host_header:  false

    tls_skip_verify: true
    allow_websockets: false
    allow_spdy: false


    policy:
    - allow:
        or:
        - claim/groups: pomerium:user
 
from here: just further routes
```

## What did you see in the logs?

```logs
no errors shown
```

aponert:
Hi @tsullyo,

thanks for your reply. I created some HAR files. 2 of them are from Chrome and I also did one on Firefox. I used incognito windows and logged in via my IDP. The HARs only contain the part where I click on "Routes"

Description of environments:

_node 1:_
The Pomerium on node1 is **directly internet-facing**. The port 4443 is destination-NATed to port 4443 of the pomerium machine, which is a standard ubuntu server running the pomerium binary natively. This machine lies in a vmWare environment with no further firewall appliances in front of it.

_node2:_
[...]

PR #5550, opened 2025-03-28
allow access to the routes portal without logging in, change status code

calebdoxsey:
## Summary
Make it so that if you want to get to the routes portal you no longer need to be logged in. Instead you will see a warning box, just like the user info page.

Also update the status codes so that when not logged in these pages return a 401.

## Related issues
- [ENG-2199](https://linear.app/pomerium/issue/ENG-2199/core-user-info-endpoint-should-return-401-when-not-logged-in)
- #5549 

## Checklist

- [ ] updated unit tests
- [x] ready for review


calebdoxsey:
I made it so that `/.pomerium/api/v1/routes` requires login again and that fixes the cli.

calebdoxsey:
We may still want the 401 status code, but we don't want the remove the login redirect.

PR #5563, opened 2025-04-03
add support for pomerium.request.headers for set_request_headers

calebdoxsey:
## Summary
Add additional `set_request_headers` templating functionality for `${pomerium.request.headers["Some-Header"]}`. This was a significant deviation from the existing syntax, so I wrote a parser to try and handle more cases. I looked at `peg` and some other parsers, but ended up just rolling one by hand.

## Related issues
- [ENG-2224](https://linear.app/pomerium/issue/ENG-2224/core-support-dynamic-upstream-request-header-substitution)
 

## Checklist

- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/headertemplate/headertemplate.go
+++ b/internal/headertemplate/headertemplate.go
@@ -9,6 +9,22 @@ func Render(src string, fn func(ref []string) string) string {
 	return p.parse()
 }
 
+// This is a hand written parser attempting to model this peg grammar:
+//
+// Grammar <- ( Variable / Text )* !.
+// Text <- .
+// Variable <- EscapedVariable / SimpleVariable / ComplexVariable
+// EscapedVariable <- '$' '$'
+// SimpleVariable <- '$' SimpleExpression
+// SimpleExpression <- identifier ( '.' identifier )*
+// ComplexVariable <- '$' '{' _ ComplexExpression _ '}'
+// ComplexExpression <- identifier _ (ComplexSelector / ComplexIndex)*
+// ComplexSelector <- '.' _ ComplexExpression _
+// ComplexIndex <- '[' _ StringLiteral _ ']' _
+// StringLiteral <- '"' (('\\'.) / [^"])* '"'
+// identifier <- [a-zA-Z0-9_] [a-zA-Z0-9_\-]*
+// _ <- ( ' ' / '\t' )*
+
 type parser struct {
 	buffer []byte
 	pos    int


internal/headertemplate/headertemplate.go
```diff
@@ -0,0 +1,239 @@
...
+
+import "strings"
+
+// Render renders a header template string.
+func Render(src string, fn func(ref []string) string) string {
+	p := newParser(src, fn)
+	return p.parse()
+}
+
+type parser struct {
```
kralicky, line 12: could you add a comment here with a reference ebnf grammar?

authorize/evaluator/headers_evaluator_test.go
```diff
@@ -218,15 +218,19 @@ func TestHeadersEvaluator(t *testing.T) {
...
-						"Client-Cert-Fingerprint": "${pomerium.client_cert_fingerprint}",
-						"Authorization":           "Bearer ${pomerium.jwt}",
-						"Foo":                     "escaped $$dollar sign",
+						"X-Custom-Header":          "CUSTOM_VALUE",
+						"X-ID-Token":               "${pomerium.id_token}",
+						"X-Access-Token":           "${pomerium.access_token}",
+						"Client-Cert-Fingerprint":  "${pomerium.client_cert_fingerprint}",
+						"Authorization":            "Bearer ${pomerium.jwt}",
+						"Foo":                      "escaped $$dollar sign",
+						"X-Incoming-Custom-Header": `From-Incoming ${pomerium.request.headers["X-Incoming-Header"]}`,
```
kralicky, line 233: Out of curiosity, what happens if you try to do a nested expression like `${pomerium.request.headers[${pomerium.request.headers["foo"]}]}`

internal/headertemplate/headertemplate.go
calebdoxsey: I don't have one. Do you want me to write one?

authorize/evaluator/headers_evaluator_test.go
calebdoxsey: That would be an invalid expression. It would end up printing out the literal contents.

calebdoxsey: If we want to support something like that should we switch to lua or some other embedded scripting language?

kralicky: No that's fine, was just curious

internal/headertemplate/headertemplate.go
kralicky: Yes that would be helpful. Just an equivalent grammar for reference in a comment somewhere here

calebdoxsey: I added a peg grammar.

kralicky (approved):

PR #5576, opened 2025-04-16
authorize: refactor logAuthorizeCheck()

kenjenkins:
## Summary

Currently, policy evaluation and authorize logging are coupled to the Envoy `CheckRequest` proto message (part of the ext_authz API). In the context of ssh proxy authentication, we won't have a `CheckRequest`. Instead, let's make the existing evaluator.Request type the source of truth for the authorize log fields.

This way, whether we populate the `evaluator.Request` struct from an ext_authz request or from an ssh proxy request, we can use the same `logAuthorizeCheck()` method for logging.

Add some additional fields to `evaluator.RequestHTTP` for the authorize log fields that are not currently represented in this struct. Move some existing conversion methods to a new package to avoid an import cycle.

## Checklist

- [ ] updated unit tests
- [ ] ready for review


kralicky (approved):
LGTM

PR #5586, opened 2025-04-24
mcp: authorize request (pt2)

wasaga:
## Summary

Once authorization request is stored in the databroker, generate a code that client can use in the `/token` exchange, that represents an encrypted authorization request ID. 
As advised by the RFC, _A maximum authorization code lifetime of 10 minutes is RECOMMENDED._
The authorization request record also needs to expire, that would be addressed in a follow up PR. 

## Related issues

Ref: https://linear.app/pomerium/issue/ENG-2320/mcp-support-oauth-21-authorization-flow

## Checklist

- [x] updated unit tests
- [x] ready for review


internal/mcp/cipher.go
```diff
@@ -0,0 +1,50 @@
...
+
+func getCipher(
+	cfg *config.Config,
+) (cipher.AEAD, error) {
+	secret, err := cfg.Options.GetSharedKey()
+	if err != nil {
+		return nil, fmt.Errorf("shared key: %w", err)
+	}
+
+	rnd := hkdf.New(sha256.New, secret, nil, nil)
```
calebdoxsey, line 23: We also use this for zero bootstrap... should we use a different salt?

internal/mcp/code.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/bufbuild/protovalidate-go"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	oauth21proto "github.com/pomerium/pomerium/internal/oauth21/gen"
+	"github.com/pomerium/pomerium/pkg/cryptutil"
+)
+
+func CreateCode(
```
calebdoxsey, line 17: Could we add some tests for these functions?

calebdoxsey (approved):

PR #5587, opened 2025-04-24
mcp: token: handle authorization_code request (pt1)

wasaga:
## Summary

Handle `authorization_code` grant call to `/token` in oauth2.1. 
only perform PKCE check that challenge matches to the authorization request that was saved on `/authorize` call step. 

## Related issues

Ref: https://linear.app/pomerium/issue/ENG-2320/mcp-support-oauth-21-authorization-flow

## Checklist

- [x] updated unit tests
- [x] ready for review


internal/mcp/handler_token.go
```diff
@@ -2,9 +2,64 @@ package mcp
...
+		return
+	}
+
+	err = AuthorizeTokenRequest(req, authReq)
+	if err != nil {
+		oauth21.ErrorResponse(w, http.StatusBadRequest, oauth21.InvalidGrant)
+		return
+	}
+
+	http.Error(w, "Not Implemented", http.StatusNotImplemented)
```
kenjenkins, line 64: Is "not implemented" still the correct status here? (What still needs to be done here?)

internal/mcp/token.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/pomerium/pomerium/internal/oauth21"
+	"github.com/pomerium/pomerium/internal/oauth21/gen"
+)
+
+func AuthorizeTokenRequest(
+	tokReq *gen.TokenRequest,
+	authReq *gen.AuthorizationRequest,
+) error {
+	if tokReq.GrantType != "authorization_code" {
```
kenjenkins, line 14: It looks like the grant type is also checked in Handler.Token() above. Should we consolidate these to a single check?

internal/mcp/token.go
```diff
@@ -0,0 +1,37 @@
...
+		return fmt.Errorf("unexpected grant type: %s", tokReq.GrantType)
+	}
+
+	if tokReq.ClientId == nil {
```
kenjenkins, line 18: Likewise, doesn't Handler.handleAuthorizationCodeToken() check this?

internal/mcp/handler_token.go
wasaga: Yes. See follow up pr https://github.com/pomerium/pomerium/pull/5589

kenjenkins (approved):

PR #5593, opened 2025-04-29
mcp: pass access token to the upstream

wasaga:
## Summary

If the upstream is an MCP Client application itself, and needs to communicate with MCP Servers protected by Pomerium, 
it needs an access token that it can set to the Authorization header. 

This PR adds `mcp: pass_upstream_access_token: true` setting for the route, that would send an access token representing user's session. 

## Related issues

Fix: https://linear.app/pomerium/issue/ENG-2336/[draft]-mcp-clients-support-passing-access-tokens

## Checklist

- [ ] updated unit tests
- [x] ready for review


wasaga:
I will probably split the settings in a follow-up PR for better clarity, so that it would look like this: 

the MCP client upstream would have
```yaml
mcp:
  client:
    pass_access_token: true
```


and MCP server upstream would have
```yaml
mcp:
  server:
    oauth2: 
```

authorize/evaluator/config.go
```diff
@@ -17,6 +19,7 @@ type evaluatorConfig struct {
 	JWTClaimsHeaders                                  config.JWTClaimHeaders
 	JWTGroupsFilter                                   config.JWTGroupsFilter
 	DefaultJWTIssuerFormat                            config.JWTIssuerFormat
+	MCPAccessTokenProvider                            func(string, time.Time) (string, error) `hash:"-"`
```
wasaga, line 22: in case of MCP server upstream, we would use this function to set the authorization header with the access token of an external oauth2 provider such as github/notion/etc

authorize/evaluator/headers_evaluator_evaluation.go
```diff
@@ -93,6 +93,27 @@ func (e *headersEvaluatorEvaluation) fillJWTClaimHeaders(ctx context.Context) er
...
+		e.request.Policy.MCP == nil ||
+		!e.request.Policy.MCP.PassUpstreamAccessToken {
+		return nil
+	}
+
+	if e.request.Session.ID == "" {
+		return nil
+	}
+
+	accessToken, err := e.evaluator.store.GetMCPAccessTokenProvider()(e.request.Session.ID, e.now.Add(time.Hour))
```
wasaga, line 108: for follow-up: need populate Session with expiration time.

calebdoxsey (approved):

PR #5618, opened 2025-05-15
config: use stable route ids for authorize matching and order xds responses

calebdoxsey:
## Summary
Update the `RouteID` to use the `policy.ID` if it is set. This makes it so that updated routes use a stable identifier between updates so if the envoy control plane is updated before the authorize service's internal definitions (or vice-versa) the authorize service will still be able to match the route.

The current behavior results in a 404 if envoy passes the old route id. The new behavior will result in inconsistency, but it should be quickly remedied. To help with debugging 4 new fields were added to the authorize check log. The `route-id` and `route-checksum` as the authorize sees it and the `envoy-route-id` and `envoy-route-checksum` as envoy sees it.

I also updated the way we send updates to envoy to try and model their recommended approach:

> In general, to avoid traffic drop, sequencing of updates should follow a make before break model, wherein:
> 
> - CDS updates (if any) must always be pushed first.
> - EDS updates (if any) must arrive after CDS updates for the respective clusters.
> - LDS updates must arrive after corresponding CDS/EDS updates.
> - RDS updates related to the newly added listeners must arrive after CDS/EDS/LDS updates.
> - VHDS updates (if any) related to the newly added RouteConfigurations must arrive after RDS updates.
> - Stale CDS clusters and related EDS endpoints (ones no longer being referenced) can then be removed.

This should help avoid 404s when configuration is being updated.

## Related issues
- [ENG-2386](https://linear.app/pomerium/issue/ENG-2386/large-number-of-routes-leads-to-404s-and-slowness)

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/policy.go
+++ b/config/policy.go
@@ -761,7 +761,15 @@ func (p *Policy) Checksum() uint64 {
 // - path
 // - regex
 // - to/redirect/response (whichever is set)
-func (p *Policy) RouteID() (uint64, error) {
+func (p *Policy) RouteID() (string, error) {
+	if p.ID != "" {
+		return p.ID, nil
+	}
+
+	return p.generateRouteID()
+}
+
+func (p *Policy) generateRouteID() (string, error) {
 	// this function is in the hot path, try not to allocate too much memory here
 	hash := hashutil.NewDigest()
 	hash.WriteStringWithLen(p.From)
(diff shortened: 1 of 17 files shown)


config/policy.go
```diff
@@ -807,12 +815,12 @@ func (p *Policy) RouteID() (uint64, error) {
 		hash.WriteInt32(int32(p.Response.Status))
 		hash.WriteStringWithLen(p.Response.Body)
 	default:
-		return 0, errEitherToOrRedirectOrResponseRequired
+		return "", errEitherToOrRedirectOrResponseRequired
 	}
-	return hash.Sum64(), nil
+	return fmt.Sprintf("%x", hash.Sum64()), nil
```
kralicky, line 820:
```suggestion
	return hex.EncodeToString(hash.Sum64()), nil
```

kralicky (commented):
LGTM just one nit

config/policy.go
calebdoxsey: `EncodeToString` expects a []byte

kralicky: oh sorry, perhaps 
```go
hex.EncodeToString(binary.LittleEndian.AppendUint64(nil, hash.Sum64()))
```

calebdoxsey: actually we can just use `hash.Sum` since it returns bytes.

PR #5622, opened 2025-05-19
Add websocket denial test

desimone:
## Summary
- fix err variable check in websocket message read
- add `TestHttp1WebsocketDenied` to ensure websockets are rejected when not allowed

## User Explanation
Adds a regression test verifying that routes without `AllowWebsockets` return an error when establishing websocket connections.

## Checklist
- [x] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/protocols_int_test.go
+++ b/config/envoyconfig/protocols_int_test.go
@@ -343,13 +343,43 @@ func TestHttp1Websocket(t *testing.T) {
 			return err
 		}
 		mt, bytes, err := conn.ReadMessage()
-		if err := err; err != nil {
+		if err != nil {
 			return err
 		}
 		assert.Equal(t, websocket.TextMessage, mt)
 		assert.Equal(t, "hello world", string(bytes))
 		return nil
 	}, upstreams.Path("/ws")))
+}
+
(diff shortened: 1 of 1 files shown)


config/envoyconfig/protocols_int_test.go
```diff
@@ -352,6 +352,36 @@ func TestHttp1Websocket(t *testing.T) {
...
+		Policy(func(p *config.Policy) {
+			p.AllowPublicUnauthenticatedAccess = true
+		})
+
+	env.AddUpstream(up)
+	env.Start()
+	snippets.WaitStartupComplete(env)
+
+	err := up.DialWS(route, func(_ *websocket.Conn) error { return nil }, upstreams.Path("/ws"))
+	require.Error(t, err)
```
kenjenkins, line 382: Tests that exercise failure modes are often tricky, because it's easy for the test to fail for the wrong reason.

Is there a specific error message we should look for here?

Or, even better, should we merge this scenario into the `TestHttp1Websocket` method above? We could define a second route to the same upstream, just without the `AllowWebsockets` option set. That way we would know that the `DialWS()` error isn't because of any misconfiguration of the upstream server.

kenjenkins (approved):

PR #5633, opened 2025-05-30
move mac build to mac-builds

gaurdro:
## Summary
In our efforts to sign and notarize binaries we want to consolidate builds to a private repo for darwin.  This moves the build to a workflow in mac-builds, and moves goreleaser mac config to a separate file. 

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/.github/goreleaser-mac.yaml
+++ b/.github/goreleaser-mac.yaml
@@ -81,3 +81,10 @@ brews:
     folder: Formula
     install: |
       bin.install "pomerium"
+    pull_request:
+      enabled: true
+      draft: true
+      base:
+        owner: pomerium
+        name: homebrew-tap
+        branch: main
--- a/.github/goreleaser.yaml
+++ b/.github/goreleaser.yaml
@@ -243,6 +243,37 @@ docker_manifests:
       - pomerium/pomerium:debug-nonroot-arm64v8-v{{ .Major }}.{{ .Minor }}
       - pomerium/pomerium:debug-nonroot-amd64-v{{ .Major }}.{{ .Minor }}
 
+brews:
+  - # Name template of the recipe
+    name: pomerium
+    # IDs of the archives to use.
+    ids:
+      - pomerium
+    # GOARM to specify which 32-bit arm version to use if there are multiple versions
+    # from the build section. Brew formulas support atm only one 32-bit version.
+    # Default is 6 for all artifacts or each id if there a multiple versions.
+    goarm: 6
+    tap:
+      owner: pomerium
+      name: homebrew-tap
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -65,6 +65,14 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           APPARITOR_GITHUB_TOKEN: ${{ secrets.APPARITOR_GITHUB_TOKEN }}
 
+      - name: trigger mac build
+        uses: peter-evans/repository-dispatch@ff45666b9427631e3450c54a1bcbee4d9ff4d7c0
+        with:
+          repository: pomeirum/mac-builds
```
kenjenkins: typo?
```suggestion
          repository: pomerium/mac-builds
```

.github/goreleaser.yaml
```diff
@@ -244,30 +243,6 @@ docker_manifests:
       - pomerium/pomerium:debug-nonroot-arm64v8-v{{ .Major }}.{{ .Minor }}
       - pomerium/pomerium:debug-nonroot-amd64-v{{ .Major }}.{{ .Minor }}
 
-brews:
```
kenjenkins: Will this end up removing the Linux builds from the homebrew tap?

.github/goreleaser-mac.yaml
```diff
@@ -0,0 +1,87 @@
...
+brews:
+  - # Name template of the recipe
+    name: pomerium
+    # IDs of the archives to use.
+    ids:
+      - pomerium
+    # GOARM to specify which 32-bit arm version to use if there are multiple versions
+    # from the build section. Brew formulas support atm only one 32-bit version.
+    # Default is 6 for all artifacts or each id if there a multiple versions.
+    goarm: 6
```
kenjenkins: I realize this was already present, but I think this won't have any effect because the only ARM build is for arm64. Let's remove it or file a ticket to clean this up later (if you want to keep this PR focused just on the mac builds change).

.github/goreleaser.yaml
gaurdro: hmm, I don't actually know.  I'll investigate.

gaurdro: I wasn't finding good information on the behavior, so I replaced the block and made both make pull requests, so a human can figure it out, if there's a problem

kenjenkins: OK, sounds good for now. We can see what the pull requests look like, but I think it's likely that each PR will delete builds from the other platform. If that's the case we can refactor this to use a separate script like we have for pomerium-cli (https://github.com/pomerium/homebrew-tap/blob/master/scripts/cli-release).

PR #5639, opened 2025-06-02
databroker: add sync-cache

calebdoxsey:
## Summary
Add a new `SyncCache`:

```go
type SyncCache interface {
	// Clear deletes all the data for the given record type in the sync cache.
	Clear(recordType string) error
	// Records yields the databroker records stored in the cache.
	Records(recordType string) iter.Seq2[*Record, error]
	// Sync syncs the cache with the databroker.
	Sync(ctx context.Context, client DataBrokerServiceClient, recordType string) error
}
```

The cache maintains databroker records in a local pebble database (which could be on-disk or in-memory). The way it's used is you first call `.Sync(ctx, client, recordType)` and then `.Records(recordType)`, which returns an iterator over all the records.

Internally we store the databroker records in a pebble key-value database. Pebble was chosen because its fast and well-tested, but any ordered key-value store would work. The first time we call `SyncLatest` to retrieve all the records. Each subsequent time we call `Sync` with the current server and record versions to retrieve only the changes. This is significantly more efficient than calling `SyncLatest` every time.

The primary use for this is in the enterprise-console as part of directory sync to improve performance with large datasets.

## Related issues
- [ENG-2401](https://linear.app/pomerium/issue/ENG-2401/enterprise-console-improve-performance-of-directory-sync-using-cached)

## Checklist

- [x] updated unit tests
- [x] ready for review


pkg/pebbleutil/pebbleutil.go
```diff
@@ -0,0 +1,115 @@
...
+// MustOpenMemory opens an in-memory pebble database. It panics if there is an error.
+func MustOpenMemory(options *pebble.Options) *pebble.DB {
+	if options == nil {
+		options = new(pebble.Options)
+	}
+	options.FS = vfs.NewMem()
+	return MustOpen("", options)
+}
+
+// MustOpen opens a pebble database. It sets options useful for pomerium.
```
wasaga:
```suggestion
// Open opens a pebble database. It sets options useful for pomerium.
```

pkg/grpc/databroker/sync_cache.go
```diff
@@ -0,0 +1,293 @@
...
+	batch := c.db.NewBatch()
+	defer batch.Close()
+
+	for {
+		res, err := stream.Recv()
+		if errors.Is(err, io.EOF) {
+			break
+		} else if status.Code(err) == codes.Aborted {
+			// the server version changed, so use sync latest
+			return c.syncLatest(ctx, client, recordType)
```
wasaga, line 162: do we need to abort the existing stream here ?

wasaga (approved):

pkg/grpc/databroker/sync_cache.go
calebdoxsey: I think the deferred context cancel above should abort the stream.

calebdoxsey: oh, but not until sync latest completes... let me refactor that.

PR #5651, opened 2025-06-11
ppl: add in string matcher

wasaga:
## Summary

Adds `in` string matcher to PPL, that matches a string with an array of strings. 

## Related issues

Related: https://linear.app/pomerium/issue/ENG-2393

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/policy/criteria/matchers.go
+++ b/pkg/policy/criteria/matchers.go
@@ -69,31 +69,7 @@ func matchStringIn(dst *ast.Body, left *ast.Term, right parser.Value) error {
 	if !ok {
 		return fmt.Errorf("in matcher requires an array of strings")
 	}
-
-	var terms []*ast.Term
-	for _, v := range arr {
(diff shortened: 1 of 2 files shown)


pkg/policy/criteria/matchers.go
```diff
@@ -63,6 +64,39 @@ func matchStringStartsWith(dst *ast.Body, left *ast.Term, right parser.Value) er
 	return nil
 }
 
+func matchStringIn(dst *ast.Body, left *ast.Term, right parser.Value) error {
```
calebdoxsey, line 67: I think you can simplify this to:

```
func matchStringIn(dst *ast.Body, left *ast.Term, right parser.Value) error {
	*dst = append(*dst, ast.Member.Expr(left, ast.NewTerm(right.RegoValue())))
	return nil
}
```

Rego has a builtin `x in y` operator, and it's called `Member` in the go package.

PR #5656, opened 2025-06-16
config: support multiple running services in addition to all-in-one mode

calebdoxsey:
## Summary
Currently you can either run a single service or all-in-one mode. This PR adds support for running any 2 or 3 services together. For example `services: authorize,proxy`.

I think I updated the conditions to make sense, but there's some risk here as there may have been some assumptions in places I'm overlooking.

## Related issues
- [ENG-2485](https://linear.app/pomerium/issue/ENG-2485/core-support-running-multiple-but-not-all-services)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/helpers.go
+++ b/config/helpers.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"slices"
 	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


config/helpers.go
```diff
@@ -20,65 +25,62 @@ const (
...
+		slices.Contains(splitServices(services), ServiceCache) ||
+		slices.Contains(splitServices(services), ServiceDataBroker)
 }
 
-// IsRegistry checks if this node should run the registry service
-func IsRegistry(s string) bool {
-	return IsDataBroker(s)
+// IsAll checks to see if we should be running all services
+func IsAll(services string) bool {
+	return slices.Contains(splitServices(services), ServiceAll)
```
gaurdro: Is it worth considering the case that someone has listed every service individually instead of just 'all' here?

wasaga (approved):

PR #5659, opened 2025-06-18
[WIP] ssh: add config options, add stub authorize grpc handlers

kralicky:
## Checklist

- [ ] updated unit tests
- [ ] ready for review


config/envoyconfig/listeners_ssh.go
```diff
@@ -0,0 +1,198 @@
...
+		Timeout: durationpb.New(0),
+		TargetSpecifier: &envoy_config_core_v3.GrpcService_EnvoyGrpc_{
+			EnvoyGrpc: &envoy_config_core_v3.GrpcService_EnvoyGrpc{
+				ClusterName: "pomerium-authorize",
+			},
+		},
+	}
+
+	var hostKeyDataSources []*envoy_config_core_v3.DataSource
+	if cfg.Options.SSHHostKeyFiles != nil {
```
kenjenkins, line 44: I think it would make sense to be consistent with how we handle TLS certificates, and use filemgr [BytesDataSource](https://github.com/pomerium/pomerium/blob/6a65c52a6c46b165845e9445c6e18405aa009abc/config/envoyconfig/filemgr/filemgr.go#L38) for the ssh keys too.

I think that would also allow us to avoid any errors from OpenSSH around file permissions issues, if Pomerium writes the temporary files itself, rather than referencing user-provided files directly?

config/envoyconfig/listeners_ssh.go
```diff
@@ -0,0 +1,198 @@
...
+						},
+					},
+				},
+			},
+		},
+	}, nil
+}
+
+func shouldStartSSHListener(options *config.Options) bool {
+	return config.IsAuthorize(options.Services)
```
kenjenkins, line 197: Should this be config.IsProxy()?

config/policy.go
```diff
@@ -930,6 +930,11 @@ func (p *Policy) IsUDPUpstream() bool {
 	return len(p.To) > 0 && p.To[0].URL.Scheme == "udp"
 }
 
+// IsSSH returns true if the route is for SSH.
+func (p *Policy) IsSSH() bool {
+	return len(p.From) > 0 && strings.HasPrefix(p.From, "ssh://")
```
kenjenkins, line 935: Is the `len(p.From) > 0` expression redundant here?

config/envoyconfig/listeners_ssh.go
kralicky: Yes

config/policy.go
kralicky: Yep :+1:

config/envoyconfig/listeners_ssh.go
kralicky: Not sure - IMO the strict openssh file permission checking is a feature not a bug. Referencing user-provided files is how I personally would want to configure this.

kenjenkins: What happens in the case of a file permission error from OpenSSH? Does this crash Envoy?

kralicky: It is validated on the pomerium side [here](https://github.com/pomerium/pomerium/blob/1badb8bdf72d64b3d2387de7c4b64eaaae96d1a5/config/options.go#L784-L791) but if it somehow still gets applied to envoy, attempting to load such keys is a fatal error, yes.

kenjenkins: OK, as long as we validate on the Pomerium side that's probably good enough for now.

PR #5665, opened 2025-06-23
ssh: implement authorization policy evaluation

kenjenkins:
## Summary

Implement the pkg/ssh.AuthInterface. Add logic for converting from the ssh stream state to an evaluator request, and for interpreting the results of policy evaluation. Refactor some of the existing authorize logic to make it easier to reuse.

## Related issues

https://linear.app/pomerium/issue/ENG-2491/enable-ssh-configuration-in-core

## Checklist

- [x] updated unit tests
- [ ] ready for review


pkg/ssh/auth.go
```diff
@@ -0,0 +1,331 @@
...
+		return KeyboardInteractiveAuthMethodResponse{}, errPublicKeyAllowNil
+	}
+
+	// Initiate the IdP login flow.
+	sessionRecordVersion, err := a.handleLogin(
+		ctx, info.Hostname, info.PublicKeyFingerprintSha256, querier)
+	if err != nil {
+		return KeyboardInteractiveAuthMethodResponse{}, err
+	}
+	info.SessionRecordVersionHint = sessionRecordVersion
```
kralicky: This field name might be confusing once we implement session recording.

pkg/ssh/auth.go
```diff
@@ -0,0 +1,331 @@
...
+	// Otherwise evaluate access policy now.
+	if err := a.EvaluateDelayed(ctx, info); err != nil {
+		// Denied.
+		return KeyboardInteractiveAuthMethodResponse{}, nil
+	}
+	// Allowed.
+	return KeyboardInteractiveAuthMethodResponse{
+		Allow: &extensions_ssh.KeyboardInteractiveAllowResponse{},
+	}, nil
+
```
kralicky: nit: extra blank line

pkg/ssh/auth.go
```diff
@@ -0,0 +1,331 @@
...
+		return 0, err
+	}
+
+	resp, err := authenticator.DeviceAuth(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	// Prompt the user to sign in.
+	go func() { // XXX: remove 'go' after rebase
```
kralicky: ^

pkg/ssh/auth.go
```diff
@@ -0,0 +1,331 @@
...
+	if info.PublicKeyAllow.Value == nil {
+		// Sanity check: this method is only valid if we already accepted a public key.
+		return KeyboardInteractiveAuthMethodResponse{}, errPublicKeyAllowNil
+	}
+
+	// Initiate the IdP login flow.
+	sessionRecordVersion, err := a.handleLogin(
+		ctx, info.Hostname, info.PublicKeyFingerprintSha256, querier)
+	if err != nil {
+		return KeyboardInteractiveAuthMethodResponse{}, err
```
kralicky, line 183: Errors returned from this method should ideally be status.Error errors, so maybe this should be `return KeyboardInteractiveAuthMethodResponse{}, status.Error(codes.Internal, err.Error())`

pkg/ssh/manager.go
```diff
@@ -30,17 +34,23 @@ func (sm *StreamManager) LookupStream(streamID uint64) *StreamHandler {
...
 	defer sm.mu.Unlock()
 	streamID := downstream.StreamId
 	writeC := make(chan *extensions_ssh.ServerMessage, 32)
 	sh := &StreamHandler{
-		auth:       sm.auth,
-		config:     cfg,
-		downstream: downstream,
-		readC:      make(chan *extensions_ssh.ClientMessage, 32),
-		writeC:     writeC,
+		auth:          NewAuth(evaluator, client, currentConfig, tracerProvider),
```
kralicky: We should move the AuthInterface argument out of NewStreamManager and into (*StreamManager).NewStreamHandler instead, so that we don't need to pass around the arguments that NewAuth needs

pkg/ssh/auth.go
kenjenkins: Reworked how we invalidate session records from the global authorize cache, so this field is no longer needed.

pkg/ssh/auth.go
kenjenkins: OK, refactored both HandleXXX() methods to return status.Error errors. I realized that maybe some error details should not be user-facing, so I added a log entry with the underlying error, and sanitized the message to be just "internal error".

pkg/ssh/auth.go
kenjenkins: Removed.

pkg/ssh/auth.go
kenjenkins: Done.

pkg/ssh/auth.go
```diff
@@ -0,0 +1,341 @@
...
+
+func (a *Auth) saveSession(
+	ctx context.Context,
+	id string,
+	claims *identity.SessionClaims,
+	token *oauth2.Token,
+) (uint64, error) {
+	now := time.Now()
+	nowpb := timestamppb.New(now)
+	sessionLifetime := a.currentConfig.Load().Options.CookieExpire
```
kralicky, line 288: I think these timestamps should be computed earlier for the public key auth response, then the session IssuedAt/ExpiresAt should use those values instead.

pkg/ssh/manager.go
kenjenkins: Reworked the initialization logic, please take a look and let me know what you think.

pkg/ssh/auth.go
kenjenkins: I think it makes sense for the session timestamps to be tied to the IdP login. I would think the session is not "issued" until a successful login, so IssuedAt should match the creation of the session record as near as possible.

kralicky: That makes sense, the expiration time is probably more important, they should match exactly

PR #5670, opened 2025-06-25
ssh: stream management api

kralicky:
## Summary

This implements the StreamManagement API defined at 
https://github.com/pomerium/envoy-custom/blob/main/api/extensions/filters/network/ssh/ssh.proto#L46-L60. Policy evaluation and authorization logic is stubbed out here, and implemented in https://github.com/pomerium/pomerium/pull/5665.

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/pkg/ssh/channel.go
+++ b/pkg/ssh/channel.go
@@ -120,7 +120,7 @@ func (ch *ChannelHandler) Run(ctx context.Context) error {
 				})
 				ch.cancel(status.Errorf(codes.Canceled, "channel closed"))
 			case ChannelEOFMsg:
-				log.Ctx(ctx).Debug().Msg("received channel EOF")
+				log.Ctx(ctx).Debug().Msg("ssh: received channel EOF")
 			default:
 				panic(fmt.Sprintf("bug: unhandled message type: %T", msg))
 			}
@@ -202,7 +202,7 @@ func (ch *ChannelHandler) handleChannelRequestMsg(ctx context.Context, msg Chann
 			ch.initiateChannelClose(err)
 		}()
 	case "env":
-		log.Ctx(ctx).Warn().Msg("env channel requests are not implemented yet")
+		log.Ctx(ctx).Warn().Msg("ssh: env channel requests are not implemented yet")
 	case "pty-req":
 		if ch.cli != nil || ch.ptyInfo != nil {
 			return status.Errorf(codes.FailedPrecondition, "unexpected channel request: %s", msg.Request)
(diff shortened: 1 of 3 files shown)


pkg/ssh/channel.go
```diff
@@ -0,0 +1,253 @@
...
+		PeersID: ch.ctrl.DownstreamChannelID(),
+	})
+}
+
+func (ch *ChannelHandler) initiateChannelClose() {
+	ch.sendChannelCloseMsgOnce.Do(func() {
+		ch.flushStdout()
+		ch.sendChannelCloseMsg()
+		// the client needs to respond to our close request
+		time.AfterFunc(1*time.Second, func() {
```
kenjenkins: How did you choose 1s as a timeout for this?

pkg/ssh/channel_impl.go
```diff
@@ -0,0 +1,198 @@
...
+		case MsgChannelWindowAdjust, MsgChannelRequest, MsgChannelSuccess, MsgChannelFailure, MsgChannelEOF, MsgChannelClose:
+			// these messages don't consume window space
+		default:
+			// NB: It is not possible for localWindow to be < msgLen, since the window
+			// size is 64x the maximum packet size, and we have already checked the
+			// packet size above. The window adjust message is sent when the window
+			// size is at half of its max value.
+			ci.localWindow -= msgLen
+			if ci.localWindow < ChannelWindowSize/2 {
+				log.Ctx(ci.stream.Context()).Debug().Msg("flow control: increasing local window size")
```
kenjenkins: Will these log messages have "ssh" somewhere in them? Let's add it if not.

pkg/ssh/cli.go
```diff
@@ -0,0 +1,244 @@
...
+) *CLI {
+	cmd := &cobra.Command{
+		Use: "pomerium",
+		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+			_, cmdIsInteractive := cmd.Annotations["interactive"]
+			switch {
+			case (ptyInfo == nil) && cmdIsInteractive:
+				return fmt.Errorf("\x1b[31m'%s' is an interactive command and requires a TTY (try passing '-t' to ssh)\x1b[0m", cmd.Use)
+			case (ptyInfo != nil) && !cmdIsInteractive:
+				return fmt.Errorf("\x1b[31m'%s' is not an interactive command (try passing '-T' to ssh, or removing '-t')\x1b[0m\r", cmd.Use)
```
kenjenkins, line 40: Does this need to be an error? I would have thought that with a TTY you could run either interactive or non-interactive commands.

pkg/ssh/channel.go
kralicky: It's arbitrary

pkg/ssh/cli.go
kralicky: You can, but if the command prints anything to stdout it will appear all messed up since it will interpret the output as raw pty data

kenjenkins (approved):

PR #5687, opened 2025-07-02
ssh: continuous authorization

kenjenkins:
## Summary

Re-evaluate ssh authorization decision on a fixed interval. If access is no longer allowed, log a new `authorize check` message and disconnect.

## Related issues

https://linear.app/pomerium/issue/ENG-2044/support-continuous-verification-with-ssh-proxying

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/pkg/ssh/manager.go
+++ b/pkg/ssh/manager.go
@@ -1,6 +1,7 @@
 package ssh
 
 import (
+	"context"
 	"sync"
 
 	extensions_ssh "github.com/pomerium/envoy-custom/api/extensions/filters/network/ssh"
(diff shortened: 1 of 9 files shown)


pkg/ssh/manager.go
```diff
@@ -39,10 +57,11 @@ func (sm *StreamManager) NewStreamHandler(
 	writeC := make(chan *extensions_ssh.ServerMessage, 32)
 	sh := &StreamHandler{
 		auth:       auth,
-		config:     cfg,
+		config:     sm.cfg,
 		downstream: downstream,
 		readC:      make(chan *extensions_ssh.ClientMessage, 32),
 		writeC:     writeC,
+		reauthC:    make(chan struct{}),
```
kralicky, line 65: This channel needs a buffer size of 1, otherwise reauth will be skipped if the stream is in the middle of handling a message

pkg/ssh/manager.go
```diff
@@ -1,48 +1,68 @@
...
+	sm.cfg = cfg
+	snapshot := make([]*StreamHandler, 0, len(sm.activeStreams))
+	for _, s := range sm.activeStreams {
+		snapshot = append(snapshot, s)
+	}
+	sm.mu.Unlock()
+
+	select {
+	case sm.reauth <- snapshot:
+	default:
```
kralicky, line 38: I think the snapshot should be computed after reading from sm.reauth - consider the following scenario:
1. N streams are active, config changes. sm.reauth contains a snapshot of N streams
2. the snapshot containing N streams is read from sm.reauth, and kicks off a re-auth for those streams in sequence.
3. During the re-auth sequence:
	a. The config is changed again, writing a snapshot of the same N streams to sm.reauth.
  b. A new stream connects and is authorized
  c. The config is changed once more, in such a way that the new stream from (b) would no longer be authorized
4. The first re-auth completes. sm.reauth is read again right away, and contains the snapshot of N streams added during (3a). The new stream from (3b) is not disconnected, as it should be. If the list of streams were computed now instead, the snapshot would contain N+1 streams as expected and the new stream would be disconnected.

kenjenkins: OK, moved the snapshot inside reauthLoop(). Please take another look.

kralicky: LGTM

kralicky (approved):

PR #5688, opened 2025-07-02
ssh: rework cached record invalidation

kenjenkins:
## Summary

Add an additional method to the `ssh.Evaluator` interface for invalidating cached databroker records. Invalidating the global cache is not sufficient, because there may be sync queriers as well.

Make sure to invalidate the User record (in addition to the Session record) during the login flow.

## Checklist

- [ ] updated unit tests
- [ ] ready for review


authorize/ssh_grpc.go
```diff
@@ -129,3 +131,8 @@ func (a *Authorize) EvaluateSSH(ctx context.Context, req *ssh.Request) (*evaluat
 
 	return res, nil
 }
+
+func (a *Authorize) InvalidateCacheForRecords(records ...*databroker.Record) {
+	ctx := a.withQuerierForCheckRequest(context.Background())
```
kralicky: Should this context be passed by the caller? Looks like it would be ignored by the sync querier but not sure what happens if sync queriers are disabled.

kenjenkins: It looked like this context was not actually used anywhere, except to store and retrieve the querier. We can pass it from the caller if you prefer.

kralicky (approved):

authorize/ssh_grpc.go
kralicky: Can't hurt I guess

Issue #5673, opened 2025-06-26
Bug: authorize_log_fields not always respected in authz logs for denied requests


nickytonline:
## What happened?

Fields listed in `authorize_log_fields` (specifically mcp-method, mcp-tool, and mcp-tool-parameters) are missing in denied authorization log entries, even though they're present for successful ones.

## What did you expect to happen?

Expected all fields listed in authorize_log_fields to be present in every authorize log entry, regardless of whether the request was allowed or denied.

## How'd it happen?

- Defined custom authorize_log_fields including MCP-relevant fields.
- Ran Pomerium built from latest main.
- Observed that denied requests were missing expected fields in the logs.

## What's your environment like?

Pomerium version (main branch as of 2025-06-25)

Server Operating System/Architecture/Cloud: Docker on DigitalOcean Ubuntu 22.04 x86_64

## What's your config.yaml?

```yaml
root@pomerium-et-all:~/pomdemo# cat config.yaml
databroker_storage_type: postgres
databroker_storage_connection_string: postgres://postgres:postgres@postgres:5432/pomerium?sslmode=disable

authenticate_service_url: ${AUTHENTICATE_SERVICE_URL}
idp_provider: ${IDP_PROVIDER}
idp_provider_url: ${IDP_PROVIDER_URL}
idp_client_id: ${IDP_CLIENT_ID}
idp_client_secret: ${IDP_CLIENT_SECRET}

cookie_domain: demo.maisonlab.dev

address: :443

runtime_flags:
  mcp: true

autocert: true

# OTEL config
# https://docs.pomerium.com/docs/reference/tracing#config-file
# running in enterprise console? go to settings/tracing, set type to otlp and
# otel_exporter_otlp_endpoint: http://jaeger:4317
otel_exporter_otlp_traces_endpoint: http://jaeger:4317
otel_exporter_otlp_protocol: grpc
otel_traces_exporter: otlp

# for MCP observability
authorize_log_fields:
  - request-id
  - user
  - email
  - mcp-method
  - mcp-tool
  - mcp-tool-parameters
  - host
routes:
  - from: https://dev.demo.maisonlab.dev
    to: http://mcp-demo-app:5173
    name: App Demo
    timeout: 60s
    mcp:
      client: {}
    policy:
      allow:
        or:
          - domain:
              is: pomerium.com
          - domain:
              is: nickyt.co

  - from: https://app.demo.maisonlab.dev
[...]

PR #5693, opened 2025-07-04
ppl: rename mcp tool criterion reasons

wasaga:
## Summary

Change MCP PPL reasons to be more human-readable when used in the denial matches. 

## Related issues

See https://github.com/pomerium/pomerium/issues/5673#issuecomment-3036879247

## Checklist

- [x] updated unit tests
- [x] ready for review


calebdoxsey (approved):
Should be ok I think. We have special handling for `device-unauthenticated`, `user-unauthenticated`, `client-certificate-required`, and `invalid-client-certificate`, but otherwise I think the reasons are just informative.

PR #5694, opened 2025-07-04
mcp: respond with jsonrpc error when MCP request is denied

wasaga:
## Summary

Individual MCP method calls may be denied (i.e. via `mcp_tool` criterion) and Pomerium has to respond with MCP protocol error, which is JSON-RPC error message, rather then with HTTP level error which seems to break some MCP clients. 

## Related issues

Fix https://linear.app/pomerium/issue/ENG-2521/pomerium-does-not-return-an-mcp-error-when-a-tool-call-is-unauthorized

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/authorize/check_response_test.go
+++ b/authorize/check_response_test.go
@@ -168,9 +168,9 @@ func TestAuthorize_handleResult(t *testing.T) {
 				},
 			},
 			&evaluator.Result{
-				Allow: evaluator.NewRuleResult(false, criteria.ReasonMCPToolUnauthorized),
+				Allow: evaluator.NewRuleResult(false, criteria.ReasonMCPToolNoMatch),
 			},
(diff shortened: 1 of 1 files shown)


internal/jsonrpc/messages.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (id *ID) UnmarshalJSON(data []byte) error {
+	if data == nil || string(data) == "null" {
+		id.value = nil
+		return nil
+	}
+
+	var v any
+	if err := json.Unmarshal(data, &v); err != nil {
```
calebdoxsey: Do JSON RPC clients ever send large numbers as IDs? Sometimes using the standard float64 JSON decoder can end up truncating large integers... would it make sense to use https://pkg.go.dev/encoding/json#Decoder.UseNumber, and maybe instead of `any` use a `RawJSON`?

PR #5696, opened 2025-07-05
authorize: add request body logging

wasaga:
## Summary

Adds an option to log request body for protocols that perform request inspection, such as MCP. 

## Related issues

Fix https://linear.app/pomerium/issue/ENG-2544/authorize-request-body-logging

## Checklist

- [ ] updated unit tests
- [ ] ready for review


calebdoxsey (approved):
Is the body always JSON? Would it be nice to use `RawJSON` so that it doesn't get double-encoded? Or is that more trouble than its worth?

There is https://pkg.go.dev/encoding/json#Valid, but I think newlines in the JSON would mess things up so it's probably not so straightforward.

wasaga:
yes all those mcp clients do tend to send json with indentation that might mess things up so lets keep it to plaintext for now

PR #5736, opened 2025-07-21
Fix typo, add more test cases for atomicutil, and mention about atomic.Pointer migration

rytswd:
## Summary

I found the `atomicutil` package used throughout the code base, but it seems to predate the addition of `atomic.Pointer[T]`. Potentially this should allow us to remove the package altogether and use stdlib directly instead. But I went ahead with starting with adding simple test cases first, so that we have better test coverage across the code base.

Also, I added an example of how `atomicutil` usage may be problematic in some specific scenario compared to how it's implemented with `atomic.Pointer[T]`. Not to mention, there is some performance gain of not using the direct casting with its generic support. (Though this is probably a minor one and the code clarity is a benefit we may want to keep rather than using other atomic values such as `atomic.Bool`.)

## Related issues

N/A

## User Explanation

This only touches the function documentation and test cases, and should not have any impact for users.

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/internal/atomicutil/value_test.go
+++ b/internal/atomicutil/value_test.go
@@ -1,6 +1,7 @@
 package atomicutil
 
 import (
+	"sync/atomic"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


internal/atomicutil/value_test.go
```diff
@@ -19,3 +20,39 @@ func TestValue(t *testing.T) {
...
+	v := NewValue(42)
+	v.Swap(33)
+	assert.Equal(t, 33, v.Load())
+}
+
+func TestCompareAndSwap(t *testing.T) {
+	v := NewValue(42)
+
+	swapped := v.CompareAndSwap(42, 33)
+	assert.Equal(t, true, swapped)
```
kenjenkins: You could also consider using the special-case [assert.True](https://pkg.go.dev/github.com/stretchr/testify/assert#True) here (and [assert.False](https://pkg.go.dev/github.com/stretchr/testify/assert#Assertions.False) on line 44).

internal/atomicutil/value_test.go
```diff
@@ -19,3 +20,39 @@ func TestValue(t *testing.T) {
...
+	assert.Equal(t, 33, v.Load())
+
+	swapped = v.CompareAndSwap(42, 33)
+	assert.Equal(t, false, swapped)
+	assert.Equal(t, 33, v.Load())
+}
+
+// This is just to illustrate how we should use atomic.Pointer[T] instead.
```
kenjenkins: I agree it would be good to standardize on atomic.Pointer wherever it makes sense, but I think it might be more helpful to explain the difference(s) between them rather than just saying "use atomic.Pointer[T] instead."

Would this be an accurate summary of this test case?
```suggestion
// Unlike atomic.Pointer[T], calling Load() on an uninitialized Value[T] will 
// return the zero value for type T.
```

rytswd: Thanks, that looks more accurate and helpful indeed!
I was under impression that, depending on how we want to handle extra utility package like this one going forward, we could deprecate the atomicutil functions and start migrating to atomic.Pointer[T], rather than adding these notes for future reference. As the number of atomicutli references isn't too large (around dozens, not hundreds or more), I think migration can happen relatively quickly. However, it would still be quite an undertaking ensuring all the edge cases are handled, and thus having clearer documentation may be the best approach for now 👍

internal/atomicutil/value_test.go
rytswd: Ah absolutely, I wasn't too familiar with testify so went with copying other pieces using assert package. This is fixed with 5a24d8c69

internal/atomicutil/value_test.go
kenjenkins: OK yes, sounds good. I've filed an issue internally to discuss migrating to atomic.Pointer throughout.

kenjenkins (approved):

PR #5737, opened 2025-07-21
[WIP] Migrate validation logic using cryptutil from databroker to configs

rytswd:
## Summary

Data Broker had a private validation function used within its initialisation, but it was only used in one place. There was a separate validation function already defined within configs module, which didn't have the cryptutil usage. The current module import ballooning in each module is probably not ideal, especially considering there are various wrapper modules already in place (configs, cryptutil, etc.).

This may have some unintended consequences if the validation is more aggressive with the config module. In such case, we should split the validation logic into multiple functions, so that at least the consuming code (in this case databroker) should not need to know too much about the implementation logic, and only rely on the specific validation it needs.

## Related issues

N/A

## User Explanation

The config validation will be more strict and correct throughout the code base, not just data broker use cases.

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -613,9 +613,12 @@ func (o *Options) Validate() error {
 		return errors.New("config: unknown databroker storage backend type")
 	}
 
-	_, err := o.GetSharedKey()
+	sharedKey, err := o.GetSharedKey()
 	if err != nil {
-		return fmt.Errorf("config: invalid shared secret: %w", err)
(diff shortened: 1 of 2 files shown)


calebdoxsey:
I think this makes sense. We also have this check in `authenticate`, `authorize` and `proxy`. I believe `Validate` on `Options` is always called (either via viper on load, or from the databroker), so we likely don't need to call it again.

PR #5739, opened 2025-07-22
contexutil: merge improvements

calebdoxsey:
## Summary
Improvements to the `contextutil.Merge` function:

- it can now take an arbitrary number of contexts rather than just 2
- it returns a `CancelCauseFunc` instead of a `CancelFunc` which should be more flexible
- `Deadline` now returns the soonest time instead of the first one defined
- use `synctest` for the tests, add a few more tests

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/contextutil/contextutil.go
+++ b/pkg/contextutil/contextutil.go
@@ -59,7 +59,7 @@ func (mc *mergedCtx) Err() error {
 }
 
 func (mc *mergedCtx) Value(key any) any {
-	// cancel cause is propogated as a value,
+	// cancel cause is propagated as a value,
 	// so we need to check the done context as well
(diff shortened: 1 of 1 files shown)


pkg/contextutil/contextutil.go
```diff
@@ -3,64 +3,74 @@ package contextutil
...
-	}
-	if err := mc.ctx2.Err(); err != nil {
-		return mc.ctx2.Err()
-	}
 	return mc.doneCtx.Err()
 }
 
 func (mc *mergedCtx) Value(key any) any {
-	if value := mc.ctx1.Value(key); value != nil {
+	// cancel cause is propogated as a value,
```
kenjenkins: spelling nit: "propogated" -> "propagated"

PR #5742, opened 2025-07-23
core: use iterators for databroker sync and sync latest

calebdoxsey:
## Summary
Refactor the databroker storage interface to use an `iter.Seq2[*databroker.Record,error]` instead of a `RecordStream`. The original implementation for the storage backend was written before iterators existed in Go and we can use the standard pattern now.

I also updated the postgres methods to return lists of records consistently, and used a much smaller batch size. (Previously only a single changed record was queried at a time) The latest query now uses `type` and `id` instead of `OFFSET` which should improve performance when there are a large number of rows.

## Related issues
- [ENG-2417](https://linear.app/pomerium/issue/ENG-2417/core-use-iterators-for-storage-backends)
- [ENG-2494](https://linear.app/pomerium/issue/ENG-2494/core-postgres-performance-improvements)

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -45,7 +45,9 @@ type Backend interface {
 	Patch(ctx context.Context, records []*databroker.Record, fields *fieldmaskpb.FieldMask) (serverVersion uint64, patchedRecords []*databroker.Record, err error)
 	// SetOptions sets the options for a type.
 	SetOptions(ctx context.Context, recordType string, options *databroker.Options) error
-	// Sync syncs record changes after the specified version.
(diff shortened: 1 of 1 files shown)


pkg/storage/storage.go
```diff
@@ -43,9 +43,9 @@ type Backend interface {
 	// SetOptions sets the options for a type.
 	SetOptions(ctx context.Context, recordType string, options *databroker.Options) error
 	// Sync syncs record changes after the specified version.
```
kenjenkins: (optional) Would it make sense to update this comment to describe how the `wait` parameter behaves?

PR #5750, opened 2025-07-24
add runtime flag for session refresh change

kenjenkins:
## Summary

Add a new runtime flag `refresh_session_at_id_token_expiration` to guard the two changes introduced by https://github.com/pomerium/pomerium/pull/5727, which are:
1. schedule session refresh based on ID token expiration (in addition to access token expiration and overall session lifetime)
2. always overwrite an existing ID token upon successful session refresh

This requires a bit of refactoring to make the runtime flag setting available in the identity manager and OIDC code:
- add a new `WithRefreshSessionAtIDTokenExpiration(bool)` option to the identity manager config
- add a new `OverwriteIDTokenOnRefresh bool` field to the oauth.Options struct

## Related issues

https://linear.app/pomerium/issue/ENG-2588/support-configuring-earlier-token-refresh-and-id-token-expiration

## Checklist

- [x] updated unit tests
- [ ] ready for review

### DIFF
--- a/pkg/identity/manager/data.go
+++ b/pkg/identity/manager/data.go
@@ -10,11 +10,16 @@ import (
 	"github.com/pomerium/pomerium/pkg/identity"
 )
 
+// RefreshSessionAtIDTokenExpiration indicates whether a session refresh should be
+// scheduled at the expiration time of the ID token.
+type RefreshSessionAtIDTokenExpiration bool
+
 func nextSessionRefresh(
 	s *session.Session,
 	lastRefresh time.Time,
 	gracePeriod time.Duration,
 	coolOffDuration time.Duration,
+	refreshAtIDTokenExpiration RefreshSessionAtIDTokenExpiration,
 ) time.Time {
 	var tm time.Time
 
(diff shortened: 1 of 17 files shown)


pkg/identity/manager/data.go
```diff
@@ -10,11 +10,16 @@ import (
 	"github.com/pomerium/pomerium/pkg/identity"
 )
 
+// RefreshSessionAtIDTokenExpiration indicates whether a session refresh should be
+// scheduled at the expiration time of the ID token.
+type RefreshSessionAtIDTokenExpiration bool
```
wasaga, line 15: unsure I understand why a dedicated type for that is necessary

wasaga (approved):

pkg/identity/manager/data.go
kenjenkins: It's certainly not necessary, but I thought it might reduce the chances that it gets mixed up with any other `bool` parameters in function arguments.

It looks like we do a similar thing in a few other places:
- https://github.com/pomerium/pomerium/blob/a3b3c7bf0e4ef9857f9842737e6b6506dca678db/config/custom.go#L307
- https://github.com/pomerium/pomerium/blob/a3b3c7bf0e4ef9857f9842737e6b6506dca678db/internal/databroker/config_source.go#L52
- https://github.com/pomerium/pomerium/blob/a3b3c7bf0e4ef9857f9842737e6b6506dca678db/config/envoyconfig/protocols.go#L49

PR #5774, opened 2025-08-07
core: on-disk databroker storage

calebdoxsey:
## Summary
This adds a new `file` databroker storage provider that stores data on-disk. The DSN format is: `file:///path/to/file`. 

This storage provider is implemented with a pebble key-value store. Pebble implements a global, sorted collection of key-value pairs. We break up the databroker storage into various "key spaces". Each key space is distinguished via a single-byte prefix. All the keys for that key space will have the same prefix.

- `Lease`: used to implement distributed locking: key is the lease name, value is the lease id and the expires at timestamp.
- `Metadata`: metadata about the store: server version and migration version.
  - The server version is randomly generated for an empty database. 
  - Migrations are used for future changes to the on-disk format.
- `Options`: options for a record type: key is the record type, value is the protobuf-encoded options.
  - Options are currently only used to set a capacity limit for a record type (for example for events).
- `RecordChange`: each record as they are updated: key is the version, value is the protobuf-encoded record.
  - Record changes are like a write-ahead log. They are used by clients via the `Sync` call to receive changes as they are made.
  - In addition to `Put`s, `Delete`s are also stored as record changes. (tombstones)
- `RecordChangeIndexByType`: an index of record type to version: key is the record type + the version, value is nil.
  - This index allows us to implement `Sync` with a record type filter efficiently.
- `Record`: the latest version of each record: key is the record type + the record id, value is the protobuf-encoded record.
- `RecordIndexByTypeVersion`: an index of record type and version to record id: key is the record type + the version, value is the record id.
  - This index is used to enforce capacity limits.
- `RegistryService`: registered services: key is kind + endpoint, value is the expires at timestamp.
  - We maintain an in-memory index by the expires at timestamp to delete services when they are expired.

We use standard protobuf wire-encoding for protobuf types. Integers are big-endian uint64s (to preserve order). Times are uint64 unix timestamps in microseconds. Keys are either fixed lengths based on the type (for example, if the first part of a key is an integer, we know its 8 bytes long), or separated with a 0x00 byte. This means strings cannot contain a null byte or the split logic will not work.

`SyncLatest` supports efficient filtering for `id` and `$index`. This is used both for typical calls to `SyncLatest`, but also for calls to `Query`. The most common type of query (used for databroker lookups when evaluating policies) looks like this:

```json
{ 
  "$or": [
    { "id": "VALUE" },
    { "$index": "VALUE" }
  ] 
}
```

This filter will be evaluated by doing a `get` on the `Record` key space, and an address lookup against an in-memory CIDR index (using the same `bart` library we use for the in-memory store). The results will then be merged together using a set-union. (since records are already ordered we can do this more efficiently than with a hash table) We may want to do further query optimization in the future, but this handles the most common queries.

## Related issues
- [ENG-2679](https://linear.app/pomerium/issue/ENG-2679/core-implement-on-disk-databroker-storage)

## Checklist

- [x] updated unit tests
- [x] ready for review


calebdoxsey:
This is mostly done. I need to clean up the code a bit and add more unit tests.

calebdoxsey:
Disabling the `RecordIndexByTypeVersion` index didn't seem to make much difference in performance. It leads to roughly 10% more disk usage, which seems marginal. I'll leave it for now.

pkg/storage/file/migrate.go
```diff
@@ -0,0 +1,41 @@
...
+		return fmt.Errorf("pebble: error getting migration version: %w", err)
+	}
+
+	for i := current; i < uint64(len(migrations)); i++ {
+		err = migrations[i]()
+		if err != nil {
+			return fmt.Errorf("pebble: error migrating to version %d: %w", i+1, err)
+		}
+
+		err = metadataKeySpace.setMigration(db, current+1)
```
wasaga: should it be `i+1` ?

pkg/storage/file/index.go
```diff
@@ -0,0 +1,163 @@
...
+	if !ok {
+		return nil
+	}
+
+	if recordType == "" {
+		return nodes
+	}
+
+	return slices.DeleteFunc(nodes, func(node recordCIDRNode) bool {
+		return node.recordType == recordType
```
wasaga: should it be `!=` ?

pkg/storage/file/pebble.go
```diff
@@ -0,0 +1,256 @@
...
+		switch u.Scheme {
+		case "memory":
+			backend.db = pebbleutil.MustOpenMemory(nil)
+		case "file":
+			backend.db, err = pebbleutil.Open(u.Path, nil)
+			if err != nil {
+				backend.initErr = fmt.Errorf("pebble: error opening database at %s: %w", u.Path, err)
+				return
+			}
+		}
```
wasaga, line 54: add `default:` for unknown schema and set `initErr` ?

pkg/storage/file/migrate.go
calebdoxsey: yes nice catch

pkg/storage/file/index.go
calebdoxsey: Yes it was wrong. Using `DeleteFunc` was also modifying the nodes stored in the index so I needed to clone the slice first. I add some unit tests for this.

pkg/iterutil/errors.go
```diff
@@ -0,0 +1,173 @@
+package iterutil
+
+import "iter"
+
+// An ErrorSeq is an iterator of values with errors.
+type ErrorSeq[E any] = iter.Seq2[E, error]
+
+// ApplyWithError creates a new ErrorSeq that applies a transformation
+// function to the simple sequence of elements.
```
kenjenkins: Would it be helpful to add a note here that the returned iterator yields the transformed elements until the first error value from the original sequence, at which point it yields the error value and stops iterating?

pkg/iterutil/iterutil_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	"github.com/pomerium/pomerium/pkg/iterutil"
+)
+
+func TestChunk(t *testing.T) {
+	t.Parallel()
+
+	assert.Equal(t,
+		[][]int{{1, 2}, {3, 4}, {5, 6}},
+		slices.Collect(iterutil.Chunk(slices.Values([]int{1, 2, 3, 4, 5, 6}), 2)))
```
kenjenkins: Would it be good to also include a test case where the last chunk is smaller than the chunk size?

e.g.
```
	assert.Equal(t,
		[][]int{{1, 2}, {3, 4}, {5}},
		slices.Collect(iterutil.Chunk(slices.Values([]int{1, 2, 3, 4, 5}), 2)))
```

kenjenkins (approved):

PR #5779, opened 2025-08-12
ssh: better handling of known unsupported channel requests

kralicky:
Add handlers for known channel requests that are unsupported in internal channels.

### DIFF
--- a/pkg/ssh/channel.go
+++ b/pkg/ssh/channel.go
@@ -188,9 +188,7 @@ func (ch *ChannelHandler) handleChannelRequestMsg(ctx context.Context, msg Chann
 			ch.cli.SetArgs(strings.Fields(execReq.Command))
 		}
 		if msg.WantReply {
-			if err := ch.ctrl.SendMessage(ChannelRequestSuccessMsg{
-				PeersID: ch.ctrl.DownstreamChannelID(),
-			}); err != nil {
+			if err := ch.sendChannelRequestSuccess(); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/ssh/channel.go
```diff
@@ -240,12 +234,37 @@ func (ch *ChannelHandler) handleChannelRequestMsg(ctx context.Context, msg Chann
...
+		action := "ignoring"
+		if msg.WantReply {
+			action = "rejecting"
+		}
+		log.Ctx(ctx).Debug().Str("request", msg.Request).Msgf("ssh: %s unsupported channel request", action)
+		if msg.WantReply {
+			if err := ch.sendChannelRequestFailure(); err != nil {
+				return err
+			}
+		}
```
kenjenkins: nit: I think it might be better to split this into two log calls, to make it easier to find this code if you search for "ssh: ignoring unsupported channel request", something like this?

```suggestion
		if msg.WantReply {
			log.Ctx(ctx).Debug().Str("request", msg.Request).Msg("ssh: rejecting unsupported channel request")
			if err := ch.sendChannelRequestFailure(); err != nil {
				return err
			}
		} else {
			log.Ctx(ctx).Debug().Str("request", msg.Request).Msg("ssh: ignoring unsupported channel request")
		}
```

kralicky: Yeah that's a good idea, agreed :+1:

PR #5780, opened 2025-08-12
authenticate: cache authenticators

calebdoxsey:
## Summary
Currently we create a new authenticator on every request. When creating sessions from incoming IdP tokens this results in HTTP requests to the OIDC discovery document and JWKS keys every time. These are unnecessary HTTP requests and should be cached.

Update the `defaultGetIdentityProvider` so that it will cache authenticators and re-use them. We actually have very similar functionality in the identity manager, but I wanted to minimize the number of changes since this will need to be backported. Future improvements would be to unify this into a single authenticator cache and also handle frequent changes to IdP settings. (as of now they will just stay cached until restart)

A separate access token validator is now being used. This lets us properly trace the requests and adds a default 10 second timeout when fetching keys. Because of the way the OIDC `RemoteKeySet` works, it doesn't use the passed-in context, but rather uses a background context, which will never be cancelled. As a result, if an HTTP request for keys were to get stuck, we'd never be able to resolve. With a timeout we should be able to recover. A future improvement would be to implement a new custom `KeySet` that uses the passed-in context rather than a background context for fetching keys. This would allow the spans to occur in the proper trace.

## Related issues
- [ENG-2701](https://linear.app/pomerium/issue/ENG-2701/authenticate-idp-access-token-verification-always-retrieves-the-openid)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/authenticate/identity.go
+++ b/authenticate/identity.go
@@ -22,7 +22,7 @@ type cachedAuthenticator struct {
 }
 
 var cachedAuthenticators = struct {
-	sync.Mutex
+	sync.RWMutex
 	*btree.BTreeG[cachedAuthenticator]
 }{
 	BTreeG: btree.NewG(2, func(a, b cachedAuthenticator) bool {
@@ -50,10 +50,17 @@ func defaultGetIdentityProvider(ctx context.Context, tracerProvider oteltrace.Tr
 
 	key.overwriteIDTokenOnRefresh = options.RuntimeFlags[config.RuntimeFlagRefreshSessionAtIDTokenExpiration]
 
+	cachedAuthenticators.RLock()
+	value, ok := cachedAuthenticators.Get(key)
+	cachedAuthenticators.RUnlock()
+	if ok {
+		return value.authenticator, nil
+	}
+
 	cachedAuthenticators.Lock()
 	defer cachedAuthenticators.Unlock()
 
-	value, ok := cachedAuthenticators.Get(key)
+	value, ok = cachedAuthenticators.Get(key)
 	if ok {
 		return value.authenticator, nil
 	}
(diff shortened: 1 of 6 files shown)


pkg/identity/oidc/azure/microsoft.go
```diff
@@ -71,6 +75,27 @@ func New(ctx context.Context, o *oauth.Options) (*Provider, error) {
 		p.AuthCodeOptions = o.AuthCodeOptions
 	}
 
+	// azure access tokens are JWTs signed with the same keys as identity tokens
+	p.getAccessTokenVerifier = sync.OnceValues(func() (*go_oidc.IDTokenVerifier, error) {
+		pp, err := p.GetProvider()
+		if err != nil {
+			return nil, err
+		}
```
wasaga: If the provider’s discovery document cannot be reached temporarily at the moment of initialization, the verifier function appears to return an error indefinitely.

authenticate/identity.go
```diff
@@ -1,25 +1,93 @@
...
 	if err != nil {
 		return nil, err
 	}
 
-	return identity.GetIdentityProvider(ctx, tracerProvider, idp, redirectURL,
-		options.RuntimeFlags[config.RuntimeFlagRefreshSessionAtIDTokenExpiration])
+	key.overwriteIDTokenOnRefresh = options.RuntimeFlags[config.RuntimeFlagRefreshSessionAtIDTokenExpiration]
+
+	cachedAuthenticators.Lock()
+	defer cachedAuthenticators.Unlock()
```
wasaga, line 61: Would using RLock() be more appropriate here, given that this code seems to run on every request?

pkg/identity/oidc/azure/microsoft.go
calebdoxsey: I changed it so this shouldn't happen anymore.

wasaga (approved):

PR #5781, opened 2025-08-12
ssh: terminate connection if corresponding session is revoked

kralicky:
If an SSH session record is deleted from the databroker, any active connections for that session will be closed.

### DIFF
--- a/pkg/ssh/manager.go
+++ b/pkg/ssh/manager.go
@@ -6,12 +6,13 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
 	extensions_ssh "github.com/pomerium/envoy-custom/api/extensions/filters/network/ssh"
 	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/internal/log"
 	"github.com/pomerium/pomerium/pkg/grpc/databroker"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 )
 
 type StreamManager struct {
@@ -58,7 +59,7 @@ func (sm *StreamManager) GetDataBrokerServiceClient() databroker.DataBrokerServi
 }
 
 // UpdateRecords implements databroker.SyncerHandler.
-func (sm *StreamManager) UpdateRecords(ctx context.Context, serverVersion uint64, records []*databroker.Record) {
+func (sm *StreamManager) UpdateRecords(ctx context.Context, _ uint64, records []*databroker.Record) {
 	sm.mu.Lock()
 	defer sm.mu.Unlock()
 	for _, record := range records {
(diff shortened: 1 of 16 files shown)


pkg/ssh/manager.go
```diff
@@ -15,6 +19,51 @@ type StreamManager struct {
 	mu            sync.Mutex
 	cfg           *config.Config
 	activeStreams map[uint64]*StreamHandler
+
+	listenersMu sync.Mutex
+	listeners   map[string][]func(*session.Session)
+}
+
+// ClearRecords implements databroker.SyncerHandler.
+func (sm *StreamManager) ClearRecords(ctx context.Context) {
```
kralicky, line 34: @calebdoxsey what is the correct way to implement this method?

calebdoxsey: The way this works is that `UpdateRecords` is called whenever records are updated, but `ClearRecords` is called if the entire sync stream has been reset and we need to start from scratch again. Probably you'd want to clear the `listeners` map.

pkg/ssh/manager.go
```diff
@@ -15,6 +19,47 @@ type StreamManager struct {
...
+			continue
+		}
+		var s session.Session
+		if err := record.GetData().UnmarshalTo(&s); err != nil {
+			panic(err)
+		}
+		for _, fn := range sm.listeners[record.Id] {
+			fn(&s)
+		}
+		sm.listeners[record.Id] = []func(*session.Session){}
```
calebdoxsey: This should probably be `delete(sm.listeners, record.Id)`? Otherwise its a memory leak?

pkg/ssh/manager.go
kralicky: Should we expect the same sessions to be retrieved again? In that case we might not want to close those connections preemptively right?

calebdoxsey: In general `ClearRecords` is only called on the initial sync, and then if the `server_version` is detected to have changed. If the `server_version` changes, that means that all of the records could've changed. In practice this happens if you restart the in-memory databroker, or connect to a new postgres database, so all the sessions would be removed entirely, though in principal it could just mean that we've fallen out of the sync window and need to do a full sync, in which case `ClearRecords` would be followed by `UpdateRecords` with much the same records we already have.

kralicky: > fallen out of the sync window

Is this an uncommon enough occurrence where it makes sense to just kick out all sessions, or should we add a grace period to see if an UpdateRecords comes through within a certain amount of time?

calebdoxsey: Its currently uncommon. With some of the databroker changes we're looking at it could become more common... though maybe we should redesign the way this syncer works if that happens. So I think just kicking out all the sessions makes sense.

calebdoxsey (approved):

PR #5784, opened 2025-08-14
databroker: add server info method

calebdoxsey:
## Summary
Add a new server info method that returns the server version, earliest record version and latest record version.

Also clean up some of the databroker code and allow gRPC reflection for the control plane.

## Related issues
- [ENG-2712](https://linear.app/pomerium/issue/ENG-2712/databroker-add-method-to-retrieve-server-info)

## Checklist

- [x] updated unit tests
- [x] ready for review


databroker/databroker.go
```diff
@@ -1,174 +1,199 @@
...
+	tracerProvider      oteltrace.TracerProvider
+	tracer              oteltrace.Tracer
 }
 
-// OnConfigChange updates the underlying databroker server whenever configuration is changed.
-func (srv *dataBrokerServer) OnConfigChange(ctx context.Context, cfg *config.Config) {
-	opts, err := srv.getOptions(cfg)
+// New creates a new databroker service.
+func New(ctx context.Context, cfg *config.Config, eventsMgr *events.Manager, options ...Option) (*DataBroker, error) {
+	localListener, err := net.Listen("tcp", "127.0.0.1:0")
```
kralicky, line 51: Is this new or refactored from somewhere else?

config/envoyconfig/listeners_grpc.go
```diff
@@ -64,6 +64,7 @@ func (b *Builder) buildGRPCHTTPConnectionManagerFilter() *envoy_config_listener_
 		"databroker.DataBrokerService",
 		"registry.Registry",
 		"grpc.health.v1.Health",
+		"grpc.reflection.v1alpha.ServerReflection",
```
kralicky, line 68: Is this used by a particular client? Only asking since there are two versions of this API and some clients can use both.

databroker/databroker.go
calebdoxsey: This should just be refactored. I renamed the variable `c` to `d`. The `databroker` used to be called `cache` and there were still some remnants of that in the code.

config/envoyconfig/listeners_grpc.go
calebdoxsey: Right that makes sense. These are basically the allowed gRPC services for the control plane, so I added `grpc.reflection.v1.ServerReflection` too.

PR #5805, opened 2025-08-29
databroker: add clear method

calebdoxsey:
## Summary
Add a new `Clear` method that erases all the records in the databroker.

## Related issues
- [ENG-2808](https://linear.app/pomerium/issue/ENG-2808/databroker-add-clear-method)

## Checklist

- [x] updated unit tests
- [x] ready for review


pkg/storage/postgres/backend.go
```diff
@@ -113,6 +116,26 @@ func (backend *Backend) Clean(ctx context.Context, options storage.CleanOptions)
...
+func (backend *Backend) Clear(ctx context.Context) error {
+	_, pool, err := backend.init(ctx)
+	if err != nil {
+		return err
+	}
+
+	backend.mu.Lock()
+	defer backend.mu.Unlock()
+
+	newServerVersion := uint64(cryptutil.NewRandomUInt32())
```
wasaga, line 129: is it intentionally calling a uint32 vs uint64 like elsewhere?

wasaga (approved):
added some comments in a ticket

pkg/storage/postgres/backend.go
calebdoxsey: Yeah, in the initial migration we set it to a uint32. I think it was written this way because `bigint` in postgres is a signed int64, so really only an int63, and the easiest way to deal with that was just to use a uint32 instead.

PR #5810, opened 2025-09-02
feat: health manager abstraction

alexandreLamarre:
## Summary

Internal abstraction for managing health checks and downstream consumers of said health checks.

## Related issues

https://linear.app/pomerium/project/healthcheck-with-graceful-teardown-e5428048d59e/overview

## Checklist

- [x] updated unit tests
- [x] ready for review


pkg/iterutil/iterutil.go
```diff
@@ -12,6 +12,24 @@ type (
 	Seq2[K, V any] = iter.Seq2[K, V]
 )
 
+func Collect[E any](seq iter.Seq[E]) []E {
```
calebdoxsey: I think this is available as `slices.Collect`?

pkg/health/channel_test.go
```diff
@@ -0,0 +1,78 @@
...
+			assert.Fail(t, "expected not to receive anything")
+		}
+	case <-time.After(timeout):
+		return
+	}
+}
+
+const dT = time.Millisecond * 5
+
+func TestChannel(t *testing.T) {
```
calebdoxsey: Possibly you could use the `synctest` package for this? There's an example in `contextutil`.

pkg/health/channel.go
```diff
@@ -0,0 +1,75 @@
...
+	tr Tracker,
+	opts ...CheckOption,
+) *ChannelProvider {
+	o := CheckOptions{}
+	o.Apply(opts...)
+
+	c := &ChannelProvider{
+		tr:           tr,
+		expected:     o.expected,
+		OnReady:      make(chan struct{}, 2),
```
calebdoxsey: could these just be closed instead of sending a value?

pkg/health/http.go
```diff
@@ -0,0 +1,202 @@
...
+	} else {
+		w.WriteHeader(http.StatusServiceUnavailable)
+	}
+	w.Header().Set("Content-Type", "application/json")
+	if r.Method == http.MethodGet {
+		_, _ = w.Write(payload)
+	}
+}
+
+func (h *HTTPProvider) LivelinessProbe(w http.ResponseWriter, r *http.Request) {
```
calebdoxsey, line 156: these all seem very similar... could they call a common function that takes the `cmpReadyHealth`, `cmpStartupHealth`, or `cmpLivelinessHealth` function?

pkg/health/http.go
```diff
@@ -0,0 +1,202 @@
...
+	Err        string `json:"error,omitempty"`
+	Attributes []Attr `json:"attributes,omitempty"`
+}
+
+type healthCmp func(Check, *Record) (httpHealtyEntry, bool)
+
+func cmpStartupHealth(c Check, r *Record) (httpHealtyEntry, bool) {
+	if r == nil {
+		return httpHealtyEntry{
+			Status: "UNAIVALABLE",
```
calebdoxsey:
```suggestion
			Status: "UNAVAILABLE",
```

pkg/health/http.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// here we intentionally do not treat errors as unsuccessful,
+	// as long as the reported state has moved out of starting
+	return entry, r.status != StatusUnknown
+}
+
+func cmpReadyHealth(c Check, r *Record) (httpHealtyEntry, bool) {
+	if r == nil {
+		return httpHealtyEntry{
+			Status: "UNAIVALABLE",
```
calebdoxsey:
```suggestion
			Status: "UNAVAILABLE",
```

pkg/health/http.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	// terminating should stop accepting connections even if all replicas in the deployment are
+	// in the terminating state
+	return entry, r.status == StatusRunning && r.err == nil
+}
+
+func cmpLivelinessHealth(c Check, r *Record) (httpHealtyEntry, bool) {
+	if r == nil {
+		return httpHealtyEntry{
+			Status: "UNAIVALABLE",
```
calebdoxsey:
```suggestion
			Status: "UNAVAILABLE",
```

pkg/health/channel.go
alexandreLamarre: I played around with this, and it might make sense to have these send values...
For example managing state transitions in a loop like:
```
for {
  select {
    case <- OnReady:
      // ready hooks
   case <- OnTerminating:
     // termination hooks
    
    
    // not implemented, but might be useful
    // case <-OnTerminated:
    // return / os.Exit(0) / os.Exit(1) / whatever
  }
}
```

Because when you close a channel the <-onReady will unblock immediately which may cause inadvertent behavior if we use something like the above but will work fine if really just want to do:
```go
<-OnReady
```

That being said, I only intended this being used in the `testenv` for stuff like integration tests

alexandreLamarre: WDYT?

calebdoxsey: ok sure, doesn't really matter if its only for testing anyway

PR #5815, opened 2025-09-03
databroker: add clustered follower server

calebdoxsey:
## Summary
Add a clustered follower databroker server. The clustered follower server is intended to be a follower node in a cluster of databroker nodes. It will forward all requests to a leader node.

Forwarding is done via a forwarding databroker server. For every method called on the forwarding server the corresponding client method will be called. Incoming metadata is copied to outgoing metadata to support passing the signed JWT or other metadata. In addition we set a `pomerium-forwarder-id` metadata value to detect cycles. (`ForwardUnary` and `ForwardStream` would be methods on `Forwarder` but Go doesn't support methods with generic arguments, so it's passed as an argument instead)

The clustered follower server also supports a `pomerium-cluster-request-mode` metadata value to explicitly request the desired forwarding behavior. `default` forwards all requests to the leader. `local` handles read-only requests with the local databroker backend. `leader` requires the node to be a leader, which means it always errors out for the follower server.

Subsequent PRs will introduce the clustered leader server and finally the clustered server which will be used when databroker cluster options are set and dispatches to the follower/leader server based on the node's role. In addition, a subsequent PR will update the clustered follower server to synchronize data from the leader to the local databroker backend.

## Related issues
- [ENG-2722](https://linear.app/pomerium/issue/ENG-2722/databroker-add-clustered-server)

## Checklist

- [x] updated unit tests
- [x] ready for review


pkg/grpcutil/serve_test.go
```diff
@@ -70,15 +70,11 @@ func TestServeWithGracefulStop(t *testing.T) {
...
 			// wait till the server is ready
-			for {
-				_, err := c.Check(ctx, &grpc_health_v1.HealthCheckRequest{
-					Service: "test",
-				})
-				if err != nil {
-					return err
-				}
-
-				break
```
wasaga: won't it introduce flakiness in CI?

calebdoxsey: The loop always terminated, so I don't think there was any difference.

PR #5817, opened 2025-09-04
databroker: implement follower sync

calebdoxsey:
## Summary
Implement sync replication of records from the clustered leader databroker server to the clustered follower databroker server.

## Related issues
- [ENG-2723](https://linear.app/pomerium/issue/ENG-2723/databroker-add-sync-replication-to-followers)

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/databroker/server_clustered_follower.go
+++ b/internal/databroker/server_clustered_follower.go
@@ -320,7 +320,7 @@ func (srv *clusteredFollowerServer) syncStep(
 			// latest records
 			return errClusteredFollowerNeedsReset
 		} else if err != nil {
-			return fmt.Errorf("error receiving sync latest message: %w", err)
+			return fmt.Errorf("error receiving sync message: %w", err)
 		}
 
 		b.Reset()


internal/databroker/server_backend.go
```diff
@@ -352,6 +352,8 @@ func (srv *backendServer) ServerInfo(ctx context.Context, _ *emptypb.Empty) (*da
 		return nil, err
 	}
 
+	log.Info().Uint64("server-version", serverVersion).Msg("SERVER INFO")
```
wasaga: debug?

internal/databroker/server_clustered_follower.go
```diff
@@ -183,3 +207,141 @@ func (srv *clusteredFollowerServer) invokeReadWrite(ctx context.Context, fn func
...
+
+	for {
+		res, err := stream.Recv()
+		if status.Code(err) == codes.Aborted {
+			return errClusteredFollowerNeedsReset
+		} else if err != nil {
+			return fmt.Errorf("error receiving sync latest message: %w", err)
+		}
+
+		versions.LatestRecordVersion = max(versions.LatestRecordVersion, res.Record.Version)
```
wasaga: we seem to assume the server version is the same - can we assert here just in case and return `errClusteredFollowerNeedsReset` ?

internal/databroker/server_clustered_follower.go
```diff
@@ -183,3 +207,141 @@ func (srv *clusteredFollowerServer) invokeReadWrite(ctx context.Context, fn func
...
+			records = append(records, res.Record)
+		}
+
+		_, err = srv.local.Put(ctx, &databrokerpb.PutRequest{
+			Records: records,
+		})
+		if err != nil {
+			return fmt.Errorf("error storing record from sync stream: %w", err)
+		}
+	}
```
wasaga: would it make sense to try and batch Put requests ?

internal/databroker/server_clustered_follower.go
```diff
@@ -183,3 +207,141 @@ func (srv *clusteredFollowerServer) invokeReadWrite(ctx context.Context, fn func
...
+
+		switch res := res.Response.(type) {
+		case *databrokerpb.SyncLatestResponse_Record:
+			_, err = srv.local.Put(ctx, &databrokerpb.PutRequest{
+				Records: []*databrokerpb.Record{res.Record},
+			})
+			if err != nil {
+				return fmt.Errorf("error storing record from sync latest stream: %w", err)
+			}
+		case *databrokerpb.SyncLatestResponse_Versions:
```
wasaga: same notice as above about batching

internal/databroker/server_clustered_follower.go
```diff
@@ -183,3 +207,141 @@ func (srv *clusteredFollowerServer) invokeReadWrite(ctx context.Context, fn func
...
+			return fmt.Errorf("error receiving sync latest message: %w", err)
+		}
+
+		versions.LatestRecordVersion = max(versions.LatestRecordVersion, res.Record.Version)
+		records := make([]*databrokerpb.Record, 0, 2)
+		records = append(records, &databrokerpb.Record{
+			Type: clusteredFollowerServerVersionsType,
+			Id:   clusteredFollowerServerVersionsID,
+			Data: protoutil.NewAny(&versions),
+		})
```
wasaga: I'm a little worried that we would be growing changelog 2x. did you consider any alternatives like storing the checkpoint out of band

internal/databroker/server_clustered_follower.go
calebdoxsey: We don't return the server version in the sync call. It's supposed to return an abort error if it doesn't match what we passed.

internal/databroker/server_clustered_follower.go
calebdoxsey: Yes we can definitely batch sync latest... not sure about sync, since changes come in as they happen. I suppose if we gave it a short 1 second delay, that'd be quick enough not to notice the delay, but long enough that it might batch a lot of records together if they're all changed at once.

internal/databroker/server_clustered_follower.go
calebdoxsey: Yeah that could definitely happen... I suppose the batching will help since we won't write it for every change. The issue with out-of-band storage is it breaks the abstraction... the only permanent store we have at this point is the databroker backend. We can't store it just in memory because if we reset the process we want it to pick up where it left off. Maybe we can make a follow up issue for this since it may not be a problem.

wasaga: To clarify: by “out-of-band,” I mean keep it on the databroker side but not as a record; access it via dedicated methods (e.g., `GetCheckpoint`/`SetCheckpoint`), letting the databroker choose the optimal storage mechanism.

calebdoxsey: That makes sense. I think pebble itself handles this ok, but it's unnecessary for us to store a changelog for these changes (and kind of annoying because we have to ignore these records when syncing).

pkg/grpc/databroker/databroker.proto
```diff
@@ -175,3 +176,17 @@ message Chunk {
...
+// A Checkpoint tracks synchronization state in a databroker.
+message Checkpoint {
+  uint64 server_version = 1;
+  uint64 record_version = 2;
+}
+
+// The CheckpointService gets and sets checkpoints.
+service CheckpointService {
+  // GetCheckpoint gets the checkpoint.
+  rpc GetCheckpoint(google.protobuf.Empty) returns (Checkpoint);
```
wasaga: would it be more future-proof to reserve an empty message instead of `google.protobuf.Empty` ?

pkg/storage/postgres/postgres.go
```diff
@@ -417,6 +436,23 @@ func putService(ctx context.Context, q querier, svc *registry.Service, expiresAt
...
+	if err != nil {
+		return err
+	}
+	err = rv.Scan(strconv.FormatUint(recordVersion, 10))
+	if err != nil {
+		return err
+	}
+	_, err = q.Exec(ctx, `
+		UPDATE `+schemaName+`.`+checkpointsTableName+`
+		SET server_version=$1, latest_record_version=$2
```
wasaga:
```suggestion
		SET server_version=$1, record_version=$2
```

as per https://github.com/pomerium/pomerium/pull/5817/files#diff-62bf548af13876630f888db4cfc0b4ab729f83712e69ae9a0d883006756503f4R180

internal/databroker/server_clustered_follower.go
```diff
@@ -183,3 +221,292 @@ func (srv *clusteredFollowerServer) invokeReadWrite(ctx context.Context, fn func
...
+	}
+
+	for {
+		res, err := stream.Recv()
+		if status.Code(err) == codes.Aborted {
+			// this indicates we need to reset and use sync latest to get the
+			// latest records
+			return errClusteredFollowerNeedsReset
+		} else if err != nil {
+			return fmt.Errorf("error receiving sync latest message: %w", err)
```
wasaga: the error is from `Sync` not from `SyncLatest` stream 


```suggestion
			return fmt.Errorf("error receiving sync message: %w", err)
```

PR #5818, opened 2025-09-04
feat: handle signals to better gracefully drain/terminate envoy

alexandreLamarre:
## Summary

Right now we explicitly do not handle OS signals in pomerium - this has the side effect of not necessarily running shutdown phases of pomerium, e.g. things that are supposed to run after a context is done, for example `context.AfterFunc(ctx, ...)`.

It works mostly as expected right now since when container orchestration frameworks like Kubernetes pass a SIGTERM to the process, any sub-processes spawned with exec.Command will inherit those signals and get sent SIGTERM as well, which according to the envoy docs will result in draining listeners according to `--drain-time` which defaults to 600s.

However, right now the pomerium.Shutdown is a utility function explicitly used by our test environment, our current pomerium.Wait doesn't handle graceful shutdown correctly, it has been amended to:
- wait for envoy server to close (either successfully or unsuccessfully)
- return errors to the parent command if and only if the first error was not surfaced via pomerium.Shutdown() or by one of the valid interrupt signals. Not handling these errors means that systemd, will handle graceful shutdown as well since it expects a zero return code on successful shutdown, which the previous iteration did not have.

An unintended side effect of catching signals is the envoy server gracePeriodSeconds on is always set to 0 (in non-test env settings) -- causing the envoy process to be SIGKILLED inadvertently by the old code. This has been fixed by respecting the default `--drain-time`.

<hr/>
In addition, we make a best effort to post an explicit graceful drain request to envoy during shutdown.

<hr/>

Some testing utilities I used to verify different termination sceanrios:
- `strace -p $(pidof pomerium) -e signal`
- `strace -p $(pidof envoy) -e signal`

should always receive SIGTERM/SIGINT and exit with status code 0. 

<hr/>

Ref:
- https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/draining
- https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-drain-time-s

## Related issues

https://linear.app/pomerium/project/healthcheck-with-graceful-teardown-e5428048d59e/overview
https://linear.app/pomerium/issue/ENG-2847/health-check-graceful-termination

## User Explanation

Users can expect that envoy's connections will be drained before the pomerium-proxy exits when it terminates gracefully, i.e. receives SIGINT/SIGTERM.
 
## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/cmd/pomerium/main.go
+++ b/cmd/pomerium/main.go
@@ -17,6 +17,7 @@ import (
 	_ "github.com/pomerium/pomerium/internal/zero/bootstrap/writers/k8s"
 	zero_cmd "github.com/pomerium/pomerium/internal/zero/cmd"
 	"github.com/pomerium/pomerium/pkg/cmd/pomerium"
+	"github.com/pomerium/pomerium/pkg/contextutil"
 	"github.com/pomerium/pomerium/pkg/envoy/files"
 	"github.com/pomerium/pomerium/pkg/telemetry/trace"
 )
@@ -32,7 +33,9 @@ func main() {
 	root.AddCommand(zero_cmd.BuildRootCmd())
 	root.PersistentFlags().StringVar(&configFile, "config", "", "Specify configuration file location")
 	log.SetLevel(zerolog.InfoLevel)
-	ctx := trace.NewContext(context.Background(), trace.NewSyncClient(nil))
+
+	sigCtx := contextutil.SetupSignals(context.Background())
+	ctx := trace.NewContext(sigCtx, trace.NewSyncClient(nil))
 	defer func() {
 		if err := trace.ShutdownContext(ctx); err != nil {
 			log.Error().Err(err).Send()
--- a/pkg/cmd/pomerium/pomerium.go
+++ b/pkg/cmd/pomerium/pomerium.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pomerium/pomerium/internal/log"
 	"github.com/pomerium/pomerium/internal/registry"
 	"github.com/pomerium/pomerium/internal/version"
+	"github.com/pomerium/pomerium/pkg/contextutil"
 	derivecert_config "github.com/pomerium/pomerium/pkg/derivecert/config"
 	"github.com/pomerium/pomerium/pkg/envoy"
 	"github.com/pomerium/pomerium/pkg/envoy/files"
@@ -85,6 +86,7 @@ type Pomerium struct {
 	startMu       sync.Mutex
 	cancel        context.CancelCauseFunc
 	envoyServer   *envoy.Server
+	envoyShutdown chan error
 	traceClientMu sync.Mutex
 }
 
@@ -93,7 +95,8 @@ func New(opts ...Option) *Pomerium {
 	options.apply(opts...)
 
 	return &Pomerium{
-		Options: options,
+		Options:       options,
+		envoyShutdown: make(chan error, 1),
 	}
 }
 
@@ -168,7 +171,7 @@ func (p *Pomerium) Start(ctx context.Context, tracerProvider oteltrace.TracerPro
 		Msg("server started")
 
 	// create envoy server
-	p.envoyServer, err = envoy.NewServer(ctx, src, controlPlane.Builder, p.envoyServerOptions...)
+	p.envoyServer, err = envoy.NewServer(ctx, p.envoyShutdown, src, controlPlane.Builder, p.envoyServerOptions...)
 	if err != nil {
 		return fmt.Errorf("error creating envoy server: %w", err)
 	}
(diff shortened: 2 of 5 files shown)


cmd/pomerium/main.go
```diff
@@ -32,7 +33,9 @@ func main() {
 	root.AddCommand(zero_cmd.BuildRootCmd())
 	root.PersistentFlags().StringVar(&configFile, "config", "", "Specify configuration file location")
 	log.SetLevel(zerolog.InfoLevel)
-	ctx := trace.NewContext(context.Background(), trace.NewSyncClient(nil))
+
+	sigCtx := contextutil.SetupSignals(context.Background())
```
calebdoxsey, line 37: We have very similar code in `internal/zero/cmd/run.go`. Could we swap that out to use this new function?

alexandreLamarre: yeah good catch

alexandreLamarre: It looks like the zero_cmd already accepts the context constructed here:
```go
    sigCtx := contextutil.SetupSignals(context.Background())
	ctx := trace.NewContext(sigCtx, trace.NewSyncClient(nil))
	defer func() {
		if err := trace.ShutdownContext(ctx); err != nil {
			log.Error().Err(err).Send()
		}
	}()
	runFn := run
	if zero_cmd.IsManagedMode(configFile) {
		runFn = zero_cmd.Run
	}
	root.RunE = func(_ *cobra.Command, _ []string) error {
		defer log.Ctx(ctx).Info().Msg("cmd/pomerium: exiting")
		return runFn(ctx, configFile)
	}

```
so I'll just remove the `withInterrupt` wrapper in Zero

pkg/envoy/envoy.go
```diff
@@ -129,41 +139,89 @@ func NewServer(ctx context.Context, src config.Source, builder *envoyconfig.Buil
...
+	q.Add("graceful", "")
+	req.URL.RawQuery = q.Encode()
+	log.Debug().Msg("requesting graceful drain from envoy")
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode/2 >= 200 {
```
kralicky: typo? !=

pkg/envoy/envoy.go
```diff
@@ -129,41 +139,89 @@ func NewServer(ctx context.Context, src config.Source, builder *envoyconfig.Buil
...
 	defer srv.mu.Unlock()
 
 	var err error
 	if srv.cmd != nil && srv.cmd.Process != nil {
-		var exited bool
-		if srv.exitGracePeriod > 0 {
+		if err := srv.Drain(); err != nil {
+			log.Error().Err(err).Msg("failed to request graceful drain from envoy")
+		}
+		log.Debug().Int("exit-grace-period-seconds", int(srv.exitGracePeriod.Seconds())).Msg("interrupting running envoy process")
```
kralicky: maybe reword this log message, it might be confusing

pkg/contextutil/signals.go
```diff
@@ -0,0 +1,28 @@
...
+	"syscall"
+
+	"github.com/pomerium/pomerium/internal/log"
+)
+
+var ErrSignalShutdown = errors.New("shutdown signal received")
+
+func SetupSignals(ctx context.Context) context.Context {
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGTERM, syscall.SIGINT)
```
kralicky, line 18: IMO we should handle SIGINT and SIGTERM differently. SIGINT should start a graceful shutdown, but SIGTERM should not

pkg/cmd/pomerium/pomerium.go
```diff
@@ -235,11 +238,12 @@ func (p *Pomerium) Shutdown(ctx context.Context) error {
 }
 
 func (p *Pomerium) Wait() error {
-	err := p.errGroup.Wait()
-	if errors.Is(err, ErrShutdown) {
-		return nil
+	errW := p.errGroup.Wait()
+	if errors.Is(errW, ErrShutdown) || errors.Is(errW, contextutil.ErrSignalShutdown) {
```
kralicky: this seems like it could be simplified. can we use one error for both with more details in the cause?

pkg/contextutil/signals.go
alexandreLamarre: `SIGTERM` is what is sent by container orchestration frameworks, when deleting a pod (e.g.) during a rollout

alexandreLamarre: `SIGTERM` is also what is used by systemd to gracefully terminate a service

alexandreLamarre: It's standard for both of these things to then send `SIGKILL` as the signal if the timeout is achieved (e.g. podTerminationGracePeriodSeconds)

kralicky: Yeah you're right, sounds like this is how we should be doing it.

pkg/envoy/envoy.go
```diff
@@ -129,41 +139,89 @@ func NewServer(ctx context.Context, src config.Source, builder *envoyconfig.Buil
...
 	if srv.cmd != nil && srv.cmd.Process != nil {
-		var exited bool
-		if srv.exitGracePeriod > 0 {
-			_ = srv.cmd.Process.Signal(os.Interrupt)
+		if err := srv.Drain(); err != nil {
+			log.Error().Err(err).Msg("failed to request graceful drain from envoy")
+		}
+		log.Debug().Int("exit-grace-period-seconds", int(srv.exitGracePeriod.Seconds())).Msg("requesting envoy to shutdown gracefully")
+		if srv.exitGracePeriodOrDefault() > 0 {
+			_ = srv.cmd.Process.Signal(syscall.SIGTERM)
```
alexandreLamarre, line 207: SIGTERM/SIGINT are handled identically in envoy, but choosing SIGTERM here, because conventions around signals is to send SIGTERM to request termination then SIGKILL if that request has not completed within the deadline

https://github.com/envoyproxy/envoy/blob/269a7227c86f8d1125b52579d60d45aee8b0ff27/source/server/server.cc#L973-L987

PR #5821, opened 2025-09-05
databroker: add ByteStream service

calebdoxsey:
## Summary
Add a new `ByteStream` gRPC service. It has a bidirectional streaming `Connect` method of `Chunk`s, which are just slices of bytes. This is used to implement the `net.Listener` and `net.Conn` interfaces. Basically implementing a bidirectional byte stream network connection over streaming gRPC.

The primary use case for this is as the network stream layer for raft. Rather than implementing the full method surface area in gRPC we can just use the built-in network transport via arbitrary bytes.

## Related issues
- [ENG-2832](https://linear.app/pomerium/issue/ENG-2832/databroker-byte-stream-grpc-service)

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/grpc/databroker/bytestream.go
+++ b/pkg/grpc/databroker/bytestream.go
@@ -158,8 +158,6 @@ func (li *byteStreamListener) Connect(stream grpc.BidiStreamingServer[Chunk, Chu
 
 	select {
 	case li.incoming <- conn: // send the connection to the accept method
-	case err := <-errCh: // the connection error'd out before we could accept it
-		return err
 	case <-li.closeCtx.Done(): // listener was closed
 		return context.Cause(li.closeCtx)
 	}


pkg/grpc/databroker/bytestream.go
```diff
@@ -0,0 +1,234 @@
...
+			err = stream.Send(chunk)
+			if err != nil {
+				return err
+			}
+		}
+	})
+	// on cancellation, cancel receiving and sending
+	eg.Go(func() error {
+		<-ctx.Done()
+		return context.Cause(ctx)
```
wasaga: it feels like in case of error returned by the receive goroutine above, the send one may be stuck in Read() as we did not 
close the connection ?

pkg/grpc/databroker/bytestream.go
```diff
@@ -0,0 +1,234 @@
...
+	li.closeCtx, li.close = context.WithCancelCause(context.Background())
+	return li
+}
+
+func (li *byteStreamListener) Connect(stream grpc.BidiStreamingServer[Chunk, Chunk]) error {
+	conn := newByteStreamConn()
+
+	// send the connection to the accept method
+	select {
+	case li.incoming <- conn:
```
wasaga: would it make sense to first start the goroutines before enqueing connection or make li.incoming buffered 1?

pkg/grpc/databroker/bytestream.go
```diff
@@ -0,0 +1,234 @@
...
+			_, err = conn.recvWriter.Write(chunk.Data)
+			if err != nil {
+				return err
+			}
+		}
+	})
+	// send data to the client
+	eg.Go(func() error {
+		for {
+			buf := make([]byte, defaultByteStreamBufferSize)
```
wasaga, line 143: is it possible to avoid allocating a new buffer on every iteration?

calebdoxsey: Unfortunately I think gRPC prevents us from doing this:

```
// It is not safe to modify the message after calling SendMsg. Tracing
// libraries and stats handlers may use the message lazily.
```

If we re-used the buffer we'd end up modifying the message sent via `Send`.

pkg/grpc/databroker/bytestream.go
calebdoxsey: Yes I think you may be right. To terminate the stream we have to return from the handler. I think the errgroup doesn't have the semantics we want for this... I will refactor.

pkg/grpc/databroker/bytestream.go
wasaga: it seems like https://github.com/grpc/grpc-go/issues/6619 might enable reuse. however it looks rather involved and if we do not expect a lot of traffic over this channel we can certainly leave it out

pkg/grpc/databroker/bytestream.go
```diff
@@ -0,0 +1,241 @@
...
+	}()
+
+	select {
+	case li.incoming <- conn: // send the connection to the accept method
+	case err := <-errCh: // the connection error'd out before we could accept it
+		return err
+	case <-li.closeCtx.Done(): // listener was closed
+		return context.Cause(li.closeCtx)
+	}
+
```
wasaga: what happens if `errCh` already has a value, do we maybe want to ensure that conn is always sent to the `incoming` so that `Accept()` won't ever hung?

wasaga (approved):

PR #5822, opened 2025-09-05
feat : http probe listeners & basic component health checks

alexandreLamarre:
## Summary

This adds basic health checks to pomerium core components to report them as running / terminating or in an error state.

The health check service reloads on configuration, and binds to a predetermined port `28080` on the machine's hostname. 

## Related issues

[ENG-2849](https://linear.app/pomerium/issue/ENG-2849/implement-core-health-checks)

## User Explanation

Users can enable health checks based on http probes, for example in Kubernetes deployments like:
```
livenessProbe:
      httpGet:
        path: /healthz
        port: 28080
      initialDelaySeconds: 3
      periodSeconds: 60
      failureThreshold: 10
startupProbe:
      httpGet:
        path: /startupz
        port: 28080
      initialDelaySeconds: 3
      periodSeconds: 60
      failureThreshold: 10
readinessProbe:
      httpGet:
        path: /readyz
        port: 28080
      initialDelaySeconds: 3
      periodSeconds: 60
      failureThreshold: 10
```
## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/pkg/cmd/pomerium/pomerium.go
+++ b/pkg/cmd/pomerium/pomerium.go
@@ -221,6 +221,7 @@ func (p *Pomerium) Start(ctx context.Context, tracerProvider oteltrace.TracerPro
 	if authorizeServer != nil {
 		p.errGroup.Go(func() error {
 			health.ReportRunning(health.AuthorizationService)
+			defer health.ReportTerminating(health.AuthorizationService)
 			return authorizeServer.Run(ctx)
 		})
 	}
@@ -228,10 +229,26 @@ func (p *Pomerium) Start(ctx context.Context, tracerProvider oteltrace.TracerPro
 		for _, check := range controlPlaneChecks {
 			health.ReportRunning(check)
 		}
+		defer func() {
+			for _, check := range controlPlaneChecks {
+				health.ReportTerminating(check)
+			}
+		}()
 		return controlPlane.Run(ctx)
 	})
 	if dataBrokerServer != nil {
 		p.errGroup.Go(func() error {
+			defer func() {
+				checks := []health.Check{
+					health.StorageBackend,
+					health.DatabrokerInitialSync,
+					health.DatabrokerBuildConfig,
+				}
+
+				for _, check := range checks {
+					health.ReportTerminating(check)
+				}
+			}()
 			return dataBrokerServer.Run(ctx)
 		})
 	}


internal/controlplane/server.go
```diff
@@ -157,11 +165,23 @@ func NewServer(
...
 	}
 
+	hostname, err := os.Hostname()
+	if err != nil {
+		return nil, err
+	}
+	// Kubernetes and other container orchestration frameworks hit health probes
+	// on externally visible endpoints, like the podIP/hostname
+	srv.HealthCheckListener, err = reuseport.Listen("tcp4", net.JoinHostPort(hostname, strconv.Itoa(cfg.Options.HealthCheckPort)))
+	if err != nil {
```
wasaga: in those environments we resolve this issue like this: 

https://github.com/pomerium/ingress-controller/blob/e358f82173846b258389928272565bc7860bed81/config/pomerium/deployment/args.yaml#L14-L24

would it make sense to retain this approach and just have some config parameter like `health_addr` ?

alexandreLamarre: yeah makes sense

internal/testenv/environment.go
```diff
@@ -1015,9 +1017,10 @@ func (e *environment) AddUpstream(up Upstream) {
...
 // ReportError implements health.Provider.
-func (e *environment) ReportError(check health.Check, err error, attributes ...health.Attr) {
+func (e *environment) ReportError(_ health.Check, _ error, _ ...health.Attr) {
 	// note: don't use e.t.Fatal here, it will deadlock
-	panic(fmt.Sprintf("%s: %v %v", check, err, attributes))
+	// TODO : re-enable when health checks can be split across test envs
+	// panic(fmt.Sprintf("%s: %v %v", check, err, attributes))
 }
 
 // ReportOK implements health.Provider.
```
wasaga:
```suggestion
// ReportStatus implements health.Provider.
```

authenticate/authenticate.go
```diff
@@ -126,6 +127,7 @@ func (a *Authenticate) OnConfigChange(ctx context.Context, cfg *config.Config) {
 	a.options.Store(cfg.Options)
 	if state, err := newAuthenticateStateFromConfig(ctx, a.tracerProvider, cfg, a.cfg, &a.outboundGrpcConn); err != nil {
 		log.Ctx(ctx).Error().Err(err).Msg("authenticate: failed to update state")
+		health.ReportError(health.AuthenticateService, fmt.Errorf("authenticate failed to update state : %w", err))
```
wasaga: how do we clear it

config/config.go
```diff
@@ -248,3 +249,52 @@ func (cfg *Config) resolveAuthenticateURL() (*url.URL, http.RoundTripper, error)
...
+func (cfg *Config) isZero() bool {
+	return cfg.ZeroClusterID != ""
+}
+
+func (cfg *Config) GetExpectedHealthChecks() (ret []health.Check) {
+	if cfg.isZero() {
+		ret = append(ret,
+			health.ZeroConnect,
+			health.ZeroBootstrapConfigSave,
+			health.ZeroRoutesReachable,
```
wasaga: this particular check may fail for a very long time due to misconfiguration, i don't feel it should hold off startup

config/config.go
```diff
@@ -248,3 +249,52 @@ func (cfg *Config) resolveAuthenticateURL() (*url.URL, http.RoundTripper, error)
...
+
+func (cfg *Config) isZero() bool {
+	return cfg.ZeroClusterID != ""
+}
+
+func (cfg *Config) GetExpectedHealthChecks() (ret []health.Check) {
+	if cfg.isZero() {
+		ret = append(ret,
+			health.ZeroConnect,
+			health.ZeroBootstrapConfigSave,
```
wasaga: this check failing is not perfect, but it does not withold application from starting up. failing zero checks are presented to the operator in the console

config/config.go
```diff
@@ -248,3 +249,52 @@ func (cfg *Config) resolveAuthenticateURL() (*url.URL, http.RoundTripper, error)
...
 }
+
+func (cfg *Config) isZero() bool {
+	return cfg.ZeroClusterID != ""
+}
+
+func (cfg *Config) GetExpectedHealthChecks() (ret []health.Check) {
+	if cfg.isZero() {
+		ret = append(ret,
+			health.ZeroConnect,
```
wasaga: i believe this health check could be flaky, as the upstream load balancer periodically resets the long-living connection, like every hour or so. also it could be blocked by some weirdness of the user network situation. however it does not block pomerium from being functional as we would just switch into periodic polling.

config/config.go
```diff
@@ -248,3 +249,52 @@ func (cfg *Config) resolveAuthenticateURL() (*url.URL, http.RoundTripper, error)
 	transport = otelhttp.NewTransport(transport)
 	return authenticateURL, transport, nil
 }
+
+func (cfg *Config) isZero() bool {
+	return cfg.ZeroClusterID != ""
+}
+
+func (cfg *Config) GetExpectedHealthChecks() (ret []health.Check) {
```
wasaga: nit: config is just config, maybe let's avoid adding runtime-specific logic in this package and move this function to `internal/controlplane`

config/config.go
```diff
@@ -248,3 +249,52 @@ func (cfg *Config) resolveAuthenticateURL() (*url.URL, http.RoundTripper, error)
...
+	if IsProxy(services) {
+		ret = append(
+			ret, health.ProxyServer,
+		)
+	}
+
+	ret = append(
+		ret,
+		// contingent on control plane
+		health.StorageBackend,
```
wasaga: i believe this one would only be reported if databroker service is running in this process

pkg/cmd/pomerium/pomerium.go
```diff
@@ -201,18 +208,26 @@ func (p *Pomerium) Start(ctx context.Context, tracerProvider oteltrace.TracerPro
...
 		p.errGroup.Go(func() error {
+			health.ReportRunning(health.AuthorizationService)
 			return authorizeServer.Run(ctx)
 		})
 	}
 	p.errGroup.Go(func() error {
+		for _, check := range controlPlaneChecks {
+			health.ReportRunning(check)
+		}
 		return controlPlane.Run(ctx)
```
wasaga: nit: shouldn't services report themselves that they are running / shutting down like i.e. envoy svc in https://github.com/pomerium/pomerium/blob/e7c456f9e252da18787c37a1812247aca5f31f42/pkg/envoy/envoy.go#L312

pkg/health/check.go
```diff
@@ -5,12 +5,25 @@ import "fmt"
...
 
 const (
+	// AuthenticationService checks if the authentication service is up and running
+	AuthenticateService = Check("authenticate.service")
+	// AuthorizationService checks if the authorization service is up and running
+	AuthorizationService = Check("authorize.service")
+	// EnvoyServer checks if the envoy server is up and running
+	EnvoyServer = Check("envoy.server")
+	// ProxyService checks if the proxy server is up and running
+	ProxyServer = Check("proxy.server")
```
wasaga: nit: I believe we call it "proxy service" elsewhere

pkg/storage/file/backend.go
```diff
@@ -51,14 +52,14 @@ type Backend struct {
 
 // New creates a new Backend.
 func New(dsn string) *Backend {
+	health.ReportRunning(health.StorageBackend)
```
wasaga: i believe it has to happen in the `init()` method. also error handling

pkg/storage/postgres/backend.go
```diff
@@ -422,6 +432,14 @@ func (backend *Backend) init(ctx context.Context) (serverVersion uint64, pool *p
...
 }
 
+func (backend *Backend) doOnceAndPeriodically(f func(ctx context.Context) error, dur time.Duration) {
+	ctx := backend.closeCtx
+	if err := f(ctx); err != nil {
+		log.Ctx(ctx).Error().Err(err).Msg("storage/postgres")
+	}
+	backend.doPeriodically(f, dur)
+}
+
```
wasaga, line 465: could you please explain why this is necessary

alexandreLamarre: this is for running the background tasks (pings, cleanup) on startup, as opposed to waiting the default 30 seconds. It's a small optimization for making the time for pods to be ready faster when doing a rollout / scaling 

For context on my machine, the containers become ready for traffic in about 17s with about 1000 routes configured making this the slowest part of the startup phase even at a respectable scale. It's also fair to expect this database to be up and running when pomerium starts, since it's not tied to the pomerium deployment

alexandreLamarre: For my local environment I do have a 14 core CPU @ 2600Mhz and a pretty fast SSD, which might skew things on the faster side for the startup phase when compared with cloud VMs

pkg/cmd/pomerium/pomerium.go
alexandreLamarre: yes although at the moment since there aren't any cleanup tasks associated with these other components, the switch between running and terminating will happen basically as the process exits in these cases. 

Probably better to include them, even if just superficially

They might be helpful to detect if future commit introduce deadlocks on context cancellation in some cases

config/config.go
alexandreLamarre: Removed zero checks from expected checks in probes

config/config.go
alexandreLamarre: Removed zero checks from expected checks in probes

config/config.go
alexandreLamarre: Removed zero checks from expected checks in probes

PR #5825, opened 2025-09-10
databroker: add clustered server

calebdoxsey:
## Summary
Implement the clustered server which will select the databroker-clustered-follower-server or the databroker-clustered-leader-server depending on configuration.

For this PR leader election is done by always selecting the first node.

## Checklist

- [ ] updated unit tests
- [ ] ready for review


internal/databroker/client.go
```diff
@@ -0,0 +1,130 @@
...
+
+	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
+
+	// if the shared key and the certificate authority hasn't changed,
+	// there's not anything to update
+	if bytes.Equal(mgr.sharedKey, sharedKey) &&
+		bytes.Equal(mgr.caPEM, caPEM) {
+		return
+	}
```
wasaga, line 68: would it make sense to do RLock and this check first ? (otherwise, why use RWMutex?)

internal/databroker/server_clustered.go
```diff
@@ -0,0 +1,314 @@
...
+
+	// get the elected leader
+	leaderID := srv.currentLeaderElector.ElectedLeaderID()
+	if !leaderID.IsValid() {
+		log.Ctx(ctx).Error().Msg("cluster has no leader")
+		srv.currentServer = NewErroringServer(databrokerpb.ErrClusterHasNoLeader)
+		return
+	}
+
+	ctx = log.Ctx(ctx).With().Str("cluster-leader-id", nodeID.String).Logger().WithContext(ctx)
```
wasaga: should it be

```suggestion
	ctx = log.Ctx(ctx).With().Str("cluster-leader-id", leaderID).Logger().WithContext(ctx)
```

?

internal/databroker/client.go
calebdoxsey: you're right, there's no advantage to the RWMutex in this case.

PR #5836, opened 2025-09-18
tests: change port allocation

calebdoxsey:
## Summary
Change how port allocation is done for tests. Instead of always picking `127.0.0.1` for a random port, pick a random loopback ip. Hopefully this will prevent address re-use issues.

## Checklist

- [ ] updated unit tests
- [ ] ready for review


kralicky:
How does DNS resolution work now in tests? Are we still using *.localhost.pomerium.io?

calebdoxsey:
> How does DNS resolution work now in tests? Are we still using *.localhost.pomerium.io?

We use names like `mock-idp-127-1-2-3.sslip.io`. This returns `127.1.2.3`.

kralicky:
Hmm, I think we still might need the dns override code then. Those requests will be un-cachable since it's a new random subdomain every time, so requests will trigger real internet dns lookups.

calebdoxsey:
> For me it takes ~100ms to resolve every new sslip.io domain, multiplied by every time we create a route for a test probably adds up to a lot. It also makes some benchmarks and stress tests impossible.

ok 🤷 . I'll move this back to draft. I'm not really sure how to approach getting rid of DNS entirely, so I'll probably end up closing this.

kralicky:
wait, what? don't close this, we just need to make some tweaks to DNS but otherwise I think this might work great

PR #5842, opened 2025-09-22
raft: skip TestStreamLayer on macOS for now

kenjenkins:
## Summary

TestStreamLayer is consistently failing on macOS. Let's skip it for now until we can figure out how best to fix it.

## Related issues

n/a

## Checklist

- [x] updated unit tests
- [ ] ready for review


calebdoxsey:
We could just pick a random port instead of listening on that address.

PR #5843, opened 2025-09-22
chore: upgrade to go 1.25

alexandreLamarre:
## Summary

DO NOT MERGE until release 0.31.0

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -15,7 +15,7 @@ jobs:
 
       - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
         with:
-          go-version: 1.25.x
+          go-version-file: .tool-versions
           cache: false
 
       - run: make deps-build
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,7 +33,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
         with:
-          go-version: 1.24.x
+          go-version-file: .tool-versions
           cache: false
 
       - name: Set up Docker
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,7 +22,7 @@ jobs:
 
       - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
         with:
-          go-version: 1.25.x
+          go-version-file: .tool-versions
           cache: false
 
       - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903
@@ -59,7 +59,7 @@ jobs:
 
       - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
         with:
-          go-version: 1.25.x
+          go-version-file: .tool-versions
 
       - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903
         with:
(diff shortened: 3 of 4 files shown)


go.mod
```diff
@@ -2,6 +2,12 @@ module github.com/pomerium/pomerium
 
 go 1.25.1
 
+// breaking change in go 1.25.0 changed the behaviour of x509.CreateCertificate:
+// https://go-review.googlesource.com/c/go/+/674477
```
alexandreLamarre, line 6: Informal proposal : https://github.com/golang/go/issues/71746

This will break the test in https://github.com/pomerium/pomerium/blob/efb100b29e46c7bf8a19acb559034eb03bf59bf3/config/config_test.go#L16, but more generally is a (probably good) change in default behavior in https://pkg.go.dev/crypto/x509#CreateCertificate 

I'm not really sure of the wider implications of this change in pomerium...

go.mod
```diff
@@ -2,6 +2,10 @@ module github.com/pomerium/pomerium
...
+godebug x509sha256skid=0
```
alexandreLamarre: Informal proposal : https://github.com/golang/go/issues/71746

This will break the test in https://github.com/pomerium/pomerium/blob/efb100b29e46c7bf8a19acb559034eb03bf59bf3/config/config_test.go#L16 but more generally is a (probably good) change in default behavior in https://pkg.go.dev/crypto/x509#CreateCertificate 

I'm not really sure of the wider implications of this change in pomerium...

alexandreLamarre: Looks like this won't affect our auto-cert functionality, and I don't think downstream consumers of these TLS certs are going to care if the hash of SubjectKeyId has changed, or depend on it being the same.

.golangci.yml
```diff
@@ -1,81 +1,89 @@
...
     - goprintffuncname
     - gosec
-    - gosimple
     - govet
     - ineffassign
     - misspell
     - nakedret
     - nolintlint
     - revive
     - staticcheck
```
kenjenkins, line 20: I would suggest we disable the staticcheck [QF1008](https://staticcheck.dev/docs/checks#QF1008) check and revert the associated changes. I think it's often clearer to be explicit about embedded fields.

(For comparison, here's the staticcheck settings from the golangci-lint configuration in Zero: https://github.com/pomerium/pomerium-zero/blob/3d33204f823a130c26fd4880155599fde8310549/.golangci.yml#L57-L71.)

alexandreLamarre: very much agree

scripts/install-lint.sh
alexandreLamarre, line 1: I added this in lieu of explicitly overriding GOTOOLCHAIN in the Makefile. Not only is this faster, but we no longer have to deal with the compile-time headaches that golangci-lint come with. 

golangci-lint has drifted away from go install & go run as valid installation mechanisms, so to prevent future mismatches I decided to do something similar similar to what we do with envoy and download golangci-lint if our version no longer matches the one provided by our github CI. 

Note that the install script puts `golangci-lint` in `./bin`

alexandreLamarre: >Error: can't load config: the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.4)
The command is terminated due to an error: can't load config: the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.4)

With go run if your toolchain was set to auto you would get these types of errors if your base go binary was 1.24

tools.mod
```diff
@@ -0,0 +1,215 @@
+module github.com/pomerium/pomerium
```
calebdoxsey: Do we need tools.mod anymore?

alexandreLamarre: oops yeah forgot to remove that sorry

.github/workflows/lint.yaml
```diff
@@ -15,14 +15,12 @@ jobs:
 
       - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
         with:
-          go-version: 1.24.x
+          go-version: 1.25.x
```
calebdoxsey: Could we use `go-version-file` and add it to `.tool-versions`?

alexandreLamarre: sure

calebdoxsey (approved):

PR #5848, opened 2025-09-23
core: replace atomicutil.Value with atomic.Pointer

calebdoxsey:
## Summary
Remove the `atomicutil.Value` and replace it with the `atomic.Pointer`.

## Related issues
- [ENG-2622](https://linear.app/pomerium/issue/ENG-2622/migrate-from-atomicutilvalue-to-atomicpointer)

## Checklist
- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/zero/reconciler/service.go
+++ b/internal/zero/reconciler/service.go
@@ -8,7 +8,6 @@ package reconciler
 import (
 	"context"
 	"sync/atomic"
-	"time"
 
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/time/rate"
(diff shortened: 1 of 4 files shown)


internal/zero/reconciler/service.go
```diff
@@ -25,7 +25,7 @@ type service struct {
 
 	fullSyncRequest        chan struct{}
 	bundleSyncRequest      chan struct{}
-	periodicUpdateInterval atomicutil.Value[time.Duration]
+	periodicUpdateInterval atomic.Pointer[time.Duration]
```
kralicky: Maybe this should be an atomic int64 instead, then cast it to/from time.Duration (same for zero/bootstrap/new.go)

PR #5855, opened 2025-09-29
ssh: pass through source ip to ppl evaluator

kralicky:
This enables source IP filtering for SSH connections.


pkg/ssh/stream.go
```diff
@@ -190,11 +191,15 @@ func (sh *StreamHandler) Run(ctx context.Context) error {
 	if sh.state != nil {
 		panic("Run called twice")
 	}
+	sourceAddr, _, err := net.SplitHostPort(sh.downstream.SourceAddress)
+	if err != nil {
+		return status.Errorf(codes.Internal, "invalid source address: %v", err)
```
kenjenkins: I think this is breaking some of the testenv tests? (https://github.com/pomerium/pomerium/actions/runs/18105592610/job/51519147932?pr=5855#step:7:284)

kralicky: oh oops, let me fix this

kenjenkins (approved):

PR #5869, opened 2025-10-02
authenticate: fix callback path

calebdoxsey:
## Summary
The authenticate callback path option wasn't being used in the authenticate service. To get this to work we needed to dynamically construct the `mux.Router` any time the config was updated. This is similar to what we do in the proxy service.

## Related issues
- [ENG-2951](https://linear.app/pomerium/issue/ENG-2951/core-fix-or-remove-the-authenticate-callback-path-setting)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/authenticate/handlers.go
+++ b/authenticate/handlers.go
@@ -19,6 +19,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	oteltrace "go.opentelemetry.io/otel/trace"
 
+	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/internal/authenticateflow"
 	"github.com/pomerium/pomerium/internal/handlers"
 	"github.com/pomerium/pomerium/internal/httputil"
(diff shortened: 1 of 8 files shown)


wasaga:
do we need to update this static reference then ? https://github.com/pomerium/pomerium/blob/30212463d8026b2fd860386162955cbe85d28ecf/internal/handlers/well_known_pomerium.go#L23

authenticate/handlers.go
```diff
@@ -32,23 +33,25 @@ import (
...
-	return r
-}
-
 // Mount mounts the authenticate routes to the given router.
 func (a *Authenticate) Mount(r *mux.Router) {
+	r.PathPrefix("/").Handler(a)
+}
+
+func (a *Authenticate) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	a.currentRouter.Load().ServeHTTP(w, r)
```
wasaga, line 42: consider checking that it's not nil and returning 503 just in case?

calebdoxsey: We always set it to at least an empty router so I think it should be ok.

PR #5870, opened 2025-10-03
envoy: fix ready check

calebdoxsey:
## Summary
I believe the expectation for the envoy ready check is that it would be performed every 30 seconds, except the first time, which would happen 1 second after starting. However the way the code was written, it simply always checked every second.

## Checklist

- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/envoy/envoy.go
+++ b/pkg/envoy/envoy.go
@@ -488,6 +488,9 @@ func (srv *Server) readiness(ctx context.Context, exited chan struct{}) {
 	ticker := time.NewTicker(time.Second * 30)
 	defer ticker.Stop()
 
+	firstTime := time.NewTicker(time.Second)
+	defer firstTime.Stop()
+
 	for {
 		select {
 		case <-exited:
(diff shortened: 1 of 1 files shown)


pkg/envoy/envoy.go
```diff
@@ -488,14 +488,17 @@ func (srv *Server) readiness(ctx context.Context, exited chan struct{}) {
 	ticker := time.NewTicker(time.Second * 30)
 	defer ticker.Stop()
 
+	firstTime := time.NewTicker(time.Second)
```
alexandreLamarre: This can be fixed using:

 `firstC := time.After(time.Second)`

to avoid opening another time.Ticker

pkg/envoy/envoy.go
```diff
@@ -488,14 +488,17 @@ func (srv *Server) readiness(ctx context.Context, exited chan struct{}) {
...
+
 	for {
 		select {
 		case <-exited:
 			return
 		case <-ctx.Done():
 			// small optimization to bring up envoy as ready
 			return
-		case <-time.After(time.Second):
+		case <-firstTime.C:
```
kenjenkins: Will this still trigger every second?

Would it be simpler to use one ticker and then call Reset(30 * time.Second) after the first tick?

pkg/envoy/envoy.go
calebdoxsey: woops

pkg/envoy/envoy.go
calebdoxsey: yeah I meant it to a timer, lol

kenjenkins (approved):

alexandreLamarre (approved):

PR #5873, opened 2025-10-06
feat: ssh authorization code flow

alexandreLamarre:
## Summary

Implements the oauth authorization code flow for native SSH connections in pomerium.

## Related issues

[ENG-3056](https://linear.app/pomerium/issue/ENG-3056/ssh-authorization-code-flow-initial-implementation)

The issue has a comment explaining some known limitations.

## User Explanation

Allows user to use native SSH without requiring IDPs to support the device code flow. Native ssh now uses the Authorization code flow, which most if not all IDPs support - notably, any existing IDP configuration for non-native ssh Pomerium should now work with native ssh.

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/ui/src/components/SessionBindingInfo.tsx
+++ b/ui/src/components/SessionBindingInfo.tsx
@@ -27,28 +27,27 @@ const SessionBindingInfoPage : FC<SessionBindingInfoProps> = ({data}) => {
                     <Table size="small" aria-label="metadata table">
                       <TableHead>
                         <TableRow>
-                          <TableCell variant="head" align="left"> Session ID </TableCell>
-                          <TableCell variant="head" align="left"> Protocol </TableCell>
-                          <TableCell variant="head" align="left"> IssuedAt </TableCell>
-                          <TableCell variant="head" align="left"> ExpiresAt </TableCell>
-                          <TableCell variant="head" align="left"> Actions </TableCell>
+                          <TableCell variant="head">Session ID</TableCell>
+                          <TableCell variant="head">Protocol</TableCell>
+                          <TableCell variant="head">IssuedAt</TableCell>
+                          <TableCell variant="head">ExpiresAt</TableCell>
+                          <TableCell variant="head">Actions</TableCell>
                         </TableRow>
                       </TableHead>
                       <TableBody>
                         {data.sessions?.map((s) => (
-                            <TableRow>
+                            <TableRow key={s.SessionID}>
                             <TableCell component="th" scope="row">{s.SessionID}</TableCell>
-                            <TableCell align="left">{s.Protocol}</TableCell>
-                            <TableCell align="left">{s.IssuedAt}</TableCell>
-                            <TableCell align="left">{s.ExpiresAt}</TableCell>
-                            <TableCell align="left">
+                            <TableCell>{s.Protocol}</TableCell>
+                            <TableCell>{s.IssuedAt}</TableCell>
+                            <TableCell>{s.ExpiresAt}</TableCell>
+                            <TableCell>
                               <Box
                                 component="form"
                                 action={s.RevokeURL}
                                 method="POST"
                                 sx={{ display: "inline-flex", gap: 1 }}
                               >
-                                {/* <input type="hidden" name="_pomerium_csrf" value={data.csrfToken} /> */}
                                 <input type="hidden" name="sessionID" value={s.SessionID} />
                                 <Button size="small" type="submit" variant="contained">
                                   Revoke
--- a/ui/src/components/SignInSuccessPage.tsx
+++ b/ui/src/components/SignInSuccessPage.tsx
@@ -18,25 +18,25 @@ const SignInSuccessPage: FC<SignInSuccessPageProps> = ({ data }) => {
   return (
     <>
       <TableContainer component={Paper} sx={{ maxWidth: 540, mx: "auto", mb: 2 }}>
-        <Table size="small" aria-label="metadata table">
+      <Table size="small" aria-label="metadata table">
           <TableHead>
             <TableRow>
               <TableCell variant="head">Field</TableCell>
-              <TableCell variant="head" align="left">Value</TableCell>
+              <TableCell variant="head">Value</TableCell>
             </TableRow>
           </TableHead>
           <TableBody>
             <TableRow>
               <TableCell component="th" scope="row">Name</TableCell>
-              <TableCell align="left">{data.user.name}</TableCell>
+              <TableCell>{data.user.name}</TableCell>
             </TableRow>
             <TableRow>
               <TableCell component="th" scope="row">Protocol</TableCell>
-              <TableCell align="left">{data.protocol}</TableCell>
+              <TableCell>{data.protocol}</TableCell>
             </TableRow>
             <TableRow>
               <TableCell component="th" scope="row">Expiry date</TableCell>
-              <TableCell align="left">{data.expiresAt}</TableCell>
+              <TableCell>{data.expiresAt}</TableCell>
             </TableRow>
           </TableBody>
         </Table>
--- a/ui/src/components/SignInVerifyPage.tsx
+++ b/ui/src/components/SignInVerifyPage.tsx
@@ -21,7 +21,7 @@ type SignInVerifyPageProps = {
   data: SignInVerifyPageData;
 };
 const SignInVerifyPage: FC<SignInVerifyPageProps> = ({ data }) => {
-  const parseToDate = (v: any): Date => {
+  const parseToDate = (v: unknown): Date => {
     if (v instanceof Date && !isNaN(v.getTime())) return v;
     if (typeof v === "number" && !isNaN(v)) return new Date(v);
     if (typeof v === "string") {
@@ -31,7 +31,7 @@ const SignInVerifyPage: FC<SignInVerifyPageProps> = ({ data }) => {
     return new Date(Date.now() + 2 * 60 * 1000);
   };
 
-  const expiresDate = parseToDate((data as any).expiresAt ?? data.expiresAt);
+  const expiresDate = parseToDate(data.expiresAt);
   const [remainingSeconds, setRemainingSeconds] = useState<number>(
     Math.max(0, Math.round((expiresDate.getTime() - Date.now()) / 1000))
   );
@@ -62,21 +62,21 @@ const SignInVerifyPage: FC<SignInVerifyPageProps> = ({ data }) => {
           <TableHead>
             <TableRow>
               <TableCell variant="head">Field</TableCell>
-              <TableCell variant="head" align="left">Value</TableCell>
+              <TableCell variant="head">Value</TableCell>
             </TableRow>
           </TableHead>
           <TableBody>
             <TableRow>
               <TableCell component="th" scope="row">Protocol</TableCell>
-              <TableCell align="left">{data.protocol}</TableCell>
+              <TableCell>{data.protocol}</TableCell>
             </TableRow>
             <TableRow>
               <TableCell component="th" scope="row">Issued at</TableCell>
-              <TableCell align="left">{parseToDate(data.issuedAt).toLocaleString()}</TableCell>
+              <TableCell>{parseToDate(data.issuedAt).toLocaleString()}</TableCell>
             </TableRow>
             <TableRow>
               <TableCell component="th" scope="row">Initiated from</TableCell>
-              <TableCell align="left">{data.sourceAddr}</TableCell>
+              <TableCell>{data.sourceAddr}</TableCell>
             </TableRow>
           </TableBody>
         </Table>


pkg/ssh/code/issuer.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	i.setupWait.L.Unlock()
+}
+
+func (i *issuer) AssociateCode(
+	ctx context.Context,
+	code CodeID,
+	sbr *session.SessionBindingRequest,
+) (CodeID, error) {
+	i.waitForSetup()
```
kralicky: If setup() returns an error and this is waiting, it will never wake up.

pkg/ssh/code/relational.go
```diff
@@ -0,0 +1,106 @@
...
+		if s.Accepted != nil {
+			// already processed
+			continue
+		}
+		ret = append(ret, T2(CodeID(rec.GetId()), s))
+	}
+
+	slices.SortFunc(ret, func(a, b Tuple2[CodeID, *session.SessionBindingRequest]) int {
+		return cmp.Compare(
+			a.B.GetCreatedAt().GetNanos(),
```
kralicky: I don't think this logic is correct since both seconds and nanos fields are part of the timestamp, you should probably convert to time.Time and compare that way instead

pkg/ssh/code/reader.go
```diff
@@ -0,0 +1,105 @@
...
+			Records: []*databroker.Record{},
+		}
+	} else if err != nil {
+		return nil, err
+	}
+
+	for _, sb := range sessBindingRecs.GetRecords() {
+		var sess session.SessionBinding
+		if err := sb.GetData().UnmarshalTo(&sess); err != nil {
+			panic(err)
```
kralicky: Perhaps this is one case where we don't actually want to panic if unmarshal fails, for example if the record is persisted in the db across a restart where the session binding proto was changed in a backwards-incompatible way. If this fails you can probably just treat it as if it was not found.

pkg/ssh/code/revoker.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	if err != nil {
+		return err
+	}
+	if sbResp.Record.GetDeletedAt() != nil {
+		return nil
+	}
+	rec := sbResp.Record
+	rec.DeletedAt = timestamppb.Now()
+	_, err = r.client.Put(ctx, &databroker.PutRequest{
```
kralicky: Maybe `Patch` would be better here (same below)

pkg/ssh/code/code_manager.go
```diff
@@ -0,0 +1,136 @@
...
+
+	slices.SortFunc(toFilter, func(a, b Status) int {
+		return a.IssuedAt.Compare(b.IssuedAt)
+	})
+
+	n := len(toFilter) - 1
+	return toFilter[n], true
+}
+
+func (c *codeManager) cleanUp() {
```
kralicky:
```suggestion
func (c *codeManager) clearExpiredCodesLocked() {
```
this could benefit from a more descriptive name

pkg/ssh/code/code_manager.go
```diff
@@ -0,0 +1,136 @@
...
+	c.cleanUp()
+	for _, record := range records {
+		codeID := record.GetId()
+
+		s := &session.SessionBindingRequest{}
+		if err := record.GetData().UnmarshalTo(s); err != nil {
+			panic(err)
+		}
+		var state State
+		if s.Accepted == nil {
```
kralicky: Would it make sense to use this enum in the proto api instead of optional bool?

pkg/ssh/code/issuer.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (i *issuer) IssueCode() CodeID {
+	code := [16]byte{}
+	_, _ = rand.Read(code[:])
+	codeStr := base64.RawURLEncoding.EncodeToString(code[:])
+	return CodeID(codeStr)
+}
+
+func (i *issuer) OnCodeDecision(ctx context.Context, code CodeID) <-chan Status {
+	ret := make(chan Status, 1)
```
kralicky, line 96: I think this logic should be delegated to the `codeManager`, where it can be implemented more efficiently. Since the code manager is the SyncerHandler, it can respond to record updates by directly notifying any watchers. When `codeByExpiration.ReplaceOrInsert` is called, it could look up any watchers registered to receive status updates for that code, and notify them directly.

alexandreLamarre: I think I need to think on this more. I recall it became a pretty difficult synchronization problem when we think about what happens when there are pending sessions across replicas

For example, at what point do we know it's safe to attach a watcher on UpdateRecords without missing updates? I don't think I can come up with a substantially better solution that works without backend changes...

I think if you look in the first commit, there was something in `pkg/ssh/pending/value.go` that did something like that, but was clunky. 

The rationale behind opting for periodically polling on a btree was because it was simple to index on many fields, which made a lot of operations semi-efficient without having to write something along the lines of "a thread-safe-durable-replay queue" (and also scale well, memory footprint-wise)

TLDR; let me sleep on it and load up on coffee first thing tomorrow ☕

pkg/ssh/code/reader.go
alexandreLamarre: I removed all panics when unmarshalling, and decided to log them at error level for now. I think when I do a more complete pass of logs,metrics and traces I'll log these differently.

pkg/ssh/code/issuer.go
alexandreLamarre: yeah in the current implementation `setup` will panic on error though, it's one of these init synchronization things that I may have to address in a follow up PR

config/testdata/basic.json
kenjenkins, line 1: Would you mind moving any formatting-only changes to a separate PR?

internal/authenticateflow/stateless.go
```diff
@@ -279,19 +280,24 @@ func (s *Stateless) RevokeSession(
...
+	if !hpke.IsEncryptedURL(vs) {
+		return maps.Clone(vs), nil
 	}
-	return idpID
+	_, requestParams, err := hpke.DecryptURLValues(s.hpkePrivateKey, vs)
+	return requestParams, err
+}
+
+func (s *Stateless) AuthenticatePendingSession(_ http.ResponseWriter, _ *http.Request, _ *sessions.Handle) error {
+	return nil
```
kenjenkins: Should these three new methods all return some "not implemented" error rather than `nil`?

internal/authenticateflow/stateful_test.go
```diff
@@ -172,13 +172,6 @@ func TestStatefulAuthenticateSignInURL(t *testing.T) {
 	})
 }
 
-func TestStatefulGetIdentityProviderIDForURLValues(t *testing.T) {
```
kenjenkins, line 175: Is it worth replacing this with a unit test for the new `Stateful.DecryptURLValues` method?

internal/authenticateflow/stateful.go
```diff
@@ -179,6 +193,281 @@ func (s *Stateful) SignIn(
 	return nil
 }
 
+func getSbrID(r *http.Request, state *sessions.Handle) string {
```
kenjenkins: naming nit: I think writing out `getSessionBindingRequestID()` might be more readable?

internal/authenticateflow/stateful.go
```diff
@@ -179,6 +193,281 @@ func (s *Stateful) SignIn(
...
+	query := r.URL.Query()
+	if query.Has(urlutil.QueryBindSession) {
+		sbrID := query.Get(urlutil.QueryBindSession)
+		if state == nil {
+			panic("bug: state missing")
+		}
```
kenjenkins: I'm a little confused by this. It looks like the `state` parameter is otherwise unused in this function? Is this more of a "validateRequest" method?

I'm also a little nervous around adding `panic`s. Would it be better to return an error from the methods that require a non-nil `state`?

internal/authenticateflow/stateful.go
```diff
@@ -179,6 +193,281 @@ func (s *Stateful) SignIn(
...
+	return query.Get("user_code")
+}
+
+func (s *Stateful) validateStateOrNew(
+	r *http.Request,
+	state *sessions.Handle,
+	sbrID string,
+	sbr *session.SessionBindingRequest,
+) (uri string, redirect bool, err error) {
+	if state != nil {
```
kenjenkins: I'm only seeing one call to `validateStateOrNew()`, and that is after a preceding call to `getSbrID()`. Given the `panic` above, is it even possible for `state` to be nil here?

internal/authenticateflow/stateful.go
```diff
@@ -349,13 +655,12 @@ func (s *Stateful) AuthenticateSignInURL(
...
 
-// GetIdentityProviderIDForURLValues returns the identity provider ID
-// associated with the given URL values.
-func (s *Stateful) GetIdentityProviderIDForURLValues(vs url.Values) string {
-	if id := vs.Get(urlutil.QueryIdentityProviderID); id != "" {
-		return id
+func (s *Stateful) DecryptURLValues(vs url.Values) (url.Values, error) {
+	vals := maps.Clone(vs)
+	if id := vs.Get(urlutil.QueryIdentityProviderID); id == "" {
+		vals.Set(urlutil.QueryIdentityProviderID, s.defaultIdentityProviderID)
```
kenjenkins: To me this feels like a slightly odd place to put this default logic. Would it be better to move this to the caller of `DecryptURLValues()`?

authenticate/state.go
```diff
@@ -33,7 +33,11 @@ type flow interface {
 	RevokeSession(ctx context.Context, r *http.Request, authenticator identity.Authenticator, h *sessions.Handle) string
 	GetUserInfoData(r *http.Request, h *sessions.Handle) handlers.UserInfoData
 	LogAuthenticateEvent(r *http.Request)
-	GetIdentityProviderIDForURLValues(url.Values) string
+	DecryptURLValues(url.Values) (url.Values, error)
```
kenjenkins: Can you help me understand the motivation for this change?

I thought maybe this was to allow decrypting URL values in other contexts, but I'm not seeing any new usage in this PR — unless I'm missing something?

config/testdata/basic.json
alexandreLamarre: It was causing pre-commit CI to fail for some reason... Let me see if I can get it to pass without running pre-commit

internal/authenticateflow/stateful.go
alexandreLamarre: Yeah I think a bad rebase re-introduced this logic, the state / redirect can/is handled by the authenticate middleware here, so I've removed this

internal/authenticateflow/stateful.go
alexandreLamarre: Same as below, I think a bad rebase re-introduced this logic, the state / redirect can/is handled by the authenticate middleware here, so I've removed this

authenticate/state.go
alexandreLamarre: I think I inherited this method change / url.Values handling from @kralicky's prototype. There's no inherent reason to introduce this IMO, besides the fact that it allows for some extensibility in the future, but yeah the new PR doesn't make use of any additional values so I'm happy to revert this

kenjenkins: My preference would be to revert this detail for now, to keep this PR focused on just the changes needed for the new flow.

pkg/ssh/ssh_int_test.go
```diff
@@ -64,6 +64,7 @@ type TemplateData struct {
 }
 
 func (s *SSHTestSuite) SetupSuite() {
+	s.T().Skip()
```
alexandreLamarre: Still working on the integration test changes

pkg/ssh/code/issuer.go
```diff
@@ -0,0 +1,226 @@
...
+
+	eg, ctxca := errgroup.WithContext(ctx)
+	// FIXME: this needs to run once everywhere we query SessionBindingRequest and SessionBinding's
+	// we want to avoid sharing a sync.Once for coordination across packages, and run this only once
+	// per pomerium instance.
+	go func() {
+		ctxT, ca := context.WithTimeout(ctxca, time.Minute*5)
+		defer ca()
+		if err := i.setup(ctxT); err != nil {
+			panic(err)
```
kenjenkins: Would it make sense to put the NewIssuer() initialization behind a runtime flag, at least for now? This makes me a little nervous.

alexandreLamarre: I changed this setup code just now to lazy load/ run only if one of the issuer methods runs, but even in a follow up PR pomerium.Run will exit with status 1 if this fails - it's very literally required on the storage backend for the ssh code to function. 

It will probably be decoupled from the issuer in the first follow up PR on this one.

ui/src/components/SessionBindingInfo.tsx
```diff
@@ -0,0 +1,68 @@
...
+                          <TableCell variant="head" align="left"> Session ID </TableCell>
+                          <TableCell variant="head" align="left"> Protocol </TableCell>
+                          <TableCell variant="head" align="left"> IssuedAt </TableCell>
+                          <TableCell variant="head" align="left"> ExpiresAt </TableCell>
+                          <TableCell variant="head" align="left"> Actions </TableCell>
+                        </TableRow>
+                      </TableHead>
+                      <TableBody>
+                        {data.sessions?.map((s) => (
+                            <TableRow>
```
nickytonline: refactor (blocking): add a `key` prop.


```suggestion
                            <TableRow key={s.SessionID}>
```

ui/src/components/SessionBindingInfo.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return (
+        <>
+            <TableContainer component={Paper} sx={{ maxWidth: 1000, mx: "auto", mb: 2 }}>
+                    <Table size="small" aria-label="metadata table">
+                      <TableHead>
+                        <TableRow>
+                          <TableCell variant="head" align="left"> Session ID </TableCell>
+                          <TableCell variant="head" align="left"> Protocol </TableCell>
+                          <TableCell variant="head" align="left"> IssuedAt </TableCell>
+                          <TableCell variant="head" align="left"> ExpiresAt </TableCell>
```
nickytonline: refactor (nit): remove white space

```suggestion
                          <TableCell variant="head" align="left">Protocol</TableCell>
                          <TableCell variant="head" align="left">IssuedAt</TableCell>
                          <TableCell variant="head" align="left">ExpiresAt</TableCell>
```

ui/src/components/SessionBindingInfo.tsx
```diff
@@ -0,0 +1,68 @@
...
+                            <TableCell align="left">{s.IssuedAt}</TableCell>
+                            <TableCell align="left">{s.ExpiresAt}</TableCell>
+                            <TableCell align="left">
+                              <Box
+                                component="form"
+                                action={s.RevokeURL}
+                                method="POST"
+                                sx={{ display: "inline-flex", gap: 1 }}
+                              >
+                                {/* <input type="hidden" name="_pomerium_csrf" value={data.csrfToken} /> */}
```
nickytonline: refactor (blocking): remove commented code.


```suggestion
```

ui/src/components/SignInSuccessPage.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+const SignInSuccessPage: FC<SignInSuccessPageProps> = ({ data }) => {
+  return (
+    <>
+      <TableContainer component={Paper} sx={{ maxWidth: 540, mx: "auto", mb: 2 }}>
+        <Table size="small" aria-label="metadata table">
+          <TableHead>
+            <TableRow>
+              <TableCell variant="head">Field</TableCell>
+              <TableCell variant="head" align="left">Value</TableCell>
```
nickytonline: refactor (blocking): remove all `align="left"`. This is the default.

ui/src/components/SignInVerifyPage.tsx
```diff
@@ -0,0 +1,153 @@
...
+  Box,
+  Checkbox,
+  Typography,
+} from "@mui/material";
+
+type SignInVerifyPageProps = {
+  data: SignInVerifyPageData;
+};
+const SignInVerifyPage: FC<SignInVerifyPageProps> = ({ data }) => {
+  const parseToDate = (v: any): Date => {
```
nickytonline: refactor (blocking): use `unknown` instead of `any`. See https://www.typescriptlang.org/docs/handbook/type-compatibility.html#any-unknown-object-void-undefined-null-and-never-assignability


```suggestion
  const parseToDate = (v: unknown): Date => {
```

ui/src/components/SignInVerifyPage.tsx
```diff
@@ -0,0 +1,153 @@
...
+    if (v instanceof Date && !isNaN(v.getTime())) return v;
+    if (typeof v === "number" && !isNaN(v)) return new Date(v);
+    if (typeof v === "string") {
+      const parsed = Date.parse(v);
+      if (!isNaN(parsed)) return new Date(parsed);
+    }
+    return new Date(Date.now() + 2 * 60 * 1000);
+  };
+
+  const expiresDate = parseToDate((data as any).expiresAt ?? data.expiresAt);
```
nickytonline: refactor (blocking): no need to cast `as any` or use `??`.


```suggestion
  const expiresDate = parseToDate(data.expiresAt);
```

PR #5888, opened 2025-10-16
endpoints: add paths

calebdoxsey:
## Summary
Add additional paths to the `endpoints` package.

## Checklist

- [x] updated unit tests
- [x] ready for review


pkg/endpoints/paths.go
```diff
@@ -2,5 +2,57 @@ package endpoints
 
 // well known paths
 const (
-	PathAuthenticateCallback = "/oauth2/callback"
+	PathAuthenticateCallback        = "/oauth2/callback"
```
desimone, line 5: I understand why we are using variables here to construct the paths, but I wish this file was more scannable / could be useful when trying to debug one of our hardcoded / well known paths.  I hate to leave a comment without a suggestion but that's the only thing that sticks out to me. It'd be great to see the raw strings here.

calebdoxsey: Updated with hardcoded strings.

PR #5891, opened 2025-10-18
controlplane: add debug server

calebdoxsey:
## Summary
Add a new setting `debug_address` that allows a user to expose the debug listener similar to the `metrics_address`. Add an index page and a `config_dump` endpoint.

The current behavior is to start the debug server on a random port only accessible to the local host. This is still the default behavior. Exposing the debug server is dangerous because there are sensitive values in the config.

## Related issues
- [ENG-3047](https://linear.app/pomerium/issue/ENG-3047/core-admin-panel)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/controlplane/server_debug.go
+++ b/internal/controlplane/server_debug.go
@@ -28,11 +28,13 @@ func (srv *debugServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 func (srv *debugServer) Update(cfg *config.Config) {
 	mux := http.NewServeMux()
 
-	// index
-	mux.HandleFunc("GET /", srv.indexHandler())
-
-	// config
-	mux.HandleFunc("GET /config_dump", srv.configDumpHandler(cfg))
+	// only enable admin endpoints if the runtime flag is set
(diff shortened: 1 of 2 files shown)


internal/controlplane/server_debug.go
```diff
@@ -0,0 +1,81 @@
...
+	mux.HandleFunc("GET /debug/pprof/", pprof.Index)
+	mux.HandleFunc("GET /debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("GET /debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("GET /debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("GET /debug/pprof/trace", pprof.Trace)
+
+	srv.mux.Store(mux)
+}
+
+func (srv *debugServer) configDumpHandler(cfg *config.Config) http.HandlerFunc {
```
kenjenkins, line 49: Could we put this behind a new runtime flag for now, with it off by default? I don't think I fully grasp the security impact, so I'd rather err on the side of caution for now.

kenjenkins (approved):
I like the debug_address option, that will certainly be useful!

PR #5896, opened 2025-10-21
feat: databroker index by record fields & query support

alexandreLamarre:
## Summary

Implements somewhat generic record field indexing on each backend. Query expressions support fetching records by these configurable indexes

## Related issues

[ENG-3057](https://linear.app/pomerium/issue/ENG-3057/databroker-indexing-records-options)
[ENG-3056](https://linear.app/pomerium/issue/ENG-3056/ssh-authorization-code-flow-implementation) / https://github.com/pomerium/pomerium/pull/5873

## User Explanation

Not an end-user facing change.

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/pkg/storage/indexer.go
+++ b/pkg/storage/indexer.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"cmp"
 	"fmt"
 	"hash/fnv"
 	"iter"
@@ -10,6 +11,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/google/btree"
 	"google.golang.org/protobuf/types/known/anypb"
 
 	"github.com/pomerium/pomerium/pkg/grpc/databroker"
@@ -179,6 +181,51 @@ func iterFromMapPtr(ptr *map[string]struct{}) iter.Seq[string] {
 		return func(_ func(string) bool) {}
 	}
 	return maps.Keys(*ptr)
+}
+
+type btreeIndexer struct {
+	accessMu *sync.RWMutex
+	btree    *btree.BTreeG[[2]string]
+}
+
+func NewBTreeIndexer(degree int) Indexer {
+	return &btreeIndexer{
+		accessMu: &sync.RWMutex{},
+		btree: btree.NewG(degree, func(a, b [2]string) bool {
+			return cmp.Or(
+				cmp.Compare(a[0], b[0]),
+				cmp.Compare(a[1], b[1]),
+			) < 0
+		}),
+	}
+}
+
+var _ Indexer = (*btreeIndexer)(nil)
+
+func (b *btreeIndexer) Put(secondaryKey string, recordID string) {
+	b.accessMu.Lock()
+	defer b.accessMu.Unlock()
+	b.btree.ReplaceOrInsert([2]string{secondaryKey, recordID})
+}
+
+func (b *btreeIndexer) Delete(secondaryKey string, recordID string) {
+	b.accessMu.Lock()
+	defer b.accessMu.Unlock()
+	b.btree.Delete([2]string{secondaryKey, recordID})
+}
+
+func (b *btreeIndexer) List(secondaryKey string) (recordIDs []string) {
+	b.accessMu.RLock()
+	defer b.accessMu.RUnlock()
+	recordIDs = []string{}
+	b.btree.AscendGreaterOrEqual([2]string{secondaryKey, ""}, func(item [2]string) bool {
+		if item[0] != secondaryKey {
+			return false
+		}
+		recordIDs = append(recordIDs, item[1])
+		return true
+	})
+	return recordIDs
 }
 
 type IndexManager struct {
--- a/pkg/storage/record_collection.go
+++ b/pkg/storage/record_collection.go
@@ -59,7 +59,7 @@ func NewRecordCollection() RecordCollection {
 		records:        make(map[string]recordCollectionNode),
 		insertionOrder: list.New(),
 		indexMgr: NewIndexManager(func() Indexer {
-			return NewFastIndexer(8)
+			return NewBTreeIndexer(2)
 		}),
 	}
 }
--- a/pkg/storage/indexer_test.go
+++ b/pkg/storage/indexer_test.go
@@ -8,9 +8,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestIndexerSimple(t *testing.T) {
-	t.Parallel()
-	idxer := NewSimpleIndexer()
+func testIndexerConformance(t *testing.T, idxer Indexer) {
+	t.Helper()
 	t.Run("crud", func(t *testing.T) {
 		testCRUD(t, idxer)
 	})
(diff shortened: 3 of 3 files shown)


pkg/storage/record_collection.go
```diff
@@ -35,6 +38,9 @@ type RecordCollection interface {
 	// Put puts a record into the collection. If the record's deleted at field is not nil, the record will
 	// be removed from the collection.
 	Put(record *databroker.Record)
+
+	// SetOptions sets databroker options that should be enforced by the collection
+	SetOptions(*databroker.Options)
```
alexandreLamarre, line 40: Temporary hack, until I refined the interface more

pkg/storage/index.go
```diff
@@ -40,6 +51,116 @@ func GetRecordIndex(msg proto.Message) *structpb.Struct {
...
+func GetIndexableFields(msg *anypb.Any, fields []string) (mappings map[string]string, err error) {
+	fieldTable := slices.Associate(fields, func(field string) (string, string) {
+		return field, ""
+	})
+	mapSt, err := anyToMap(msg)
+	if err != nil {
+		return nil, err
+	}
+
+	err = handleMap(mapSt, fieldTable, fields)
```
alexandreLamarre, line 99: @kralicky  @calebdoxsey 

This is not a great way to deref fields on protobuf messages encoded into an *anypb.Any. Maybe there is a better/more efficient way? I couldn't find any in the functions in the official google packages that would do the same thing

pkg/storage/index.go
```diff
@@ -40,6 +51,116 @@ func GetRecordIndex(msg proto.Message) *structpb.Struct {
...
+	out := make(map[string]any)
+	m.Range(func(fd protoreflect.FieldDescriptor, v protoreflect.Value) bool {
+		out[string(fd.Name())] = fieldToValue(fd, v)
+		return true
+	})
+	return out
+}
+
+func anyToMap(a *anypb.Any) (map[string]any, error) {
+	msg, err := a.UnmarshalNew()
```
alexandreLamarre, line 157: one idea I had  was here after we unmarshal the message we can check if the underlying proto message struct implements an interface like :
```go
type  Indexable Interface {
   IndexFieldMappings() map[string]string
}
```

to speed up extract field -> value pairs.

For example :
```go
func (s *Session) IndexFieldMappings() map[string]string {
   return map[string]string{
     "user_id" : s.UserId,
   }
}
```

If it doesn't, proceed to the generic case. Thoughts?

pkg/storage/indexer.go
alexandreLamarre, line 1: I spent some time here making a lock free thread-safe implementation for the in-memory indexer, because I thought it would be great if pebble had that one as well to speed up lookups. The simple index implementation allows for less concurrent operations, but still fast lookups (it's just a map[string]string), but comes with a greater memory impact that specialized datastructures like Btrees because of the pointer allocation under the hood.

I ended up not using an in-memory implementation for pebble, and I think it might be hard to see the (concurrent) speed up given by the lock-free implementation. I'm open to greatly simplifying this, because it's only applicable to the in-memory backend (which in my mind won't be scaling to the moon)...

<hr/>
Some other considerations:

- I ended up not going with your traditional Btree, arc tree, or skip list data structures that databases tend to use because they didn't deal well with `key -> mutable set{values}` problem... 
- The one data structure that might be a good idea to default here instead of the current ones is the Btree implementation since it will save on memory, at the cost of having to iterate over records instead of direct lookups.

pkg/storage/file/backend.go
```diff
@@ -708,6 +699,18 @@ func (backend *Backend) setOptionsLocked(
 	recordType string,
 	options *databrokerpb.Options,
 ) error {
+	existingOpts := backend.options[recordType]
+	if existingOpts != nil && len(existingOpts.GetIndexableFields()) > 0 {
```
calebdoxsey: If we added an indexable field, and then removed it, shouldn't we also delete the associated index values? So I think we'd either need to always delete.. or maybe just check if the indexed fields changed?

alexandreLamarre: oh yeah. I implemented methods on the keyspace to do exactly this, but ended up never using them... let me fix that

pkg/storage/indexer.go
calebdoxsey: I believe you could implement a btree index like this:

```go
type btreeIndexer struct {
	btree btree.BTreeG[[2]string]
}

func newBTreeIndexer(degree int) *btreeIndexer {
	return &btreeIndexer{
		btree: *btree.NewG(degree, func(a, b [2]string) bool {
			return cmp.Or(
				cmp.Compare(a[0], b[0]),
				cmp.Compare(a[1], b[1]),
			) < 0
		}),
	}
}

func (idx *btreeIndexer) Put(secondaryKey string, recordID string) {
	idx.btree.ReplaceOrInsert([2]string{secondaryKey, recordID})
}

func (idx *btreeIndexer) Delete(secondaryKey string, recordID string) {
	idx.btree.Delete([2]string{secondaryKey, recordID})
}

func (idx *btreeIndexer) List(secondaryKey string) (recordIDs []string) {
	recordIDs = []string{}
	idx.btree.AscendGreaterOrEqual([2]string{secondaryKey, ""}, func(item [2]string) bool {
		if item[0] != secondaryKey {
			return false
		}
		recordIDs = append(recordIDs, item[1])
		return true
	})
	return recordIDs
}
```

That's not thread-safe though and maybe performance is a lot worse 🤷 .

alexandreLamarre: This is probably the way to go, even if we wrap it with a RWMutex. 👍 

It was hard to get the "theoretical" speed of the lock-free impl into the backend implementation (because of how cleanly we/pebble wraps the RW transactions)

My feeling is it's probably better to use the BTree by default for the in-memory implementation

pkg/storage/index.go
alexandreLamarre: I think I can improve here slightly by traversing the proto-reflect path only for each field, instead of basically unmarshalling the whole struct...

pkg/storage/file/keyspaces.go
```diff
@@ -26,8 +26,119 @@ const (
 	prefixRecordKeySpace
 	prefixRecordIndexByTypeVersionKeySpace
 	prefixRegistryServiceKeySpace
+	prefixIndexableFieldsKeySpace
 )
 
+// indexable fields key space
+//
+//	keys: prefix-reference-graph | 0x00 |{record-type as bytes} | {from-field as bytes} | 0x00 | {from-value as bytes} | 0x00 | {record_id as bytes}
```
calebdoxsey: The prefix is a single byte so there's no `0x00` after it. But there should be one after the record type.

```suggestion
//	keys: prefix-reference-graph | {record-type as bytes}| 0x00 | {from-field as bytes} | 0x00 | {from-value as bytes} | 0x00 | {record_id as bytes}
```

pkg/storage/file/keyspaces.go
```diff
@@ -26,8 +26,119 @@ const (
...
+	it, err := r.NewIter(opts)
+	if err != nil {
+		return func(yield func(string, error) bool) {
+			yield("", err)
+		}
+	}
+	return func(yield func(string, error) bool) {
+		for ok := it.First(); ok; ok = it.Next() {
+			k := it.Key()
+			for i := len(k) - 1; i >= 0; i-- {
```
calebdoxsey: would it be possible to just use `bytes.LastIndexByte` for this?

pkg/storage/index.go
calebdoxsey: Not sure it's any better, but we have some similar code in pkg/logutil/scrub.go

alexandreLamarre: yeah it's similar, but a little different. I think since we know the "indexable" fields ahead of time, I can improve on this first pass by just traversing paths in the protoreflect .Range selectively, instead of doing the whole message

calebdoxsey (approved):

pkg/storage/indexer.go
alexandreLamarre: Based on my "ad-hoc" benchmarking + profiling,  I found that the btree was:
- reasonably similarly to the SimpleIndexer which is just a map[string]string - with the exception that it ends up allocating much less memory to do its operations and using less memory (50% of simple indexer), which is expected because that's what it is optimized to do. 
- it's actually faster at huge amounts of updates in a short period of time than the map[string]string{}, probably because it doesn't have to resize nearly as often
- lookups are slower, but I think this is OK. Lookups MIGHT be really slow if we reach millions of records being indexed, but at that point I don't think you really want to be using the in-memory backend

PR #5898, opened 2025-10-21
config: add resolvers to dns options

wasaga:
## Summary

Adds `dns_resolvers` core config option, that allows to provide a list of TCP or UDP DNS resolvers as a list of IP-only URLs like `udp://1.1.1.1:53` 

## Related issues

Ref: 
 - https://linear.app/pomerium/issue/ENG-3067

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/envoyconfig.go
+++ b/config/envoyconfig/envoyconfig.go
@@ -126,15 +126,14 @@ func buildIPAddressFromURL(src string) (*envoy_config_core_v3.Address, error) {
 	if err != nil {
 		return nil, fmt.Errorf("parse url: %w", err)
 	}
-	if u.Scheme != "tcp" && u.Scheme != "udp" {
-		return nil, fmt.Errorf("unsupported url scheme: %s", u.Scheme)
-	}
 	var protocol envoy_config_core_v3.SocketAddress_Protocol
 	switch u.Scheme {
 	case "tcp":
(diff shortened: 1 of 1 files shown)


calebdoxsey (approved):
LGTM. We will need protobuf types too, but could do that as a follow up.

agocs (approved):

config/envoyconfig/envoyconfig.go
```diff
@@ -121,6 +121,47 @@ func buildUDPAddress(hostport string, defaultPort uint32) *envoy_config_core_v3.
...
+		return nil, fmt.Errorf("must include an IP address: %s", src)
+	}
+	if _, err := netip.ParseAddr(host); err != nil {
+		return nil, fmt.Errorf("host must be an IP address: %s", host)
+	}
+	portStr := u.Port()
+	if portStr == "" {
+		return nil, fmt.Errorf("must include a port: %s", src)
+	}
+	port, err := strconv.ParseUint(portStr, 10, 16)
```
agocs, line 149: Nice, I was about to ask a question about ports > 65535 but you've handled it here.

calebdoxsey:
> > We will need protobuf types too, but could do that as a follow up.
> 
> do you see those settings configurable via enterprise console? if incorrectly set, that could brick core

This is true but we've generally exposed all config options in the protobuf even if they can't be set in the enterprise console.

PR #5905, opened 2025-10-29
ssh: port forward management

kralicky:
This adds SSH port forward management logic for upcoming reverse tunnel functionality.


pkg/ssh/portforward/manager.go
```diff
@@ -0,0 +1,331 @@
...
+//
+// When a dynamic forwarded-tcpip channel is closed, the virtual port is
+// released so that it can be reused in the future. Channels using static ports
+// use the same port for all requests, but if configuration is changed such that
+// a static port is no longer used by the server, any open channels which
+// previously requested to port-forward with that static port are closed. The
+// client may remain connected though, and if the configuration is changed to
+// re-enable the port, port-forwards will be once again be allowed using the
+// original permission; clients do not need to reconnect.
+type Manager struct {
```
kenjenkins, line 150: Just to make sure I understand— there will be one instance of Manager per connection, is that right? (not a global Manager for all connections?)

kralicky: Correct

PR #5907, opened 2025-10-29
test: add composite filter to storagetest filtering

alexandreLamarre:
## Summary

Adds a conformance test for backends to correctly handled composite filter expressions like:
(`b.Id == '1'` OR `b.id==2`) AND (`b.cidr` matches `192.168.0.1`)

using our storage filter language.

Fixes bugs in the memory backend where:
- a cidr index couldn't be matched to more than one record
- composite expressions were using pointer equality implicity with the `comparable` interface, making them not match anything since records are copied when matched

## Related issues

Part of https://github.com/pomerium/pomerium/pull/5896; I was pretty sure I broke something with querying in the pebble backend - and when the tests passed I dug a little deeper here... and found that the tests weren't testing composite ANDs + ORs

## User Explanation

N/A

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/pkg/storage/record_collection.go
+++ b/pkg/storage/record_collection.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"cmp"
 	"container/list"
 	"fmt"
 	"maps"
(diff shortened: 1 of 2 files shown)


pkg/storage/record_collection.go
```diff
@@ -258,46 +272,46 @@ func (c *recordCollection) lookupAddr(addr netip.Addr) []*databroker.Record {
...
 	}
 	return final
 }
 
-func union[T comparable](xs [][]T) []T {
-	var final []T
-	seen := set.New[T](0)
+func union(xs [][]*databroker.Record) []*databroker.Record {
+	var final []*databroker.Record
+	seen := set.New[string](0)
```
calebdoxsey: I think you could just use a `[2]string` as a key and then avoid having to do `recordToID(e)` but do `[2]string{e.GetType(),e.GetId()}` instead?

alexandreLamarre: TIL [2]string is a valid comparable, and doesn't just check for pointer equality

calebdoxsey (approved):

PR #5908, opened 2025-10-31
switch to npm

calebdoxsey:
## Summary
Yarn isn't really any better than NPM these days. So let's just use NPM.

## Related issues
- [ENG-2781](https://linear.app/pomerium/issue/ENG-2781/move-from-yarn-to-npm-for-repos-that-have-javascript-in-them)

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -60,6 +60,7 @@
     "importOrderSortSpecifiers": true
   },
   "engines": {
+    "node": ">=22.0.0",
     "yarn": "YARN NO LONGER USED - use npm instead."
   },
   "packageManager": "npm@10.9.4"
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -39,6 +39,7 @@
         "typescript": "^5.1.3"
       },
       "engines": {
+        "node": ">=22.0.0",
         "yarn": "YARN NO LONGER USED - use npm instead."
       }
     },


calebdoxsey:
Yeah the UI here is minimal so this is one of the easier projects to do. After this we can try upgrading some deps which are probably fairly ancient by now.

.github/workflows/test.yaml
```diff
@@ -28,8 +28,6 @@ jobs:
       - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444
         with:
           node-version: ${{ matrix.node-version }}
-          cache: yarn
-          cache-dependency-path: "**/yarn.lock"
```
nickytonline, line 32: refactor (blocking): add the npm cache, i.e. `cache: 'npm'`

.github/workflows/test.yaml
```diff
@@ -66,8 +64,6 @@ jobs:
       - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444
         with:
           node-version: ${{ matrix.node-version }}
-          cache: yarn
-          cache-dependency-path: ui/yarn.lock
```
nickytonline, line 70: refactor (blocking): add the npm cache, i.e. `cache: 'npm'`

ui/package.json
```diff
@@ -57,5 +58,9 @@
     ],
     "importOrderSeparation": true,
     "importOrderSortSpecifiers": true
-  }
+  },
+  "engines": {
+    "yarn": "YARN NO LONGER USED - use npm instead."
```
nickytonline, line 64: refactor (blocking): add a minimal node engine of 20, ideally 22 (current LTS)

nickytonline (changes requested):

.github/workflows/test.yaml
calebdoxsey: the latest setup-node actually does this by default now

PR #5915, opened 2025-11-04
ssh: initial implementation of reverse tunnel EDS

kralicky:
This implements the reverse tunnel Endpoint Discovery Service endpoint. 

There are still more tests to be written but those will be added in a follow-up.

### DIFF
--- a/pkg/ssh/manager.go
+++ b/pkg/ssh/manager.go
@@ -59,7 +59,7 @@ type activeStream struct {
 
 type StreamManager struct {
 	endpointv3.UnimplementedEndpointDiscoveryServiceServer
-	lg                 *zerolog.Logger
+	logger             *zerolog.Logger
 	auth               AuthInterface
 	reauthC            chan struct{}
 	initialSyncDone    bool
@@ -185,7 +185,7 @@ func (sm *StreamManager) SetSessionIDForStream(ctx context.Context, streamID uin
 
 func NewStreamManager(ctx context.Context, auth AuthInterface, cfg *config.Config) *StreamManager {
 	sm := &StreamManager{
-		lg:                   log.Ctx(ctx),
+		logger:               log.Ctx(ctx),
 		auth:                 auth,
 		waitForInitialSync:   make(chan struct{}),
 		reauthC:              make(chan struct{}, 1),
@@ -279,7 +279,7 @@ func (sm *StreamManager) onStreamHandlerClosed(streamID uint64) {
 	}
 
 	if len(info.Endpoints) > 0 {
-		sm.lg.Debug().
+		sm.logger.Debug().
 			Uint64("stream-id", streamID).
 			Any("endpoints", info.Endpoints).
 			Msg("clearing endpoints for closed stream")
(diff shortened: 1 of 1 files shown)


pkg/ssh/portforward/manager.go
```diff
@@ -3,7 +3,8 @@ package portforward
 import (
 	"context"
 	"errors"
-	"slices"
+	"maps"
+	stdslices "slices"
```
kenjenkins: Is there a reason this package needs to be renamed? I'm not seeing another "slices" in this file.

pkg/ssh/portforward/manager.go
```diff
@@ -155,7 +156,7 @@ type Manager struct {
 	ownedStaticPorts map[uint]context.CancelCauseFunc
 
 	cachedAuthorizedRoutes []RouteInfo
-	cachedEndpoints        []RoutePortForwardInfo
+	cachedEndpoints        map[string]RoutePortForwardInfo
```
kenjenkins: What is the map key here? Should we add a comment to mention this?

pkg/ssh/portforward/manager.go
```diff
@@ -178,14 +180,18 @@ func (pfm *Manager) AddUpdateListener(l UpdateListener) {
 	defer pfm.mu.Unlock()
 	pfm.updateListeners = append(pfm.updateListeners, l)
 	l.OnRoutesUpdated(pfm.cachedAuthorizedRoutes)
-	l.OnPermissionsUpdated(pfm.permissions)
-	l.OnClusterEndpointsUpdated(pfm.cachedEndpoints)
+	var entries []Permission
+	for p := range pfm.permissions.AllEntries() {
+		entries = append(entries, *p)
+	}
```
kenjenkins: I see this snippet repeated a few times. Would it make sense to add a method on the PermissionSet type for this, maybe something like this?
```
func (ps *PermissionSet) CloneEntries() []Permission {
	entries := make([]Permission, len(ps.entries))	
	for i, p := range ps.entries {
		entries[i] = *p
	}
	return entries
}
```

pkg/ssh/portforward/manager.go
kralicky: Oops this is old

pkg/ssh/portforward/manager.go
kralicky: Yes - this is the cluster ID, I will add a note here

pkg/ssh/portforward/manager.go
kralicky: Actually I think I can have `AllEntries` use `iter.Seq[Permission]` instead, that would be easier.

pkg/ssh/manager.go
```diff
@@ -1,26 +1,65 @@
...
-	Handler *StreamHandler
-	Session *string
+	Handler   *StreamHandler
+	Session   *string
+	Endpoints map[string]struct{}
 }
 
 type StreamManager struct {
+	endpointv3.UnimplementedEndpointDiscoveryServiceServer
+	lg                 *zerolog.Logger
```
kenjenkins: (optional) naming nit: I think we usually name a Logger field just `log` or `logger`

PR #5916, opened 2025-11-04
ssh: reverse tunnel cluster config

kralicky:
This adds logic to configure route clusters when the upstream_tunnel policy is set.

### DIFF
--- a/config/envoyconfig/listeners_test.go
+++ b/config/envoyconfig/listeners_test.go
@@ -104,7 +104,7 @@ func TestBuildListeners(t *testing.T) {
 							}]
 						}
 					}
-				}`, httpConfig.Get("httpFilters.7").String(),
+				}`, httpConfig.Get("httpFilters.8").String(),
 					"should add alt-svc header")
 			case "quic-ingress":
 				hasQUIC = true
--- a/config/envoyconfig/testdata/main_http_connection_manager_filter.json
+++ b/config/envoyconfig/testdata/main_http_connection_manager_filter.json
@@ -122,6 +122,41 @@
         }
       },
       {
+        "name": "envoy.filters.http.set_filter_state",
+        "typedConfig": {
+          "@type": "type.googleapis.com/envoy.extensions.filters.http.set_filter_state.v3.Config",
+          "onRequestHeaders": [
+            {
+              "formatString": {
+                "textFormatSource": {
+                  "inlineString": "%DOWNSTREAM_REMOTE_ADDRESS%"
+                }
+              },
+              "objectKey": "pomerium.extensions.ssh.downstream_source_address",
+              "sharedWithUpstream": "ONCE"
+            },
+            {
+              "formatString": {
+                "textFormatSource": {
+                  "inlineString": "%REQUESTED_SERVER_NAME%"
+                }
+              },
+              "objectKey": "pomerium.extensions.ssh.requested_server_name",
+              "sharedWithUpstream": "ONCE"
+            },
+            {
+              "formatString": {
+                "textFormatSource": {
+                  "inlineString": "%PATH(NQ)%"
+                }
+              },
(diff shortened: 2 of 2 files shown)


config/envoyconfig/clusters.go
```diff
@@ -243,6 +243,16 @@ func (b *Builder) buildPolicyCluster(ctx context.Context, cfg *config.Config, po
...
 	cluster.CircuitBreakers = buildRouteCircuitBreakers(cfg, policy)
 
+	// Keep last
+	if policy.UpstreamTunnel != nil {
+		if cfg.Options.IsRuntimeFlagSet(config.RuntimeFlagSSHUpstreamTunnel) {
+			configureUpstreamTunnelCluster(policy, cluster)
+		} else {
+			log.Ctx(ctx).Warn().
+				Msgf("upstream_tunnel support is disabled; enable it with the runtime flag %s",
+					config.RuntimeFlagSSHUpstreamTunnel)
```
kenjenkins, line 253: Will this warning show once for each cluster? Would it make sense to move it to Opitons.Validate(), so we could log it just once per configuration update?

config/envoyconfig/upstream_tunnel.go
```diff
@@ -0,0 +1,169 @@
...
+		TypedConfig: marshalAny(&envoy_extensions_transport_sockets_internal_upstream_v3.InternalUpstreamTransport{
+			TransportSocket: proto.Clone(socket).(*envoy_config_core_v3.TransportSocket),
+		}),
+	}
+	socket.Name = "envoy.transport_sockets.internal_upstream"
+	socket.ConfigType = wrappedTypedConfig
+}
+
+func SetConnectionStateFilter() *envoy_extensions_filters_network_http_connection_manager.HttpFilter {
+	// TODO: this would probably be better implemented via http PerFilterConfig
```
kenjenkins, line 108: Should we file an issue tracking this cleanup?

config/envoyconfig/upstream_tunnel.go
```diff
@@ -0,0 +1,169 @@
...
+
+	extensions_ssh "github.com/pomerium/envoy-custom/api/extensions/filters/network/ssh"
+	"github.com/pomerium/pomerium/config"
+)
+
+func configureUpstreamTunnelCluster(policy *config.Policy, cluster *envoy_config_cluster_v3.Cluster) {
+	cluster.UpstreamBindConfig = nil
+	protocolOptions := cluster.TypedExtensionProtocolOptions["envoy.extensions.upstreams.http.v3.HttpProtocolOptions"]
+	if protocolOptions == nil {
+		panic("bug: invalid cluster config: expected HttpProtocolOptions to be set")
```
kenjenkins, line 23: Will a panic here be recovered?

Would it be safer to let this method return an error, that we could propagate to the return value of `buildPolicyCluster()`?

config/envoyconfig/clusters.go
kralicky: Good idea

config/envoyconfig/upstream_tunnel.go
kralicky: Yes I'll add one

config/envoyconfig/upstream_tunnel.go
kralicky: This is intended to be a fatal error. This makes some assumptions about how the policy clusters are configured, so if that ever changes this code will need to be updated.

config/envoyconfig/clusters.go
kralicky: There isn't an obvious way to check this in Validate, and the runtime flag is supposed to be temporary anyway so I'll leave this warning here for now.

kenjenkins (approved):

PR #5921, opened 2025-11-07
ssh: basic reverse tunnel tui

kralicky:
This adds the initial structure and components for the reverse tunnel tui. Existing code for the routes portal was also moved to the new tui package as well.

### DIFF
--- a/pkg/ssh/cli.go
+++ b/pkg/ssh/cli.go
@@ -120,6 +120,11 @@ func (s *sshEnviron) Getenv(key string) string {
 
 var _ termenv.Environ = (*sshEnviron)(nil)
 
+const (
+	ptyWidthMax  = 512
+	ptyHeightMax = 512
+)
+
 func (cli *CLI) AddTunnelCommand(ctrl ChannelControlInterface) {
 	cli.AddCommand(&cobra.Command{
 		Use:    "tunnel",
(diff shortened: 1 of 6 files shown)


pkg/ssh/cli.go
```diff
@@ -121,53 +176,35 @@ func (cli *CLI) AddPortalCommand(ctrl ChannelControlInterface) {
...
-			l.SetFilteringEnabled(false)
-			l.Styles.Title = titleStyle
-			l.Styles.PaginationStyle = paginationStyle
-			l.Styles.HelpStyle = helpStyle
-
-			cli.SendTeaMsg(tea.WindowSizeMsg{Width: int(cli.ptyInfo.WidthColumns), Height: int(cli.ptyInfo.HeightRows)})
-			close(cli.msgQueue)
-
-			cli.tui = tea.NewProgram(model{list: l},
+			prog := tui.NewPortalProgram(cmd.Context(), routes, int(cli.ptyInfo.WidthColumns-2), int(cli.ptyInfo.HeightRows-2),
```
wasaga: if a client would send us width/height as 1/0 we will overflow and allocate huge buffer here ?

pkg/ssh/cli.go
```diff
@@ -101,6 +100,62 @@ func (cli *CLI) AddWhoamiCommand(ctrl ChannelControlInterface) {
...
+			prog := tui.NewTunnelStatusProgram(cmd.Context(),
+				tea.WithInput(cli.stdin),
+				tea.WithOutput(termenv.NewOutput(cli.stdout, termenv.WithEnvironment(env), termenv.WithUnsafe())),
+				tea.WithEnvironment(env.Environ()),
+			)
+			cli.tui = prog.Program
+
+			go func() {
+				cli.SendTeaMsg(tea.WindowSizeMsg{Width: int(cli.ptyInfo.WidthColumns), Height: int(cli.ptyInfo.HeightRows)})
+				ctrl.AddPortForwardUpdateListener(prog)
```
wasaga, line 157: will it be eventually garbage collected ?

pkg/ssh/cli.go
kralicky: Yes good catch, there are a few places where we need to clamp when converting from ptyInfo here.

pkg/ssh/cli.go
kralicky: This is tied to the lifetime of the connection, and once the tui exits it is a no-op to send update messages to it. However, it's probably a good idea to explicitly remove this when the cli ends since you can close the tui without disconnecting.

PR #5933, opened 2025-11-17
fix: standardize resolving sessions

alexandreLamarre:
## Related issues

[ENG-3172](https://linear.app/pomerium/issue/ENG-3172/session-binding-expiry-not-handled-correctly-in-all-cases)

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/authorize/ssh_grpc.go
+++ b/authorize/ssh_grpc.go
@@ -95,6 +95,13 @@ func (a *Authorize) ServeChannel(stream extensions_ssh.StreamManagement_ServeCha
 func (a *Authorize) EvaluateSSH(ctx context.Context, streamID uint64, req *ssh.Request) (*evaluator.Result, error) {
 	ctx = a.withQuerierForCheckRequest(ctx)
 
+	sessionID := ""
+	// this checks to make sure if the binding ID was never set,
+	// then the criteria will return un-authenticated
+	if req.SessionBindingID != "" && req.SessionID != "" {
+		sessionID = req.SessionID
+	}
+
 	evalreq := evaluator.Request{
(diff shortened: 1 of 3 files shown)


authorize/ssh_grpc.go
```diff
@@ -95,6 +95,13 @@ func (a *Authorize) ServeChannel(stream extensions_ssh.StreamManagement_ServeCha
 func (a *Authorize) EvaluateSSH(ctx context.Context, streamID uint64, req *ssh.Request) (*evaluator.Result, error) {
 	ctx = a.withQuerierForCheckRequest(ctx)
 
+	sessionID := ""
+	// this checks to make sure if the binding ID was never set,
+	// then the criteria will return un-autenticated
```
kralicky:
```suggestion
	// then the criteria will return un-authenticated
```

pkg/ssh/auth.go
```diff
@@ -95,9 +95,14 @@ func (a *Auth) handlePublicKeyMethodRequest(
 ) (PublicKeyAuthMethodResponse, error) {
 	sessionID, err := a.resolveSessionIDFromFingerprint(ctx, req.PublicKeyFingerprintSha256)
 	if err != nil {
-		if !databroker.IsNotFound(err) {
-			return PublicKeyAuthMethodResponse{}, err
+		st, ok := status.FromError(err)
+		if ok && (st.Code() == codes.NotFound || st.Code() == codes.InvalidArgument) {
```
kralicky: InvalidArgument is too likely to cause regressions here IMO (since it is much more common), we should stick with NotFound only and change the expiration error to be NotFound.

PR #5936, opened 2025-11-18
authorize: allow overload EvaluateSSH func

wasaga:
## Summary

Allows programmatic option to overload `authorize.EvaluateSSH` function.
 
## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/authorize/ssh_grpc.go
+++ b/authorize/ssh_grpc.go
@@ -99,6 +99,13 @@ func (a *Authorize) EvaluateSSH(ctx context.Context, streamID uint64, req *ssh.R
 
 	ctx = a.withQuerierForCheckRequest(ctx)
 
+	sessionID := ""
+	// this checks to make sure if the binding ID was never set,
+	// then the criteria will return un-authenticated
+	if req.SessionBindingID != "" && req.SessionID != "" {
+		sessionID = req.SessionID
+	}
+
 	evalreq := evaluator.Request{
 		HTTP: evaluator.RequestHTTP{
 			Hostname: req.Hostname,
@@ -109,7 +116,7 @@ func (a *Authorize) EvaluateSSH(ctx context.Context, streamID uint64, req *ssh.R
 			PublicKey: req.PublicKey,
 		},
 		Session: evaluator.RequestSession{
-			ID: req.SessionID,
+			ID: sessionID,
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


scripts/check-component-versions
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/bin/bash
```
wasaga, line 1: mac

authorize/authorize.go
```diff
@@ -44,10 +47,22 @@ type Authorize struct {
...
 	outboundGrpcConn grpc.CachedOutboundGRPClientConn
+
+	// customSSHEvaluate allows overriding the default SSH evaluation logic
+	customSSHEvaluate SSHEvaluateFunc
+}
+
+type Option func(*Authorize)
+
+// WithSSHEvaluateFunc sets a custom SSH evaluation function that will be used instead of the default implementation.
+func WithSSHEvaluateFunc(fn SSHEvaluateFunc) Option {
```
kralicky, line 58: Overriding this with an option is fine, but I think we should swap out the whole Evaluator instead

authorize/ssh_grpc.go
```diff
@@ -123,8 +127,12 @@ func (a *Authorize) EvaluateSSH(ctx context.Context, streamID uint64, req *ssh.R
 	}
 
 	if req.UseUpstreamTunnelPolicy {
-		// XXX: temporary stub
-		log.Ctx(ctx).Debug().Msg("evaluating upstream tunnel policy")
+		if req.PortForwardPolicy != nil {
+			log.Ctx(ctx).Debug().
+				Str("policy_from", req.PortForwardPolicy.From).
+				Msg("evaluating upstream tunnel policy")
+			evalreq.Policy = req.PortForwardPolicy
```
kralicky, line 134: I don't think this has any effect, the policy object is the same as the default

wasaga: see https://github.com/pomerium/pomerium/blob/8f373929a958fa421b830706b58c66868aae91b4/config/options.go#L1040 

it would only find a policy if its schema is `ssh`, which is not true for tunnel routes

kralicky: Not necessarily, ssh routes can also be tunnel routes

PR #5937, opened 2025-11-18
feat (ssh) : override ssh evaluator logic

alexandreLamarre:
## Related issues

https://github.com/pomerium/pomerium/pull/5936

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/pkg/grpc/databroker/client.go
+++ b/pkg/grpc/databroker/client.go
@@ -1,5 +1,14 @@
 package databroker
 
+import (
+	"context"
+
+	grpc "google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	status "google.golang.org/grpc/status"
+	emptypb "google.golang.org/protobuf/types/known/emptypb"
+)
+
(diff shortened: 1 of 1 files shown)


pkg/grpc/databroker/client.go
```diff
@@ -17,3 +26,90 @@ type staticClientGetter struct {
...
+func GetDatabrokerClient(ctx context.Context) DataBrokerServiceClient {
+	c, ok := ctx.Value(databrokerClientKey{}).(DataBrokerServiceClient)
+	if !ok {
+		c = &nilClient{
+			defaultErr: status.Error(codes.Unavailable, "not yet initialized"),
+		}
+	}
+	return c
+}
+
```
alexandreLamarre, line 45: @wasaga what do you think about injecting the databroker client into a context, that way we don't have to couple the interface and overrides to include the databroker client directly in signatures?

Just a thought experiment on my part, but maybe this will work fine

PR #5940, opened 2025-11-19
ssh: show an error if login not enabled

kenjenkins:
## Summary

Do not generate an ssh login link if we know that login is not supported. Instead, show an error message to the user.

## Related issues

https://linear.app/pomerium/issue/ENG-3200/core-do-not-generate-ssh-login-url-for-stateless-authenticate-flow

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/ssh/auth.go
+++ b/pkg/ssh/auth.go
@@ -237,7 +237,7 @@ func (a *Auth) handleLogin(
 ) error {
 	cfg := a.currentConfig.Load()
 	if cfg.Options.UseStatelessAuthenticateFlow() {
-		return errors.New("ssh login is not currently enabled")
+		return status.Error(codes.FailedPrecondition, "ssh login is not currently enabled")
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/ssh/auth.go
```diff
@@ -235,6 +235,11 @@ func (a *Auth) handleLogin(
 	publicKeyFingerprint []byte,
 	querier KeyboardInteractiveQuerier,
 ) error {
+	cfg := a.currentConfig.Load()
+	if cfg.Options.UseStatelessAuthenticateFlow() {
+		return errors.New("ssh login is not currently enabled")
```
alexandreLamarre: This will be `codes.Internal` error by default. Perhaps we want `codes.FailedPrecondition` / `codes.Unimplemented` ? I'm not sure what makes the most sense

kenjenkins: Sure, I think `FailedPrecondition` would fit here.

PR #5943, opened 2025-11-20
feat(ssh): add single setup databroker indexing

alexandreLamarre:
## Summary

The indexing on databroker records will happen server-side, so no client api calls will have to be synchronized across pomerium services.

## Related issues

[ENG-3139](https://linear.app/pomerium/issue/ENG-3139/ssh-authorization-code-flow-wait-for-init-single-setup-for-databroker)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/databroker/server_backend.go
+++ b/internal/databroker/server_backend.go
@@ -584,7 +584,7 @@ func (srv *backendServer) getBackend(ctx context.Context) (backend storage.Backe
 		backend = srv.backend
 		var err error
 		if backend == nil {
-			backend, err = srv.newBackendLocked(ctx)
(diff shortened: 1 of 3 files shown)


pkg/cmd/pomerium/pomerium.go
alexandreLamarre, line 1: I don't like changing all these contructor type contracts...

I'm wondering if instead of changing the constructor type contracts, it might be preferable to embed the initializer on config / config.Options, although I know we want to avoid making that stateful-y

Instead, we could also follow the Options pattern to avoid changing all the construct type contracts...

Maybe there's a cleaner alternative I'm not thinking of?

PR #5945, opened 2025-11-21
authorize: ssh upstream policy support

kenjenkins:
## Summary

Add a field to the UpstreamTunnel config struct accepting a PPL policy to control which ssh connections can serve as upstreams for a route.

Update the policy evaluator to store a separate upstream policy query. When building a policy evaluator, prepare both the existing queries and this new upstream policy query.

Add a field to the RequestSSH struct to signal whether to evaluate the regular policy or the upstream policy.

Refactor the existing EvaluatePortForward and EvaluateRoute methods to accept a `*config.Policy`, for compatibility with the policy evaluator setup.

## Related issues

https://linear.app/pomerium/issue/ENG-2955/enforce-upstreamtunnel-policy-on-reverse-tunnel-attachments

## Checklist

- [x] updated unit tests
- [ ] ready for review

### DIFF
--- a/authorize/ssh_grpc.go
+++ b/authorize/ssh_grpc.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"io"
 
+	"github.com/rs/zerolog"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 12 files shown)


authorize/ssh_grpc.go
```diff
@@ -140,26 +121,67 @@ func (a *Authorize) EvaluateSSH(ctx context.Context, streamID uint64, req ssh.Au
...
+	evalreq := baseEvaluatorRequestFromSSHRequest(req)
+	evalreq.Policy = route
+
+	res, err := a.state.Load().evaluator.Evaluate(ctx, evalreq)
+	if err != nil {
+		log.Ctx(ctx).Error().Err(err).Msg("EvaluateUpstreamTunnel: error during OPA evaluation")
+		return nil, err
+	}
+
+	a.fetchSessionAndLogAuthorizeCheck(ctx, req.SessionID, evalreq, res)
```
kralicky: I think we should log these decisions differently (maybe at trace level, and in such a way that this isn't called at all if it is disabled). It could be hard to differentiate between these and downstream requests, and could be very noisy if there are a lot of routes. Failing an authorize check here is also not really a noteworthy event like it would be normally.

kralicky: We could also skip this entirely, since there will be logs emitted elsewhere if a stream's list of authorized routes changes based on these results.

kenjenkins: I think it's still important to log these checks. If the policy is defined in PPL I would expect to see all the usual allow/deny reason strings in the log output.

I've added a `level` parameter to the logAuthorizeCheck() method. The downstream checks will still be logged at the info level, while I've changed these new checks to be logged at the debug level instead.

I have a small refactor in a separate PR (https://github.com/pomerium/pomerium/pull/5956) that will avoid the step of querying for a User record if the log message is not going to be written.

kralicky: Sounds good! Thanks

PR #5947, opened 2025-11-24
feat/fix(databroker): add options to sync latest stream

alexandreLamarre:
## Summary

The SyncLatest / Sync stream used to not sync databroker record options. However with the introduction of databroker record indexing, which is required for SSH authorization code flow, options should be synced across replicas in clustered mode.

## Related issues

[ENG-3234](https://linear.app/pomerium/issue/ENG-3234/databroker-clustered-mode-does-not-sync-options)

## User Explanation

N/A

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/databroker/server_clustered_follower.go
+++ b/internal/databroker/server_clustered_follower.go
@@ -284,11 +284,10 @@ func (srv *clusteredFollowerServer) sync(ctx context.Context, b backoff.BackOff,
 	// 	 - put the options
 	ch1 := make(chan clusteredFollowerServerBatchStepPayload, 1)
 	ch2 := make(chan clusteredFollowerServerPutStepPayload, 1)
-	ch3 := make(chan *databrokerpb.TypedOptions, 1)
 	eg, ctx := errgroup.WithContext(ctx)
 	eg.Go(func() error { defer close(ch1); return srv.syncStep(ctx, b, ch1, wait) })
-	eg.Go(func() error { defer close(ch2); defer close(ch3); return srv.batchStep(ctx, ch1, ch2, ch3) })
-	eg.Go(func() error { return srv.putStep(ctx, ch2, ch3) })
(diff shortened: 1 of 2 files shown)


internal/databroker/server_backend.go
```diff
@@ -461,17 +462,45 @@ func (srv *backendServer) Sync(req *databrokerpb.SyncRequest, stream databrokerp
...
+	_, err, ok := next()
+	if !ok {
+		return status.Error(codes.Internal, "sync never returned a first message")
+	}
+	if err != nil {
+		return err
+	}
+	// FIXME: this only syncs the databroker options once per stream...
+	// We need to either periodically poll and compute changes or implement SyncOptions()
+	// on storage backends.
```
alexandreLamarre, line 478: WDYT? 

I can always follow up in a separate PR for this one, but in practice for how we use the Options API I don't think it currently matters much to do a "near real-time" sync of options

calebdoxsey: Sure that's fine. Since we set the options when we start I don't think it will matter for now.

internal/databroker/server_clustered_follower.go
calebdoxsey, line 602: could we just add the options to the `PutStepPayload` instead of using separate channels?

pkg/storage/inmemory/backend.go
```diff
@@ -52,11 +54,12 @@ type Backend struct {
...
 
-	mu       sync.RWMutex
-	lookup   map[string]storage.RecordCollection
-	capacity map[string]*uint64
-	changes  *btree.BTree
-	leases   map[string]*lease
+	mu            sync.RWMutex
+	lookup        map[string]storage.RecordCollection
+	lookupOptions map[string]*databroker.Options
+	capacity      map[string]*uint64
```
calebdoxsey: do we need `capacity` as a separate map? Or is it easier to just use `lookupOptions` for that too?

PR #5948, opened 2025-11-24
ssh: policy index api + in-memory implementation

kralicky:
This adds a new PolicyIndexer interface, which is responsible for processing updates to configuration, session records, and active ssh streams. All active streams can subscribe to receive incremental updates as state changes. 

This allows reverse tunnel routes to scale, since each reverse tunnel connection does not need to independently compute its own set of authorized upstream tunnel routes. The implementation can be swapped out to support different environments.

An in-memory implementation is included, as well as a conformance test suite that can be run against any implementation of the PolicyIndexer interface to validate that it works as expected in a variety of scenarios.

### DIFF
--- a/pkg/ssh/ssh_export_test.go
+++ b/pkg/ssh/ssh_export_test.go
@@ -32,6 +32,6 @@ func (sm *StreamManager) UnexportedWaitForInitialSync(ctx context.Context) error
 	return sm.waitForInitialSync(ctx)
 }
 
-func (i *InMemoryPolicyIndexer) UnexportedState() *inMemoryIndexerState { //nolint:unexported-return
+func (i *InMemoryPolicyIndexer) UnexportedState() *inMemoryIndexerState { //revive:disable-line:unexported-return
 	return &i.state
 }


pkg/ssh/manager.go
```diff
@@ -183,8 +206,14 @@ func (sm *StreamManager) UpdateRecords(ctx context.Context, _ uint64, records []
 	defer sm.mu.Unlock()
 	for _, record := range records {
 		if record.DeletedAt == nil {
+			var s session.Session
+			if err := record.Data.UnmarshalTo(&s); err != nil {
+				panic(err.Error())
```
alexandreLamarre: Probably better to log an error and continue

pkg/ssh/stream.go
```diff
@@ -509,9 +507,9 @@ func (sh *StreamHandler) handleAuthRequest(ctx context.Context, req *extensions_
 		// methods have a valid response in the state
 		sh.state.InitialAuthComplete = true
 		// Initialize the port forward manager
-		sh.portForwards = portforward.NewManager(ctx, sh)
-		sh.portForwards.OnConfigUpdate(sh.config)
-		sh.portForwards.AddUpdateListener(sh)
+		// sh.portForwards = portforward.NewManager(ctx, sh)
+		// sh.portForwards.OnConfigUpdate(sh.config)
+		// sh.portForwards.AddUpdateListener(sh)
```
alexandreLamarre: Commented out?

pkg/ssh/policy_index.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+type PolicyIndexer interface {
+	ProcessConfigUpdate(cfg *config.Config)
+	OnStreamAuthenticated(streamID uint64, req AuthRequest)
+	OnSessionCreated(session *session.Session)
+	OnSessionDeleted(sessionID string)
+	AddStream(streamID uint64, sub PolicyIndexSubscriber)
+	RemoveStream(streamID uint64, sub PolicyIndexSubscriber)
+	Shutdown()
```
alexandreLamarre: I think some of these interface methods would benefit from comments

pkg/ssh/manager.go
```diff
@@ -217,30 +246,33 @@ func (sm *StreamManager) updateRecordsBinding(ctx context.Context, _ uint64, rec
...
+	if sm.bindingStreams[req.SessionBindingID] == nil {
+		sm.bindingStreams[req.SessionBindingID] = map[uint64]struct{}{}
+	}
+	sm.sessionStreams[req.SessionID][streamID] = struct{}{}
+	sm.bindingStreams[req.SessionBindingID][streamID] = struct{}{}
+
+	activeStream.Session = new(string)
+	*activeStream.Session = req.SessionID
+	activeStream.SessionBindingID = new(string)
+	*activeStream.SessionBindingID = req.SessionBindingID
```
alexandreLamarre, line 273: is there a reason not to directly indirect `&req.Session` and `&req.SessionBindingID`

pkg/ssh/manager.go
```diff
@@ -217,30 +246,33 @@ func (sm *StreamManager) updateRecordsBinding(ctx context.Context, _ uint64, rec
...
+	if err := sm.waitForInitialSync(ctx); err != nil {
+		return err
 	}
-	defer sm.mu.Unlock()
-	if sm.sessionStreams[sessionID] == nil {
-		sm.sessionStreams[sessionID] = map[uint64]struct{}{}
+	activeStream := sm.activeStreams[streamID]
+	if activeStream.Session != nil || activeStream.SessionBindingID != nil {
+		return status.Errorf(codes.Internal, "stream %d already has an associated session", streamID)
 	}
```
alexandreLamarre, line 259: I think this while handling the TODO from EvaluateSSH
> TODO: only do this once, not on re-evaluate

Might cause bugs when there are multiple binding IDs associated to a session.

I think there's a good chance I'm reading the code changes here wrong, but will verify

pkg/ssh/policy_index.go
wasaga: can we remove Shutdown() method and instead add Run(context.Context) instead (which is what current implementation follows)

pkg/ssh/manager.go
kralicky: This is just a hedge against possible future bugs. I'd prefer not to store pointers to interior fields since the parameter was recently refactored from `*AuthRequest` to `AuthRequest` and since I don't remember exactly why it was a pointer to begin with, it is possible that it may change back to a pointer in the future, which could introduce a subtle bug here if we point to actual struct fields.

pkg/ssh/manager.go
kralicky: Hmm, could you be more specific (like order of operations)? I can write a test to verify this.

pkg/ssh/manager.go
kralicky: Yeah probably a good idea

pkg/ssh/policy_index.go
kralicky: Yeah I don't really like needing Shutdown() here either, it was mainly used for tests but I can try to find a better way. `Run` might be weird to have in the interface though. How were you thinking of using it?

kralicky: Added some comments, lmk if they look good

alexandreLamarre: Yeah excellent, super clear thanks

pkg/ssh/manager.go
alexandreLamarre: I think we don't currently have tests with multiple ssh client keys associated with the same session ID, if it isn't too much trouble to add in this PR - otherwise I'll follow up in a PR since it's more so to do with SSH auth code flow in a sense

kralicky: There is only one tracked Session and SessionBindingID per active stream, would that need to change?

alexandreLamarre: no

alexandreLamarre: yeah this shouldn't cause problems never mind

pkg/ssh/manager.go
```diff
@@ -204,43 +235,44 @@ func (sm *StreamManager) updateRecordsBinding(ctx context.Context, _ uint64, rec
...
-	sm.activeStreams[streamID].SessionBindingID = &sessionBindingID
+	sm.sessionStreams[req.SessionID][streamID] = struct{}{}
+	sm.bindingStreams[req.SessionBindingID][streamID] = struct{}{}
+
+	activeStream.Session = new(string)
+	*activeStream.Session = req.SessionID
+	activeStream.SessionBindingID = new(string)
+	*activeStream.SessionBindingID = req.SessionBindingID
+
+	activeStream.PortForwardManager.AddUpdateListener(activeStream.Handler)
```
wasaga, line 275: we hold lock ni `sm.waitForInitialSync` but we do not lock here, why?

pkg/ssh/policy_index_inmemory.go
```diff
@@ -0,0 +1,381 @@
...
+				lg := log.Ctx(ctx).With().Uint64("stream-id", event.streamID).Logger()
+				if stream, ok := i.state.KnownStreams[event.streamID]; ok {
+					lg.Debug().Msg("policy indexer: tracked stream removed")
+					if stream.SessionID != "" {
+						session := i.state.KnownSessions[stream.SessionID] // must be non-nil
+						if len(i.state.EnabledStaticPorts) > 0 {
+							// Note: if static ports were removed after having been enabled
+							// previously, we would have already notified this subscriber
+							// at the time the ports were removed.
+							stream.Subscriber.UpdateEnabledStaticPorts(nil)
```
wasaga, line 228: add check the subscriber is not nil

pkg/ssh/policy_index_inmemory.go
```diff
@@ -0,0 +1,381 @@
...
+	AllTunnelEnabledRoutes []knownRoute
+}
+
+func (i *InMemoryPolicyIndexer) recomputeSessionAuthorizedRoutes(ctx context.Context, session *knownSession) {
+	var authorizedRoutes []knownRoute
+	if session.Record != nil && session.AuthRequest != nil {
+		authorizedRoutes = make([]knownRoute, 0, len(session.AuthorizedRoutes))
+		for _, route := range i.state.AllTunnelEnabledRoutes {
+			result, err := i.evaluator.EvaluateUpstreamTunnel(ctx, *session.AuthRequest, route.Route)
+			if err == nil && result.Allow.Value && !result.Deny.Value {
```
wasaga: log error?

pkg/ssh/policy_index_inmemory.go
```diff
@@ -0,0 +1,381 @@
...
+					for _, stream := range i.state.KnownStreams {
+						if stream.SessionID != "" && stream.Subscriber != nil {
+							stream.Subscriber.UpdateEnabledStaticPorts(i.state.EnabledStaticPorts)
+						}
+					}
+				}
+
+				clear(i.state.AllTunnelEnabledRoutes)
+				i.state.AllTunnelEnabledRoutes = i.state.AllTunnelEnabledRoutes[:0]
+				for route := range options.GetAllPolicies() {
```
wasaga:
```suggestion
				i.state.AllTunnelEnabledRoutes = i.state.AllTunnelEnabledRoutes[:0]
				for route := range options.GetAllPolicies() {
```

pkg/ssh/policy_index_inmemory.go
```diff
@@ -0,0 +1,381 @@
...
+				if session.AuthRequest == nil {
+					lg.Debug().Msg("policy indexer: computing session authorized routes")
+					// If the session is not known from a previous stream, compute its
+					// authorized routes now. We don't need to do this if e.g. the same
+					// user disconnects and reconnects with the same (valid) session.
+					reqCopy := event.authRequest
+					session.AuthRequest = &reqCopy
+					i.recomputeSessionAuthorizedRoutes(ctx, session)
+				} else if *session.AuthRequest != event.authRequest {
+					panic("bug: inconsistent session auth state")
```
wasaga, line 180: log & error, dont panic

pkg/ssh/manager.go
kralicky: Good catch, this is missing a lock.

pkg/ssh/policy_index_inmemory.go
kralicky: If this ever happens, something has gone very wrong, i.e. the same stream was authenticated twice with different users.

pkg/ssh/manager.go
```diff
@@ -289,9 +325,29 @@ func NewStreamManager(ctx context.Context, auth AuthInterface, cfg *config.Confi
 	}
 
 	sm.bindingSyncer = bindingSyncer
+
+	go sm.indexer.ProcessConfigUpdate(cfg)
```
wasaga: flagging for potential race condition in config versions delivery

pkg/ssh/manager.go
```diff
@@ -333,13 +389,9 @@ func (sm *StreamManager) Run(ctx context.Context) error {
 }
 
 func (sm *StreamManager) OnConfigChange(cfg *config.Config) {
-	sm.mu.Lock()
-	sm.cfg = cfg
-	for _, s := range sm.activeStreams {
-		s.Handler.portForwards.OnConfigUpdate(cfg)
-	}
-	sm.mu.Unlock()
+	go sm.indexer.ProcessConfigUpdate(cfg)
```
wasaga: flagging for potential race condition in config versions delivery

pkg/ssh/manager.go
wasaga: as agreed, it would be responsibility of the consumer to not block

kralicky: We should make all the callbacks synchronous then no?

PR #5955, opened 2025-12-02
envoy: store version to avoid importing files package

calebdoxsey:
## Summary
When we import the `files` package under `envoy` it requires the envoy files to have been downloaded since they aren't stored in git. This ended up making any package which imports pomerium also have to download the files and be built with the `embed_pomerium` tag.

Instead we can store the version of envoy we download separately and only import that, which should make it possible to import pomerium without requiring the build tag.

## Checklist

- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/envoy/envoyversion/envoyversion.go
+++ b/pkg/envoy/envoyversion/envoyversion.go
@@ -1,4 +1,38 @@
 package envoyversion
 
-// Version is the envoy version
-const Version = "1.36.2-rc1"
+import (
+	"runtime/debug"
+	"strings"
+
+	"golang.org/x/mod/module"
+
+	_ "github.com/pomerium/envoy-custom/api/extensions/filters/network/ssh" // to get version info
+)
+
+// Version returns the envoy version.
+func Version() string {
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return ""
+	}
+
+	var modVersion string
+	for _, dep := range info.Deps {
+		if dep.Path == "github.com/pomerium/envoy-custom" {
+			if dep.Replace != nil {
+				modVersion = dep.Replace.Version
+			} else {
+				modVersion = dep.Version
+			}
+			break
+		}
+	}
+
+	if module.IsPseudoVersion(modVersion) {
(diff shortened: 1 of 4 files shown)


pkg/envoy/envoyversion/envoyversion.go
```diff
@@ -0,0 +1,4 @@
+package envoyversion
+
+// Version is the envoy version
+const Version = "1.36.2-rc1"
```
kralicky: Let's implement this by reading the build info. this should return the version in the go.mod and strip the pseudo-version if any (e.g. `v1.36.2-rc1.0.20251120234524-51bf1e57025a` -> `v1.36.2-rc1`)
```go
package envoyversion

import (
	"runtime/debug"
	"strings"

	_ "github.com/pomerium/envoy-custom/api/extensions/filters/network/ssh"
	"golang.org/x/mod/module"
)

func GetEnvoyVersion() string {
	info, ok := debug.ReadBuildInfo()
	if !ok {
		return ""
	}
	var modVersion string
	for _, dep := range info.Deps {
		if dep.Path == "github.com/pomerium/envoy-custom" {
			if dep.Replace != nil {
				modVersion = dep.Replace.Version
			} else {
				modVersion = dep.Version
			}
			break
		}
	}
	if module.IsPseudoVersion(modVersion) {
		if base, err := module.PseudoVersionBase(modVersion); err == nil {
			modVersion = base
		}
	}
	return strings.TrimPrefix(modVersion, "v")
}

```

kralicky: It needs to import some path from envoy-custom, but I can add a blank main.go in the top level folder so we don't have to import an arbitrary package

calebdoxsey: `get-envoy` is using this version info to download the actual files. It wouldn't support arbitrary commits.

kralicky: Right that's why we can strip out the commit info, PseudoVersionBase gives you the tag part of the version.

PR #5957, opened 2025-12-02
feat(ssh): ssh RLS filter and grpc implementation

alexandreLamarre:
## Summary

See related issue for context.

Adds a grpc implementation of the RLS service. It is attached to the ssh listener, so it can act on incoming SSH requests before they are authenticated.

Here is a description of the service : https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ratelimit/v3/rls.proto#envoy-v3-api-msg-service-ratelimit-v3-ratelimitresponse.

There are some dynamic access log fields available on the ssh listener
https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log, but for the ssh listener the only one with relevant information on the downstream connection was `%DOWNSTREAM_DIRECT_REMOTE_ADDRESS_WITHOUT_PORT%`

Example config to enable:
```yaml
ssh_rls_enabled : true
ssh_rls_additonal_entries:
  - ["foo", "bar"]
```

There is an example implementation by the envoy maintainers here : https://github.com/envoyproxy/ratelimit.

<hr/>

This implementation is also (maybe temporarily) attached to the authorize evaluator so that any of the go ssh-code can make very specific requests to the RLS implementation.  Each hook should have its own `Domain` to uniquely identify the hook to differentiate it from requests being made from the network filter and from each other.

Each domain should be handled in a unique manner by the underlying implementation.

## Related issues

[ENG-3058](https://linear.app/pomerium/issue/ENG-3058/ssh-rate-limit-incoming-pending-sessions)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/envoyconfig/listeners_ssh.go
+++ b/config/envoyconfig/listeners_ssh.go
@@ -104,10 +104,10 @@ func buildSSHListener(cfg *config.Config) (*envoy_config_listener_v3.Listener, e
 	filters := []*envoy_config_listener_v3.Filter{}
 
 	if cfg.Options.SSHRLSEnabled {
-		additionalEntries := slices.Map(cfg.Options.SSHRLSAdditonalEntries, func(el [2]string) *envoy_common_ratelimit_v3.RateLimitDescriptor_Entry {
+		additionalEntries := slices.Map(cfg.Options.SSHRLSAdditonalEntries, func(el config.GenericKeyVal) *envoy_common_ratelimit_v3.RateLimitDescriptor_Entry {
 			return &envoy_common_ratelimit_v3.RateLimitDescriptor_Entry{
-				Key:   el[0],
-				Value: el[1],
+				Key:   el.Key,
+				Value: el.Value,
 			}
 		})
 
--- a/config/options.go
+++ b/config/options.go
@@ -256,7 +256,7 @@ type Options struct {
 	SSHRLSEnabled bool `mapstructure:"ssh_rls_enabled" yaml:"ssh_rls_enabled,omitempty"`
 	// SSHRLSAdditonalEntries Specifies [2]{Key, Value} pairs of RLS entries
 	// https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/rate_limit_filter#substitution-formatting
-	SSHRLSAdditonalEntries [][2]string `mapstructure:"ssh_rls_additional_entries" yaml:"ssh_rls_additional_entries"`
+	SSHRLSAdditonalEntries []GenericKeyVal `mapstructure:"ssh_rls_additional_entries" yaml:"ssh_rls_additional_entries"`
 	// DownstreamMTLS holds all downstream mTLS settings.
 	DownstreamMTLS DownstreamMTLSSettings `mapstructure:"downstream_mtls" yaml:"downstream_mtls,omitempty"`
 
@@ -310,6 +310,11 @@ type certificateFilePair struct {
 	// CertFile and KeyFile is the x509 certificate used to hydrate TLSCertificate
 	CertFile string `mapstructure:"cert" yaml:"cert,omitempty"`
 	KeyFile  string `mapstructure:"key" yaml:"key,omitempty"`
+}
+
+type GenericKeyVal struct {
+	Key   string `mapstructure:"key" yaml:"key,omitempty"`
+	Value string `mapstructure:"value" yaml:"value,omitempty"`
 }
 
 // DefaultOptions are the default configuration options for pomerium
--- a/pkg/ssh/ratelimit/ratelimit.go
+++ b/pkg/ssh/ratelimit/ratelimit.go
@@ -13,8 +13,6 @@ const (
 	// DomainSSHInbound is the domain corresponding to incoming requests before they are authenticated
 	// and must go through the bespoke oauth code flow
 	DomainSSHInbound = "ssh-inbound"
-	// DomainSSHInFlight is the domain corresponding to requests made during active ssh connections
-	DomainSSHInFlight = "ssh-inflight"
 )
 
 var _ envoy_service_ratelimit_v3.RateLimitServiceServer = (*RateLimiter)(nil)


pkg/envoy/envoy.go
```diff
@@ -273,7 +273,8 @@ func (srv *Server) run(ctx context.Context, cfg *config.Config) error {
 
 	args := []string{
 		"-c", configFileName,
-		"--log-level", srv.logLevel.ToEnvoy(),
+		"--log-level", "debug",
+		"--component-log-level", "filter:debug",
```
wasaga: please restore use of configured log level

config/options.go
```diff
@@ -252,7 +252,13 @@ type Options struct {
 	// String contents of SSH key used to sign ephemeral certificate keys for
 	// upstream authentication. Mutually exclusive with ssh_user_ca_key_file.
 	SSHUserCAKey string `mapstructure:"ssh_user_ca_key" yaml:"ssh_user_ca_key,omitempty"`
-
+	// SSHRLSEnabled Enable the RLS service for ssh connections
+	SSHRLSEnabled bool `mapstructure:"ssh_rls_enabled" yaml:"ssh_rls_enabled,omitempty"`
+	// SSHRLSRemoteAddress An (optional) remote address of an RLS server to be called from the SSH RLS server
+	SSHRLSRemoteAddress string `mapstructure:"ssh_rls_remote_address" yaml:"ssh_rls_remote_address,omitempty"`
```
wasaga: it doesn't look like this field is accessed elsewhere

alexandreLamarre: oh yeah good catch from a previous implementation.

config/envoyconfig/listeners_ssh.go
```diff
@@ -79,40 +102,83 @@ func buildSSHListener(cfg *config.Config) (*envoy_config_listener_v3.Listener, e
...
+				GrpcService: &envoy_config_core_v3.GrpcService{
+					TargetSpecifier: &envoy_config_core_v3.GrpcService_EnvoyGrpc_{
+						EnvoyGrpc: &envoy_config_core_v3.GrpcService_EnvoyGrpc{
+							ClusterName: "pomerium-ssh-ratelimit",
+						},
+					},
+				},
+			},
+		}
+		rlsCfg, err := anypb.New(rl)
```
kralicky: Can use the helper `marshalAny` here; if this fails, it should panic

config/envoyconfig/clusters.go
```diff
@@ -121,6 +121,15 @@ func (b *Builder) BuildClusters(ctx context.Context, cfg *config.Config) ([]*env
 		envoyAdminCluster,
 	}
 
+	if cfg.Options.SSHRLSEnabled {
+		log.Ctx(ctx).Info().Msg("enabling RLS rate limit for ssh address")
+		cluster, err := b.buildInternalCluster(ctx, cfg, "pomerium-ssh-ratelimit", grpcURLs, upstreamProtocolHTTP2, Keepalive(false))
```
kralicky: Would it be better to reuse pomerium-control-plane-grpc for this?

internal/testenv/environment.go
```diff
@@ -661,10 +664,11 @@ func (e *environment) Start() {
 	e.ports.ALPN.Resolve(atoi(ports[11]))
 	cfg.AllocatePorts(*(*[6]string)(ports[6:]))
 
+	cfg.Options.SSHRLSEnabled = true
 	cfg.Options.AutocertOptions = config.AutocertOptions{Enable: false}
 	cfg.Options.Services = "all"
 	cfg.Options.LogLevel = config.LogLevelDebug
-	cfg.Options.ProxyLogLevel = config.LogLevelInfo
+	cfg.Options.ProxyLogLevel = config.LogLevelDebug
```
kralicky: Let's keep this info for now

config/options.go
```diff
@@ -252,7 +252,11 @@ type Options struct {
 	// String contents of SSH key used to sign ephemeral certificate keys for
 	// upstream authentication. Mutually exclusive with ssh_user_ca_key_file.
 	SSHUserCAKey string `mapstructure:"ssh_user_ca_key" yaml:"ssh_user_ca_key,omitempty"`
-
+	// SSHRLSEnabled Enable the RLS service for ssh connections
+	SSHRLSEnabled bool `mapstructure:"ssh_rls_enabled" yaml:"ssh_rls_enabled,omitempty"`
+	// SSHRLSAdditonalEntries Specifies [2]{Key, Value} pairs of RLS entries
+	// https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/rate_limit_filter#substitution-formatting
+	SSHRLSAdditonalEntries [][2]string `mapstructure:"ssh_rls_additional_entries" yaml:"ssh_rls_additional_entries"`
```
wasaga:
```suggestion
	SSHRLSAdditionalEntries [][2]string `mapstructure:"ssh_rls_additional_entries" yaml:"ssh_rls_additional_entries"`
```

authorize/authorize.go
```diff
@@ -91,7 +103,8 @@ func New(ctx context.Context, cfg *config.Config, opts ...Option) (*Authorize, e
 		return nil, err
 	}
 	a.state.Store(state)
-
+	rls := ratelimit.NewRateLimiter(trace.NewTracerProvider(ctx, "RLS"), o.rls)
```
wasaga, line 106: why not reuse `tracerProvider`?

internal/testenv/environment.go
```diff
@@ -661,6 +664,7 @@ func (e *environment) Start() {
 	e.ports.ALPN.Resolve(atoi(ports[11]))
 	cfg.AllocatePorts(*(*[6]string)(ports[6:]))
 
+	cfg.Options.SSHRLSEnabled = true
```
wasaga, line 667: do we need it enabled for 100% of tests?

config/options.go
wasaga: [][2]string is not very readable in code and in config file; did you consider making it an explicit struct of {Key,Value string} and then doing a slice of structs?

pkg/ssh/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/pomerium/pomerium/pkg/telemetry/trace"
+)
+
+const (
+	// DomainSSHInbound is the domain corresponding to incoming requests before they are authenticated
+	// and must go through the bespoke oauth code flow
+	DomainSSHInbound = "ssh-inbound"
+	// DomainSSHInFlight is the domain corresponding to requests made during active ssh connections
+	DomainSSHInFlight = "ssh-inflight"
```
wasaga: unused?

authorize/authorize.go
alexandreLamarre: because it's scoped to 	
```
tracerProvider := trace.NewTracerProvider(ctx,"Authorize")`
```
which is accurate from a code location perspective but in the chain of events, the RLS filter happens before any authorize code ever runs, I though it might be more clear if it were it's own separate service from the trace perspective

internal/testenv/environment.go
alexandreLamarre: no although I have to come up with a way to inject startup config values:
```go
func (e *environment) Start() {
	//...

	cfg := &config.Config{
		Options: config.NewDefaultOptions(),
	}

     // 
```
which might be non trivial since there's no good mechanism at the moment. It should only impact ssh integration tests at the moment, but I can create a backlog ticket to improve that if that's alright.

config/options.go
alexandreLamarre: changed
```yaml
ssh_rls_additonal_entries:
  - ["foo", "bar"]
```

to:
```yaml
ssh_rls_additonal_entries:
  - key: "foo"
    value: "bar"
```

internal/testenv/environment.go
wasaga: lets defer that as a follow-up with @kralicky

wasaga (approved):

PR #5963, opened 2025-12-03
feat(ssh/UI) SSH session binding UI improvements

alexandreLamarre:
## Related issues

Requires https://github.com/pomerium/pomerium/pull/5961
[ENG-3135](https://linear.app/pomerium/issue/ENG-3135/ssh-auth-code-flow-ui-tuneups)
[ENG-3179](https://linear.app/pomerium/issue/ENG-3179/sshauth-ui-session-lifetime-info-not-accurate)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/authenticateflow/stateful.go
+++ b/internal/authenticateflow/stateful.go
@@ -8,6 +8,8 @@ import (
 	"maps"
 	"net/http"
 	"net/url"
+	"slices"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/authenticateflow/stateful.go
```diff
@@ -383,7 +394,8 @@ func (s *Stateful) GetSessionBindingInfo(w http.ResponseWriter, r *http.Request,
 
 	renderData := []handlers.SessionBindingData{}
 
-	for sessionBindingID, p := range pairs {
+	for sessionBindingID := range iterutil.SortedUnion(strings.Compare, maps.Keys(pairs)) {
```
calebdoxsey: I think `maps.Keys` doesn't return the keys in sorted order, so `SortedUnion` wouldn't behave properly

calebdoxsey: Oh it's only a single sequence? Maybe just use `slices.Sorted` then?

alexandreLamarre: oh yeah I see. something like a `slices.Collect` + `sort.Strings` would probably work better. Let me change a couple places where I used this pattern mistakenly

alexandreLamarre: I think I had some tests in https://github.com/pomerium/pomerium/pull/5961 where wrongly I assumed those functions would chain together

calebdoxsey (approved):

PR #5967, opened 2025-12-03
pebble: fix secure fs wrapper

calebdoxsey:
## Summary
Fix the secure file system wrapper for pebble. The arguments passed to `ReuseForWrite` were misinterpreted. It should've been `(oldname, newname)`, but we were expecting `(newname, oldname)`.

Because the `chmod` failed and the call to `ReuseForWrite` returned an error, pebble would call `logger.Fatalf` with the error. Since our logger wrapper just ignored `Fatalf`, this resulted in a confusing panic "fatal error: sync: unlock of unlocked mutex". (the code expected that `Fatal` would exit the function immediately, so the subsuqent unlock should never have been called)

So now the logger is setup to actually log things, except `Info` will be sent to `Debug` instead, since these logs are pretty noisy.

No AI was used for this PR.

## Related issues
- [ENG-3266](https://linear.app/pomerium/issue/ENG-3266/corepebble-fatal-error-sync-unlock-of-unlocked-mutex)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/pebbleutil/pebbleutil.go
+++ b/pkg/pebbleutil/pebbleutil.go
@@ -84,7 +84,7 @@ func MustOpenMemory(options *pebble.Options) *pebble.DB {
 	if options == nil {
 		options = new(pebble.Options)
 	}
-	options.FS = NewSecureFS(vfs.Default)
+	options.FS = vfs.NewMem()
 	return MustOpen("", options)
 }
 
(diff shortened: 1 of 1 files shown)


agocs (approved):
smart move renaming `name` to `newname`

PR #5973, opened 2025-12-05
feat(ssh): add auth code flow telemetry signals

alexandreLamarre:
## Summary

Adds useful sub-spans, structured logs, and metrics for ssh auth code flow.

## Related issues

[ENG-3134](https://linear.app/pomerium/issue/ENG-3134/ssh-auth-code-flow-telemetry-logs-metrics-traces)
Indirectly related to [ENG-3058](https://linear.app/pomerium/issue/ENG-3058/ssh-rate-limit-incoming-pending-sessions)

## User Explanation

There is now structured telemetry around SSH authorization code flow.

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/ssh/auth.go
+++ b/pkg/ssh/auth.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/rs/zerolog"
-	"go.opentelemetry.io/otel/attribute"
 	otelcode "go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
 	noopmetric "go.opentelemetry.io/otel/metric/noop"
@@ -135,9 +134,12 @@ func (a *Auth) HandlePublicKeyMethodRequest(
 	info StreamAuthInfo,
 	req *extensions_ssh.PublicKeyMethodRequest,
 ) (PublicKeyAuthMethodResponse, error) {
+	ctx, span := a.tracer.Start(ctx, "authorize.ssh.HandlePublicKeyMethodRequest")
+	defer span.End()
 	resp, err := a.handlePublicKeyMethodRequest(ctx, info, req)
 	if err != nil {
 		log.Ctx(ctx).Error().Err(err).Msg("ssh publickey auth request error")
+		span.SetStatus(otelcode.Error, "internal error")
 		return resp, status.Error(codes.Internal, "internal error")
 	}
 	return resp, err
@@ -157,10 +159,10 @@ func (a *Auth) handlePublicKeyMethodRequest(
 ) (PublicKeyAuthMethodResponse, error) {
 	ctx, span := a.tracer.Start(ctx, "authorize.ssh.handlePublicKeyMethodRequest")
 	defer span.End()
-	span.SetAttributes(attribute.String(telemetryFingerprintAttribute, fingerprintAsStrAttribute(req.PublicKeyFingerprintSha256)))
 	sessionID, err := a.resolveSessionIDFromFingerprint(ctx, req.PublicKeyFingerprintSha256)
 	if err != nil {
 		if st, ok := status.FromError(err); ok && st.Code() == codes.NotFound {
+			span.SetStatus(otelcode.Ok, "must reauthenticated")
 			return PublicKeyAuthMethodResponse{
 				Allow:                    publicKeyAllowResponse(req.PublicKey),
 				RequireAdditionalMethods: []string{MethodKeyboardInteractive},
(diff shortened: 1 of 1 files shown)


pkg/ssh/auth.go
```diff
@@ -201,6 +264,13 @@ func (a *Auth) handleKeyboardInteractiveMethodRequest(
 	info StreamAuthInfo,
 	querier KeyboardInteractiveQuerier,
 ) (KeyboardInteractiveAuthMethodResponse, error) {
+	fingerprintAttrVal := fingerprintAsStrAttribute(info.PublicKeyFingerprintSha256)
+	ctx, span := a.tracer.Start(ctx, "authorize.ssh.handleKeyboardInteractiveMethodRequest")
+	defer span.End()
+	span.SetAttributes(
+		attribute.String(telemetryFingerprintAttribute, fingerprintAttrVal),
```
agocs: Are you concerned about how adding a high cardinality attribute will affect the memory footprint of Pomerium? I once more-than-doubled the memory footprint of a Go service by adding a high cardinality attribute.

pkg/ssh/code/metrics.go
```diff
@@ -0,0 +1,110 @@
...
+		SSHAuthCodeRequestsTotal:        codeReqTotal,
+		SSHAuthCodeRequestFailuresTotal: codeRequestFailuresTotal,
+		SSHActivePendingSessions:        activePendingSessions,
+		SSHIssueCodeDuration:            issueCodeDuration,
+		SSHUserCodeDecisionDuration:     userCodeDecision,
+	}, nil
+}
+
+func (m *Metrics) PendingSessionInc() {
+	m.SSHActivePendingSessions.Add(context.TODO(), 1)
```
agocs: Intentionally left as context.TODO?

pkg/ssh/auth.go
alexandreLamarre: That's a good point, I'm not sure how the in-memory footprint of core will change with traces

alexandreLamarre: I'm not even sure if we configure sampling in the tracer provider or if that's an exercise left to the end-user (inside a collector), I'll look into this

pkg/ssh/code/metrics.go
alexandreLamarre: yeah, the context used in the stream can be cancelled or timeout, so I left all the spots where we use these metrics as `context.TODO()` since we don't have a parent context available in the authorize service besides context.Background

pkg/ssh/auth.go
```diff
@@ -234,19 +302,33 @@ func (a *Auth) handleLogin(
...
 		return status.Error(codes.FailedPrecondition, "ssh login is not currently enabled")
 	}
 
+	l := log.Ctx(ctx).With().
+		Str("protocol", "ssh").
+		Str("source-addr", sourceAddr).
+		Str(telemetryFingerprintAttribute, fingerprintAsStrAttribute(publicKeyFingerprint)).
+		Logger()
+
+	a.codeMetrics.SSHAuthCodeRequestsTotal.Add(context.TODO(), 1)
```
kenjenkins: Can you help me understand the usage of context.TODO() here? This method already has a context parameter. If I'm reading https://github.com/open-telemetry/opentelemetry-go/pull/4671 correctly, a canceled context will not prevent the metric from being incremented.

alexandreLamarre: oh I see. I guess I've been using the SDK for so long I didn't realize they changed that behavior. I can update all the context's then 👍

pkg/ssh/code/metrics.go
alexandreLamarre: Addressed as part of https://github.com/pomerium/pomerium/pull/5973#discussion_r2599581181

pkg/ssh/auth.go
alexandreLamarre: Yeah without any kind of SDK side sampling, this could have a pretty substantial impact, I wrote a very very imperfect benchmark in `pkg/telemetry/trace/provider_test.go` which shows that the memory footprint has at least a (likely) linear relationship with attribute cardinality

alexandreLamarre: Removed the granular attributes, and added some slightly informative status code messages based on login outcomes. My experience is that these traces are not super helpful while debugging if there's not at least some amount of internal information being surfaced, but I'm open to just leaving the spans as bare-bones as possible which is what we do in most other services

pkg/telemetry/trace/provider_test.go
```diff
@@ -0,0 +1,87 @@
...
+		n,
+		end.HeapAlloc-start.HeapAlloc,
+		end.HeapInuse-start.HeapInuse,
+		end.HeapObjects-start.HeapObjects,
+		end.Frees-start.Frees,
+	)
+	b.Logf("ratio (in use / samples) : %f", float64(end.HeapInuse-start.HeapInuse)/float64(numSamples))
+}
+
+func BenchmarkCardinality(b *testing.B) {
```
agocs, line 59: I love a good benchmark

agocs (approved):
This is good stuff

PR #5979, opened 2025-12-09
ssh: refactor in-memory policy indexer

kralicky:
This refactors some of the internals of the in-memory policy indexer to properly account for multiple streams authorized to the same session, but with different auth requests. Within a single session, upstream tunnel authorized routes are evaluated and cached separately for each unique AuthRequest within that session. If multiple streams connect and have the same parameters in the AuthRequest, the cached routes will be shared as expected.

There is a limit to how many different sets of authorized routes will be cached (currently 5). If more streams with unique AuthRequests connect, the oldest cached entries will be evicted.

### DIFF
--- a/pkg/ssh/test/policy_index_suite.go
+++ b/pkg/ssh/test/policy_index_suite.go
@@ -821,16 +821,16 @@ func (s *PolicyIndexConformanceSuite[T]) TestMultipleAuthRequestsForSession() {
 
 	subs := make([]*mock_ssh.MockPolicyIndexSubscriber, NumAuthRequests)
 	for i := range uint64(NumAuthRequests) {
+		select {
+		case <-s.wait:
+			return
+		default:
+		}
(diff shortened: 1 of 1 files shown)


pkg/ssh/policy_index_inmemory.go
```diff
@@ -81,41 +253,59 @@ type inMemoryIndexerState struct {
 	AllTunnelEnabledRoutes []knownRoute
 }
 
-func (i *InMemoryPolicyIndexer) recomputeSessionAuthorizedRoutes(ctx context.Context, session *knownSession) {
-	var authorizedRoutes []knownRoute
-	if session.Record != nil && session.AuthRequest != nil {
-		authorizedRoutes = make([]knownRoute, 0, len(session.AuthorizedRoutes))
+const MaxCachedAuthRequestsPerStream = 5
```
wasaga: rename as its per-session

PR #5980, opened 2025-12-10
ssh: add internal cli controller abstraction

kralicky:
This adds support for customizing the behavior of the internal CLI.


pkg/ssh/cli.go
```diff
@@ -51,57 +63,71 @@ func NewCLI(
...
+func (cli *internalCLI) Stdout() io.Writer {
+	return cli.stdout
+}
+
+// SendTeaMsg implements InternalCLI.
+func (cli *internalCLI) SendTeaMsg(msg tea.Msg) {
+	select {
+	case <-cli.programDone:
+	case cli.msgQueue <- msg:
+	}
```
wasaga, line 109: consider adding `default` clause to ensure this function is non-blocking

kralicky: Dropping messages might lead to very confusing bugs in some cases. If we run into any issues here we can revisit but this is intended to block

PR #5982, opened 2025-12-10
ci: add nodejs to tool versions

calebdoxsey:
## Summary
Add nodejs to tool versions and update github actions to use it. Also fix the go version. asdf doesn't support `.x` syntax.

## Related issues
- [ENG-3294](https://linear.app/pomerium/issue/ENG-3294/ci-use-tool-versions-for-node)

## Checklist

- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/.github/workflows/update-dependencies.yaml
+++ b/.github/workflows/update-dependencies.yaml
@@ -1,7 +1,6 @@
 name: Update Dependencies
 
 on:
-  pull_request:
   schedule:
     - cron: "40 1 * * *"
   workflow_dispatch:


.tool-versions
```diff
@@ -1,2 +1,3 @@
 golangci-lint 2.7.2
-golang 1.25.x
+golang 1.25.5
```
agocs, line 2: You might want to bump `FROM golang:1.25.4-bookworm` in the Dockerfile as well. I haven't been able to figure out a good way to keep those in sync.

kenjenkins, line 2: Is this worth a wider discussion with the team? Or maybe just a heads up.

When we stuck to 1.25.x in the setup-go action, we'd automatically get any security patches from Go itself when we cut a new release. If we specify the patch number too, we'll need to remember to bump this version whenever there is a new Go patch release.

Personally, I could see an argument either way. On the one hand it's convenient not to need to update this as frequently. On the other hand, this change makes the CI build more hermetic.

calebdoxsey: Maybe we can get dependabot to do those updates daily?

calebdoxsey: Doesn't look like dependabot has support for it. As an alternative we have a nightly update script in most repos. We could update it to bump the go version too if there's a new one?

The downside of the `.x` approach was that it just got picked up incidentally, so if there were no commits on a repo, we wouldn't rebuild anyway I don't think.

kenjenkins: > The downside of the .x approach was that it just got picked up incidentally, so if there were no commits on a repo, we wouldn't rebuild anyway I don't think.

Yes, that's a good point. Maybe an update script is a good idea.

PR #5996, opened 2025-12-18
feat(health) : stream health updates via gRPC

alexandreLamarre:
## Summary

Implements a grpc stream that batches updates from the health manager abstraction that currently manages health check integrations.

This allows downstream applications with the proper authorization (e.g. with the shared secret) to consume the health updates of pomerium instances.

## Related issues

Part of [ENG-3312](https://linear.app/pomerium/issue/ENG-3312/console-show-cluster-health-check)

## User Explanation

End users can now use `pomerium health watch -p <pomerium-grpc-port>` to open a TUI that displays health checks 

## Checklist

- [X] updated unit tests
- [X] ready for review


pkg/grpc/health/health.proto
```diff
@@ -0,0 +1,39 @@
...
+  Running     = 1;
+  Terminating = 2;
+}
+
+enum OverallStatus {
+  StatusUnknown     = 0;
+  StatusStarting    = 1;
+  StatusRunning     = 2;
+  StatusTerminating = 3;
+}
```
calebdoxsey, line 24: Convention is:


```suggestion
enum HealthStatus {
  HEALTH_STATUS_UNKNOWN     = 0;
  HEALTH_STATUS_RUNNING     = 1;
  HEALTH_STATUS_TERMINATING = 2;
}

enum OverallStatus {
  OVERALL_STATUS_UNKNOWN     = 0;
  OVERALL_STATUS_STARTING    = 1;
  OVERALL_STATUS_RUNNING     = 2;
  OVERALL_STATUS_TERMINATING = 3;
}
```

pkg/grpc/health/health.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+
+package health;
+
+import "google/protobuf/empty.proto";
+
+option go_package = "github.com/pomerium/pomerium/pkg/grpc/health";
+
+service HealthNotifier {
+  rpc SyncHealth(google.protobuf.Empty) returns (stream HealthMessage);
```
calebdoxsey: I think it's generally considered safer to make an empty type here instead of `google.protobuf.Empty` in case we ever want to add fields.

pkg/health/grpc_stream.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+}
+
+func (g *GRPCStreamProvider) SyncHealth(_ *emptypb.Empty, server grpc.ServerStreamingServer[healthpb.HealthMessage]) error {
+	ctx := server.Context()
+	t := time.NewTicker(time.Second * 90)
+	defer t.Stop()
+	// always send computed state on connect
+	id := uuid.New().String()
+	l := log.With().Str("stream-id", id).Logger()
```
calebdoxsey: You could add this to the current context logger. Something like:

```
ctx = log.Ctx(server.Context()).With().Str("stream-id", id).WithContext(ctx)
```

And then just use `log.Ctx(ctx).Debug().Msg("sending initial health message")` etc

pkg/health/grpc_stream.go
```diff
@@ -0,0 +1,259 @@
...
+
+	sp := &GRPCStreamProvider{
+		tr:             tr,
+		requiredChecks: slices.Collect(maps.Keys(defaultOpts.expected)),
+		batchWindow:    batchWindow,
+		batch:          make(chan struct{}, 1),
+		subscribers:    make(map[string]chan struct{}),
+		subscribersMu:  &sync.RWMutex{},
+	}
+	eg, ctx := errgroup.WithContext(parentCtx)
```
calebdoxsey: what's the purpose of the errgroup here? It seems to only start one goroutine?

alexandreLamarre: oh yeah sorry that's left over from a previous iteration, let me clean this up. I was working on some changes last Friday / this morning for gracefully closing remote connections

PR #6004, opened 2025-12-23
ssh: reverse tunnel status tui improvements

kralicky:
This PR includes significant changes to the structure and configuration of the reverse tunnel status TUI, as well as new components and functionality.

The TUI is much more configurable than it was before, although the current configuration is such that it doesn't look much different.

High level changes:
- Theme and colors are now fully configurable and all styling is derived from the theme instead of hard-coded. 
- Size/layout/presence of components is now configurable within the CLI controller.
- New persistent header bar with highly configurable segments
- UI elements have been organized into a widget library with consistent code style
- New highly configurable context menus and dialogs
- Support for inline editing of table data
- Basic user preference management system
- Support for active route health checks
- Copy-to-clipboard support (with terminal capability detection)

### DIFF
--- a/pkg/ssh/models/item_model_test.go
+++ b/pkg/ssh/models/item_model_test.go
@@ -4,10 +4,11 @@ import (
 	"slices"
 	"testing"
 
-	"github.com/pomerium/pomerium/pkg/ssh/models"
-	mock_models "github.com/pomerium/pomerium/pkg/ssh/models/mock"
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/mock/gomock"
+
+	"github.com/pomerium/pomerium/pkg/ssh/models"
+	mock_models "github.com/pomerium/pomerium/pkg/ssh/models/mock"
 )
 
 type TestData struct {


pkg/ssh/tui/tunnel_status/components/layout.go
```diff
@@ -0,0 +1,237 @@
...
+	for i, component := range components {
+		widgets[i] = &GroupComponentWidget{
+			Component: component,
+			Widget:    factories.NewComponentWidget(component),
+		}
+	}
+	// sort components in row-major order
+	slices.SortStableFunc(components, func(a, b Component) int {
+		return cmp.Or(cmp.Compare(a.RowHint(), b.RowHint()), cmp.Compare(a.ColumnHint(), b.ColumnHint()))
+	})
```
alexandreLamarre: The sort doesn't seem to have any impact, perhaps it's intended to happen before the widgets slice construction?

kralicky: great catch :+1: :+1:

alexandreLamarre (commented):
Overall this looks fine, I'm just a little nervous about some of the panics in the model / in other code paths - I understand it fails loudly for development for scenarios that shouldn't happen (and the health event panic messages completely make sense), but given how much TUI implementation there is and how little test coverage realistically makes sense to write for this implementation, I would feel much better if we have something like

`WithPanicHandler(cb func(...))` 

when we serve the TUI in the channel handlers to not crash the process, and this handler could be flexible enough to serve a TUI error page or something along those lines.

WDYT @kralicky ?

kralicky:
Yes it's built into the tea program by default. There is an option `tea.WithoutCatchPanics()` that you can add to disable it (but we leave it enabled)

PR #6024, opened 2025-12-31
core/authenticate: clear querier cache when revoking sessions

calebdoxsey:
## Summary
Update the authenticate stateful flow to clear the querier cache when revoking sessions.

Pomerium maintains a short-lived cache for databroker records for the authorize service. If a session is deleted from the databroker, the authorize service may continue to allow use of that session until the cache expires. Since the cache TTL is short, this is generally not an issue. 

However if a user logs out of one route and then immediately refreshes another route associated with the same session, the cookie for the first route will have been removed, but the cookie for the second route still exists, so authorization is done with the now-deleted session. Since that session is in the authorize service's cache, access is granted. If this happens with the enterprise console, it will result in an error, because the enterprise console retrieves user information directly from the databroker without the cache, and the associated session doesn't actually exist.

In all-in-one mode, with a single instance of Pomerium, we use a global cache for the authorize and proxy services, so it's possible for us to clear the cache when we revoke sessions from the authenticate service. This doesn't solve the problem universally but it does solve it for the most common case.

I also moved the sign out handlers to separate files while I was in here.

## Related issues
- [ENG-3421](https://linear.app/pomerium/issue/ENG-3421/console-breaks-when-user-returns-after-route-logout)

## Checklist

- [x] updated unit tests
- [x] ready for review


pkg/storage/querier_client.go
```diff
@@ -25,3 +27,39 @@ func (q *clientQuerier) Query(ctx context.Context, in *databroker.QueryRequest,
 }
 
 func (*clientQuerier) Stop() {}
+
+// DeleteDataBrokerRecord deletes a databroker record from the databroker and invalidates any cache.
```
agocs, line 31: Great refactor

PR #6032, opened 2026-01-06
databroker: add VersionedConfig record type

kenjenkins:
## Summary

Add a new VersionedConfig databroker record type, to act as a conditional config source.

## Related issues

https://linear.app/pomerium/issue/ENG-3178/zero-need-mechanism-for-serving-different-config-for-different

## User Explanation

n/a

## Checklist

- [x] updated unit tests
- [ ] ready for review


internal/databroker/config_source.go
```diff
@@ -122,6 +128,43 @@ func (src *ConfigSource) rebuild(ctx context.Context, firstTime firstTime) {
 	metrics.SetConfigInfo(ctx, cfg.Options.Services, "databroker", cfg.Checksum(), true)
 }
 
+// allDBConfigsLocked returns an iterator over the union of values in dbConfigs
+// and dbVersionedConfigs, in an unspecified order. The mutex must be held.
```
agocs, line 132: When you say "the mutex must be held", does that mean the caller must hold an RLock to prevent writing? I'm trying to find a way to enforce that, whether by contract or not. Would it be reasonable to call `mu.TryLock()`, expecting `false` before line 134 to ensure the mutex is locked? This might be one of those [correct uses of TryLock.](https://pkg.go.dev/sync#RWMutex:~:text=whether%20it%20succeeded.-,Note%20that%20while%20correct%20uses%20of%20TryLock%20do%20exist%2C%20they%20are%20rare%2C%20and%20use%20of%20TryLock%20is%20often%20a%20sign%20of%20a%20deeper%20problem%20in%20a%20particular%20use%20of%20mutexes.,-func%20(*RWMutex))

agocs: I see more `...Locked` functions. That's beginning to feel like an antipattern. A larger refactor might be in order. Let me know your thoughts.

agocs: At the very least, we should be clear about what sort of lock the caller is expected to hold.

pkg/grpc/config/config.go
```diff
@@ -30,3 +34,42 @@ func (x *Settings) HasBrandingOptions() bool {
...
+	}
+	return true
+}
+
+func parseSemVer(version string) *semver.Version {
+	v, _ := semver.NewVersion(version)
+	if v == nil {
+		return nil
+	}
+	v2, _ := v.SetPrerelease("")
```
agocs, line 73: What should `SetPrerelease("")` be doing here? I assume it's stripping off any e.g. `-beta-rc2` suffix from the version, but I wrote a test to validate that assumption and it fails. 


```
func Test_parseSemVer(t *testing.T) {
	type args struct {
		version string
	}
	tests := []struct {
		name           string
		args           args
		wantMajor      uint64
		wantMinor      uint64
		wantPatch      uint64
		wantPre        string
		wantStr        string
		shouldEqualStr string
	}{
		{
			name: "basic",
			args: args{
				version: "v0.1.0",
			},
			wantMajor:      0,
			wantMinor:      1,
			wantPatch:      0,
			wantPre:        "",
			wantStr:        "0.1.0",
			shouldEqualStr: "v0.1.0",
		},
		{
			name: "prerelease passes",
			args: args{
				version: "v0.1.0-beta-rc2",
			},
			wantMajor:      0,
			wantMinor:      1,
			wantPatch:      0,
			wantPre:        "beta-rc2",
			wantStr:        "0.1.0-beta-rc2",
			shouldEqualStr: "v0.1.0-beta-rc2",
		},
		{
			name: "prerelease that I think should pass but doesn't",
			args: args{
				version: "v0.1.0-beta-rc2",
			},
			wantMajor:      0,
			wantMinor:      1,
			wantPatch:      0,
			wantPre:        "",
			wantStr:        "0.1.0",
			shouldEqualStr: "v0.1.0",
		},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			v, _ := semver.NewVersion(tt.args.version)
			assert.Equal(t, tt.wantMajor, v.Major())
			assert.Equal(t, tt.wantMinor, v.Minor())
			assert.Equal(t, tt.wantPatch, v.Patch())
			assert.Equal(t, tt.wantPre, v.Prerelease())
			assert.Equal(t, tt.wantStr, v.String())
			shouldEqual, _ := semver.NewVersion(tt.shouldEqualStr)
			assert.True(t, v.Equal(shouldEqual))
		})
	}
}
```

kenjenkins: Indeed v.SetPrerelease("") clears any prerelease tag. You can see we do the same thing for the existing version checks in Zero: https://github.com/pomerium/pomerium-zero/blob/99e5de0e1b3dcedbbb431f0a2afde67abb2665fb/internal/coreversion/version.go#L50.

I can add a test case to exercise this method directly. The test case in your comment does not appear to call this method or to call SetPrerelease(), so I'm unclear on what it is meant to demonstrate.

internal/databroker/config_source.go
kenjenkins: The mutex can be held either for reading or for writing. Neither of these two new methods writes to any of the fields protected by the mutex.

I don't think TryLock() would be appropriate here. A return value of `false` does not indicate whether or not the current goroutine was the one that called Lock().

Could you help me understand what feels like an antipattern to you? My thought in naming these methods was to aim for consistency with the existing code.

pkg/grpc/config/config.go
agocs: 🤦 My mistake. I think I was trying to figure out what SetPrerelease does and what parseSemVer does at the same time and I got my wires crossed and tested neither.

internal/databroker/config_source.go
agocs: I feel like either there should be some contractural way of ensuring the caller must hold the lock (although I don't see an easy way to implement that), or all of the `...Locked` functions should at least try to obtain an RLock themselves (unless the caller swears up and down it has a WLock???). Just naming a function `...Locked` and making callers pinky promise they have a lock seems to be choosing the most dangerous option by default. 

I'm probably overthinking it, and if this is a pattern we use in pomerium/internal, good enough.

kenjenkins: If it helps at all to put your mind at ease, these methods are all unexported, which makes it much easier to reason about the callers. If you use something like VSCode's "Show Call Hierarchy" function, I think it's fairly easy to verify that all of these trace back to the rebuild() method, which does call mu.Lock().

PR #6037, opened 2026-01-08
config: derive signing key from shared secret

kenjenkins:
## Summary

If a JWT signing key is not specified explicitly, derive one from the shared secret.

## Related issues

https://linear.app/pomerium/issue/ENG-3436/core-derive-jwt-signing-key-from-shared-secret-if-not-set-explicitly

## Checklist

- [x] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -4,9 +4,12 @@ import (
 	"bytes"
 	"cmp"
 	"context"
+	"crypto/elliptic"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"iter"
@@ -19,11 +22,13 @@ import (
 	"strings"
 	"time"
 
+	"filippo.io/keygen"
 	"github.com/go-viper/mapstructure/v2"
 	goset "github.com/hashicorp/go-set/v3"
 	"github.com/rs/zerolog"
 	"github.com/spf13/viper"
 	"github.com/volatiletech/null/v9"
+	"golang.org/x/crypto/hkdf"
 	"google.golang.org/protobuf/types/known/durationpb"
 
 	"github.com/pomerium/pomerium/config/otelconfig"
@@ -1448,6 +1453,10 @@ func (o *Options) GetSigningKey() ([]byte, error) {
 		return nil, nil
 	}
 
+	if o.SigningKey == "" && o.SigningKeyFile == "" {
+		return o.deriveSigningKey()
+	}
+
 	rawSigningKey := o.SigningKey
 	if o.SigningKeyFile != "" {
 		bs, err := os.ReadFile(o.SigningKeyFile)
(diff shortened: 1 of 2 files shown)


config/options.go
```diff
@@ -795,6 +800,20 @@ func (o *Options) Validate() error {
 		}
 	}
 
+	if o.SigningKey == "" && o.SigningKeyFile == "" {
```
kenjenkins: This is maybe not the best place to put this logic, but I wasn't exactly sure where it could go instead. Definitely open to other suggestions.

config/options.go
```diff
@@ -19,11 +22,13 @@ import (
 	"strings"
 	"time"
 
+	"filippo.io/keygen"
```
desimone, line 25: TIL. cool

config/options.go
desimone: It's probably worth putting this into a separate function. I'm not sure where either though. It'd be nice to have a some sets of unit tests around it. Although you already have The big ones covered. I'm not really sure how an error case would happen here.

config/options.go
```diff
@@ -795,6 +800,20 @@ func (o *Options) Validate() error {
...
+		priv, err := keygen.ECDSALegacy(elliptic.P256(), r)
+		if err != nil {
+			return fmt.Errorf("couldn't derive JWT signing key: %w", err)
+		}
+		der, err := x509.MarshalECPrivateKey(priv)
+		if err != nil {
+			return fmt.Errorf("couldn't marshal derived JWT signing key: %w", err)
+		}
+		o.SigningKey = base64.StdEncoding.EncodeToString(
+			pem.EncodeToMemory(&pem.Block{Type: "EC PRIVATE KEY", Bytes: der}))
```
desimone: my preference as an operator would be that anything derived I'd want an explicit log message in case we get any sort of surprising non determinism to let users know that the key was derived (maybe the thought they set this value but it failed for whatever reason).  Im not sure what level makes sense, I'll leave that to you.

config/options.go
```diff
@@ -795,6 +800,20 @@ func (o *Options) Validate() error {
 		}
 	}
 
+	if o.SigningKey == "" && o.SigningKeyFile == "" {
+		r := hkdf.New(sha256.New, sharedSecret, nil, []byte("derived-jwt-signing-key"))
```
desimone:
```suggestion
		r := hkdf.New(sha256.New, sharedSecret, nil, []byte("pomerium/attestation-jwt/signing-key/v1"))
```

or something similar might be useful if we want to rotate it (for reasons I can't think of right now) in a later version

config/options.go
calebdoxsey: I tried to get the side effects out of Validate because its a footgun. We could have a `GetSigningKey` function that lazily would do it.

config/options.go
kenjenkins: I think it doesn't matter too much, as we can always switch to a format like that later. The important thing is for different keys to use different values for the context bytes. (I thought about centralizing all of the context values in one place, but I think it's probably not worth it for now.)

config/options.go
kenjenkins: I agree that a log message would be nice to have, but ideally I'd want to see it logged only once (or at most once per config reload). Now that I've moved the key derivation back to GetSigningKey(), if we log from that method we'll see the message appear multiple times for each config reload.

But if that's OK, I can go ahead and add a log message.

config/options.go
kenjenkins: Moved this out of Validate() and into GetSigningKey().

PR #6039, opened 2026-01-08
core/autocert: improve test performance

calebdoxsey:
## Summary
Switch from a minio docker container to https://github.com/johannesboyne/gofakes3 for testing s3 storage with autocert. This allows tests to run faster.

Also implement the `TryLocker` interface so we can avoid blocking some of the time, and support `username:password` in urls. (which allows us to avoid setting environment variables for the test)

## Related issues
- [ENG-3445](https://linear.app/pomerium/issue/ENG-3445/coreautocert-tests-are-slow)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/autocert/manager_test.go
+++ b/internal/autocert/manager_test.go
@@ -35,6 +35,7 @@ import (
 
 	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/internal/log"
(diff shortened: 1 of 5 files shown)


calebdoxsey:
I added a test and found a couple other things that were making tests slow. I think this is ready to review now.

PR #6049, opened 2026-01-14
mcp: implement refresh token support

wasaga:
## Summary

Add MCP refresh token functionality that allows MCP clients to refresh their access tokens without requiring user re-authentication.

Key implementation details:
- Store refresh token metadata in databroker (`type.googleapis.com/oauth21.MCPRefreshToken`)
- Implement token rotation on refresh (old token revoked, new one issued)
- Recreate Pomerium session from upstream IdP refresh token

## User Explanation

MCP clients can now use the `refresh_token` grant type to obtain new access tokens when the current one expires, without requiring the user to re-authenticate through the browser flow.

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/authenticate/authenticate_test.go
+++ b/authenticate/authenticate_test.go
@@ -87,12 +87,6 @@ func TestNew(t *testing.T) {
 	emptyProviderURL.Provider = "oidc"
 	emptyProviderURL.ProviderURL = ""
 
-	goodSigningKey := newTestOptions(t)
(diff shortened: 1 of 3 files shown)


wasaga:
## Refresh Token Rotation

Per OAuth 2.1 draft Section 4.3.1, the old refresh token should be invalidated but **not deleted**:

> *Refresh token rotation:* the authorization server issues a new refresh token with every access token refresh response. **The previous refresh token is invalidated but information about the relationship is retained by the authorization server.** If a refresh token is compromised and subsequently used by both the attacker and the legitimate client, one of them will present an invalidated refresh token, which will inform the authorization server of the breach. The authorization server cannot determine which party submitted the invalid refresh token, but it will revoke the active refresh token as well as the access authorization grant associated with it.

This is why we set `Revoked = true` on the old token rather than deleting it from the databroker.

PR #6050, opened 2026-01-14
core/storage: add additional filter operators, order by

calebdoxsey:
## Summary
Add additional filter expression operators besides just `$eq`. These operators aren't implemented in the existing code (using them will end up returning an error), but are in some projects that use this code. The `OrderBy` is the same and was also added.

## Related issues
- [ENG-3457](https://linear.app/pomerium/issue/ENG-3457/corestorage-add-additional-filter-operators)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/storage/filter.go
+++ b/pkg/storage/filter.go
@@ -156,6 +156,10 @@ func MustEqualsFilterExpression(path string, value any) FilterExpression {
 
 // ValueAsString returns the value as a string.
 func (expr SimpleFilterExpression) ValueAsString() string {
+	if expr.Value == nil {
+		return ""
+	}
+
 	switch v := expr.Value.Kind.(type) {
 	case *structpb.Value_BoolValue:
 		return fmt.Sprint(v.BoolValue)
--- a/pkg/storage/indexer.go
+++ b/pkg/storage/indexer.go
@@ -296,6 +296,10 @@ func (i *IndexManager) compareAndReindex(incomingFields []string, recordsF func(
 }
 
 func (i *IndexManager) GetRelatedIDs(expr SimpleFilterExpression) ([]string, error) {
+	if expr.Operator != FilterExpressionOperatorEquals {
+		return nil, fmt.Errorf("%w: %v", ErrOperatorNotSupported, expr.Operator)
+	}
+
 	keyPath := strings.Join(expr.Fields, ".")
 	i.indexMu.RLock()
 	idxer, ok := i.genericIndices[keyPath]


pkg/storage/indexer.go
```diff
@@ -295,15 +295,15 @@ func (i *IndexManager) compareAndReindex(incomingFields []string, recordsF func(
 	i.fields = incomingFields
 }
 
-func (i *IndexManager) GetRelatedIDs(expr EqualsFilterExpression) ([]string, error) {
+func (i *IndexManager) GetRelatedIDs(expr SimpleFilterExpression) ([]string, error) {
```
wasaga, line 298: given it ignores `expr.Operator`, it probably should have a similar err check like https://github.com/pomerium/pomerium/blob/2f2abf9047e31d34f0575bfb07402be52b8eb991/pkg/storage/record_collection.go#L130-L133

pkg/storage/filter.go
```diff
@@ -134,10 +128,155 @@ type AndFilterExpression []FilterExpression
...
+func MustEqualsFilterExpression(path string, value any) FilterExpression {
+	v, err := structpb.NewValue(value)
+	if err != nil {
+		panic(err)
+	}
+	return SimpleFilterExpression{Fields: strings.Split(path, "."), Operator: FilterExpressionOperatorEquals, Value: v}
+}
+
+// ValueAsString returns the value as a string.
+func (expr SimpleFilterExpression) ValueAsString() string {
```
wasaga, line 158:
```suggestion
func (expr SimpleFilterExpression) ValueAsString() string {
    if expr.Value == nil {
        return ""
    }  
```

pkg/storage/filter.go
```diff
@@ -77,18 +95,20 @@ func filterExpressionFromStruct(path []string, s *structpb.Struct) (FilterExpres
...
-			if err != nil {
-				return nil, err
+			if op, ok := FilterExpressionOperatorFromField(f); ok {
+				and = append(and, SimpleFilterExpression{
+					Fields:   path,
+					Operator: op,
+					Value:    v,
+				})
+			} else {
+				expr, err := filterExpressionFromValue(append(path, f), v)
```
alexandreLamarre, line 106: If the filter expression is not in the map of supported expressions, we're checking here for the nested value for expressions (it makes sense to check nested expressions for `$and` and `$or` operators), but here it feels a little un-ituitive to consumers of the filter expressions

Are there existing implementations relying on this?

calebdoxsey: Yes, the behavior is that if there's not an operator we treat it as an equals:

```
{"k1": "v1", "k2": "v2"}
```

and:

```
{"$eq": { "k1": "v1", "k2": "v2" }}
```

both translate too:

```obj["k1"] === "v1" && obj["k2"] === "v2"```

alexandreLamarre (approved):

PR #6051, opened 2026-01-14
authenticate: remove unused 'jwk' fields

kenjenkins:
## Summary

The `authenticateState` type and `authenticateflow.Stateless` both currently build a JSONWebKeySet from the JWT signing key(s), but this doesn't appear to be used. Let's remove the `jwk` fields and corresponding initialization logic.

## Related issues

n/a

## Checklist

- [x] updated unit tests
- [ ] ready for review

### DIFF
--- a/authenticate/authenticate_test.go
+++ b/authenticate/authenticate_test.go
@@ -87,12 +87,6 @@ func TestNew(t *testing.T) {
 	emptyProviderURL.Provider = "oidc"
 	emptyProviderURL.ProviderURL = ""
 
-	goodSigningKey := newTestOptions(t)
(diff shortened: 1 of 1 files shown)


authenticate/state.go
```diff
@@ -53,8 +52,6 @@ type authenticateState struct {
 	sessionHandleWriter sessions.HandleWriter
 
 	csrf *csrfCookieValidation
-
-	jwk *jose.JSONWebKeySet
```
kenjenkins, line 57: Looks like authenticateState has the same logic? I'll go ahead and remove it from this one as well if that sounds good.

PR #6052, opened 2026-01-14
handlers: refactor JWKSHandler

kenjenkins:
## Summary

Currently `JWKSHandler` will parse and serialize the JWT signing key(s) for each new incoming request to the JWKS endpoint. Let's refactor the handler to do the parsing and serialization only once.

Add a couple more unit test cases and a benchmark.

Benchmark results when run on my machine:
```
goos: darwin
goarch: arm64
pkg: github.com/pomerium/pomerium/internal/handlers
cpu: Apple M1 Max
               │ bench-old.txt │           bench-new.txt            │
               │    sec/op     │   sec/op     vs base               │
JWKSHandler-10    22.041µ ± 3%   2.958µ ± 5%  -86.58% (p=0.002 n=6)

               │ bench-old.txt │            bench-new.txt            │
               │     B/op      │     B/op      vs base               │
JWKSHandler-10   10.958Ki ± 0%   6.659Ki ± 0%  -39.23% (p=0.002 n=6)

               │ bench-old.txt │           bench-new.txt           │
               │   allocs/op   │ allocs/op   vs base               │
JWKSHandler-10     103.00 ± 0%   30.00 ± 0%  -70.87% (p=0.002 n=6)
```

## Related issues

n/a

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/internal/handlers/jwks.go
+++ b/internal/handlers/jwks.go
@@ -3,7 +3,6 @@ package handlers
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"hash/fnv"
 	"net/http"
@@ -39,7 +38,8 @@ func marshalJWKS(signingKey []byte) (body []byte, hash uint64, err error) {
 	if len(signingKey) > 0 {
 		ks, err := cryptutil.PublicJWKsFromBytes(signingKey)
 		if err != nil {
-			return nil, 0, httputil.NewError(http.StatusInternalServerError, errors.New("bad signing key"))
+			return nil, 0, httputil.NewError(http.StatusInternalServerError,
+				fmt.Errorf("bad signing key: %w", err))
 		}
 		for _, k := range ks {
 			jwks.Keys = append(jwks.Keys, *k)


internal/handlers/jwks.go
```diff
@@ -18,34 +18,42 @@ import (
...
 }
+
+func marshalJWKS(signingKey []byte) (body []byte, hash uint64, err error) {
+	var jwks struct {
+		Keys []any `json:"keys"`
+	}
+	if len(signingKey) > 0 {
+		ks, err := cryptutil.PublicJWKsFromBytes(signingKey)
+		if err != nil {
+			return nil, 0, httputil.NewError(http.StatusInternalServerError, errors.New("bad signing key"))
```
agocs: Would you want to wrap the error from cryptoutil in your new error here?

agocs (approved):
Impressive results

internal/handlers/jwks.go
kenjenkins: I can never remember if this will expose the wrapped error message in the HTML response — but looking at the HTTPError.ErrorResponse() method it looks like it does not. I think that means it should be fine to wrap the cryptutil error here.

PR #6056, opened 2026-01-20
core/config: add additional config options

calebdoxsey:
## Summary
Add additional config options.

- `id`, `name`, `namespace_id`, `originator_id`, `created_at`, `modified_at` as standard fields. All optional. (some of these already existed, and were required, but I believe its wire-compatible to change a required field to optional)
- add `reserved` in several places where we had merely commented out a field
- add reference fields to allow references between entities (like `policy_ids` to reference a policy instead of embedding one directly). Loading this data will be implemented in a separate PR.
- add a `KeyPair` entity
- add a `ConfigService` and CRUD operations for it. This is not implemented yet.

## Related issues
- [ENG-3447](https://linear.app/pomerium/issue/ENG-3447/coregrpc-add-config-service-stub)

## Checklist

- [x] updated unit tests
- [x] ready for review


config/policy.go
```diff
@@ -1126,3 +1126,11 @@ func UpstreamTunnelToProto(t *UpstreamTunnel) *configpb.UpstreamTunnel {
 	}
 	return &configpb.UpstreamTunnel{}
 }
+
+func nilOnZero[T comparable](val T) *T {
```
agocs, line 1130: I swear to god, I could have used this pattern so many times in the last decade. I'm going to remember it.

PR #6058, opened 2026-01-21
feat(health) : add metrics integration

alexandreLamarre:
## Summary

Adds metrics as a health check integration

## User Explanation

metrics will be exposed about current internal health check state, how many times a state was reported for each check, and the duration for each component to start successfully.

## Checklist

- [X] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/controlplane/server.go
+++ b/internal/controlplane/server.go
@@ -16,6 +16,7 @@ import (
 	"github.com/libp2p/go-reuseport"
 	"github.com/rs/zerolog"
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	"go.opentelemetry.io/otel"
 	oteltrace "go.opentelemetry.io/otel/trace"
 	coltracepb "go.opentelemetry.io/proto/otlp/collector/trace/v1"
 	"golang.org/x/net/nettest"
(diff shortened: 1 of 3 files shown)


pkg/health/metrics.go
```diff
@@ -0,0 +1,189 @@
...
+			),
+		)
+	}
+}
+
+func (m *MetricsProvider) ReportStatus(check Check, status Status, attributes ...Attr) {
+	commonAttr := attributeSet(
+		attributes,
+		attribute.String("check", string(check)),
+	)
```
alexandreLamarre, line 149: expected cardinality for these metrics are
```
numChecks x numStatuses x numAttribute combinations
```
which taking the maximum value for attribute pairs is:
```
12 x 4 x 3 = 144
```
although only the storage backend check has 3 unique attribute pairs, and the rest mostly have none with some exceptions so the likely total number of series is ~50-60 and I don't expect there will be much of a scale problem with these fields in the future

internal/controlplane/server.go
```diff
@@ -49,6 +50,24 @@ import (
 	"github.com/pomerium/pomerium/pkg/telemetry/trace"
 )
 
+type Options struct {
+	startTime time.Time
+}
+
+func (o *Options) Apply(opts ...Option) {
+	for _, opt := range opts {
```
wasaga, line 62: add default WithStartTime(time.Now()) ?

pkg/health/metrics_test.go
```diff
@@ -0,0 +1,191 @@
...
+	err := setup.reader.Collect(setup.ctx, &rm)
+	require.NoError(t, err)
+
+	require.Len(t, rm.ScopeMetrics, 1)
+	scopeMetrics := rm.ScopeMetrics[0]
+
+	var totalCount int64
+	var runningCount int64
+	var terminatingCount int64
+	var errorCount int64
```
wasaga, line 69: doesn't seem to be checked?

PR #6060, opened 2026-01-22
tidy: clustered databroker execution flow

alexandreLamarre:
## Summary

Tidied up a couple of things that looked a little out of place:

- Signals will now be unbound in the data-broker clustered leader. The mutex behaviour will also not be undefined by switching the signal to a pointer.
- Leader will wait to stop before starting as a follower 
- ~~Send explicit close frames to the server from client side streams, it is more verbose and doesn't appear to be a source of bugs currently~~

## User Explanation

N/A

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/internal/databroker/server_clustered_leader.go
+++ b/internal/databroker/server_clustered_leader.go
@@ -19,6 +19,7 @@ type clusteredLeaderServer struct {
 	onChange *signal.Signal
 
 	cancel context.CancelCauseFunc
+	done   chan struct{}
 }
 
 // NewClusteredLeaderServer creates a new clustered leader databroker server.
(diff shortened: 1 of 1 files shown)


internal/signal/signal.go
```diff
@@ -26,6 +57,7 @@ func (s *Signal) Broadcast(ctx context.Context) {
 		select {
 		case ch <- ctx:
 		default:
+			s.logger.Warn().Ctx(ctx).Msg("failed to broadcast signal update, buffer full")
```
alexandreLamarre, line 60: Normally I agree with our logging conventions not to use `warn` level, but I think it's appropriate here. It would get lost in `debug` / `info`, and could become spammable at the `error` level.

If `warn` isn't applicable here, I'd rather `error`

internal/controlplane/server_debug.go
```diff
@@ -163,6 +163,7 @@ func (srv *debugServer) serveDatabrokerOptionsIndex(w http.ResponseWriter, r *ht
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	defer func() { _ = stream.CloseSend() }()
```
calebdoxsey: My understanding is that the correct way to do this is to cancel the context that was used to start the stream.

PR #6061, opened 2026-01-24
authenticate: add server-side PKCE support for authorization code flow

### LINKED ISSUE #5269: Feature request: support PKCE for OAuth authorization code flow

PKCE ("Proof Key for Code Exchange") is an optional part of the OAuth authorization code flow. While primarily intended for mobile apps and single-page web applications, it may also be used to enhance the security of traditional server-side web applications.

Let's add support for PKCE to Pomerium.

n/a

The draft "OAuth 2.0 Security Best Current Practice" document has this to say:

> For confidential clients, the use of PKCE [[RFC7636](https://www.rfc-editor.org/info/rfc7636)] is RECOMMENDED, as it provides strong protection against misuse and injection of authorization codes as described in [Section 4.5.3.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#pkce_as_injection_protection) and, as a side-effect, prevents CSRF even in the presence of strong attackers as described in [Section 4.7.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#csrf_countermeasures).

(from https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-2.1.1-2.2)

**Additional context**

- The [x/oauth2](https://pkg.go.dev/golang.org/x/oauth2) library now supports PKCE (as of https://go-review.googlesource.com/c/oauth2/+/463979).
- IdP support for PKCE should be indicated by the [`code_challenge_methods_supported`](https://www.rfc-editor.org/rfc/rfc8414.html#:~:text=code_challenge_methods_supported%0A%20%20%20%20%20%20OPTIONAL.%20%20JSON,not%20support%20PKCE.) property of the /.well-known/openid-configuration discovery endpoint.
- [RFC 7636](https://www.rfc-editor.org/rfc/rfc7636) Proof Key for Code Exchange by OAuth Public Clients

desimone:
## Summary

- Implements server-side PKCE ([RFC 7636](https://www.rfc-editor.org/rfc/rfc7636), S256) for the OAuth authorization code flow, per [RFC 9700 (BCP 240)](https://www.rfc-editor.org/rfc/rfc9700) and [OAuth 2.1 draft](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1) guidance
- Binds a per-request verifier to OAuth `state` via an encrypted per-state cookie
- Auto-enables PKCE when the IdP advertises `code_challenge_methods_supported: S256` via OIDC discovery; 
- OAuth-only providers are unaffected

## Test plan

- [x] `make test` passes (unit tests for cookie store, providers, context helpers)
- [x] `make lint` passes
- [ ] Manual verification: login flow works unchanged for providers without PKCE discovery (e.g., GitHub)

Resolves #5269
Resolves ENG-919


authenticate/pkce.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// shouldUsePKCE reports whether the authenticator's IdP advertises S256 PKCE support.
+func shouldUsePKCE(authenticator identity.Authenticator) bool {
+	provider, ok := authenticator.(pkce.MethodsProvider)
+	if !ok {
+		return false
+	}
+	for _, m := range provider.PKCEMethods() {
+		if strings.EqualFold(m, "S256") {
```
kenjenkins: I think this can just be `m == "S256"` — unless I'm missing something I think there's no reason this should be case-insensitive.

kenjenkins: FYI this is specified in RFC 7636, section 6.2.1 (emphasis added):

>   Code Challenge Method Parameter Name:
>      The name requested (e.g., "example").  Because a core goal of this
>      specification is for the resulting representations to be compact,
>      it is RECOMMENDED that the name be short -- not to exceed 8
>      characters without a compelling reason to do so.  **This name is
>      case-sensitive.**  Names may not match other registered names in a
>      case-insensitive manner unless the Designated Expert(s) states
>      that there is a compelling reason to allow an exception in this
>      particular case.

https://datatracker.ietf.org/doc/html/rfc7636#section-6.2.1

PR #6062, opened 2026-01-26
test(acceptance): add E2E acceptance test harness

desimone:
## Summary
- Add comprehensive E2E acceptance test suite using Playwright
- Docker Compose orchestration with Keycloak IdP
- 59 focused tests covering security-critical behaviors
- Single source of truth for test users via fixtures/users.json
- CI workflow for automated testing

### Authentication (`authn/`)
- Login flow with OIDC
- Logout with HMAC-signed URLs
- CSRF/state protection
- Token refresh behavior

### Authorization (`authz/`)
- Email domain policies
- Group-based policies (single, multiple, compound)
- Explicit deny rules
- mTLS client certificates
- Internal endpoint security (path traversal, public vs protected)

### Headers (`headers/`)
- Cookie security flags
- CORS preflight handling
- JWT assertion headers
- WebSocket upgrade

## Test Plan
- [x] All 59 tests pass locally
- [x] TypeScript type checking passes
- [x] Docker compose config validates
- [x] CI workflow passes

## Architecture

```
internal/acceptance/
├── browser/           # Playwright tests
│   ├── fixtures/      # Test data (users.ts, test-data.ts)
│   ├── helpers/       # Shared utilities
│   └── tests/         # Test specs (authn/, authz/, headers/)
├── fixtures/          # Single source of truth (users.json)
├── pomerium/          # Pomerium config
├── keycloak/          # Keycloak realm config
├── scripts/           # Shell scripts for setup
└── ws-server/         # WebSocket/CORS test server
```

### DIFF
--- a/internal/acceptance/acceptance_test.go
+++ b/internal/acceptance/acceptance_test.go
@@ -22,7 +22,6 @@
 // Prerequisites:
 // - Docker and Docker Compose
 // - Node.js 18+
-// - /etc/hosts entries for *.localhost.pomerium.io -> 127.0.0.1
 package acceptance
 
 import (
--- a/internal/acceptance/browser/helpers/authn-flow.ts
+++ b/internal/acceptance/browser/helpers/authn-flow.ts
@@ -58,12 +58,13 @@ export async function login(page: Page, options: LoginOptions): Promise<void> {
   expect(currentUrl).toContain("keycloak.localhost.pomerium.io");
   expect(currentUrl).toContain("/realms/pomerium-e2e/protocol/openid-connect/auth");
 
-  // Fill in credentials
-  await page.locator("#username").fill(keycloakUsername);
-  await page.locator("#password").fill(user.password);
+  // Fill in credentials using accessible selectors
+  // Use exact match for Password to avoid matching "Show password" button
+  await page.getByLabel(/username/i).fill(keycloakUsername);
+  await page.getByLabel("Password", { exact: true }).fill(user.password);
 
   // Submit the form
-  await page.locator("#kc-login").click();
+  await page.getByRole("button", { name: /sign in/i }).click();
 
   if (waitForRedirect) {
     // Wait for redirect back to Pomerium
--- a/internal/acceptance/browser/tests/authn/csrf-protection.spec.ts
+++ b/internal/acceptance/browser/tests/authn/csrf-protection.spec.ts
@@ -82,12 +82,12 @@ test.describe("CSRF/State Protection", () => {
     // Tamper with CSRF cookie
     await tamperCSRFCookie(page);
 
-    // Complete login
+    // Complete login using accessible selectors
     const user = testUsers.alice;
     const keycloakUsername = getKeycloakUsername(user);
-    await page.locator("#username").fill(keycloakUsername);
-    await page.locator("#password").fill(user.password);
-    await page.locator("#kc-login").click();
+    await page.getByLabel(/username/i).fill(keycloakUsername);
+    await page.getByLabel("Password", { exact: true }).fill(user.password);
+    await page.getByRole("button", { name: /sign in/i }).click();
 
     // Wait for response
     await page.waitForLoadState("domcontentloaded");
@@ -123,12 +123,12 @@ test.describe("CSRF/State Protection", () => {
     // Remove CSRF cookie before completing login
     await removeCSRFCookie(page);
 
-    // Complete login
+    // Complete login using accessible selectors
     const user = testUsers.alice;
     const keycloakUsername = getKeycloakUsername(user);
-    await page.locator("#username").fill(keycloakUsername);
-    await page.locator("#password").fill(user.password);
-    await page.locator("#kc-login").click();
+    await page.getByLabel(/username/i).fill(keycloakUsername);
+    await page.getByLabel("Password", { exact: true }).fill(user.password);
+    await page.getByRole("button", { name: /sign in/i }).click();
 
     // Wait for response
     await page.waitForLoadState("domcontentloaded");
--- a/internal/acceptance/browser/tests/authn/login.spec.ts
+++ b/internal/acceptance/browser/tests/authn/login.spec.ts
@@ -78,11 +78,11 @@ test.describe("Auth Code Flow Login", () => {
       timeout: 15000,
     });
 
-    // Complete login
+    // Complete login using accessible selectors
     const keycloakUsername = getKeycloakUsername(user);
-    await page.locator("#username").fill(keycloakUsername);
-    await page.locator("#password").fill(user.password);
-    await page.locator("#kc-login").click();
+    await page.getByLabel(/username/i).fill(keycloakUsername);
+    await page.getByLabel("Password", { exact: true }).fill(user.password);
+    await page.getByRole("button", { name: /sign in/i }).click();
 
     // Should return to original target path
     await page.waitForURL((url) => {
--- a/internal/acceptance/browser/tests/authn/logout.spec.ts
+++ b/internal/acceptance/browser/tests/authn/logout.spec.ts
@@ -132,10 +132,11 @@ test.describe("Logout", () => {
     });
 
     // The page should have some form of logout confirmation UI
-    // Check for common logout button text variants
-    const logoutButton = page.locator('button, a, input[type="submit"]').filter({
-      hasText: /log\s*out|sign\s*out|confirm/i,
-    }).first();
+    // Use getByRole for accessibility-first selectors
+    const logoutButton = page
+      .getByRole("button", { name: /log\s*out|sign\s*out|confirm/i })
+      .or(page.getByRole("link", { name: /log\s*out|sign\s*out|confirm/i }))
+      .first();
 
     // If there's a confirmation button, click it
     if (await logoutButton.isVisible()) {


.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,151 @@
...
+      - 'internal/handlers/**'
+      - 'pkg/endpoints/**'
+      - 'internal/acceptance/**'
+      - '.github/workflows/acceptance.yaml'
+  push:
+    branches: [main]
+    paths:
+      - 'internal/acceptance/**'
+      - '.github/workflows/acceptance.yaml'
+  workflow_dispatch:
```
calebdoxsey, line 8: Is there a reason we wouldn't want to always run the tests? Code for the different pomerium services is fairly distributed across the codebase as well as imported packages.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,151 @@
...
+        with:
+          node-version-file: .tool-versions
+          cache: 'npm'
+          cache-dependency-path: internal/acceptance/browser/package-lock.json
+
+      - name: Set up Go
+        uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c
+        with:
+          go-version-file: .tool-versions
+          cache: true
```
calebdoxsey: I think cache is enabled by default?

internal/acceptance/Makefile
```diff
@@ -0,0 +1,111 @@
...
+.PHONY: help deps up down wait seed cleanup-users test test-headed test-debug test-authn test-authz test-headers test-go logs status report artifacts clean
+
+COMPOSE_PROJECT ?= acceptance
+COMPOSE_FILE ?= docker-compose.yml
+BROWSER_DIR ?= browser
+ARTIFACTS_DIR ?= artifacts
+
+COMPOSE := docker compose -f $(COMPOSE_FILE) -p $(COMPOSE_PROJECT)
+
+export RUN_ID ?= $(shell date +%s)
```
wasaga: Deferred expansion (`?=`) causes `date +%s` to be re-evaluated each time RUN_ID is used, resulting in different timestamps between seed (creates `test-user-<ts1>-alice`) and tests (looks for `test-user-<ts2>-alice`). Use immediate expansion:

```suggestion
# Use immediate expansion to ensure RUN_ID is set once at make invocation time
ifndef RUN_ID
  RUN_ID := $(shell date +%s)
endif
export RUN_ID
```

internal/acceptance/Makefile
```diff
@@ -0,0 +1,111 @@
...
+	@echo "Environment: RUN_ID (default: timestamp)"
+
+# Dependencies
+deps:
+	cd $(BROWSER_DIR) && npm ci
+	cd $(BROWSER_DIR) && npx playwright install chromium
+
+# Start the test stack
+up:
+	$(COMPOSE) up -d --build --wait
```
wasaga: `--wait` causes docker compose to return non-zero when the `stack-ready` one-shot container exits (even with code 0). Remove it since the `test` target depends on `seed` → `wait` which handles readiness via `wait-for-ready.sh`.

```suggestion
	$(COMPOSE) up -d --build
```

internal/acceptance/Makefile
```diff
@@ -0,0 +1,111 @@
...
+# Stop the test stack
+down:
+	$(COMPOSE) down -v
+
+# Wait for stack readiness
+wait:
+	./scripts/wait-for-ready.sh
+
+# Seed Keycloak with test users
+seed: wait
```
wasaga, line 57: Keycloak 26 enforces HTTPS for admin API calls from outside the container. Run the seed script inside the Docker network where HTTP works:

```suggestion
	docker run --rm --network acceptance_acceptance \
		-v $(CURDIR)/scripts:/scripts:ro \
		-v $(CURDIR)/fixtures:/fixtures:ro \
		-e KEYCLOAK_URL=http://keycloak.localhost.pomerium.io:8080 \
		-e USERS_FILE=/fixtures/users.json \
		-e RUN_ID=$(RUN_ID) \
		alpine:3.21 sh -c "apk add --no-cache curl jq bash >/dev/null 2>&1 && bash /scripts/seed-keycloak.sh seed"
```

internal/acceptance/browser/global-setup.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Check if a URL is reachable.
+ * Certificate validation is handled via NODE_TLS_REJECT_UNAUTHORIZED env var.
+ */
+function checkUrl(url: string, name: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const isHttps = url.startsWith("https://");
+    const client = isHttps ? https : http;
+
+    const req = client.get(url, (res) => {
```
wasaga: Node's native `https` module does not respect Playwright's `ignoreHTTPSErrors` setting. Self-signed certs need explicit `rejectUnauthorized: false`:

```suggestion
    // Disable certificate validation for self-signed certs in test environment
    const options = isHttps ? { rejectUnauthorized: false } : {};

    const req = client.get(url, options, (res) => {
```

internal/acceptance/docker-compose.yml
```diff
@@ -0,0 +1,136 @@
...
+  keycloak:
+    image: quay.io/keycloak/keycloak:26.0
+    command:
+      - start-dev
+      - --import-realm
+      - --health-enabled=true
+      - --http-port=8080
+      - --hostname-strict=false
+    environment:
+      KEYCLOAK_ADMIN: admin
```
wasaga: Keycloak 26 deprecates `KEYCLOAK_ADMIN`/`KEYCLOAK_ADMIN_PASSWORD`. Use new env vars and add `KC_HOSTNAME_STRICT: false` for HTTP compatibility:

```suggestion
      KC_BOOTSTRAP_ADMIN_USERNAME: admin
      KC_BOOTSTRAP_ADMIN_PASSWORD: admin
      KC_HTTP_ENABLED: "true"
      KC_HOSTNAME: keycloak.localhost.pomerium.io
      KC_HOSTNAME_STRICT: "false"
      KC_PROXY_HEADERS: xforwarded
```

internal/acceptance/README.md
```diff
@@ -0,0 +1,104 @@
...
+- `/etc/hosts` entries for local DNS
+
+Add these host entries once:
+```
+127.0.0.1 keycloak.localhost.pomerium.io
+127.0.0.1 authenticate.localhost.pomerium.io
+127.0.0.1 app.localhost.pomerium.io
+127.0.0.1 admin.localhost.pomerium.io
+127.0.0.1 mtls.localhost.pomerium.io
+```
```
wasaga: for host its not really required as normal DNS resolution would suffice

desimone: Not sure I understand this comment?

kenjenkins: There's a real DNS record for *.localhost.pomerium.io that resolves to 127.0.0.1. These /etc/hosts entries should not be necessary.

internal/acceptance/docker-compose.yml
desimone: I should upate keycloak too.

internal/acceptance/acceptance_test.go
```diff
@@ -0,0 +1,213 @@
...
+	// defaultTimeout for the entire test suite
+	defaultTimeout = 10 * time.Minute
+
+	// browserDir is the relative path to the Playwright test directory
+	browserDir = "browser"
+)
+
+// TestAcceptance runs the Playwright E2E acceptance tests.
+func TestAcceptance(t *testing.T) {
+	if testing.Short() {
```
kenjenkins, line 48: It looks like there is an `acceptance` build tag for these tests, so maybe we don't also need to skip these for the `-short` flag?

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,137 @@
...
+      - name: Configure /etc/hosts
+        run: |
+          sudo tee -a /etc/hosts <<EOF
+          127.0.0.1 keycloak.localhost.pomerium.io
+          127.0.0.1 authenticate.localhost.pomerium.io
+          127.0.0.1 app.localhost.pomerium.io
+          127.0.0.1 admin.localhost.pomerium.io
+          127.0.0.1 mtls.localhost.pomerium.io
+          EOF
+          cat /etc/hosts
```
kenjenkins: Is this step also unnecessary?

internal/acceptance/Makefile
```diff
@@ -0,0 +1,126 @@
...
+	@echo "Acceptance test targets:"
+	@echo "  make deps         - Install npm deps and Playwright browser"
+	@echo "  make up           - Start the test stack"
+	@echo "  make seed         - Seed Keycloak users"
+	@echo "  make test         - Run all Playwright tests"
+	@echo "  make test-headed  - Run tests with a visible browser"
+	@echo "  make test-debug   - Run tests in debug mode"
+	@echo "  make test-authn   - Run authn tests"
+	@echo "  make test-authz   - Run authz tests"
+	@echo "  make test-headers - Run headers tests"
```
kenjenkins: It looks like `test-headed` and `test-debug` are different test execution strategies, while `test-authn`, `test-authz`, and `test-headers` are different categories of tests? This seems like an inconsistent naming convention.

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,137 @@
...
+        working-directory: internal/acceptance
+        run: ./scripts/seed-keycloak.sh seed
+
+      - name: Run Playwright tests
+        id: playwright
+        working-directory: internal/acceptance/browser
+        run: npm test
+        env:
+          CI: true
+        continue-on-error: true
```
kenjenkins: Could all of these steps be condensed into one step that runs `make test`?

.github/workflows/acceptance.yaml
```diff
@@ -0,0 +1,137 @@
...
+          retention-days: 7
+
+      - name: Post PR comment
+        if: always() && github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
+        uses: marocchino/sticky-pull-request-comment@52423e01640425a022ef5fd42c6fb5f633a02728
+        with:
+          header: acceptance-summary
+          path: internal/acceptance/artifacts/playwright/summary.md
+
+      - name: Fail if tests failed
```
kenjenkins: Is this needed? Won't a workflow fail if any step fails?

internal/acceptance/keycloak/realm.json
```diff
@@ -0,0 +1,270 @@
...
+  "enabled": true,
+  "displayName": "Pomerium E2E Testing",
+  "sslRequired": "none",
+  "registrationAllowed": false,
+  "loginWithEmailAllowed": true,
+  "duplicateEmailsAllowed": false,
+  "resetPasswordAllowed": false,
+  "editUsernameAllowed": false,
+  "bruteForceProtected": false,
+  "accessTokenLifespan": 5,
```
kenjenkins: I think this will just not work reliably with Pomerium as it is now. Pomerium has a hard-coded 10 second backoff on session refresh operations. If the access token lifespan is shorter than this I think there's going to be a race condition here.

(I saw the AI comment about keycloak version but I'm not currently convinced by that explanation. I think regardless of the keycloak version this is not going to work reliably without some additional changes.)

desimone: Same. I got pulled off before trying to fix it but I think claude is just wrong here.

internal/acceptance/browser/tests/authn/logout.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+    // Navigate to sign_out - Pomerium handles this and shows confirmation page
+    await page.goto(`${urls.app}${paths.signOut}`, {
+      waitUntil: "domcontentloaded",
+    });
+
+    // The page should have some form of logout confirmation UI
+    // Check for common logout button text variants
+    const logoutButton = page.locator('button, a, input[type="submit"]').filter({
+      hasText: /log\s*out|sign\s*out|confirm/i,
+    }).first();
```
nickytonline: refactor (blocking): use `getByRole`. Ideally something laser focused so whatever the button or link text is, e.g. `page.getByRole("button", { name: /^logout$/i })`. The LLM is just covering all the bases. I imagine there's either a button or a link, not both on the page. In the meantime, the locator suggestion is just converted to getByRole.


```suggestion
    const logoutButton = page
      .getByRole("button", { name: /log\s*out|sign\s*out|confirm/i })
      .or(page.getByRole("link", { name: /log\s*out|sign\s*out|confirm/i }))
      .first();
```

internal/acceptance/browser/tests/authn/login.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    // Should redirect to Keycloak
+    await page.waitForURL((url) => url.toString().includes("keycloak"), {
+      timeout: 15000,
+    });
+
+    // Complete login
+    const keycloakUsername = getKeycloakUsername(user);
+    await page.locator("#username").fill(keycloakUsername);
+    await page.locator("#password").fill(user.password);
+    await page.locator("#kc-login").click();
```
nickytonline: refactor (nit): should be using `getByRole` like elsewhere. Normally this would be blocking, but I imagine these Keycloak element id's won't be changing anytime soon.

internal/acceptance/browser/tests/authn/csrf-protection.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    // Tamper with CSRF cookie
+    await tamperCSRFCookie(page);
+
+    // Complete login
+    const user = testUsers.alice;
+    const keycloakUsername = getKeycloakUsername(user);
+    await page.locator("#username").fill(keycloakUsername);
+    await page.locator("#password").fill(user.password);
+    await page.locator("#kc-login").click();
```
nickytonline: refactor (nit): should be using getByRole like elsewhere. Normally this would be blocking, but I imagine these Keycloak element id's won't be changing anytime soon.

internal/acceptance/browser/helpers/authn-flow.ts
```diff
@@ -0,0 +1,234 @@
...
+  });
+
+  // Verify we're on the Keycloak login page
+  const currentUrl = page.url();
+  expect(currentUrl).toContain("keycloak.localhost.pomerium.io");
+  expect(currentUrl).toContain("/realms/pomerium-e2e/protocol/openid-connect/auth");
+
+  // Fill in credentials
+  await page.locator("#username").fill(keycloakUsername);
+  await page.locator("#password").fill(user.password);
```
nickytonline: refactor (nit): should be using getByRole like elsewhere. Normally this would be blocking, but I imagine these Keycloak element id's won't be changing anytime soon.

PR #6064, opened 2026-01-26
core/protoutil: implement sorting of arbitrary protobuf messages

calebdoxsey:
## Summary
Add support for sorting a slice of arbitrary protobuf messages.

This is implemented via field masks. Each field mask path is used to reference a field of a protobuf message. For example, with:

```proto
message Example {
  string id   = 1;
  string name = 2;
}
```

You can use a field mask path of `name` to sort a list of `Example` messages by name. Sorting of scalar types uses the standard `cmp.Compare` function, as well as `bytes.Compare` for `[]byte`, duration and timestamp support, and `proto.Marshal` for any other message type. A final repeated field is supported by comparing lists of values.

The field mask path can support embedded messages. For example, with:

```proto
message Inner {
  string name = 1;
}
message Outer {
  Inner inner = 1;
}
```

You can use a field mask path of `inner.name` to sort a list of `Outer` messages. A repeated inner message is not supported, nor are map types, though support for these could probably be added in the future.

## Related issues
- [ENG-3495](https://linear.app/pomerium/issue/ENG-3495/corestorage-implement-arbitrary-sorting-for-protobuf-messages)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/storage/order_by.go
+++ b/pkg/storage/order_by.go
@@ -54,7 +54,7 @@ func (orderBy OrderBy) String() string {
 	return sb.String()
 }
 
-// SortStable sorts a list of protobuf messages based on the order by.
+// Sort sorts a list of protobuf messages based on the order by.
 func Sort[T any, TMsg interface {
 	*T
 	proto.Message


pkg/storage/order_by.go
```diff
@@ -45,3 +53,60 @@ func (orderBy OrderBy) String() string {
 	}
 	return sb.String()
 }
+
+// SortStable sorts a list of protobuf messages based on the order by.
+func Sort[T any, TMsg interface {
```
alexandreLamarre: docstring typo

PR #6071, opened 2026-01-29
fix(authorize): sync_querier can have an inconsistent state 

alexandreLamarre:
## Summary

In the authorize service the sync_querier can have an inconsistent state, which leads it to thinking it can serve records, when it in fact, cannot.

This issue can manifest itself as an infinite redirect loop after authentication with Pomerium.

## Related issues

Part of [ENG-2657](https://linear.app/pomerium/issue/ENG-2657/core-infinite-redirect-loop-after-switching-databroker-backend)

## User Explanation

fixed a rare race condition where pomerium would sometimes redirect infinitely after login

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/pkg/grpc/databroker/syncer.go
+++ b/pkg/grpc/databroker/syncer.go
@@ -29,6 +29,9 @@ type SyncerOption func(cfg *syncerConfig)
 
 func getSyncerConfig(options ...SyncerOption) *syncerConfig {
 	cfg := new(syncerConfig)
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxElapsedTime = 0
+	cfg.BackOff = bo
 	WithSyncerTracerProvider(noop.NewTracerProvider())(cfg)
 	for _, option := range options {
 		option(cfg)
@@ -96,11 +99,6 @@ func NewSyncer(ctx context.Context, id string, handler SyncerHandler, options ..
 	closeCtx, closeCtxCancel := context.WithCancel(context.WithoutCancel(ctx))
 
 	cfg := getSyncerConfig(options...)
-	if cfg.BackOff == nil {
-		bo := backoff.NewExponentialBackOff()
-		bo.MaxElapsedTime = 0
-		cfg.BackOff = bo
-	}
 
 	if DebugUseFasterBackoff.Load() {
 		bo := backoff.NewExponentialBackOff(


pkg/storage/querier_sync_test.go
alexandreLamarre, line 1: These tests aren't completely exhaustive (and I can always omit them and the mockgen stuff from this PR), but will be useful as a starting point in a potential follow up PR where we refactor the querier_sync.go to use the databroker.SyncerHandler

authorize/state.go
alexandreLamarre, line 1: @calebdoxsey I think when we were looking at this file, it looked like the logic was wrong here w.r. to the maps not being cleared, but upon closer inspection they are...

I took the liberty of trying to make the logic more readable here.

pkg/grpc/databroker/syncer.go
```diff
@@ -89,22 +95,26 @@ var DebugUseFasterBackoff atomic.Bool
 func NewSyncer(ctx context.Context, id string, handler SyncerHandler, options ...SyncerOption) *Syncer {
 	closeCtx, closeCtxCancel := context.WithCancel(context.WithoutCancel(ctx))
 
-	var bo *backoff.ExponentialBackOff
+	cfg := getSyncerConfig(options...)
+	if cfg.BackOff == nil {
+		bo := backoff.NewExponentialBackOff()
+		bo.MaxElapsedTime = 0
+		cfg.BackOff = bo
+	}
```
calebdoxsey: I think this logic might be better put in the `getSyncConfig` function itself. Like:

```
	cfg := new(syncerConfig)
    bo := backoff.NewExponentialBackOff()
    bo.MaxElapsedTime = 0
    WithBackoff(bo)(cfg)
	WithSyncerTracerProvider(noop.NewTracerProvider())(cfg)
	for _, option := range options {
		option(cfg)
	}
	return cfg
```

pkg/grpc/databroker/syncer.go
```diff
@@ -89,22 +95,26 @@ var DebugUseFasterBackoff atomic.Bool
...
+		bo := backoff.NewExponentialBackOff(
 			backoff.WithInitialInterval(10*time.Millisecond),
 			backoff.WithMultiplier(1.0),
 			backoff.WithMaxElapsedTime(100*time.Millisecond),
 		)
 		bo.MaxElapsedTime = 0
-	} else {
-		bo = backoff.NewExponentialBackOff()
-		bo.MaxElapsedTime = 0
+		cfg.BackOff = bo
```
calebdoxsey, line 110: And maybe `DebugUseFasterBackoff` isn't needed anymore? Because the option can be passed explicitly?

alexandreLamarre: Yeah I think I tried to unpack removing this, but it's set by the testenv so that it's passed to every syncer that runs within the testenv which I remember being not so obvious to change.

Let me take another look

alexandreLamarre: I can't really come up with a good way to remove this without injecting that behavior in other constructors, or injecting a global default backoff that is then passed to all pomerium components where each constructor would inherit that default, both of which seem worse than the flag?

PR #6078, opened 2026-01-29
feat(grpc): ensure "coordination" between keepalives in http/2 protocols in internal pomerium grpc impls

alexandreLamarre:
## Summary

For gRPC:
- Ensure on both clients and servers we have:
  - `PermitWithoutStream : true` ensures http/2 pings are sent without any active stream RPCs occurring.
 - Ensure that the outbound TCP listener closes unreachable connections after 20s using TCP_USER_TIMEOUT on the socket
   - For consistency, this matches our http/2 ping timeouts across the board
- Ensure that http/2 pings from envoy to our servers are not sent more frequently than at a 5 minute interval - to prevent grpc server from sending GOAWAY with error=ENHANCE_YOUR_CALM to envoy, abruptly closing the envoy http2 stream, but resulting in outbound clients with a "half-closed" stream receiving RST_STREAM errors.
- Fixes a potential race condition with expiring contexts resulting in grpc.ClientConn getting shut on a fresh & valid connection, causing `ErrClientConnectionClose` on all subsequent connections

## Related issues

[ENG-2657](https://linear.app/pomerium/issue/ENG-2657/core-infinite-redirect-loop-after-switching-databroker-backend)
[ENG-3437](https://linear.app/pomerium/issue/ENG-3437/clustered-databroker-follower-health-graceful-shutdown-error-from-sync)
[ENG-3536](https://linear.app/pomerium/issue/ENG-3536/bug-http2-stream-terminated-by-rst-stream-with-error-code-no-error)
[ENG-3526](https://linear.app/pomerium/issue/ENG-3526/prod-console-is-not-currently-syncing-to-the-proxy)

## User Explanation

N/A

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/envoyconfig/listeners_grpc.go
+++ b/config/envoyconfig/listeners_grpc.go
@@ -3,7 +3,6 @@ package envoyconfig
 import (
 	"context"
 	"fmt"
-	"time"
 
 	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_config_listener_v3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
@@ -134,9 +133,6 @@ func (b *Builder) buildGRPCHTTPConnectionManagerFilter() *envoy_config_listener_
 		Routes:  routes,
 	}})
 
-	// !! Must not send pings more frequent than grpc.Server keepalive policy enforcement. Default 5mins
-	http2ProtocolOpts := WithKeepalive(http2ProtocolOptions, time.Minute*6, 15)
-
 	return b.HTTPConnectionManagerFilter(&envoy_http_connection_manager.HttpConnectionManager{
 		CodecType:  envoy_http_connection_manager.HttpConnectionManager_AUTO,
 		StatPrefix: "grpc_ingress",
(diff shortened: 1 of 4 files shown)


config/envoyconfig/protocols.go
alexandreLamarre, line 1: The changes in this file and how we pass them to `listeners_grpc.go` and `outbound.go` look a little weird because I'm still unsure about how exactly we want to propagate specific keepalive settings.

These http2 protocol options are shared across all envoy cluster creations so changing the previous keepalive settings to play nicely specifically with the gRPC-go policy enforcement might have adverse side effects on pomerium routes that aren't intended with that usecase in mind  (That being said I can't think of any real problems of increasing keepalive ping intervals from ~1min to ~6mins)

Is it worth taking the time of splitting out how these http2 protocol settings are passed across the board? Or should we keep it to one set of defaults?

config/envoyconfig/listeners_grpc.go
```diff
@@ -132,6 +134,9 @@ func (b *Builder) buildGRPCHTTPConnectionManagerFilter() *envoy_config_listener_
 		Routes:  routes,
 	}})
 
+	// !! Must not send pings more frequent than grpc.Server keepalive policy enforcement. Default 5mins
+	http2ProtocolOpts := WithKeepalive(http2ProtocolOptions, time.Minute*6, 15)
```
calebdoxsey: This is the same as in outbound... should it be pulled into a function or a constant?

alexandreLamarre: yeah it's actually the same as the global default WithKeepalive, I'm not sure if it makes sense to split them out, or apply them consistently to all clusters (that have keepalive)?

config/envoyconfig/protocols.go
calebdoxsey: I think it's fine to change it for everything.

alexandreLamarre:
Looks like the ECONNREFUSED isn't consistently reproducible. I left a couple background scenarios running letting them go idle > 2 hours, and it didn't pop up again.

PR #6098, opened 2026-02-05
debug: config overrides for ports and envoy admin sock

alexandreLamarre:
## Summary

Set config.yaml options for;
```
envoy_admin_address_sock_name: envoy_admin_a.sock
port_overrides:
  grpc: "22220"
  http: "22221"
  outbound: "22222"
  metrics: "22223"
  debug: "22224"
  acme_tls_alpn: "22225"
  connect: "22226"
```

port_overrides are useful for capturing network packets and understanding which envoy/pomerium ports they are associated to

`envoy_admin_address_sock_name` is required for running a clustered databroker on the same host machine.

## Related issues

N/A

## User Explanation

N/A

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -19,6 +19,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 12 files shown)


calebdoxsey:
These will also have to be added to the protobuf definitions and conversion code.

config/options.go
```diff
@@ -323,6 +328,148 @@ type GenericKeyVal struct {
...
+	ACMETLSALPNPort uint32 `mapstructure:"acme_tls_alpn" yaml:"acme_tls_alpn,omitempty"`
+	// ConnectPort overrides the dynamically allocated connect server port.
+	ConnectPort uint32 `mapstructure:"connect" yaml:"connect,omitempty"`
+}
+
+// portEntries returns all port overrides as ordered name/value pairs.
+// The order must match the field order in PortOverrides and Config.AllocatePorts.
+func (p *PortOverrides) portEntries() []struct {
+	name string
+	port uint32
```
agocs, line 355: Would it make sense for `port` to be `uint16`? You could do without `Validate`

agocs (approved):
Quick suggestion about using uint16 for port numbers, but feel free to tell me no. It seems like if you used uint16 instead of uint32 all over the place, you could use the type system to ensure valid port numbers.

config/options.go
agocs: In retrospect, there's no protobuf unit16 type, so never mind. You can ignore this.
https://protobuf.dev/programming-guides/proto3/#scalar

alexandreLamarre: I also opted not to use `uint16` for fat finger syndrome, where one might type 633333 and then suddenly spend too much time debugging

alexandreLamarre: which often happens to me

PR #6100, opened 2026-02-06
mcp: add upstream MCP token storage

wasaga:
## Summary

Add persistent databroker storage for OAuth tokens acquired from remote MCP server authorization servers. Tokens are bound to `(user_id, route_id, upstream_server)` and support basic CRUD operations.

- New `UpstreamMCPToken` protobuf message with composite key fields, token fields, timestamps, and metadata
- `PutUpstreamMCPToken`, `GetUpstreamMCPToken`, `DeleteUpstreamMCPToken` on `handlerStorage` interface
- Input validation rejects empty key components
- Indexed `user_id` and `route_id` for future bulk queries
- Fix pre-existing panic in `fieldToValue` for repeated proto fields

## Related issues

- ENG-3556

## User Explanation

No user-facing changes. This adds internal storage infrastructure for upcoming MCP upstream token management.

## Checklist

- [x] updated unit tests
- [ ] ready for review

### DIFF
--- a/internal/mcp/storage.go
+++ b/internal/mcp/storage.go
@@ -313,8 +313,12 @@ func (storage *Storage) DeleteMCPRefreshToken(
 }
 
 // upstreamMCPTokenID builds the composite key for an UpstreamMCPToken record.
-func upstreamMCPTokenID(userID, routeID, upstreamServer string) string {
-	return databroker.CompositeRecordID(map[string]any{"user_id": userID, "route_id": routeID, "upstream_server": upstreamServer})
+// All three fields must be non-empty.
+func upstreamMCPTokenID(userID, routeID, upstreamServer string) (string, error) {
+	if userID == "" || routeID == "" || upstreamServer == "" {
+		return "", fmt.Errorf("upstream MCP token requires non-empty user_id, route_id, and upstream_server")
+	}
+	return databroker.CompositeRecordID(map[string]any{"user_id": userID, "route_id": routeID, "upstream_server": upstreamServer}), nil
 }
 
 // PutUpstreamMCPToken stores or updates an upstream MCP token record.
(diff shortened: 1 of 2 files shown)


internal/mcp/storage.go
```diff
@@ -302,6 +305,90 @@ func (storage *Storage) DeleteMCPRefreshToken(
 	return nil
 }
 
+// upstreamMCPTokenID builds the composite key for an UpstreamMCPToken record.
+func upstreamMCPTokenID(userID, routeID, upstreamServer string) string {
+	return fmt.Sprintf("%s|%s|%s", userID, routeID, upstreamServer)
```
kenjenkins: Is it possible for any of these strings parts to contain a `|` character? Should we escape the inputs somehow?

e.g. to prevent an ID collision between
- userID = "foo|bar" routeID = "baz"
- userID = "foo" routeID = "bar|baz"

(or something along those lines)

pkg/storage/index.go
```diff
@@ -106,21 +106,25 @@ func fieldToValue(fd protoreflect.FieldDescriptor, v protoreflect.Value) any {
...
 			return true
 		})
 		return res
 	}
 
-	switch fd.Kind() {
+	return kindToValue(fd.Kind(), v)
+}
+
+func kindToValue(kind protoreflect.Kind, v protoreflect.Value) any {
```
kenjenkins, line 126: Could this method be replaced with calling [Interface()](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect#Value.Interface) on the protoreflect.Value?

internal/mcp/storage.go
wasaga: yes its not perfect, if we want to address it it'll have to be changed for some other types as well for consistency; i'll start a discussion thread about it

pkg/storage/index.go
wasaga: msgToMap(v.Message()) is called for MessageKind, but Interface() would return a protoreflect.Message

PR #6108, opened 2026-02-11
ci: update dependencies

apparitor:
This PR updates dependencies not managed by dependabot.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -52,13 +52,13 @@ linters:
         text: Potential hardcoded credentials
       - linters:
           - gosec
-        text: 'G112:'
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -59,6 +59,10 @@ linters:
       - linters:
           - gosec
         text: 'G402: TLS MinVersion too low.'
+      - linters:
+          - revive
+        source: "^package (diff|encoding|hpke|httputil|internal|log|metrics|parser|signal|slices|ssh|testenv|token|trace|user|version)$"
```
kenjenkins: Should we just drop this rule entirely?

PR #6109, opened 2026-02-11
fix(mcp): CORS header consistency and e2e tests

wasaga:
## Summary

Fixes CORS header inconsistencies across MCP proxy layers, corrects the 401 response reason phrase for MCP routes, and adds comprehensive e2e tests that validate CORS headers on responses **as they come out of Pomerium**.

### Changes

1. **`authorize/check_response.go`** — Three fixes in `requireLoginResponse`:
   - **Scoped "Unauthenticated" → "Unauthorized" rename** for MCP server routes only. RFC 9110 §15.5.2 defines the status as "401 Unauthorized" — this is the canonical reason phrase. The MCP Inspector client-side `is401Error` detection ([`useConnection.ts`](https://github.com/modelcontextprotocol/inspector/blob/1f424f252546b8611e275a158d1756fd30517d87/client/src/lib/hooks/useConnection.ts#L387)) string-matches `"Unauthorized"` in error messages; "Unauthenticated" didn't match, so the OAuth flow was never triggered. Non-MCP routes retain `"Unauthenticated"` to preserve existing Pomerium behavior.
   - **Propagate `SetWWWAuthenticateHeader` error** instead of suppressing it with `_ =`, matching the existing error-handling pattern in `handleResultDenied`.
   - **Add CORS headers** to the MCP 401 response via `SetCORSHeaders`.

2. **`internal/mcp/handler.go`** — Add `"authorization"` to CORS `AllowedHeaders` on OAuth handler endpoints. The token endpoint supports `client_secret_basic` authentication which sends credentials in the `Authorization` header (RFC 6749 §2.3.1, OAuth 2.1 §3.2).

3. **`internal/mcp/e2e/mcp_cors_test.go`** *(new)* — E2e test `TestMCPCORSHeaders` with 6 subtests validating CORS headers across all three proxy layers:
   - **ext_authz layer** (`cors.go`): unauthenticated 401 response + OPTIONS preflight
   - **Metadata handler** (`handler_metadata.go`): `.well-known/oauth-protected-resource` + `.well-known/oauth-authorization-server`
   - **OAuth handler** (`handler.go`): dynamic client registration endpoint
   - **Full stack**: complete OAuth flow → authenticated request reaches upstream

4. **`internal/mcp/cors.go`** — Added spec reference comments (MCP Transport spec, RFC 9728 §5).

### Spec references

| Endpoint | CORS | Spec |
|----------|------|------|
| Token (`/token`) | SHOULD | OAuth 2.1 §3.2 |
| Registration (`/register`) | SHOULD | OAuth 2.1 §3.2 |
| Metadata (`.well-known/*`) | SHOULD | OAuth 2.1 §3.2, RFC 9700 |
| Authorization (`/authorize`) | MUST NOT | OAuth 2.1 §3.2 (accessed via redirect) |
| MCP upstream (401/preflight) | Yes | MCP Transport spec |

**401 reason phrase:** RFC 9110 §15.5.2 defines the status as "401 Unauthorized". RFC 9110 §11.6.1 requires a `WWW-Authenticate` header on 401 responses. The MCP SDK (`streamableHttp.ts:496`) triggers auth on `response.status === 401`, while the MCP Inspector client (`useConnection.ts:387`) additionally string-matches `"Unauthorized"` in error messages — the previous "Unauthenticated" text prevented the OAuth flow from triggering. The rename is scoped to `IsMCPServer()` routes only to avoid changing behavior for existing non-MCP clients.

## Related issues

- [ENG-3570: Bug: MCP Inspector does not immediately connect to Pomerium-fronted servers](https://linear.app/pomerium/issue/ENG-3570/bug-mcp-inspector-does-not-immediatelly-connects-to-pomerium-fronted)

## User Explanation

Browser-based MCP clients (such as MCP Inspector) now receive correct CORS headers and the standard "Unauthorized" reason phrase on 401 responses from Pomerium, enabling seamless connection to Pomerium-fronted MCP servers without CORS errors or missed OAuth triggers.

## Checklist

- [x] updated unit tests
- [ ] ready for review


internal/mcp/cors.go
```diff
@@ -0,0 +1,22 @@
...
+// (401/403) and OPTIONS preflight responses at the Envoy ext_authz layer,
+// before requests reach the Go HTTP handler's own CORS middleware.
+//
+// Spec references:
+//   - MCP Transport spec: servers MUST validate the Origin header on all incoming
+//     connections; browser-based clients require CORS on auth error responses.
+//   - RFC 9728 §5: WWW-Authenticate with resource_metadata must be exposed to
+//     browser clients via Access-Control-Expose-Headers.
+func SetCORSHeaders(dst http.Header) {
+	dst.Set("Access-Control-Allow-Origin", "*")
```
desimone, line 31: more of a gut reaction but ... `Access-Control-Allow-Origin: *` is very permissive preflight responses. Can we scope this to validated origins?

internal/mcp/cors.go
```diff
@@ -0,0 +1,22 @@
...
+// This is used by the authorize service to add CORS headers to denied responses
+// (401/403) and OPTIONS preflight responses at the Envoy ext_authz layer,
+// before requests reach the Go HTTP handler's own CORS middleware.
+//
+// Spec references:
+//   - MCP Transport spec: servers MUST validate the Origin header on all incoming
+//     connections; browser-based clients require CORS on auth error responses.
+//   - RFC 9728 §5: WWW-Authenticate with resource_metadata must be exposed to
+//     browser clients via Access-Control-Expose-Headers.
+func SetCORSHeaders(dst http.Header) {
```
desimone, line 30: Please add `Vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers` when setting CORS headers to avoid cache/proxy cross-origin policy leakage.

internal/mcp/cors.go
```diff
@@ -0,0 +1,22 @@
...
+	dst.Set("Access-Control-Allow-Origin", "*")
+	dst.Set("Access-Control-Allow-Methods", "GET, POST, DELETE, OPTIONS")
```
desimone, line 32: This includes `DELETE`, but OAuth handler CORS currently allows only `GET, POST, OPTIONS` (`internal/mcp/handler.go`). Can we unify these policies or document why ext_authz intentionally needs a broader method set?

internal/mcp/cors.go
```diff
@@ -0,0 +1,22 @@
...
+	dst.Set("Access-Control-Allow-Headers", "Authorization, Content-Type, Accept, MCP-Protocol-Version, MCP-Session-Id, Last-Event-ID")
```
desimone, line 33: again, gut but this feels very broad. can we minimize to required headers per endpoint/path, or split transport-vs-OAuth CORS policies?

internal/mcp/cors.go
wasaga: `*` is intentional here:

- **OAuth 2.1 §3.2** says token/registration/metadata endpoints should "support the necessary CORS headers to allow the responses to be visible to the application" — no origin restriction specified.
- This is a **public OAuth AS** with dynamic client registration (RFC 7591) — any browser-based client is a valid MCP client by design. We don't know origins in advance.
- `AllowCredentials` is **not set** (and can't be with `*`), so no cookie/credential leakage.
- The MCP Transport spec's "MUST validate the Origin header" is specifically about **DNS rebinding protection on the upstream MCP server**, not CORS policy at the auth/ext_authz layer.
- `handler.go` and `handler_metadata.go` already use `AllowedOrigins: []string{"*"}` via `rs/cors`.

Scoping to validated origins would require maintaining an allowlist of MCP client origins, which breaks the dynamic registration model.

internal/mcp/cors.go
wasaga: Not a mismatch — these are **different request paths**:

- **ext_authz layer** (`cors.go`): handles **MCP upstream routes** where the MCP Streamable HTTP transport uses `DELETE` to terminate sessions, `POST` to send JSON-RPC, and `GET` to open SSE streams.
- **OAuth handler** (`handler.go`): handles `/.pomerium/mcp/*` OAuth endpoints where only `GET`/`POST` are used.

The ext_authz layer fires before routing and only applies to MCP server policy routes (guarded by `IsMCPServer()`), so it needs the full MCP transport method set. Updated the doc comment to explain this split.

internal/mcp/cors.go
wasaga: added

internal/mcp/cors.go
wasaga: see above

PR #6112, opened 2026-02-12
feat(acceptance): add MCP E2E acceptance tests

ssveta7ak:
## Summary
- Add comprehensive E2E acceptance tests for Pomerium's MCP (Model Context Protocol) support
- Cover full OAuth 2.1 authorization flow (PKCE, dynamic client registration, token exchange)
- Test both Streamable HTTP and SSE transports through Pomerium proxy
- Validate MCP Inspector and MCPJam Inspector UI tool discovery
- MCP stack isolated via Docker Compose override file — existing tests unaffected
- Dedicated CI workflow with path-filtered triggers

### Infrastructure
- `mcp-server/` — Minimal TypeScript MCP test server (echo, add, get_time tools) with SSE + Streamable HTTP transports
- `docker-compose.mcp.yml` — Self-contained overlay with mcp-server, MCP Inspector, MCPJam Inspector, pomerium network alias, and mcp-node-modules volume
- `pomerium/config-mcp.yaml` — Full Pomerium config with `runtime_flags: { mcp: true }`

### Test suites (18 tests total: 16 active + 2 skipped)
| Suite | Tests | What it validates |
|-------|-------|-------------------|
| SDK HTTP | 8 | OAuth metadata, client registration, auth gating, initialize, tools/list, echo, add, get_time |
| SDK SSE | 3 | Auth gating, SSE endpoint event, tools/list via SSE session |
| MCP Inspector UI | 3 (+2 skip) | UI load, Streamable HTTP direct, SSE direct |
| MCPJam Inspector UI | 2 | UI load, HTTP transport direct |

### CI isolation
- `docker-compose.yml` has zero MCP references — `docker compose up` never starts MCP services
- `docker-compose.mcp.yml` is self-contained: all MCP services, volumes, and config overrides
- Makefile parameterized with `SUITE` variable:
  - `make test` — runs 59 base tests (authn, authz, headers)
  - `make test SUITE=mcp` — runs 18 MCP tests with MCP infrastructure
  - `make test SUITE=all` — runs all 77 tests
- `npm test` scoped to `tests/authn/ tests/authz/ tests/headers/` — Playwright never discovers MCP specs in base CI
- Dedicated `.github/workflows/acceptance-mcp.yaml` workflow with path-filtered triggers, summary generation, and PR comments

### Integration
- `feature-map.json`: MCP category with 4 test file entries
- `acceptance_test.go`: `TestAcceptanceMCP` Go wrapper
- `package.json`: `test`, `test:mcp`, `test:all` npm scripts

## Test plan
- [x] All 18 MCP tests pass (16 passed, 2 skipped) in ~37s
- [x] All 59 existing tests (authn, authz, headers) pass — zero regressions
- [x] All 77 tests pass together (75 passed, 2 skipped) in ~1.6m
- [x] Docker isolation verified: `docker compose up` does NOT start mcp-server or inspectors
- [x] `npm test` discovers exactly 59 tests — zero from `tests/mcp/`
- [x] `acceptance.yaml` workflow is unmodified
- [x] Inspector Pomerium tests intentionally skipped (proxy can't do MCP OAuth flow; already covered by SDK tests)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nickytonline (commented):
Consider adding tests focused on resources and error passthrough as well.

MCP clients may also use `resources/*` methods, and we should verify:

* Pomerium doesn't break other JSON-RPC methods (resources/list, resources/read)
* Error responses pass through intact (proxy doesn't mangle JSON-RPC errors)

These are proxy behavior tests, not MCP server implementation tests. Since the test server already implements a server-info resource, we just need to verify Pomerium proxies it correctly.

i.e.

For HTTP Streamable (mcp-sdk-http.spec.ts):
* `resources/list` should work through proxy
* `resources/read` should work through proxy
* error responses should pass through proxy correctly

For SSE (mcp-sdk-sse.spec.ts):

* `resources/list` via SSE should work through proxy
* `resources/read` via SSE should work through proxy
* error responses via SSE should pass through correctly

Example test:

```typescript
test("resources/list should work through proxy", async ({ page }) => {
  const user = testUsers.alice;
  const { tokens } = await acquireMcpToken(page, user);

  await mcpInitialize(page, tokens.access_token);

  const result = await mcpRequest(page, tokens.access_token, {
    jsonrpc: "2.0",
    id: 4,
    method: "resources/list",
  });

  expect(result.status).toBe(200);
  expect(result.body.result).toBeDefined();
  expect(result.body.result.resources).toBeDefined();
  expect(Array.isArray(result.body.result.resources)).toBe(true);
  
  const resourceUris = result.body.result.resources.map((r: unknown) => {
    return (typeof r === "object" && r != null && "uri" in r) ? r.uri : null;
  });
  
  expect(resourceUris).toContain("info://server");
});
```

internal/acceptance/docker-compose.yml
```diff
@@ -109,6 +110,38 @@ services:
       retries: 10
       start_period: 10s
 
+  # MCP test server (SSE + Streamable HTTP transports)
+  # ISOLATED via profile — only starts with: docker compose --profile mcp up
```
kenjenkins: Does it make sense to use both a profile and an override file? Could we put this `mcp-server` service and `mcp-node-modules` volume in the override file and skip the profile?

Or for that matter, why do we want to isolate the MCP config to its own set of Makefile targets? Are we going to need to add 4 new Makefile targets whenever we want to test another feature?

PR #6114, opened 2026-02-13
databroker: add CompositeRecordID utility for building record IDs

wasaga:
## Summary

Add `CompositeRecordID(map[string]any) string` utility to `pkg/grpc/databroker` that builds deterministic, URL-safe composite record IDs using `url.Values.Encode()`. Migrate all pipe-separated (`|`) composite key constructions in MCP storage to use dedicated helper functions backed by the new utility.

## User Explanation

No user-facing changes.

## Checklist

- [x] updated unit tests
- [ ] ready for review

### DIFF
--- a/pkg/grpc/databroker/databroker_test.go
+++ b/pkg/grpc/databroker/databroker_test.go
@@ -17,6 +17,33 @@ import (
 	"github.com/pomerium/pomerium/pkg/protoutil"
 )
 
+func TestCompositeRecordID(t *testing.T) {
+	t.Parallel()
+
+	t.Run("deterministic regardless of map iteration", func(t *testing.T) {
+		// Keys are sorted alphabetically, so output is always the same.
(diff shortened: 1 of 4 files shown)


pkg/grpc/databroker/databroker_test.go
```diff
@@ -17,6 +17,33 @@ import (
...
+	})
+
+	t.Run("non-string values", func(t *testing.T) {
+		id := CompositeRecordID(map[string]any{"count": 42, "enabled": true})
+		assert.Equal(t, "count=42&enabled=true", id)
+	})
+
+	t.Run("url-encodes special characters", func(t *testing.T) {
+		id := CompositeRecordID(map[string]any{"host": "example.com:443", "path": "/a&b=c"})
+		assert.Contains(t, id, "host=example.com%3A443")
```
kenjenkins: Would it be better to use assert.Equal() here and compare the entire expected output?

PR #6115, opened 2026-02-14
simplify acceptance test users setup

kenjenkins:
## Summary

Simplify acceptance test framework a little:
- import realm and users from a Keycloak export, rather than populating users and attributes via a bash script
- rework fixtures/users.ts to read user data from the same Keycloak json file, to maintain a single source of truth
- get rid of RUN_ID plumbing and username prefixing
- simplify docker compose file
- get rid of wait-for-ready.sh bash script

## Related issues

https://linear.app/pomerium/issue/ENG-3587/core-acceptance-tests-user-seeding-and-cleanup-does-not-appear-to-work

## User Explanation

n/a

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/.github/workflows/acceptance.yaml
+++ b/.github/workflows/acceptance.yaml
@@ -7,11 +7,6 @@ on:
     branches: [main]
   workflow_dispatch:
 
-env:
-  RUN_ID: ${{ github.run_id }}
-  # Accept self-signed certificates in test environment
-  NODE_TLS_REJECT_UNAUTHORIZED: "0"
-
 permissions:
   contents: read
   pull-requests: write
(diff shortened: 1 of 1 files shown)


internal/acceptance/keycloak/pomerium-e2e-realm.json
kenjenkins, line 1: Most of the added lines are from this file. This file and the -users-0.json file were generated by running the kc.sh `export` command after seeding the users (see https://www.keycloak.org/server/importExport#_exporting_a_realm_to_a_directory).

We could probably trim this file down significantly if we want.

desimone (approved):

PR #6123, opened 2026-02-17
ci: upgrade to Go 1.26.0

wasaga:
## Summary

Upgrade Go from 1.25.x to 1.26.0 in go.mod, .tool-versions, and Dockerfiles.

Go 1.26 [reordered](https://github.com/golang/go/commit/2a71af11fc7e45903be9ab84e59c668bb051f528) the fields in `net/url.URL`, which changes the output of `hashstructure`-based checksums used in route configuration. The checksums are non-cryptographic hashes used for change detection, so the value change is benign. Updated expected `route_checksum` values in envoyconfig tests accordingly.

## User Explanation

No user-facing changes.

## Checklist

- [x] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/route_configurations_test.go
+++ b/config/envoyconfig/route_configurations_test.go
@@ -99,7 +99,7 @@ func TestBuilder_buildMainRouteConfiguration(t *testing.T) {
 								"checkSettings": {
 									"contextExtensions": {
 										"internal": "false",
-										"route_checksum": "10993227095879864081",
+										"route_checksum": "14103096999915755832",
 										"route_id": "5fbd81d8f19363f4"
 									}
 								}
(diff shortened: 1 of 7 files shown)


config/envoyconfig/route_configurations_test.go
```diff
@@ -99,7 +99,7 @@ func TestBuilder_buildMainRouteConfiguration(t *testing.T) {
 								"checkSettings": {
 									"contextExtensions": {
 										"internal": "false",
-										"route_checksum": "10993227095879864081",
+										"route_checksum": "14103096999915755832",
```
kenjenkins, line 102: Do we understand why these checksums would be different after updating to Go 1.26?

wasaga: yes it is mentioned in the PR description

PR #6128, opened 2026-02-18
databroker: add VersionedConfig proto and syncer (#6032)

kenjenkins:
Manual backports of:
- https://github.com/pomerium/pomerium/commit/4dcfb4b609601916819476442d165d8b4a15ad37 from #6032
- https://github.com/pomerium/pomerium/commit/a4eac2221b49d46ef4570ae803db50c1bd7b68a3 from #6110

### DIFF
--- a/internal/version/components.json
+++ b/internal/version/components.json
@@ -1,5 +1,6 @@
 {
-  "config": "v0.11.0",
+  "config": "v0.6.0",
+  "hosted-authenticate-oidc": "v0.1.0",
   "mcp": "v0.12.0",
   "ssh": "v0.9.0"
 }


internal/version/components.json
```diff
@@ -1,5 +1,5 @@
 {
-  "config": "v0.6.0",
+  "config": "v0.11.0",
```
kenjenkins: I'm not totally sure if this is correct. Originally I wanted to bump the `config` feature version to indicate support for versioned config, but maybe this would have been better as an entirely separate component feature?

I'll need to check whether there were any significant changes from 0.6.0 to 0.10.0.

wasaga (approved):

internal/version/components.json
kenjenkins: Looks like there was indeed a significant change to the `config` feature in v0.7.0 (see https://github.com/pomerium/pomerium/pull/6056), so I think maybe better to keep this at v0.6.0 for now.

I've also added in the new version component from https://github.com/pomerium/pomerium/pull/6110.

PR #6137, opened 2026-02-20
docs: add MCP proxying architecture design document

wasaga:
## Summary

Adds `internal/mcp/DESIGN.md` documenting the MCP proxying architecture.

## User Explanation

No user-facing changes. Internal design documentation.

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/mcp/DESIGN.md
+++ b/internal/mcp/DESIGN.md
@@ -19,7 +19,6 @@ injecting upstream OAuth tokens and intercepting auth challenges transparently.
 - [Storage Model](#storage-model)
 - [Configuration and Wiring](#configuration-and-wiring)
 - [Endpoint Map](#endpoint-map)
-- [Key Files](#key-files)
 
 ---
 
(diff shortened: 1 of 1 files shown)


internal/version/components.json
kenjenkins, line 1: Is there a way we can avoid incrementing these for unrelated changes? It makes backports more difficult if there are non-semantic changes to these versions.

Or if we really can't bypass the precommit check, could we increment the patch number instead of the minor version number? That might make things a little clearer when trying to decipher what these version numbers actually represent.

kenjenkins (approved):

internal/version/components.json
wasaga: https://linear.app/pomerium/issue/ENG-3653

calebdoxsey: We can get rid of the automated check if it's not useful. I suspect no one will manually update these version numbers but it's unclear if we will ever actually use them anyway.

PR #6138, opened 2026-02-21
feat: schema-based blob chunking for resume uploads

alexandreLamarre:
## Summary

This PR implements the storage layer primitives to perform immutable writes of data with simple in-flight integrity checks for the session recording API & protocol.

The chunk writer:
- writes metadata in proto and json format (one for querying, one for human-readability)
- acts as an append only writer for the recording content
- finalizes the recording by persisting a signature (provenance & integrity) and manifest (chunk metadata for querying & lookup)

It handles several conflict cases that should abort immediately:
- metadata already exists
- chunk gap
- chunks already signed   

The chunk reader:
- Implements some basic iteration over content chunks
- Reads metadata
- Implements size, manifest and last modified primitives for later use at query time

<hr/>

The PR also implements a conformance suite that can be run against blob stores that verifies the append-only and immutable behavior end-to-end.

<hr/>

See the issue's linked RFC for more details.

## Related issues

[ENG-3609](https://linear.app/pomerium/issue/ENG-3609/blob-storage-chunked-reading-and-writing)

## User Explanation
N/A

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/pkg/storage/blob/chunk.go
+++ b/pkg/storage/blob/chunk.go
@@ -36,10 +36,7 @@ type chunkReader struct {
 
 type chunkWriter struct {
 	bucket *blob.Bucket
-
-	writeCtx context.Context
-	writeCa  context.CancelFunc
-	schema   SchemaV1WithKey
+	schema SchemaV1WithKey
 
 	// meant to guard for reading via the CurrentManifest() method
 	// in separate go-routine while writing
@@ -55,17 +52,13 @@ var (
 // Write methods
 
 func NewChunkWriter(ctx context.Context, schema SchemaV1WithKey, bucket *blob.Bucket) (ChunkWriter, error) {
-	ctxca, ca := context.WithCancel(ctx)
 	cw := &chunkWriter{
-		bucket:   bucket,
-		schema:   schema,
-		writeCtx: ctxca,
-		writeCa:  ca,
+		bucket: bucket,
+		schema: schema,
 	}
 
 	locked, err := cw.isLockedForAppend(ctx)
 	if err != nil {
-		_ = cw.Abort()
 		return nil, fmt.Errorf("check for signature: %w", err)
 	}
 	if locked {
@@ -73,7 +66,6 @@ func NewChunkWriter(ctx context.Context, schema SchemaV1WithKey, bucket *blob.Bu
 	}
 
 	if err := cw.loadManifest(ctx); err != nil {
-		_ = cw.Abort()
 		return nil, fmt.Errorf("failed to load chunk manifest: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/storage/blob/providers/providers.go
alexandreLamarre, line 1: There is a ton of drivers / config surface area here to handle - I'd like to keep this package light for development and revisit last when we're near feature complete for session recording

pkg/storage/blob/types.go
```diff
@@ -0,0 +1,54 @@
...
+	// The ChunkManifest is updated to reflect the newly written data.
+	WriteChunk(ctx context.Context, data []byte, checksum [16]byte) error
+	// CurrentManifest returns the current state of the chunk manifest,
+	// containing metadata about all chunks written so far.
+	CurrentManifest() *recording.ChunkManifest
+	// Finalize signs the aggregate contains of the written data with information
+	// about the data itself and its provenance
+	Finalize(ctx context.Context, sig *recording.RecordingSignature) error
+	// Abort cancels the in flight write operations
+	Abort(ctx context.Context) error
```
calebdoxsey: `ctx` can be tricky if its actually used and is the `ctx` passed to other functions. If the reason an operation fails is because the `ctx` is cancelled and/or times out, and `Abort` uses it to make a remote call, that call will fail immediately... 

We could probably do `.WithoutCancel` when we call `.Abort` to make sure that doesn't happen (if we still want tracing or other context values).

pkg/storage/blob/chunk.go
```diff
@@ -0,0 +1,365 @@
...
+		return nil, fmt.Errorf("failed to load chunk manifest: %w", err)
+	}
+
+	return cw, nil
+}
+
+func (c *chunkWriter) CurrentManifest() *recording.ChunkManifest {
+	c.manifestMu.RLock()
+	defer c.manifestMu.RUnlock()
+	return proto.Clone(c.manifest).(*recording.ChunkManifest)
```
calebdoxsey:
```suggestion
	return proto.CloneOf(c.manifest)
```

pkg/storage/blob/types.go
alexandreLamarre: Yeah that's a really good point.  The context isn't used for any cancellable operations, so perhaps we should just remove the context parameter to prevent a future interface from doing something like that

alexandreLamarre: It's a little weird because currently the implementation of the ChunkWriter doesn't use the passed in context evenly. 

I think it's left over from my first iteration where abort would cancel the chunkWriter's write context, preventing writes. But the intended effect of Abort is somewhat delegated to the follow-up PR with the recording server.

If we do need to Abort, it's because either the bucket config has changed and all operations need to be invalidated or the current context for the writer is no longer valid.

I'll just opt to remove the Abort implementation and pass in context's evenly to remote operations in the implementation

agocs (approved):

PR #6140, opened 2026-02-22
feat: session recording server

alexandreLamarre:
## Summary

Requires: https://github.com/pomerium/pomerium/pull/6138

Implements the session recording API & protocol for chunked writing of recordingh data and adds conformance tests and on config change tests for some edgecases.

The PR intentionally omits registering the API to any listeners as that will be followed up in a separate PR.

## Related issues

[ENG-3649](https://linear.app/pomerium/issue/ENG-3649/recording-server-api-and-protocol-implementation)

Partially addresses: [ENG-3612](https://linear.app/pomerium/issue/ENG-3612/blob-storage-and-recording-server-configuration)
## Checklist

- [X] updated unit tests
- [ ] ready for review

### DIFF
--- a/pkg/grpc/config/config.pb.go
+++ b/pkg/grpc/config/config.pb.go
@@ -7,10 +7,6 @@
 package config
 
 import (
-	reflect "reflect"
-	sync "sync"
-	unsafe "unsafe"
-
 	_ "github.com/envoyproxy/protoc-gen-validate/validate"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
@@ -18,6 +14,9 @@ import (
 	structpb "google.golang.org/protobuf/types/known/structpb"
 	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
 	wrapperspb "google.golang.org/protobuf/types/known/wrapperspb"
+	reflect "reflect"
+	sync "sync"
+	unsafe "unsafe"
 )
 
 const (
@@ -2486,7 +2485,7 @@ func (x *Policy) GetNamespaceName() string {
 	return ""
 }
 
-// Next ID: 177.
+// Next ID: 180
 type Settings struct {
 	state              protoimpl.MessageState `protogen:"open.v1"`
 	Id                 *string                `protobuf:"bytes,158,opt,name=id,proto3,oneof" json:"id,omitempty"`
@@ -8502,7 +8501,7 @@ const file_config_proto_rawDesc = "" +
 	"\v_source_pplB\x0e\n" +
 	"\f_explanationB\x0e\n" +
 	"\f_remediationB\x11\n" +
-	"\x0f_namespace_nameJ\x04\b\x04\x10\x05\"\xf6_\n" +
+	"\x0f_namespace_nameJ\x04\b\x04\x10\x05\"\xb6a\n" +
 	"\bSettings\x12\x14\n" +
 	"\x02id\x18\x9e\x01 \x01(\tH\x00R\x02id\x88\x01\x01\x12'\n" +
 	"\fnamespace_id\x18\x9f\x01 \x01(\tH\x01R\vnamespaceId\x88\x01\x01\x12#\n" +


internal/recording/dynamic_semaphore.go
alexandreLamarre, line 1: This is an admittedly far from perfect and heavy-handed mechanism for hot reloading the allowed max concurrent streams. I tried several different things and ended up with this, I couldn't think of a great way to avoid using a mutex here

I avoided `sync/atomic` counters / pointers to avoid spin lock cases, which could be lead to 100% CPU utilization under load

internal/recording/server.go
```diff
@@ -0,0 +1,367 @@
...
+			if actual != recvCheckSum {
+				err := status.Error(codes.DataLoss, "checksum did not match")
+				log.Ctx(ctx).Err(err).Msg("checksum did not match")
+				return err
+			}
+			select {
+			case send <- AccumulatedChunk{
+				checksum: recvCheckSum,
+				data:     accumulated,
+			}:
```
calebdoxsey: ~Might need a `case <-ctx.Done()` here?~

Nevermind, I read the tabs wrong... there is one

alexandreLamarre: It's kind of an indication that this should be split up into smaller functions, it's not super readable

internal/recording/dynamic_semaphore.go
alexandreLamarre: Re-reviewing this myself, this implementation doesn't make any sense

I think I started off with a non-reloadable sempahore.Weighted for convenience, but then tunnel visioned on that too much 

wrapping it in a mutex.Lock is equivalent to wrapping two integers in a mutex.Lock

internal/recording/server.go
alexandreLamarre: I split out some of those functions so I could re-read this control flow a little better

internal/recording/dynamic_semaphore.go
```diff
@@ -0,0 +1,41 @@
...
+		return false
+	}
+	s.current++
+	return true
+}
+
+func (s *dynamicSemaphore) Release() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.current > 0 {
```
kralicky: If s.current == 0 this should panic instead

internal/recording/server.go
```diff
@@ -0,0 +1,420 @@
...
+	cfgMu sync.RWMutex
+
+	blobCfg   atomic.Pointer[blob.StorageConfig]
+	config    atomic.Pointer[config.RecordingServerConfig]
+	bucket    atomic.Pointer[gblob.Bucket]
+	bucketErr error
+}
+
+func NewRecordingServer(ctx context.Context, cfg *config.Config) Server {
+	limit := 8
```
kralicky: I wonder if there is a better way to limit this than by max concurrent streams. Maybe we should not impose concurrency limits here, and instead apply bandwidth limits on the envoy side instead. That way we can take into account pending buffered data from all streams. Since we will have a separate fixed (small) thread pool processing recordings, that should moderate the number of concurrent reads we would actually receive from the client, I think. (i.e. we could have many concurrent recordings going, but they wouldn't all be writing full blast at once).

internal/recording/server.go
```diff
@@ -0,0 +1,420 @@
...
+	//nolint:gosec
+	actual := md5.Sum(data)
+	if actual != incomingChecksum {
+		err := status.Error(codes.DataLoss, "checksum did not match")
+		log.Ctx(ctx).Err(err).Msg("checksum did not match")
+		return err
+	}
+	select {
+	case send <- AccumulatedChunk{
+		checksum: incomingChecksum,
```
kralicky, line 347: Looking at the code in pkg/storage/blob/chunk.go, do we actually pass this checksum to the blob writer? It looks like it calculates the md5 itself again before uploading.

internal/recording/dynamic_semaphore.go
alexandreLamarre: it should not, for example if:
- server holds 8 streams
- user resizes down to 4
- panic ensues, once more than 4 of the 8 streams finish

internal/recording/server.go
alexandreLamarre: yes this was an omission we should explicitly pass it to the blob writer options, good catch

internal/recording/server.go
alexandreLamarre: That seems fine to me, but the reason I left it in is for the case explicitly if we needed to pull the recording server out of core (for resource saturation or ingest latency reasons) to serve multiple envoy client instances

alexandreLamarre: But I can yank this out

internal/recording/server.go
```diff
@@ -0,0 +1,420 @@
...
+			case <-eCtx.Done():
+				return
+			}
+			if err != nil {
+				return
+			}
+		}
+	}()
+
+	pipe := make(chan AccumulatedChunk, 1)
```
agocs, line 104: This smells like a potential race condition. `sendWithWaitAndCancel` might not block when writing to `send` which means it's possible for `Finalize()` to be called before writeStage writes the last chunk.

internal/recording/server.go
alexandreLamarre: @kralicky Should we enforce a hard limit like 10000 or similar number on the server using a simple semaphore that is not configurable by end-users?

internal/recording/server.go
```diff
@@ -0,0 +1,420 @@
...
+			r.bucketErr = err
+			r.bucket.Store(nil)
+		} else {
+			r.bucket.Store(bucket)
+			r.bucketErr = nil
+			health.ReportRunning(health.BlobStorage)
+		}
+	}
+
+	if cfg == nil {
```
agocs, line 145: Do you want to check for cfg==nil at the beginning of this funciton?

internal/recording/server.go
alexandreLamarre: Yeah we should send the signature to the writer goroutine as well

internal/recording/server.go
alexandreLamarre: Ah yes probably, I see the logic isn't correct w.r. to the incoming config being nil for `hasConfigChanged` and `isConfigDifferent`, let me also add some additional tests

alexandreLamarre: I fixed the issues with that control flow and attempted to make that part more readable

internal/recording/server.go
alexandreLamarre: this should be fixed now

internal/recording/server.go
alexandreLamarre: this should be fixed now

internal/recording/server.go
alexandreLamarre: Will get to this in the morning

internal/recording/dynamic_semaphore.go
kralicky: Once the user resizes to 4, current would still be 8 until those streams finish. If limit is < current, it just won't let you create new ones

internal/recording/server.go
kralicky: Yeah that makes sense, we can keep the limit configs here and set the defaults to something large just as a sanity measure.

internal/recording/dynamic_semaphore.go
alexandreLamarre: Ah I see the problem now

pkg/grpc/config/config.proto
```diff
@@ -499,6 +499,9 @@ message Settings {
   optional google.protobuf.Duration directory_provider_refresh_interval = 173;
   optional google.protobuf.Duration directory_provider_refresh_timeout  = 174;
 
+  optional RecordingServerSettings recording_server = 177;
+  optional BlobStorageSettings     blob_storage     = 178;
```
calebdoxsey: Can you update the "Next ID" comment above?

internal/recording/server.go
alexandreLamarre: https://github.com/pomerium/envoy-custom/pull/160

kralicky (approved):

PR #6166, opened 2026-03-02
ssh: refactor StreamHandler state management

kralicky:
This refactors some of the state management in `ssh.StreamHandler` to fix potential data races between `StreamHandler.Run()` and `StreamHandler.ServeChannel()`. 

Instead of keeping state as a field in StreamHandler itself, it is kept local to the `Run` loop. Methods which previously accessed the state field now accept the state as an argument instead and were converted to free functions if possible.

Operations outside of the `Run` loop which need to read the state (currently, creating internal channels from ServeChannel, and requesting handoff) now interact with the `Run` loop via new channels specifically for those requests. The `Run` loop can then reply with a _copy_ of the state and/or other needed info, or an error (if the request can fail).

Additionally, the username+hostname fields which used to be in the stream state as individual fields were pulled out into a separate type `api.UserRequest`. This struct holds a username and hostname, and is only modifiable from an Update method. It encapsulates the validation logic for username and hostname updates and should be less error-prone to use.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -66,6 +66,9 @@ linters:
       - linters:
           - revive
         text: "var-naming: avoid package names that conflict with Go standard library package names"
+      - linters:
+          - revive
+        text: "var-naming: avoid meaningless package names"
     paths:
       - third_party$
       - builtin$


pkg/ssh/stream.go
```diff
@@ -163,15 +205,17 @@ func NewStreamHandler(
 ) *StreamHandler {
 	writeC := make(chan *extensions_ssh.ServerMessage, 32)
```
wasaga, line 206: what is the purpose of this channel? it doesn't seem to be used other than closed later

pkg/ssh/stream.go
```diff
@@ -72,60 +74,102 @@ type EndpointDiscoveryInterface interface {
...
+	readC                   chan *extensions_ssh.ClientMessage
+	reauthC                 chan struct{}
+	terminateC              chan error
+	internalChannelRequestC chan InternalChannelRequest
+	handoffRequestC         chan HandoffRequest
+
+	close   func()
+	runOnce bool
 
 	expectingInternalChannel bool
```
wasaga: seems to be accessed from multiple goroutines? maybe use atomic.Bool instead?

pkg/ssh/stream_test.go
```diff
@@ -1801,10 +1811,10 @@ func (s *StreamHandlerSuite) TestServeChannel_Session_RoutesPortal_Select() {
...
 		case 2:
 			currentFrame++
-			s.mockAuth.EXPECT().EvaluateDelayed(Any(), Any()).
-				DoAndReturn(func(_ context.Context, info ssh.StreamAuthInfo) error {
-					s.Equal(info.Username, ptr("test"))
-					s.Equal(info.Hostname, ptr("host2"))
+			s.mockAuth.EXPECT().EvaluateDelayed(Any(), Any(), Any()).
+				DoAndReturn(func(_ context.Context, _ ssh.StreamAuthInfo, user api.UserRequest) error {
+					s.Equal(user.Username, ptr("test"))
+					s.Equal(user.Hostname, ptr("host2"))
```
wasaga, line 1817: these are funcs 
```go
func (u api.UserRequest) Username() string
```

kralicky: This test is disabled in this branch, It is fixed in https://github.com/pomerium/pomerium/pull/6167

pkg/ssh/stream.go
kralicky: Good catch thx

pkg/ssh/stream.go
kralicky: Oh that's weird... this definitely didn't work correctly before. It was either supposed to be used as the writeC field in the struct below, or I had changed it to never close writeC and forgot to remove this. I will try to figure out which one it's supposed to be

kralicky: Should be fixed now

pkg/ssh/stream.go
```diff
@@ -309,41 +354,84 @@ func (sh *StreamHandler) Run(ctx context.Context) error {
...
+		req.Err <- status.Error(codes.InvalidArgument, err.Error())
+		return
+	}
+	err := sh.auth.EvaluateDelayed(ctx, state.StreamAuthInfo, req.User)
+	if err != nil {
+		lg.Debug().Err(err).Msg("ssh: handoff request denied")
+		req.Err <- status.Error(codes.PermissionDenied, err.Error())
+		return
+	}
+	state.CurrentUser = pendingUser
```
wasaga, line 407: I'm not sure i understand the orchestration here - is `EvaluateDelayed` supposed to use pendingUser or req.User ?

kralicky: state.CurrentUser should have a non-empty username to start, and an empty hostname. For this request to be valid, req.User should have a matching username, and a non-empty hostname. 
The goal is to "promote" the empty hostname into a non-empty hostname, leaving the username the same. If PromoteFrom() succeeds, then both pendingUser and req.User will become identical. But we don't want to directly mutate state.CurrentUser yet in case EvaluateDelayed fails. We also don't want to run evaluation if the user request is invalid.
I'll add a comment here, and I suppose it does make more sense to pass pendingUser to EvaluateDelayed even though it's the same as req.User.

PR #6185, opened 2026-03-10
core/config: add support for ssh policy ids

calebdoxsey:
## Summary
Because of the way the upstream tunnel ssh policy was implemented, it only supported hardcoded PPL, but the consolidated API uses a dedicated policy type, and the enterprise console and zero generate rego from PPL. This PR adds additional upstream tunnel fields to support this.

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/options_test.go
+++ b/config/options_test.go
@@ -1596,8 +1596,11 @@ func TestRoute_FromToProto(t *testing.T) {
 		}
 
 		// if the upstream tunnel ssh policy is set, reset it to a valid policy
-		if pb.UpstreamTunnel != nil && pb.UpstreamTunnel.SshPolicy != nil {
-			pb.UpstreamTunnel.SshPolicy.Raw = []byte(`{"allow":{"or":[{"email":{"is":"user1@example.com"}}]}}`)
+		if pb.UpstreamTunnel != nil {
+			pb.UpstreamTunnel.SshPolicyId = nil
+			if pb.UpstreamTunnel.SshPolicy != nil {
+				pb.UpstreamTunnel.SshPolicy.Raw = []byte(`{"allow":{"or":[{"email":{"is":"user1@example.com"}}]}}`)
+			}
 		}
 
 		return pb


internal/version/components.json
```diff
@@ -1,5 +1,5 @@
 {
-  "config": "v0.22.0",
+  "config": "v0.23.0",
   "hosted-authenticate-oidc": "v0.1.0",
   "mcp": "v0.22.0",
   "ssh": "v0.11.0"
```
kenjenkins: Should we bump the `ssh` component version too? (Although I'm not entirely sure myself.)

authorize/evaluator/policy_evaluator.go
```diff
@@ -201,7 +201,16 @@ func NewPolicyEvaluator(
...
+
+	if configPolicy.UpstreamTunnel != nil && len(configPolicy.UpstreamTunnel.SSHPolicyRego) > 0 {
+		for _, r := range configPolicy.UpstreamTunnel.SSHPolicyRego {
+			q := policyQuery{script: r}
+			if err := q.prepareQuery(ctx, store); err != nil {
+				return nil, err
+			}
+			e.upstreamTunnelQuery = append(e.upstreamTunnelQuery, q)
+		}
+	} else if ppl := configPolicy.UpstreamTunnelPPL(); ppl != nil {
```
kenjenkins, line 213: Will we ever need to support both PPL and Rego at the same time? i.e. should this be its own `if` instead of an `else if`?

kenjenkins (approved):
Thanks for adding this!

authorize/evaluator/policy_evaluator.go
calebdoxsey: Its a little weird, but at least with how we do normal policies the source ppl is what was used to generate the rego, and actually it can't be generated in core because enterprise/zero have criteria that don't exist in core, so I wanted to make it work the same way.

Issue #6180, opened 2026-03-09
Proxy protocol breaks ACME HTTP01 challenges


akhayyat:
## What happened?

I'm seeing the same behavior described in #5403, on OCI (Oracle cloud): enabling proxy protocol while using the network load-balancer causes http (non-https) requests to return 400 errors, breaking ACME HTTP01 challenges.

## What did you expect to happen?

ACME HTTP01 challenge URLs get routed to the challenge solver.

## How'd it happen?

1. Set `useProxyProtocol: true` in the `Pomerium` `global` resource.
2. Enable proxy protocol in the OCI network load-balancer
3. Observe cert-manager `Challenge` status:

    ```
    Status:
      Presented:   true
      Processing:  true
      Reason:      Waiting for HTTP-01 challenge propagation: wrong status code '400', expected '200'
      State:       pending
    ```

Also:

```
curl http://<internal-cm-acme-http-solver-endpoint>/.well-known/acme-challenge/<token> -H "Host: <FQDN>" -i
HTTP/1.1 200 OK

curl http://<load-balancer-ip>/.well-known/acme-challenge/<token> -H "Host: <FQDN>" -i
HTTP/1.1 400 Bad Request
```

## What's your environment like?

- Pomerium version: v0.32.0
- k3s on OCI (Oracle cloud)

## What's your config?

```
apiVersion: ingress.pomerium.io/v1
kind: Pomerium
metadata:
  name: global
spec:
  authenticate:
    url: https://<pomerium-url>
  certificates:
  - pomerium/pomerium-tls
  identityProvider:
    provider: oidc
    scopes:
    - openid
    - profile
    - email
    secret: pomerium/pomerium-oidc
    url: https://<oidc-url>
  secrets: pomerium/bootstrap
  useProxyProtocol: true
```

## What did you see in the logs?

pomerium-proxy
```
{"level":"error","ts":"2026-03-09T02:01:18Z","msg":"Reconciler error","controller":"pomerium-ingress","controllerGroup":"networking.k8s.io","controllerKind":"Ingress","Ingress":{"name":"some-service","namespace":"some-service"},"namespace":"some-service","name":"some-service","reconcileID":"a3ec7640-45f2-446e-b101-59c2e66a99e9","error":"fetch ingress related resources: tls: get secret some-service/some-service-tls: Secret \"some-service-tls\" not found"}
```

cm-acme-http-solver: only a single log record
```
[...]

PR #6191, opened 2026-03-17
core/autocert: detect use_proxy_protocol changes in http redirect server

calebdoxsey:
## Summary
For the http redirect server, in addition to detecting changes to the http redirect address, also detect changes to the `use_proxy_protocol` option.

Previously if only the `use_proxy_protocol` option was changed, we wouldn't create a new proxy protocol listener, resulting in the server expecting standard HTTP traffic, leading to 400 errors.

## Related issues
- #6180 

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/autocert/manager.go
+++ b/internal/autocert/manager.go
@@ -54,6 +54,7 @@ type Manager struct {
 	acmeMgr             atomic.Pointer[certmagic.ACMEIssuer]
 	srv                 *http.Server
 	srvAddr             string
+	srvLi               net.Listener
 	srvUseProxyProtocol bool
 
 	acmeTLSALPNLock     sync.Mutex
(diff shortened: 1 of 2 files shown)


internal/autocert/manager.go
```diff
@@ -316,53 +319,67 @@ func (mgr *Manager) updateAutocert(ctx context.Context, cfg *config.Config) erro
...
+	}
+	mgr.srvAddr = srvAddr
+	mgr.srvUseProxyProtocol = srvUseProxyProtocol
+
+	// close any existing server
 	if mgr.srv != nil {
-		// nothing to do if the address hasn't changed
-		if mgr.srv.Addr == cfg.Options.HTTPRedirectAddr {
-			return
-		}
```
calebdoxsey, line 323: This was the code that was only detecting the change to the address. It now checks both the address and the `UseProxyProtocol` option.

PR #6203, opened 2026-03-18
feat: add allow_upgrades configuration option

namelivia:
## Summary
- Added `allow_upgrades` field to route configuration to support custom HTTP upgrade types (e.g. `tailscale-control-protocol`).
- Updated Envoy configuration generator to include these custom upgrades in `UpgradeConfigs`.
- Ensured stream idle timeouts are disabled and upstream protocol is forced to `HTTP/1.1` when custom upgrades are used.

## Test Plan
- Unit tests in `config/policy_test.go` verify policy validation and proto mapping.
- Unit tests in `config/envoyconfig/routes_test.go` verify correct Envoy route configuration.
- Unit tests in `config/envoyconfig/clusters_test.go` verify ALPN protocols for upstream connections.
- Updated existing tests to account for policy checksum changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:22.22.1-bookworm@sha256:b501c082306a4f528bc4038cbf2fbb58095d583d0419a259b2114b5ac53d12e9 AS ui
+FROM node:22.22.1-bookworm@sha256:f90672bf4c76dfc077d17be4c115b1ae7731d2e8558b457d86bca42aeb193866 AS ui
 WORKDIR /build
 
 COPY .git ./.git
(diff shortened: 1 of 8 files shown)


config/envoyconfig/protocols.go
```diff
@@ -163,8 +163,8 @@ func buildUpstreamALPN(upstreamProtocol upstreamProtocolConfig) []string {
 
 func getUpstreamProtocolForPolicy(_ context.Context, policy *config.Policy) upstreamProtocolConfig {
 	upstreamProtocol := upstreamProtocolAuto
-	if policy.AllowWebsockets {
-		// #2388, force http/1 when using web sockets
+	if policy.AllowWebsockets || len(policy.AllowUpgrades) > 0 {
+		// #2388, force http/1 when using web sockets or custom upgrades
```
namelivia, line 167: Forcing `HTTP/1.1` ensures that the standard, widely-supported `101 Switching Protocols` handshake reaches the upstream server exactly as it expects.

PR #6216, opened 2026-03-26
core/config: add allow_upgrades

calebdoxsey:
## Summary
Add new `allow_upgrades` global and per-route options. This can be used for arbitrary upgrade types, in addition to the existing `allow_websockets` and `allow_spdy`.

These changes include a new code generator that takes the config protobuf and generates local types and conversion functions. The two types are `RouteOptions` and `GlobalOptions` which are embedded with a `squash` tag in the existing `Options` and `Policy` structs. This way they are populated via viper. The generated conversion functions handle converting from and to the corresponding protobuf type.

## Related issues
- [ENG-3766](https://linear.app/pomerium/issue/ENG-3766/supporting-routing-tailscaleheadscale-traffic-by-allowing-custom-http)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/envoyconfig/routes_test.go
+++ b/config/envoyconfig/routes_test.go
@@ -1193,7 +1193,7 @@ func Test_buildPolicyRoutes(t *testing.T) {
 					"upgradeConfigs": [
 						{ "enabled": false, "upgradeType": "websocket"},
 						{ "enabled": false, "upgradeType": "spdy/3.1"},
-						{ "enabled": true, "upgradeType": "CONNECT", "connectConfig": {} }
+						{ "enabled": true, "upgradeType": "connect", "connectConfig": {} }
 					]
 				},
 				"requestHeadersToRemove": [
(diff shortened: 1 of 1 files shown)


internal/httputil/upgrade.go
```diff
@@ -0,0 +1,9 @@
+package httputil
+
+// upgrade types
+const (
+	UpgradeTypeConnectTCP = "CONNECT"
+	UpgradeTypeConnectUDP = "CONNECT-UDP"
+	UpgradeTypeWebsocket  = "websocket"
```
desimone, line 7: HTTP upgrade types are case-insensitive per RFC 7230.

In a few places we use slices.Contains directly. Maybe we could normalize the value at parse time or change the checks to use strings.EqualFold?

calebdoxsey: Updated to be case-insensitive.

PR #6224, opened 2026-03-31
core/pebble: use pebbleutil.Iterate for iterators, add reversed

calebdoxsey:
## Summary
To ensure that the pebble iterator is always closed, use `pebbleutil.Iterate` everywhere and add a `pebbleutil.IterateReversed` for going in reverse order.

## Related issues
For [ENG-3806](https://linear.app/pomerium/issue/ENG-3806/corepebble-iterator-not-always-closed)

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -359,3 +359,8 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20260203192932-546029d2fa20 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+ignore (
+	./ui/node_modules
+	./ui/src
+)


pkg/pebbleutil/pebbleutil.go
```diff
@@ -15,6 +15,51 @@ import (
...
+			return it.First()
+		}
+		return it.Last()
+	}
+	seekNext := func(it *pebble.Iterator) bool {
+		if forward {
+			return it.Next()
+		}
+		return it.Prev()
+	}
```
kenjenkins: Would it be any simpler if we pulled the `if forward` checks out of `seekFirst` and `seekNext`, like this?

```suggestion
	var seekFirst, seekNext func(*pebble.Iterator) bool
	if forward {
		seekFirst = (*pebble.Iterator).First
		seekNext = (*pebble.Iterator).Next
	} else {
		seekFirst = (*pebble.Iterator).Last
		seekNext = (*pebble.Iterator).Prev
	}
```

PR #6238, opened 2026-04-02
ci: skip Cloudsmith publish for release candidates

desimone:
## Summary

- RC versions were being published to apt/rpm repos via Cloudsmith, causing `apt-get upgrade` to install pre-release builds
- Uses `github.event.release.prerelease` flag to skip Cloudsmith publish for any prerelease (RC, beta, alpha, etc.) -- relies on GitHub release metadata rather than tag string matching
- Mac-builds dispatch is intentionally NOT gated -- prereleases still need macOS binaries; the mac-builds workflow itself gates the Homebrew tap via `get-latest-release`
- Docker `latest` tag filtering was already in place and is unchanged
- Existing RC packages are left in place to avoid breaking users who already installed them

Related: pomerium/cli#655

## Related PRs

- pomerium/cli#656 -- same Cloudsmith fix for CLI
- pomerium/mac-builds#16 -- Homebrew tap guard for CLI releases
- pomerium/desktop-client#502 -- Homebrew tap guard for desktop releases

## Test plan

- [x] `actionlint` passes
- [x] CI passes on this PR
- [ ] Next RC release skips Cloudsmith publish in Actions log
- [ ] Next stable release still publishes to Cloudsmith normally

AI-assisted: Claude drafted the workflow change; verified with actionlint.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -81,7 +81,7 @@ jobs:
           APPARITOR_GITHUB_TOKEN: ${{ secrets.APPARITOR_GITHUB_TOKEN }}
 
       - name: trigger mac build
-        if: github.event_name == 'release'
+        if: github.event_name == 'release' && !github.event.release.prerelease
         uses: peter-evans/repository-dispatch@28959ce8df70de7be546dd1250a005dd32156697
         with:
           repository: pomerium/mac-builds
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -81,7 +81,7 @@ jobs:
           APPARITOR_GITHUB_TOKEN: ${{ secrets.APPARITOR_GITHUB_TOKEN }}
 
       - name: trigger mac build
-        if: github.event_name == 'release'
+        if: github.event_name == 'release' && !github.event.release.prerelease
```
kenjenkins, line 84: We should still build mac binaries here though, right?

Issue #6246, opened 2026-04-08
HTTP/3 / QUIC + macOS + SO_REUSEPORT + concurrency issue


IngmarStein:
## What happened?

I've enabled HTTP/3 by setting

```
codec_type: http3
```

so that Envoy listens on both TCP and UDP. At startup, I receive this error:

```
{"level":"error","service":"envoy","name":"config","time":"2026-04-08T12:53:48+02:00","message":"gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) quic-ingress: Listening on UDP when concurrency is > 1 without the SO_REUSEPORT socket option results in unstable packet proxying. Configure the reuse_port listener option or set concurrency = 1.\\n"}
```

## What did you expect to happen?

#5066 disabled `SO_REUSEPORT` for all listeners on non-Linux OSes. https://github.com/envoyproxy/envoy/blob/e166852c46ba694ce270f062716f41df887252aa/api/envoy/config/listener/v3/listener.proto#L354-L376 mentions that on macOS "For QUIC/H3, SW routing will send packets to other workers.".

Should `EnableReusePort` only be disabled for TCP listeners on Mac and Windows?

## How'd it happen?

1. Ran `/opt/homebrew/bin/pomerium --config /opt/homebrew/etc/pomerium.yaml`
2. Checked the logs
3. Saw the error

## What's your environment like?

- Pomerium version (retrieve with `pomerium --version`): 0.32.4+8beb67872
- Server Operating System/Architecture/Cloud: macOS 26.4, installed via Homebrew

## What's your config.yaml?

```config.yaml
address: "[::]:9443"
metrics_address: "127.0.0.1:9091"
codec_type: http3

routes:
  - from: https://grafana.example.com
    to: http://[::1]:3000
[…]
 ```

## What did you see in the logs?

```logs
{"level":"error","service":"envoy","name":"config","time":"2026-04-08T12:53:48+02:00","message":"gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) quic-ingress: Listening on UDP when concurrency is > 1 without the SO_REUSEPORT socket option results in unstable packet proxying. Configure the reuse_port listener option or set concurrency = 1.\\n"}
```

## Additional context

`GOMAXPROCS` is unset, i.e. I haven't manually configured concurrency.

IngmarStein:
[...]

PR #6259, opened 2026-04-10
feat: add unix socket support for target resources

IngmarStein:
## Summary

Add support for unix:// and unix+https:// URL schemes to route to Unix domain sockets as upstream destinations.

## Related issues

- #6246

## User Explanation

Pomerium currently only supports TCP/UDP host-based upstreams. Many services (Docker, PostgreSQL, Redis, Nginx, etc.) expose Unix domain sockets as an alternative to network ports.

Benefits:

1. Direct socket access - Route to Unix sockets like unix:///var/run/docker.sock without needing a network port, reducing attack surface and latency.
2. TLS support - The unix+https:// scheme enables TLS-encrypted connections to Unix socket services that support HTTPS over Unix sockets (e.g., nginx with proxy_protocol).
3. Protocol auto-detection - unix:// uses HTTP/1.1, while unix+https:// leverages existing ALPN-based protocol detection (HTTP/1.1 or HTTP/2).
4. Security - Unix sockets can use filesystem permissions for access control instead of network-level restrictions.
5. Consistent with existing patterns - Follows the same scheme mixing rules as TCP/UDP upstreams.

Example use case:
```
routes:
  - from: https://secure.internal.example.com
    to: https+unix:///run/nginx-tls.sock
```

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/envoyconfig/clusters.go
+++ b/config/envoyconfig/clusters.go
@@ -336,7 +336,7 @@ func (b *Builder) buildPolicyTransportSocket(
 	policy *config.Policy,
 	dst url.URL,
 ) (*envoy_config_core_v3.TransportSocket, error) {
-	if dst.Scheme != "https" {
+	if dst.Scheme != "https" && dst.Scheme != "unix+https" {
 		return nil, nil
 	}
 
@@ -523,14 +523,19 @@ func (b *Builder) buildLbEndpoints(endpoints []Endpoint) ([]*envoy_config_endpoi
 			u.Host = strings.ReplaceAll(e.url.Host, "localhost", "127.0.0.1")
 		}
 
-		endpoint := &envoy_config_endpoint_v3.Endpoint{
-			Hostname: e.url.Host,
+		var addr *envoy_config_core_v3.Address
+		if urlutil.IsUnixScheme(e.url.Scheme) {
+			addr = buildPipeAddress(e.url.Path, 0o600)
+		} else {
+			addr, _ = buildTCPListenAddresses(u.Host, defaultPort)
 		}
-		endpoint.Address, _ = buildTCPListenAddresses(u.Host, defaultPort)
 
 		lbe := &envoy_config_endpoint_v3.LbEndpoint{
 			HostIdentifier: &envoy_config_endpoint_v3.LbEndpoint_Endpoint{
-				Endpoint: endpoint,
+				Endpoint: &envoy_config_endpoint_v3.Endpoint{
+					Address:  addr,
+					Hostname: e.url.Host,
+				},
 			},
 			LoadBalancingWeight: e.loadBalancerWeight,
 		}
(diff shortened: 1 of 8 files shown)


config/envoyconfig/envoyconfig.go
```diff
@@ -148,6 +148,22 @@ func buildTCPListenAddresses(
...
+				Mode: mode,
+			},
+		},
+	}
+}
+
+func buildTCPAddress(hostport string, defaultPort uint32) *envoy_config_core_v3.Address {
+	mainAddress, _ := buildListenAddresses(envoy_config_core_v3.SocketAddress_TCP, hostport, defaultPort)
+	return mainAddress
+}
```
calebdoxsey:
```suggestion
```

I think this function isn't used.

config/envoyconfig/envoyconfig.go
```diff
@@ -148,6 +148,22 @@ func buildTCPListenAddresses(
 	return buildListenAddresses(envoy_config_core_v3.SocketAddress_TCP, hostport, defaultPort)
 }
 
+func buildPipeAddress(path string, mode uint32) *envoy_config_core_v3.Address {
+	return &envoy_config_core_v3.Address{
+		Address: &envoy_config_core_v3.Address_Pipe{
+			Pipe: &envoy_config_core_v3.Pipe{
+				Path: path,
+				Mode: mode,
```
calebdoxsey: I think `mode` is only used for listeners... and apparently its an error if its used with abstract pipes. Would it work to just leave it unset?

config/envoyconfig/envoyconfig.go
IngmarStein: Oh thanks, I think this was a leftover from a merge conflict.

config/envoyconfig/envoyconfig.go
IngmarStein: Done

.claude/settings.local.json
calebdoxsey, line 1: Could we remove this file?

IngmarStein: Definitely! Done

calebdoxsey (approved):
LGTM

config/envoyconfig/clusters.go
```diff
@@ -336,7 +336,7 @@ func (b *Builder) buildPolicyTransportSocket(
 	policy *config.Policy,
 	dst url.URL,
 ) (*envoy_config_core_v3.TransportSocket, error) {
-	if dst.Scheme != "https" {
+	if dst.Scheme != "https" && dst.Scheme != "unix+https" {
```
kenjenkins: At least for "From" URLs, I think we use `tcp+https` to mean "TCP over HTTPS". If this is HTTPS over a Unix socket should this be `https+unix` instead?

I was trying to find any prior art and I did come across this library which seems to use a similar convention: https://pypi.org/project/requests-unixsocket/.

IngmarStein: There's no standard, as far as I can tell . Cloudflare uses `unix+tls`: https://developers.cloudflare.com/cloudflare-one/networks/connectors/cloudflare-tunnel/routing-to-tunnel/protocols/. Happy to adjust to whatever you think works best for this codebase.

kenjenkins: Yes, looks like there are many different variations. I also did a quick GitHub code search:

| query | results |
| --- | ---: |
| "unix+http://"      | 197 files  |
| "unix+https://"   | 548 files   |
| "http+unix://" | 4,300 files |
| "https+unix://" | 222 files |

@calebdoxsey do you have a preference on this? To me `https+unix` feels more consistent with what we have so far, but it's not a strong preference.

calebdoxsey: Sure we could change it to make it more consistent with TCP.

IngmarStein: SGTM. Done

calebdoxsey:
Thanks @IngmarStein for your contribution!

kralicky:
I suppose we could now implement the `forwarded-streamlocal` openssh extension to allow reverse tunneling sockets :thinking:

PR #6265, opened 2026-04-13
feat(blob) : middleware and reading util

alexandreLamarre:
## Summary

Wraps Read, Write and List operations with middleware that allows for customizing per-operation blob options using the BeforeRead, BeforeWrite and BeforeList hooks, as well as injecting context values for the context used by blob operations.

Adds default middleware for adding `User-Agent` (or-equivalent) headers identifying the pomerium version and the access_id,for correlation with enterprise audit logs. This information is configured to also be stored in each cloud provider's blob audit logs.
 
## Related issues

Part of [ENG-3804](https://linear.app/pomerium/issue/ENG-3804/recordings-audit-logging)

## User Explanation

Adds audit logging middleware for blob storage.

## Checklist

- [X] updated unit tests
- [X] ready for review


pkg/storage/blob/middleware/middleware.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+type (
+	ListMiddleware   func(*ListOp) error
+	WriterMiddleware func(*WriteOp) error
+	ReadMiddleware   func(*ReadOp) error
+)
+
+var (
+	defaultMu               sync.Mutex
```
calebdoxsey: This mutex would need to be held on read too for it to be effective...

alexandreLamarre: I think that is avoided by only registering default middleware in `init` blocks

alexandreLamarre: instead of having that pattern I can always just write functions like `DefaultXXXMiddleware() []XXXMiddleware`. It means console will have to pass additional middleware to each constructor if it wishes to set defaults, but maybe that is more readable. WDYT?

pkg/storage/blob/providers/middleware.go
```diff
@@ -0,0 +1,174 @@
...
+		base := o.HTTPClient
+		if base == nil {
+			base = http.DefaultClient
+		}
+		o.HTTPClient = &withExtraParams{
+			base: base,
+			params: [][2]string{
+				{key, val},
+			},
+		}
```
calebdoxsey, line 100: One thing that's bit me before is that anything that uses the OAuth2 library uses a custom http client... do we have to worry about that?

calebdoxsey (approved):

pkg/storage/blob/middleware/middleware.go
calebdoxsey: Can we just remove the mutex?

pkg/storage/blob/providers/middleware.go
alexandreLamarre: maybe. Let me take another look at the s3 SDK because I do believe they use a different default than `http.DefaultClient`. My understanding is that that client should always be set anyways in our setup when we get to the point of modifying it (the url Opener handles the default http client construction based on whatever credentials are supplied), but http.DefaultClient is liable to cause issues if we do hit that code path

pkg/storage/blob/middleware/middleware.go
alexandreLamarre: oh right there's no data-race because init's are sequential, yes let's do that

pkg/storage/blob/providers/middleware.go
alexandreLamarre: I replaced this with the default AWS SDK buildable client, which in theory should allow for the SDK configurations to be correctly propagated in all cases.

I'll make a ticket to verify a variety of configurations with s3 storage as we wrap up session recording

PR #6266, opened 2026-04-13
fix: validate GCS metadata identity response before using as token

desimone:
## Summary

- reject non-200 responses from the GCP metadata identity endpoint instead of using the response body as a bearer token
- reject empty metadata tokens after trimming whitespace
- include the audience and a truncated response body in errors to improve diagnostics
- add regression coverage around unexpected metadata responses and header generation

## Context

While investigating an environment-specific failure mode, we found that unexpected responses from the metadata identity endpoint could flow into the authorization-header path. This PR hardens that code path so bad metadata responses are treated as errors instead of token material.

This is a correctness and defense-in-depth fix. It should not be read as a claim that this PR alone explains every aspect of the incident being investigated.

## Test plan

- [x] success path still returns a token
- [x] non-200 metadata responses return an error
- [x] empty metadata responses return an error
- [x] header-generation path does not emit headers when token retrieval fails
- [x] existing evaluator coverage still passes
- [x] `make build`
- [x] `make test`
- [x] `make lint`

## AI Assistance

AI helped draft the initial diagnosis and test coverage. I manually reviewed the change, validated the behavior against a live environment, and reran the repository validation commands.

### DIFF
--- a/authorize/evaluator/google_cloud_serverless_test.go
+++ b/authorize/evaluator/google_cloud_serverless_test.go
@@ -7,41 +7,102 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/oauth2"
 )
 
-func withMockGCP(t *testing.T, f func()) {
+func withMockGCP(t *testing.T, handler http.HandlerFunc) (cleanup func()) {
+	t.Helper()
 	originalGCPIdentityDocURL := GCPIdentityDocURL
-	defer func() {
-		GCPIdentityDocURL = originalGCPIdentityDocURL
-		GCPIdentityNow = time.Now
-	}()
+	originalNow := GCPIdentityNow
 
 	now := time.Date(2020, 1, 1, 1, 0, 0, 0, time.UTC)
 	GCPIdentityNow = func() time.Time {
 		return now
 	}
 
-	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		assert.Equal(t, "Google", r.Header.Get("Metadata-Flavor"))
-		assert.Equal(t, "full", r.URL.Query().Get("format"))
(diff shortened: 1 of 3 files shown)


authorize/evaluator/google_cloud_serverless_test.go
```diff
@@ -7,41 +7,99 @@ import (
...
+		// This is the exact scenario that caused the staging 503:
+		// metadata server returns 404, body gets used as Bearer token,
+		// Envoy rejects the invalid header value.
+		withMockGCP(t, func(w http.ResponseWriter, _ *http.Request) {
+			w.WriteHeader(http.StatusNotFound)
+			_, _ = w.Write([]byte("not found\n"))
+		})
+
+		headers, err := getGoogleCloudServerlessHeaders("", "https://example.run.app")
+		assert.Error(t, err)
```
kenjenkins: I think we can even verify the expected error message?

```suggestion
		assert.ErrorContains(t, err, "error retrieving google cloud serverless token: "+
			`metadata identity endpoint returned HTTP 404 for audience "https://example.run.app": `+
			"not found")
```

authorize/evaluator/google_cloud_serverless_test.go
```diff
@@ -7,41 +7,99 @@ import (
...
+
+		token, err := src.Token()
+		assert.Nil(t, token)
+		assert.ErrorContains(t, err, "empty token")
+	})
+
+	t.Run("non-200 prevents invalid Authorization header", func(t *testing.T) {
+		// This is the exact scenario that caused the staging 503:
+		// metadata server returns 404, body gets used as Bearer token,
+		// Envoy rejects the invalid header value.
```
kenjenkins: I think this might not make sense without some additional context. What about something like this?

```suggestion
		// If the token response contains an error, it should be
		// propagated (not set as the token value).
```

kenjenkins (approved):
Nice fix!

PR #6272, opened 2026-04-15
fix: apply TLSUpstreamServerName to health checks via transport_socket_match_criteria

IngmarStein:
## Summary

When `tls_upstream_server_name` is configured on a route with health checks, ensure Envoy's health checker uses the correct TLS SNI by setting transport_socket_match_criteria on the health check configuration.

## User Explanation

This prevents SSL errors like 'no alternative certificate subject name matches target hostname' when the upstream certificate is issued for the route's domain rather than localhost.

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/envoyconfig/clusters.go
+++ b/config/envoyconfig/clusters.go
@@ -251,7 +251,24 @@ func (b *Builder) buildPolicyCluster(ctx context.Context, cfg *config.Config, po
 	cluster.OutlierDetection = policy.OutlierDetection.ToEnvoy()
 	cluster.HealthChecks = nil
 	for _, hc := range policy.HealthChecks {
-		cluster.HealthChecks = append(cluster.HealthChecks, hc.ToEnvoy())
+		envoyHC := hc.ToEnvoy()
+
+		// If TLSUpstreamServerName is set and we have TLS transport socket matches,
+		// set transport_socket_match_criteria so health checks use the correct TLS SNI
+		if policy.TLSUpstreamServerName != "" && len(cluster.TransportSocketMatches) > 0 {
+			for _, tsm := range cluster.TransportSocketMatches {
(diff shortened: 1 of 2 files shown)


kenjenkins (commented):
I'm not sure I understand... I tried to reproduce this scenario using a toy Go server that logs incoming connection SNI and request paths, together with a Pomerium route configured like this:

```yaml
routes:
 - from: https://demo.localhost.pomerium.io
   to: https://localhost:7443
   allow_any_authenticated_user: true
   tls_upstream_server_name: sni.localhost.pomerium.io
   health_checks:
     - timeout: 15s
       interval: 15s
       unhealthy_threshold: 3
       healthy_threshold: 1
       http_health_check:
         path: /healthz
         codec_client_type: HTTP2
```

I observed that the `tls_upstream_server_name` applied to both proxied requests and for health check requests.

Are you seeing a different behavior? What does your route configuration look like?

IngmarStein:
@kenjenkins apologies, I think this looks like a user error.

I combined `allow_websockets: true` (which restricts the route to HTTP/1.1) with `http_health_check.codec_client_type: HTTP2` which sets ALPN: h2:

```
  - from: https://grafana.example.com
    to: https://localhost:3000
    name: "grafana"
    description: "Grafana monitoring dashboards"
    tls_upstream_server_name: grafana.example.com
    allow_websockets: true
    preserve_host_header: true
    pass_identity_headers: true
    health_checks:
      - timeout: 10s
        interval: 60s
        healthy_threshold: 1
        unhealthy_threshold: 3
        http_health_check:
          path: /api/health
          codec_client_type: HTTP2
          expected_statuses:
            - start: 200
              end: 400
```

Consequently, the route is working, but not the health check and the `no alternative certificate subject name matches target hostname` error might have been a red herring.

PR #6277, opened 2026-04-17
Add envoy_concurrency config option

IngmarStein:
## Summary

Adds a new `envoy_concurrency` configuration option that allows setting Envoy's `--concurrency` flag independently of GOMAXPROCS. This is useful for platforms like macOS with Homebrew where the launchd environment cannot be modified and GOMAXPROCS cannot be controlled.

Fixes the QUIC/http3 issue on macOS where Envoy requires concurrency=1 due to SO_REUSEPORT not being supported for UDP listeners.

## Related issues

- #6246

## User Explanation

On macOS with Homebrew, you can now set `envoy_concurrency: 1` in your config to enable QUIC/http3 support without needing to modify the launchd environment.

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/options.go
+++ b/config/options.go
@@ -290,6 +290,11 @@ type Options struct {
 	EnvoyAdminAddress            string    `mapstructure:"envoy_admin_address" yaml:"envoy_admin_address"`
 	EnvoyBindConfigSourceAddress string    `mapstructure:"envoy_bind_config_source_address" yaml:"envoy_bind_config_source_address,omitempty"`
 	EnvoyBindConfigFreebind      null.Bool `mapstructure:"envoy_bind_config_freebind" yaml:"envoy_bind_config_freebind,omitempty"`
+	// EnvoyConcurrency sets the number of worker threads for the Envoy process.
+	// This can be used to work around issues on platforms where SO_REUSEPORT
+	// is not supported for certain listener types (e.g., QUIC on macOS).
+	// If unset, Envoy will use its default (number of CPU cores).
+	EnvoyConcurrency *uint32 `mapstructure:"envoy_concurrency" yaml:"envoy_concurrency,omitempty"`
 
 	// ProgrammaticRedirectDomainWhitelist restricts the allowed redirect URLs when using programmatic login.
 	ProgrammaticRedirectDomainWhitelist []string `mapstructure:"programmatic_redirect_domain_whitelist" yaml:"programmatic_redirect_domain_whitelist,omitempty" json:"programmatic_redirect_domain_whitelist,omitempty"`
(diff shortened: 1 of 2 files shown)


kralicky:
Having this option may be useful still, but is the underlying issue that we are incorrectly configuring udp listeners on macos? Having to limit envoy to 1 worker doesn't sound ideal either.

IngmarStein:
@kralicky maybe that belongs to https://github.com/pomerium/pomerium/issues/6246, but as far as I can tell, it's the least bad option for now (I would love not having to restrict concurrency, though!).

Let's look at https://github.com/envoyproxy/envoy/blob/e166852c46ba694ce270f062716f41df887252aa/api/envoy/config/listener/v3/listener.proto#L354-L376:

```
  // * On macOS, reuse_port for TCP does not do what it does on Linux. Instead of load balancing,
  //   the last socket wins and receives all connections/packets. For TCP, reuse_port is force
  //   disabled and the user is warned. For UDP, it is enabled, but only one worker will receive
  //   packets. For QUIC/H3, SW routing will send packets to other workers. For "raw" UDP, only
  //   a single worker will currently receive packets.
  // * On Windows, reuse_port for TCP has undefined behavior. It is force disabled and the user
  //   is warned similar to macOS. It is left enabled for UDP with undefined behavior currently.
```

Sounds like it should never be enabled on Windows, but for QUIC/H3,`SO_REUSEPORT` might actually work thanks to "SW routing"? If so, https://github.com/pomerium/pomerium/pull/5066 may have been too coarse.

wasaga:
#6284 runs http3 downstream on macOS 26.4 arm64 with default concurrency — passes on current main, fails when #6260 is reverted. Maybe the knob isn't needed on macOS after all?

IngmarStein:
Thanks for the integration test! I knew it worked in a smoke test, but I was worried how Envoy behaves under load when distributing packets among workers. Given the warning message, it seemed safest to restrict concurrency, but maybe the "For QUIC/H3, SW routing will send packets to other workers." part addresses that concern fully.

wasaga:
Hi @IngmarStein we are trying to keep config surface lean and have less knobs when possible. Are fixes in `main` working for you or you believe this config option is still required?

IngmarStein:
@wasaga that's a good goal. I can't say if the fixes in `main` are working until this repo imports a newer envoy version which includes https://github.com/pomerium/envoy-custom/pull/221 which is supposed to fix https://github.com/pomerium/pomerium/issues/6301. If you don't see any utility of a `envoy_concurrency` setting beyond the use case mentioned here, I'm happy to close this PR.

PR #6283, opened 2026-04-20
feat(envoyconfig):  load dynamic extensions, configure session recording policy & serve recording server

alexandreLamarre:
## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/internal/recording/protocol.go
+++ b/internal/recording/protocol.go
@@ -1,0 +1,370 @@
+package recording
+
+import (
+	"context"
+	//nolint:gosec
+	"crypto/md5"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	gblob "gocloud.dev/blob"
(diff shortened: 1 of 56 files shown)


internal/recording/protocol.go
```diff
@@ -0,0 +1,282 @@
...
+		default:
+		}
+		msg, err := t.Recv(ctx)
+		if errors.Is(err, io.EOF) {
+			return nil
+		}
+		if err != nil {
+			log.Ctx(ctx).Err(err).Msg("failed to receive message from session recording client")
+			goto RETRY
+		}
```
alexandreLamarre: Still need to look into what errors to handle in each transport impl that are recoverable and maybe surface them as strongly typed errors here to distinguish retryable versus non-retryable

alexandreLamarre:
Superceded by other PRs.

PR #6286, opened 2026-04-21
feat(mcp/configapi): shared library exposing ConfigService as MCP tools

wasaga:
Adds a library that exposes pomerium's ConfigService proto as Model Context Protocol tools, with redaction of fields marked sensitive.

Every ConfigService RPC becomes an MCP tool. A new `(sensitive) = true` field option marks values that the library scrubs from responses, preserves across partial updates, and omits from generated input schemas.

The library has no transport, authentication, or listener of its own. It is consumed by the zero and console MCP servers. The core integration that mounts an MCP server inside pomerium will land in a follow-up.

Linear: [ENG-2506](https://linear.app/pomerium/issue/ENG-2506/create-an-mcp-server-for-the-pomerium-api)


internal/controlplane/server_mcp.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/pkg/httputil"
+	"github.com/pomerium/pomerium/pkg/mcp/configapi"
+)
+
+// runMCPSupervisor owns the MCP ConfigService listener lifecycle. It
+// reconciles the active listener against cfg.Options.MCPAddress on every
+// signal from update(): binds/rebinds/unbinds transparently. The goroutine
+// terminates when ctx is canceled.
```
kenjenkins: What does "The goroutine" refer to here? Should this say something more like "This method returns when ctx is canceled."?

internal/controlplane/server.go
```diff
@@ -417,6 +422,9 @@ func (srv *Server) Run(ctx context.Context) error {
 		}
 	})
 
+	// manage the optional MCP ConfigService listener
```
kenjenkins: The debug listener and metrics listener are also optional, aren't they? Can you help me understand why the MCP listener needs a special supervisor to handle it?

internal/controlplane/server_mcp.go
```diff
@@ -0,0 +1,138 @@
...
+// in-memory ConfigService request as a Pomerium-prefixed bearer token. This
+// lets the downstream securedServer.authorize pass without bypassing auth.
+//
+// Any failure to obtain or sign the shared key is a gross misconfiguration
+// (no shared key configured, signing primitive broken). The stamp returns
+// a non-nil error in that case; the caller refuses the tool dispatch and
+// surfaces a structured error to the MCP client, instead of silently
+// sending an unauthenticated request that the downstream rejects with a
+// generic auth failure the operator cannot diagnose.
+func (srv *Server) newSharedKeyStamp() configapi.RequestStamp {
```
kenjenkins: newSharedKeyStamp() and signSharedKeyJWT() look fairly similar to the logic in grpcutil.WithSignedJWT(). Is it feasible to reuse or consolidate some of this?

wasaga:
I reduced this PR to shared code only, will open a dedicated PR for core-side MCP listener.

internal/controlplane/server.go
wasaga: they are different; they always start at random ports and then we update envoy xds config to either proxy to them or not. but a fair point.

pkg/mcp/configapi/handler.go
```diff
@@ -0,0 +1,252 @@
...
+	stamps                  []RequestStamp
+	skip                    map[string]bool
+	metaContributors        []MetaContributor
+	errMappers              []ErrorMapper
+	serverMutators          []ServerMutator
+	inputSchemaContributors []InputSchemaContributor
+	preCalls                []PreCall
+}
+
+// RequestStamp mutates an in-memory Connect request before it is dispatched.
```
kenjenkins: Is "stamp" a standard MCP term? I might expect this be called a "RequestModifier" (or something to make it clear that it modifies the request — "stamp" to me sounds more like the thing you would attach to a request, like an authentication JWT).

pkg/mcp/configapi/annotations.go
```diff
@@ -0,0 +1,49 @@
...
+// annotationsForMethod returns MCP tool annotations based on the RPC method name prefix.
+// The naming convention in ConfigService is consistent:
+//   - Get*, List* → read-only
+//   - Create*     → additive (not destructive)
+//   - Update*     → idempotent mutation (not destructive)
+//   - Delete*     → destructive, idempotent
+func annotationsForMethod(methodName string) *mcp.ToolAnnotations {
+	f := false
+	t := true
+	openWorld := false
```
kenjenkins: Would it be more readable to just inline `new(true)` and `new(false)` as needed below?

pkg/mcp/configapi/configapi_test.go
```diff
@@ -0,0 +1,695 @@
...
+
+	tools, err := session.ListTools(ctx, nil)
+	require.NoError(t, err)
+
+	names := map[string]*mcp.Tool{}
+	for _, tool := range tools.Tools {
+		names[tool.Name] = tool
+	}
+
+	expected := []string{
```
kenjenkins, line 68: If we add methods to the API, do we intend to keep this `expected` list up to date with the full set of methods?

If so, should we replace the `Contains` and `NotContains` assertions with a single assertion like this?
```
assert.ElementsMatch(t, expected, slices.Collect(maps.Keys(names)))
```

kenjenkins (approved):

PR #6287, opened 2026-04-21
feat : minor recording storage schema improvements

alexandreLamarre:
## Summary

- Adds a missing version prefix for the schema. 

- Collapses root level metadata/sig/manifest into the object_dir to reduce the number of objects that needed to be listed & makes the objects stored more human readable.

new schema looks like:
```
	<cluster-id>/<recording-type>/<version>/<key>
	|-- metadata.json
	|-- metadata.proto
	|-- manifest
	|-- recording_0000000000.json
	|-- recording_XXXXXXXXXX.json
```

- Adds helpers to protoutil package for marshalling and unmarshalling arrays of protobuf / protojson arrays. These used to be defined in pomerium-console, but make more sense in core for some upcoming work.

## Related issues

Pre-cursor to [ENG-3848](https://linear.app/pomerium/issue/ENG-3848/session-recordings-contents-need-to-be-searchable-indexable-by-storage).

We'll probably want to rewrite in https://github.com/pomerium/pomerium/pull/6283 proto-delimited arrays to protojson delimited arrays

## User Explanation

N/A

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/pkg/protoutil/delim.go
+++ b/pkg/protoutil/delim.go
@@ -1,0 +1,98 @@
+package protoutil
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"reflect"
+
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/encoding/protowire"
+	"google.golang.org/protobuf/proto"
+)
+
+func MarshalLenghDelimitedProtojson[T proto.Message](msgs []T) ([]byte, error) {
+	var buf []byte
+	opts := protojson.MarshalOptions{
+		Multiline:         false,
+		Indent:            "",
+		AllowPartial:      false,
+		UseProtoNames:     true,
+		UseEnumNumbers:    false,
+		EmitUnpopulated:   true,
+		EmitDefaultValues: true,
+	}
+	for _, msg := range msgs {
+		data, err := opts.Marshal(msg)
+		if err != nil {
+			return nil, err
+		}
+		buf = append(buf, data...)
+		buf = append(buf, []byte("\n")...)
+	}
+	return buf, nil
+}
+
+func UnmarshalLengthDelimitedProtojson[T proto.Message](data []byte) ([]T, error) {
+	s := bufio.NewScanner(bytes.NewReader(data))
+	var msgs []T
+	for s.Scan() {
+		line := s.Bytes()
+		if len(bytes.TrimSpace(line)) == 0 {
+			continue
+		}
+		msg := newProtoMessage[T]()
+		if err := protojson.Unmarshal(line, msg); err != nil {
+			return nil, err
+		}
+		msgs = append(msgs, msg)
+	}
+	return msgs, nil
+}
+
+// MarshalLengthDelimited encodes a slice of proto messages into a single byte
+// buffer using varint-length-delimited framing.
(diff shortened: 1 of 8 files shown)


pkg/protoutil/delim.go
```diff
@@ -0,0 +1,98 @@
...
+		Multiline:         false,
+		Indent:            "",
+		AllowPartial:      false,
+		UseProtoNames:     true,
+		UseEnumNumbers:    false,
+		EmitUnpopulated:   true,
+		EmitDefaultValues: true,
+	}
+	for _, msg := range msgs {
+		data, err := opts.Marshal(msg)
```
calebdoxsey, line 26: Should you use MarshalAppend?

pkg/protoutil/delim.go
```diff
@@ -0,0 +1,98 @@
...
+	"bytes"
+	"fmt"
+	"reflect"
+
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/encoding/protowire"
+	"google.golang.org/protobuf/proto"
+)
+
+func MarshalLenghDelimitedProtojson[T proto.Message](msgs []T) ([]byte, error) {
```
calebdoxsey, line 14: There's a typo... but it should be `MarshalNewLineDelimitedProtoJSON`?

pkg/protoutil/delim.go
```diff
@@ -0,0 +1,98 @@
...
+			return nil, err
+		}
+		buf = append(buf, data...)
+		buf = append(buf, []byte("\n")...)
+	}
+	return buf, nil
+}
+
+func UnmarshalLengthDelimitedProtojson[T proto.Message](data []byte) ([]T, error) {
+	s := bufio.NewScanner(bytes.NewReader(data))
```
calebdoxsey, line 37: The `bufio.Scanner` has an unfortunate behavior in that it will error out on lines that are too long... probably better just to use indexing

pkg/protoutil/delim.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	return msgs, nil
+}
+
+// MarshalLengthDelimited encodes a slice of proto messages into a single byte
+// buffer using varint-length-delimited framing.
+func MarshalLengthDelimited[T proto.Message](msgs []T) ([]byte, error) {
+	var buf []byte
+	for _, msg := range msgs {
+		data, err := proto.Marshal(msg)
```
calebdoxsey, line 58: Probably should use `MarshalAppend`

calebdoxsey (changes requested):

pkg/protoutil/delim.go
alexandreLamarre: There's apparently a `protodelim` package that abstracts away this behaviour, so I'll remove the raw proto delim marshalling with that standard package

pkg/protoutil/delim.go
alexandreLamarre: good catch

pkg/protoutil/delim.go
alexandreLamarre: https://pkg.go.dev/google.golang.org/protobuf/encoding/protodelim

calebdoxsey (approved):

kralicky (approved):

PR #6303, opened 2026-04-26
fix(mcp): loopback redirect URI matching per RFC 8252 §7.3

### LINKED ISSUE #6209: MCP OAuth: loopback redirect URIs should ignore port per RFC 8252 Section 7.3
## Summary

MCP OAuth authorization fails for CLI clients (e.g. Claude Code) because `ValidateRedirectURI` performs strict string comparison and does not implement RFC 8252 Section 7.3 loopback redirect URI port-agnostic matching.

## Steps to Reproduce

1. Configure Pomerium with MCP enabled (`runtimeFlags: { mcp: true }`) and `mcpAllowedClientIdDomains: ["claude.ai"]`
2. Create an MCP server Ingress with `ingress.pomerium.io/mcp_server: "true"`
3. Connect from Claude Code CLI:

   ```
   
   ```
4. 
5. Claude Code initiates OAuth authorization with:

   ```
   redirect_uri=http://localhost:58831/callback
   ```
6. Pomerium fetches the CIMD from `https://claude.ai/oauth/claude-code-client-metadata` which declares:

   ```json
   "redirect_uris": ["http://localhost/callback", "http://127.0.0.1/callback"]
   ```
7. Authorization fails with `invalid_grant` error: "client redirect URI does not match registered redirect URIs"

## Root Cause

In [`internal/mcp/client_id_metadata.go`](<https://github.com/pomerium/pomerium/blob/main/internal/mcp/client_id_metadata.go>), `ValidateRedirectURI` uses exact string comparison:

```go
func (doc *ClientIDMetadataDocument) ValidateRedirectURI(redirectURI string) error {
    for _, allowed := range doc.RedirectURIs {
        if allowed == redirectURI {
            return nil
        }
    }
    return fmt.Errorf(...)
}
```

This does not account for [RFC 8252 Section 7.3](<https://datatracker.ietf.org/doc/html/rfc8252#section-7.3>) which states:

> The authorization server MUST allow any port to be specified at the time of the request for loopback IP redirect URIs, to accommodate clients that obtain an available ephemeral port from the operating system at the time of the request.

## Expected Behavior

When the registered redirect URI uses a loopback address (`localhost`, `127.0.0.1`, `[::1]`) and does not specify a port, any port in the incoming `redirect_uri` should be accepted. For example:

* Registered: `http://localhost/callback`
* Incoming: `http://localhost:58831/callback`
* Result: **should match** (port ignored for loopback)

## Suggested Fix

Parse both URIs and, when the registered URI is a loopback address, compare scheme + host + path while ignoring the port:

```go
func (doc *ClientIDMetadataDocument) ValidateRedirectURI(redirectURI string) error {
    for _, allowed := range doc.RedirectURIs {
        if allowed == redirectURI {
            return nil
        }
        if matchesLoopbackRedirect(allowed, redirectURI) {
            return nil
        }
    }
    return fmt.Errorf(...)
}

func matchesLoopbackRedirect(registered, actual string) bool {
    r, err := url.Parse(registered)
    if err != nil { return false }
    a, err := url.Parse(actual)
    if err != nil { return false }

    if !isLoopback(r.Hostname()) { return false }
    return r.Scheme == a.Scheme &&
        r.Hostname() == a.Hostname() &&
        r.Path == a.Path
}
```

## Environment

* Pomerium ingress-controller: v0.32.5 (core v0.32.4)
* Client: Claude Code CLI v2.1.81
* CIMD URL: `https://claude.ai/oauth/claude-code-client-metadata`

artgas1:
Fixes #6209.

## Summary

OAuth 2.1 §3.1.2.2 and [RFC 8252 §7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3) require authorization servers to allow any port for loopback redirect URIs (`http://localhost`, `http://127.0.0.1`, `http://[::1]`) regardless of the port specified in the registered `redirect_uris`. MCP CLI clients (e.g. Claude Code) declare port-less loopback `redirect_uris` in their CIMD but request specific ephemeral ports at runtime, so a strict string comparison always fails.

## Changes

Two validation paths required the fix:

1. **`internal/mcp/client_id_metadata.go::ValidateRedirectURI`** — used by the MCP authorization handler when loading the client metadata document.

2. **`internal/oauth21/validate_client.go::ValidateAuthorizationRequestRedirectURI`** — generic OAuth 2.1 `redirect_uri` check called from the MCP authorization request validator.

Both now fall back to scheme + host + path comparison when both URIs are loopback HTTP redirects, matching the §7.3 requirement. Exact-match path is unchanged.

## Test Plan

- [ ] Add unit tests covering loopback port-mismatch case for both `ValidateRedirectURI` and `ValidateAuthorizationRequestRedirectURI`
- [x] Manually verified end-to-end with Claude Code 2.1.119 against a self-hosted Pomerium MCP route:
  - Without the fix: `[invalid_grant] client redirect URI does not match registered redirect URIs`
  - With the fix: OAuth authorization completes through to the token exchange step (verified via `journalctl -u pomerium`)

## Notes

- Branch is based on `v0.32.5` tag — happy to rebase onto `main` if needed.
- Tests still need to be added in `internal/mcp/client_id_metadata_test.go` and a new test file for `internal/oauth21/`. Posting now to confirm direction before fleshing out the test suite — let me know if the approach looks right and I'll add tests.
- Implementation matches the [Issue #6209](https://github.com/pomerium/pomerium/issues/6209) repro steps verbatim.


artgas1:
Closing as duplicate — I just rebased and saw that #6251 already merged the same fix into `main`. Apologies for not checking the merged PRs first; I started from `v0.32.5` tag and didn't see #6251 there yet.

Thanks @kenjenkins / @wasaga for getting #6251 in!

One small observation in case it's useful: the merged implementation matches `localhost` and `127.0.0.1`, but not `[::1]` (IPv6 loopback). On macOS \`localhost\` often resolves to \`::1\` first, so some MCP CLI clients may end up sending \`redirect_uri=http://[::1]:NNNN/callback\`. Happy to open a tiny follow-up PR adding \`::1\` to \`isLoopbackHost\` if you'd want that — let me know.

(Also confirmed end-to-end with Claude Code 2.1.119 against a self-hosted Pomerium 0.32.5 + #6251 cherry-pick on top — the redirect_uri validation now passes. We hit a separate downstream issue with GitHub OAuth Apps not supplying refresh tokens, but that's not related to this fix.)

PR #6323, opened 2026-05-05
feat(upstream): add HTTP/3 (QUIC) support for upstream HTTPS connections

IngmarStein:
## Summary

This change enables Pomerium to use HTTP/3 (QUIC) protocol for outbound connections to HTTPS backends when explicitly configured. HTTP/3 discovery occurs automatically via `alt-svc` response headers.

HTTP/3 is opt-in via the `enable_http3_upstream configuration` flag. All other protocol behaviors (HTTP/1.1, HTTP/2, h2c) remain unchanged.

## Related issues

- #6327

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/envoyconfig/bootstrap.go
+++ b/config/envoyconfig/bootstrap.go
@@ -184,6 +184,11 @@ func (b *Builder) BuildBootstrapLayeredRuntime(ctx context.Context, cfg *config.
 				"min_flush_spans": minFlushSpans,
 			},
 		},
+		"envoy": map[string]any{
+			"reloadable_features": map[string]any{
+				"http3_happy_eyeballs": true,
+			},
+		},
 	})
 	if err != nil {
 		return nil, fmt.Errorf("envoyconfig: failed to create layered runtime layer: %w", err)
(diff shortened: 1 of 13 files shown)


IngmarStein:
@calebdoxsey would you directionally be interested in this feature? If so, I'll go ahead and resolve the merge conflicts.

calebdoxsey:
My initial thoughts are that when I implemented HTTP/3 support for downstream I was surprised to learn that the protocol has a lot of rough edges and trade-offs that make it difficult to use universally and it's unclear how widespread adoption is ever going to be. In particular the arguments for HTTP/3 are stronger for an internet-facing service as opposed to an internal-facing one. Features like connection migration and better handling of packet loss are largely irrelevant on a reliable, fast, local network, which is typically what would be sitting behind Pomerium, so upstream HTTP/3 support didn't seem like a priority (especially given its experimental nature in Envoy itself).

But adding support may be a relatively small lift with no real downside, in which case we can probably add it. I will take a look at this this week.

PR #6335, opened 2026-05-11
feat: update session recording protocol & disable session recording

alexandreLamarre:
## Summary

Implements the new unix pipe-based protocol.

Disables the recording server entirely, it is not loaded in any pomerium code paths in this PR. Removes the recording server from authorize.

When paired with the follow up PR that configures dynamic extensions, the new server will live in `pkg/envoy`, since whenever envoy hot restarts a new set of pipes must be constructed and the transport needs to be informed/reloaded.

Changes with previous protocol handler:

- gRPC integration is removed.
- Each protocol handler is long-lived, and should live as long as the process lives or envoy hot restarts
- Each chunk of data has to be tagged with an ID, since the protocol no longer handles one recorded session per client
- `protocol_test.go` holds detailed step by step conformance the client should respect.
- gRPC is no longer the default. GRPC cannot be configured as a transport in the initial release of the feature, as pipes will the default for IPC.
- OnConfigChange was separated into `OnConfigChange` and `OnTransportChange`:
  - `OnConfigChange` handles the hot reloadable configuration tied to the blob store
  - `OnTransportChange` handles when the transport change (pipes), and is triggered when envoy hot restarts

## Related issues

Requires : https://linear.app/pomerium/issue/ENG-3971/envoy-build-dynamic-link-to-libc-when-building-for-macos

[ENG-3933](https://linear.app/pomerium/issue/ENG-3933/session-recording-server-unix-pipe-transport)

Follow up : https://github.com/pomerium/pomerium/pull/6283

## User Explanation

N/A

## AI disclosure

Claude PR review

## Checklist

- [X] updated unit tests
- [X] ready for review


pkg/storage/blob/schema.go
alexandreLamarre, line 1: Missed a spot to flatten signature in object_dir in previous PR, this fixes that

calebdoxsey (approved):

PR #6336, opened 2026-05-11
feat(controlplane): in-process MCP ConfigService server on a Unix socket

wasaga:
Adds an in-process MCP server that exposes pomerium's ConfigService as MCP tools, bound on a Unix domain socket at startup. Operators expose it externally by adding a regular pomerium route with `to: unix://<path>` and the policy of their choice.

The socket is always bound at `$TMPDIR/pomerium-mcp` — the same convention used by the Envoy admin socket. It is chmod'd to 0600 so only same-uid processes — pomerium itself and the Envoy it manages — can connect; the route's policy gates everyone else. Bind failures log at error level and leave the feature unreachable, rather than blocking startup.

Authentication on the in-process Connect dispatch reuses pomerium's existing shared-key bearer pattern. The HS256 signing primitive is factored into `grpcutil.SignSharedKey` so the gRPC metadata path (`WithSignedJWT`) and this new HTTP-header path share the same implementation.

```yaml
routes:
  - from: https://mcp.example.com
    to: unix:///tmp/pomerium-mcp
    mcp:
      server: {}
    policy:
      allow:
        or:
          - email:
              is: admin@example.com
```

Stacked on top of #6286.

Linear: [ENG-2506](https://linear.app/pomerium/issue/ENG-2506/create-an-mcp-server-for-the-pomerium-api)


internal/controlplane/server_mcp.go
```diff
@@ -0,0 +1,100 @@
...
+	// limit so a borderline-long path fails identically across platforms.
+	maxUnixSocketPathLen = 104
+)
+
+// bindMCPConfigAPIListener binds a Unix domain socket for the in-process
+// configapi MCP server, removing any stale file at the path and forcing
+// 0o600 permissions so only same-uid processes can connect. The socket
+// is always bound at startup; operators opt in or out by writing a
+// route that targets it.
+func (srv *Server) bindMCPConfigAPIListener() (net.Listener, error) {
```
kenjenkins, line 32: Should we make this listener optional? Maybe add a separate runtime flag for it? I figure it would be useful to limit the surface area.

wasaga: other, similar, listeners like `debug_addr` always start on random port and then we either enable / disable it by creating envoy listener on config-defined address or not. but they are always running. same with envoy admin socket i think. so this is kind of the same mechanism - just instead of adding a new config option, an admin can either create such a route or not. 

I can add a runtime flag but then it'll require a similar dance to enable/disable listener as existed in the previous PR if we want to disable the socket creation. 

I'd argue in vanilla core, person who is creating routes in the config files is superadmin so we don't need additional controls - if the admin adds a route to internal socket this is the enablement. For zero & console we probably need to similarly restrict unix socket upstreams to global admins.

kenjenkins (approved):

nickytonline (approved):

nickytonline:
I mixed up my tabs when I approved this. 😅 Can't dismiss my review for some reason, but I see Ken already approved.

PR #6340, opened 2026-05-12
core/envoy: use abstract namespace for unix sockets on linux

calebdoxsey:
## Summary
Use the abstract socket namespace on linux for the envoy admin socket.

## Related issues

Follow up from #6339 

## AI disclosure
none

## Checklist

- [x] reference any related issues
- [x] updated unit tests
- [ ] add appropriate label (`enhancement`, `bug`, `breaking`, `dependencies`, `ci`)
- [x] disclosed AI usage (or wrote "none") per AI_POLICY.md
- [x] ready for review

PR #6347, opened 2026-05-14
core/envoy: abstract socket improvements

calebdoxsey:
## Summary
Set `POMERIUM_SOCKET_DIRECTORY` in test env to fix tests.

Also move the socket naming to `netutil`. 

## Related issues
- #6340 

## AI disclosure
none

## Checklist

- [x] updated unit tests
- [x] ready for review


internal/testenv/environment.go
```diff
@@ -639,6 +639,7 @@ func (e *environment) Start() {
 	e.advanceState(Starting)
 	e.t.Cleanup(e.onTestCleanup)
 	e.t.Setenv("TMPDIR", e.TempDir())
+	e.t.Setenv("POMERIUM_SOCKET_DIRECTORY", e.TempDir())
```
kenjenkins, line 642: Are we going to run into the same socket name length limit if we go back to putting this in the test temp dir?

Or is this a different temp dir than what we were using before the abstract socket change?

calebdoxsey: Yeah we probably could. I'm working on a change that would give a new name every time and then we could remove this... but that would be a subsequent PR.

alexandreLamarre (approved):

PR #6352, opened 2026-05-14
feat(envoy): configure dynamic extensions for envoy

alexandreLamarre:
## Summary

- Adds dynamic extensions as a configuration target
- Parses and constructs ssh dynamic extension config
- Enables/disables the recording server depending on the presence of the session recording extension
- Loads the extensions into envoy
- Smartly reconstructs the pipe transports for the recording protocol when envoy hot-restarts

## Related issues

[ENG-3945](https://linear.app/pomerium/issue/ENG-3945/core-configure-envoy-dynamic-extensions)

Splits out the logic for configuring dynamic extensions in from https://github.com/pomerium/pomerium/pull/6283

Requires:
- [x] https://github.com/pomerium/pomerium/pull/6335
- [x] https://github.com/pomerium/envoy-custom/pull/219

## User Explanation

End users can configure enterprise dynamic extensions by setting the `envoy_dynamic_extensions` field in the config file.

Example:
```yaml
envoy_dynamic_extensions:
  - <path-to-extension>/<extension>.so
```

## AI disclosure

none

## Checklist

- [X] updated unit tests
- [ ] ready for review

### DIFF
--- a/config/envoyconfig/bootstrap.go
+++ b/config/envoyconfig/bootstrap.go
@@ -47,7 +47,7 @@ func (b *Builder) BuildBootstrap(
 	ctx, span := trace.Continue(ctx, "envoyconfig.Builder.BuildBootstrap")
 	defer span.End()
 	bootstrap = new(envoy_config_bootstrap_v3.Bootstrap)
-	if dynCfg != nil {
+	if dynCfg != nil && len(dynCfg.ExtensionsToLoad) > 0 {
 		bootstrap.BootstrapExtensions = append(bootstrap.BootstrapExtensions, dynCfg.DynamicExtensions)
 		b.extensionsToLoad = dynCfg.ExtensionsToLoad
 	}
--- a/config/envoyconfig/bootstrap.go
+++ b/config/envoyconfig/bootstrap.go
@@ -26,18 +26,31 @@ import (
 	"github.com/pomerium/pomerium/pkg/telemetry/trace"
 )
 
+const (
+	ExtensionSSHSessionRecording = "pomerium.ssh.session_recording"
+)
+
 const maxActiveDownstreamConnections = 50000
+
+type DynamicExtensionsConfig struct {
+	ExtensionsToLoad  []string
+	DynamicExtensions *envoy_config_core_v3.TypedExtensionConfig
+}
 
 // BuildBootstrap builds the bootstrap config.
 func (b *Builder) BuildBootstrap(
 	ctx context.Context,
 	cfg *config.Config,
 	fullyStatic bool,
+	dynCfg *DynamicExtensionsConfig,
 ) (bootstrap *envoy_config_bootstrap_v3.Bootstrap, err error) {
 	ctx, span := trace.Continue(ctx, "envoyconfig.Builder.BuildBootstrap")
 	defer span.End()
-
 	bootstrap = new(envoy_config_bootstrap_v3.Bootstrap)
+	if dynCfg != nil {
+		bootstrap.BootstrapExtensions = append(bootstrap.BootstrapExtensions, dynCfg.DynamicExtensions)
+		b.extensionsToLoad = dynCfg.ExtensionsToLoad
+	}
 
 	bootstrap.Admin, err = b.BuildBootstrapAdmin(cfg)
 	if err != nil {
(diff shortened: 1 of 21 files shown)


config/envoyconfig/bootstrap.go
```diff
@@ -26,18 +26,31 @@ import (
...
+
 const maxActiveDownstreamConnections = 50000
 
+type DynamicExtensionsConfig struct {
+	ExtensionsToLoad  []string
+	DynamicExtensions *envoy_config_core_v3.TypedExtensionConfig
+}
+
 // BuildBootstrap builds the bootstrap config.
 func (b *Builder) BuildBootstrap(
```
alexandreLamarre, line 41: I added the dynamic extensions to the BuildBootstrap signature instead of New() since the call site for New is static and configuration can be passed dynamically on hot-restart, for example session_recording_concurrency=\<number-worker-threads\>

pkg/envoy/envoy.go
```diff
@@ -269,7 +279,21 @@ func (srv *Server) run(ctx context.Context, cfg *config.Config) error {
 	// cancel any process monitor since we will be killing the previous process
 	srv.monitorProcessCancel()
 
-	if err := srv.writeConfig(ctx, cfg); err != nil {
+	dynCfg, err := srv.configureDynamicExtensions(ctx, cfg, srv.ServerOptions.dynamicExtensionPaths)
```
kralicky, line 282: we should only add the bootstrap extension if there are >0 paths configured

pkg/grpc/config/config.proto
```diff
@@ -702,6 +702,8 @@ message Settings {
   optional bool auto_apply_changesets = 180;
   // Allows HTTP upgrade requests to be forwarded upstream.
   optional StringList allow_upgrades = 181;
+  // Maps dynamic extension id to the path where it should be loaded from
+  StringList envoy_dynamic_extensions = 182;
```
kralicky: What do you think about naming this `dynamic_extensions` (or maybe just `extensions`) instead?

pkg/envoy/dynamic_extensions.go
```diff
@@ -0,0 +1,117 @@
...
+	// TODO : in follow-up PR that sets ssh specific config options
+	conc := 8
+	workers, err := ipc.NewPipeWorkers[*xrecording.RecordingData, *xrecording.RecordingCheckpoint](conc)
+	if err != nil {
+		return nil, fmt.Errorf("configuring %s extension: %w", extID, err)
+	}
+
+	sshCfg := &xssh.Config{
+		UploadConfig: &xssh.UploadConfig{
+			DefaultBufferSize: 1024 * 1024 * 32,
```
kralicky: Let's set this to 8MB to start. I think we need to add a couple more fields to this config though anyway (max chunk size and flush timeout)

config/envoyconfig/listeners_ssh.go
```diff
@@ -139,6 +140,14 @@ func buildSSHListener(cfg *config.Config) (*envoy_config_listener_v3.Listener, e
 		})
 	}
 
+	var enabledChannelFilters []*envoy_config_core_v3.TypedExtensionConfig
+	if slices.Contains(extensionsToLoad, ExtensionSSHSessionRecording) {
+		enabledChannelFilters = append(enabledChannelFilters, &envoy_config_core_v3.TypedExtensionConfig{
+			Name:        "session_recording",
+			TypedConfig: marshalAny(&wrapperspb.StringValue{}),
```
kralicky: Ah I forgot about this, we should probably use an actual named message type here (it can be empty though)

pkg/envoy/envoy.go
alexandreLamarre: The guard is in the BuildBootstrap method, but can be moved here as well

config/envoyconfig/listeners_ssh.go
alexandreLamarre: what do you suggest?

pkg/grpc/config/config.proto
alexandreLamarre: I think `envoy_extensions` make this most sense here, especially following the plugin discussion

alexandreLamarre: Or perhaps `proxy_extensions`

pkg/envoy/envoy.go
kralicky: oh yeah i see it. that's fine then

config/envoyconfig/listeners_ssh.go
kralicky: It will have to be part of the x/recording api, I'll add something there

kralicky (approved):

PR #6353, opened 2026-05-14
core/config: add internal-address, envoy admin to config

calebdoxsey:
## Summary
Add a new `netutil.InternalAddress` type to represent an address suitable for communication between Pomerium components or between Pomerium and Envoy. This type has several methods:

- `Listen`: to start a network listener on the address
- `Dial`: to create a network connection to the listener
- `EnvoyAddress`: to convert the address into an envoy address protobuf type

Internal addresses are represented by a URL. There are currently 3 supported address types:

- Unix socket addresses: `unix:/example`
- Abstract unix socket addresses: `unix-abstract:example` (same as `@example`)
- TCP addresses: `tcp://127.0.0.1:1234`

The default behavior is to create an abstract UUID socket on linux and a filesystem UUID socket (in the temp dir, or `POMERIUM_SOCKET_DIRECTORY`) on macos.

This internal socket is used for the envoy admin interface. The listener will be created by envoy and exposed by envoy if the admin address is set. It is also used for draining connections.

## AI disclosure
none

## Checklist

- [x] updated unit tests
- [x] ready for review


pkg/envoy/envoy.go
```diff
@@ -143,23 +150,23 @@ func NewServer(
 	return srv, nil
 }
 
-func (srv *Server) envoyAdminClient() *http.Client {
+func (*Server) envoyAdminClient(adminAddress netutil.InternalAddress) *http.Client {
```
kenjenkins: If this method does not use the `*Server` receiver, should we make it a standalone function instead?

Then I think `Drain()` also would not need its `*Server` receiver and could likewise be made a standalone function.

kenjenkins (approved):

PR #6357, opened 2026-05-15
core/databroker: add a cache invalidator and invalidate cache when the storage backend changes

calebdoxsey:
## Summary
Clear the cache when the databroker backend changes.

- There's a new cache invalidator that adds gRPC interceptors. For any message that returns the server version, if we detect that the server version changes, we invalidate the cache.
- Also invalidate the cache in the databroker server when a new backend is created.
- Since the cache is global, invalidating it in all-in-one mode is sufficient to invalidate it for the proxy and authorize services. In split mode detecting the new server version from gRPC methods should hopefully also clear the cache fairly quickly since we do `Sync` and `SyncLatest` streams. When the backend server version changes `Sync` should abort, and we should try to do a `SyncLatest` which should return a `Versions` message where we can pick up the new server version.

## Related issues
- [ENG-4016](https://linear.app/pomerium/issue/ENG-4016/pomerium-does-not-recover-from-a-runtime-databroker-storage-type-swap)

## AI disclosure
none

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/grpc/databroker/cache.go
+++ b/pkg/grpc/databroker/cache.go
@@ -1,0 +1,63 @@
+package databroker
+
+import (
+	"context"
+	"sync/atomic"
+
+	"google.golang.org/grpc"
+)
+
+// A CacheInvalidator has interceptors that can be used to clear the cache when
+// the server version changes.
+type CacheInvalidator struct {
+	StreamClientInterceptor grpc.StreamClientInterceptor
+	UnaryClientInterceptor  grpc.UnaryClientInterceptor
+}
+
+type clientStreamRecvMsgWrapper struct {
+	grpc.ClientStream
+	recvMsg func(m any) error
+}
+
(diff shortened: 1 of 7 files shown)


pkg/grpc/databroker/cache.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"google.golang.org/grpc"
+)
+
+// A CacheInvalidator has interceptors that can be used to clear the cache when
+// the server version changes.
+type CacheInvalidator struct {
+	StreamClientInterceptor grpc.StreamClientInterceptor
+	UnaryClientInterceptor  grpc.UnaryClientInterceptor
+}
```
alexandreLamarre, line 15: Does it make sense to implement the interceptors as methods on the struct?
That way currentServerVersion can be encapsulated on the struct itself instead of variable captured in the respective closures?

alexandreLamarre (approved):

PR #6363, opened 2026-05-20
core/config: add update_mask support

calebdoxsey:
## Summary
Add update mask support to the update methods for config entities. This allows a user to only update a subset of the fields.

## Related issues
- [ENG-3987](https://linear.app/pomerium/issue/ENG-3987/add-patch-method-to-pomerium-zero-routes-endpoint)

## AI disclosure
No AI was used.

## Checklist

- [x] updated unit tests
- [x] ready for review


wasaga (approved):
it seems like it can be used to override server-managed fields like created_at - should that be prevented?

PR #6380, opened 2026-05-27
Revert "core/config: add internal-address, envoy admin to config"

calebdoxsey:
Reverts pomerium/pomerium#6353


calebdoxsey:
I think this is not needed.

PR #6382, opened 2026-05-27
core/zero: default envoy admin internal address in bootstrap config

wasaga:
## Summary

The Pomerium Zero data plane crash-loops on startup:

```
panic: unsupported internal address:
pkg/netutil.(*InternalAddress).EnvoyAddress            pkg/netutil/internal.go:103
config/envoyconfig.(*Builder).buildEnvoyAdminCluster   config/envoyconfig/clusters_envoy_admin.go:22
```

#6353 added `Config.EnvoyAdminInternalAddress` but defaulted it only in `config.New()`. The zero bootstrap source builds its config via `new(config.Config)` instead, so the field stays empty and `EnvoyAddress()` hits its `default: panic` branch when building the envoy admin cluster. Only Zero mode is affected; the file/env path uses `config.New()`.

## Fix

Apply the `*netutil.NewInternalAddress()` default in `setConfigDefaults`, matching `config.New()`. Test `TestNewSetsEnvoyAdminInternalAddress` is red before / green after.

## AI disclosure

Investigation and fix drafted with Claude Code.

## Checklist

- [x] reference any related issues (ENG-4068)
- [x] updated unit tests
- [ ] ready for review (draft)

### DIFF
--- a/internal/zero/bootstrap/new.go
+++ b/internal/zero/bootstrap/new.go
@@ -35,7 +35,7 @@ type Source struct {
 
 // New creates a new bootstrap config source
 func New(secret []byte, fileCachePath *string, writer writers.ConfigWriter, api *sdk.API) (*Source, error) {
-	cfg := new(config.Config)
+	cfg := config.New(config.NewDefaultOptions())
 
 	err := setConfigDefaults(cfg)
 	if err != nil {
@@ -76,8 +76,6 @@ func New(secret []byte, fileCachePath *string, writer writers.ConfigWriter, api 
 }
 
 func setConfigDefaults(cfg *config.Config) error {
-	cfg.Options = config.NewDefaultOptions()
-
 	ports, err := netutil.AllocatePorts(7)
 	if err != nil {
 		return fmt.Errorf("allocating ports: %w", err)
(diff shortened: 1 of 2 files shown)


calebdoxsey (approved):

internal/zero/bootstrap/new.go
```diff
@@ -85,6 +85,10 @@ func setConfigDefaults(cfg *config.Config) error {
 
 	cfg.AllocatePorts(*(*[7]string)(ports))
 
+	// Apply the same envoy admin address default as config.New, which this
+	// path bypasses by building from a zero-value config.Config.
+	cfg.EnvoyAdminInternalAddress = *netutil.NewInternalAddress()
```
kenjenkins: Would it be better to update line 38 to call `config.New(config.NewDefaultOptions())` instead?

(I think that way, we'll automatically inherit any future changes to the config initialization here.)

wasaga: Good call — done. Now builds via `config.New(config.NewDefaultOptions())` and dropped the now-redundant `Options`/admin-address assignments from `setConfigDefaults`. Verified `config.New` is a pure constructor (no extra startup side effects) and `setConfigDefaults` has only this one caller, so the zero path is unchanged apart from inheriting the default.

PR #6387, opened 2026-05-28
core: refactor config, databroker proto packages, add go.mods

calebdoxsey:
## Summary
For the `config` and `databroker` gRPC packages, give them their own `go.mod` files and setup a `go.work` for the repo so that they can be used by other projects without having to pull in all of core's other dependencies.

To do this we needed to remove references to other core packages, so I created a `databrokerutil` package for much of that functionality.

Unfortunately `-modfile=` is not compatible with workspaces, so existing tools had to be put back into the root `go.mod` file and I removed the `internal/tools` folder

## Related issues
- [ENG-4071](https://linear.app/pomerium/issue/ENG-4071/core-setup-workspace-and-gomod-files-for-config-and-databroker)

## AI disclosure
No AI was used for this PR.

## Checklist

- [x] updated unit tests
- [x] ready for review


kenjenkins (commented):
I'm trying to think through the implications for a release... if we want tagged releases for these submodules we'll need to also add tags like pkg/grpc/config/v0.X.Y and pkg/grpc/databroker/v0.X.Y.
- Do we want tagged release versions at all? Or do we leave these untagged and update using pseudo-versions?
- If we do want tagged release versions, do we need any changes to release.yaml? (Or maybe not, if we push tags for the submodules but do not create corresponding github releases?)
- Or maybe we could automate it so the release workflow also pushes new matching tags for the submodules?

Any thoughts?

calebdoxsey:
Since these are in the same repo, I think the version tag on the repo should work right?

calebdoxsey:
> I might be misunderstanding, but I think modules in a subdirectory do not inherit the version tags from the root module? There's some documentation about this here but it's a bit sparse: https://go.dev/doc/modules/managing-source#multiple-module-source.
> 
> But I guess this is only relevant for other repos that import from this one? Because this repo has the go.work file that will end up treating the new modules as part of the root module?

Oh I didn't realize it uses a different tag. So yes it would have a different version. Right now it just uses the SHA pseudo version in other repos...

kenjenkins (approved):

PR #6399, opened 2026-05-28
fix(session_recording): validate that the bucket is accessible before propagating it to the recording server

alexandreLamarre:
## Summary

Adds some bucket validation before propagating it to the recording server for use

## Related issues

Related to [ENG-4074](https://linear.app/pomerium/issue/ENG-4074/session-recording-ui-feedback)

## User Explanation

N/A

## AI disclosure

none

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/internal/recording/server.go
+++ b/internal/recording/server.go
@@ -177,7 +177,7 @@ func (r *recordingServer) handleBlobChange(ctx context.Context, cfg *blob.Storag
 	if newBucketURI != "" {
 		log.Ctx(ctx).Debug().Str("blob-uri", newBucketURI).Msg("opening new bucket")
 		// Open the new bucket.
-		bucket, err := providers.OpenBucket(ctx, newBucketURI)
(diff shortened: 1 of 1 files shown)


internal/recording/server.go
```diff
@@ -196,6 +196,26 @@ func (r *recordingServer) handleBlobChange(ctx context.Context, cfg *blob.Storag
...
+	}
+
+	ok, err := bucket.IsAccessible(ctxT)
+	if !ok {
+		st := ""
+		if err != nil {
+			st = ": " + err.Error()
+		}
+		_ = bucket.Close()
+		return nil, fmt.Errorf("bucket is not accessible%s", st)
```
calebdoxsey: Should this be wrapping the error with `%w`?

alexandreLamarre: yeah I can the change if err != nil {} block it looks kinda weird right now

calebdoxsey (approved):

PR #6426, opened 2026-06-04
feat(ssh): add ssh routes to portal

alexandreLamarre:
## Summary

there are some PPL policies that cannot be evaluated by the route portal like requiring specific ssh usernames & ssh keys.

## Related issues

[ENG-2541](https://linear.app/pomerium/issue/ENG-2541/ssh-routes-are-not-displayed-on-the-routes-portal)

## User Explanation

N/A

## AI disclosure

none

## Checklist

- [X] updated unit tests
- [X] ready for review


kralicky:
I don't know if we _should_ do this necessarily, since ssh has its own routes portal. The SSH routes portal is also currently gated behind a runtime flag, so if we do want to do this we should also gate this feature behind the same flag.

alexandreLamarre:
@kralicky It's done just for product consistency and not necessarily completeness, seeing as all other route types appear here.

Alternatively, we can route to the routes portal, which currently spits out strange errors if the routes portal feature flag is not configured

kralicky:
Actually yeah what if we put the ssh command here that you could copy/paste to get to the ssh routes portal? it can just be `ssh <pomerium address> [-p <ssh port>]`

alexandreLamarre:
yeah I think that is a better solution. We should maybe consider enabling routes portal by default? Let me put this in draft for now

PR #6431, opened 2026-06-08
fix(mcp): encode authorization code with base64url so it survives a form-encoded token request

### LINKED ISSUE #6430: MCP authorization code can fail with `invalid_grant` when it contains a `+` (StdEncoding vs form-urlencoded body)
## What happened?

The MCP OAuth authorization code that Pomerium issues at `/.pomerium/mcp/authorize`
is encoded with `base64.StdEncoding`, whose alphabet includes `+` and `/`. The code
is then redeemed by the client at `POST /.pomerium/mcp/token` in an
`application/x-www-form-urlencoded` request body.

If an MCP client puts the `+` character literally into the form body (instead of
percent-encoding it as `%2B`), Go's form parser (`url.ParseQuery` / `r.ParseForm`,
per the `application/x-www-form-urlencoded` spec) decodes `+` back to a space. The
resulting `code` value is corrupted, fails to base64-decode, and the token endpoint
returns `invalid_grant`.

Because only codes that happen to contain a `+` are affected, the failure is
intermittent and client-dependent: roughly the share of codes whose Base64 output
contains at least one `+` will fail, while codes without a `+` succeed. This makes it
look like a flaky/transient auth error rather than a deterministic bug.

Observed with an MCP client that sends `+` literally (not percent-encoded) in the
form-encoded token request body.

The authorization code should round-trip through an
`application/x-www-form-urlencoded` token request unchanged, regardless of whether
the client percent-encodes individual characters, and the token exchange should
succeed every time.

## How'd it happen?

1. Start an MCP OAuth flow against a Pomerium route with MCP enabled.
2. Complete authorization so Pomerium issues a code via the
   `/.pomerium/mcp/authorize` 302 `Location`.
3. Have the client redeem the code at `POST /.pomerium/mcp/token` with a
   `application/x-www-form-urlencoded` body, sending the `code` value with any `+`
   character left literal (not percent-encoded).
4. When the issued code contains a `+`, the token endpoint returns `invalid_grant`;
   when it does not, the exchange succeeds. Repeat a few times to see the
   intermittent pattern.

## Root cause

`internal/mcp/code.go`: `CreateCode` / `CreateCodeWithRecordVersion` encode the
encrypted code with `base64.StdEncoding`. The Standard Base64 alphabet contains `+`
and `/`, neither of which is safe to carry unescaped in an
`application/x-www-form-urlencoded` body — `+` in particular is interpreted as a
space by form parsers. Since the code is delivered through a form-encoded POST body,
a URL-safe alphabet should be used instead.

## Suggested fix

Encode the code with `base64.RawURLEncoding`, whose alphabet uses `-` and `_` (never
`+`, `/`, or `=`), so the code is safe in a form body and in URLs without any
escaping. For backward compatibility during rollout, `DecryptCode` can try
`RawURLEncoding` first and fall back to `StdEncoding`, keeping any codes issued just
before the change redeemable through their short lifetime. Authorization codes are
opaque to clients per OAuth 2.1, so changing the encoding is transparent to them.

- Pomerium version: reproducible on `main` (current `internal/mcp/code.go`).
- Server Operating System/Architecture/Cloud: not environment-specific; the bug is
  in the code-encoding path and is independent of OS/arch.

## AI disclosure

This report was drafted with AI assistance (Claude Code) and reviewed and edited by
the author. The author has reproduced the failure and confirmed the root cause in
`internal/mcp/code.go`.

makvitaly:
## Summary

The MCP OAuth authorization code is issued in a `/.pomerium/mcp/authorize` 302
`Location` and redeemed at `POST /.pomerium/mcp/token` in an
`application/x-www-form-urlencoded` body. It was encoded with `base64.StdEncoding`,
whose alphabet includes `+` and `/`.

**Problem.** When a client sends the `code` value with a literal `+` (not
percent-encoded) in the form-encoded body, Go's form parser decodes `+` back to a
space (per the `application/x-www-form-urlencoded` rules). The corrupted code then
fails to base64-decode and the token endpoint returns `invalid_grant`. Only codes
that happen to contain a `+` are affected, so the failure is intermittent and
client-dependent — it looks like a flaky auth error rather than a deterministic bug.
Observed with an MCP client that does not percent-encode `+` in the form-encoded
token body.

**Root cause.** `internal/mcp/code.go`: `CreateCode` /
`CreateCodeWithRecordVersion` emit the encrypted code with `base64.StdEncoding`. The
Standard Base64 alphabet (`+`, `/`, `=`) is not safe to carry unescaped through a
form body.

**Fix.**
- `CreateCode` now emits `base64.RawURLEncoding` — the URL-safe alphabet (`-`, `_`,
  no padding) never produces `+`, `/`, or `=`, so the code is safe in a form body and
  in URLs without any escaping.
- `DecryptCode` tries `RawURLEncoding` first and falls back to `StdEncoding`. This
  keeps any codes issued just before the cutover (still using the legacy alphabet)
  redeemable through their short lifetime, so the change can roll out without
  invalidating in-flight codes.

All three code types (authorization, refresh, access) route through these two
functions, so the fix covers every MCP code.

**Compatibility.** Authorization codes are opaque to clients per OAuth 2.1, so
changing the encoding alphabet is transparent — clients treat the code as an opaque
string and never parse it. The `StdEncoding` decode fallback covers the brief
transition window; no client or stored-data migration is required.

## Test coverage

Added to `internal/mcp/code_test.go`:
- `TestCreateCodeFormURLEncodingSafe` — asserts the emitted code contains no `+`,
  `/`, or `=`, and survives a `url.Values` form-urlencoded round-trip unchanged.
- `TestDecryptCodeLegacyStdEncoding` / `...WithPlus` — assert that a legacy
  `StdEncoding` code (including one containing `+`) still decodes via the fallback,
  proving the backward-compatible decode path.
- Adjusted the existing "invalid base64" case input to a `!`-containing string
  (invalid under both alphabets) so it still exercises the base64-decode error path
  now that `-` is a valid `RawURLEncoding` character.

`go test ./internal/mcp/` passes.

## Related issues

Fixes #6430

## User Explanation

MCP OAuth token exchanges that previously failed intermittently with `invalid_grant`
— because the authorization code happened to contain a `+` that a client sent
un-escaped in the form-encoded token request — now succeed reliably. The code is
encoded with a URL-safe alphabet, so it is safe to carry in form bodies and URLs.
There is no user-facing configuration change; existing in-flight codes continue to
work via a decode fallback.

## AI disclosure

This change was implemented with AI assistance (Claude Code): AI drafted the
encoding change, the backward-compatible decode fallback, and the unit tests. The
author reviewed and edited all code and the commit messages, reproduced the original
`invalid_grant` failure, confirmed the root cause in `internal/mcp/code.go`, and can
explain and defend the change without the AI in the loop. Per AI_POLICY.md.

## Checklist

- [x] updated unit tests
- [x] ready for review


makvitaly:
Good catch — updated the `codeNoExpiryStr` fixture to `base64.RawURLEncoding` in 9157e6499 so it matches what production now emits and is independent of the legacy StdEncoding fallback. Thanks!

makvitaly:
Closing this one from my side.

We have moved to a small shim in front of Pomerium that re-encodes the authorization code, so we no longer run a patched build and I cannot keep this branch maintained or rebased against main.

The underlying issue is still real, for whoever picks it up: `CreateCode` emits standard base64, and a client that does not percent-encode `+` in a form-encoded token request gets a spurious `invalid_grant`. The change here is small — `RawURLEncoding` with a legacy `StdEncoding` decode fallback so existing codes keep working — and the tests are in the diff. Please feel free to take it over.

Thanks for the review feedback.

PR #6442, opened 2026-06-09
authorize: surface underlying error for unresolved MCP access token

wasaga:
Wrap the underlying `err` (e.g. decrypt/expired/type-mismatch) in the "no session found for access token" path so the real cause is logged instead of a generic session-not-found.

### DIFF
--- a/authorize/grpc.go
+++ b/authorize/grpc.go
@@ -183,7 +183,7 @@ func (a *Authorize) getMCPSession(
 	accessToken := auth[len(prefix):]
 	sessionID, sessionRecordVersion, err := a.state.Load().mcp.GetSessionAndVersionFromAccessToken(accessToken)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


authorize/grpc.go
```diff
@@ -183,7 +183,7 @@ func (a *Authorize) getMCPSession(
 	accessToken := auth[len(prefix):]
 	sessionID, sessionRecordVersion, err := a.state.Load().mcp.GetSessionAndVersionFromAccessToken(accessToken)
 	if err != nil {
-		return nil, fmt.Errorf("no session found for access token: %w", sessions.ErrNoSessionFound)
+		return nil, fmt.Errorf("no session found for access token: %w: %w", err, sessions.ErrNoSessionFound)
```
alexandreLamarre, line 186: shouldn't this be the other way around `%w:%w, sessions.ErrNoSessionFound, err`?

alexandreLamarre (approved):

PR #6446, opened 2026-06-10
databroker: redact sensitive fields in debug browser

wasaga:
The databroker debug browser rendered records as raw protojson, including session OAuth tokens, ID-token JWTs, and IdP client secrets. Redact fields annotated `[(pomerium.config.sensitive) = true]` (replacing string/bytes values with `[REDACTED]`, descending into `Any` payloads) and annotate the session and identity record types.

Linear: [ENG-4132](https://linear.app/pomerium/issue/ENG-4132/redact-sensitive-fields-in-the-databroker-debug-browser)


go.work.sum
kenjenkins, line 1: Is this an unrelated change?

pkg/protoutil/redact_test.go
```diff
@@ -0,0 +1,144 @@
...
+	assert.Equal(t, []byte("public-cert"), settings.GetCertificates()[0].GetCertBytes())
+	assert.Equal(t, []byte("[REDACTED]"), settings.GetCertificates()[0].GetKeyBytes())
+}
+
+func TestRedactSensitive_MessageKindCleared(t *testing.T) {
+	t.Parallel()
+
+	// ssh_host_keys is a sensitive message-kind field (StringList) carrying
+	// inline SSH host private keys: no placeholder representation exists, so
+	// it must be cleared entirely.
```
kenjenkins, line 75: Do we not want to redact the individual elements of a StringList? Or replace the entire StringList with a single "[REDACTED]" element?

kenjenkins (approved):

internal/controlplane/server_debug.go
alexandreLamarre, line 1: Just a thought, but it might make sense to have some grpc interceptors in the future for clients that make requests that should be redacted. This way other grpc clients could be a little more resilient to potential sensitive field leaks in new records / proto messages.

Overkill for this PR for sure

PR #6448, opened 2026-06-12
ssh: fix connection buffer limits

### LINKED ISSUE #6285: Add config parameter to set value of 'connectionBufferLimit' in envoyconfig - issues with git clone via ssh
When using native ssh-access via pomerium to clone from Gitlab via ssh, we faced issues on a repo with large files in it. The connection was stalled at some point and the connection was just terminated, leaving the user not able to clone the repo.

Verified by building the pomerium docker image with updated parameter for connectionBufferLimit
	connectionBufferLimit            uint32 = 1024 * 1024

After this change the clone was successful.

I'm thinking of adding a parameter into the config file that is read on start and used in the bootstrap of envoy.

related config is at config/envoyconfig/protocols.go:30
const (
	connectionBufferLimit            uint32 = 32 * 1024 <---- 
	maxConcurrentStreams             uint32 = 100
	initialStreamWindowSizeLimit     uint32 = 64 * 1024
	initialConnectionWindowSizeLimit uint32 = 1 * 1024 * 1024
)

kralicky:
## Summary

For SSH clusters, set connection buffer high watermark limits to the minimum safe value of 524288. The same limit is also used for the ssh listener (it was previously unset, which used the 1MB default). 

See the linked envoy-custom PR for more details on where that number comes from.

## Related issues

Fixes https://github.com/pomerium/pomerium/issues/6285

Related: https://github.com/pomerium/envoy-custom/pull/251/changes

## User Explanation

The default connection buffer limit, which was applied globally, is 32768. This is suitable for HTTP, but is too small for SSH connections. This could cause connections to get stuck in some use cases, such as during large file transfer.

## AI disclosure

none

## Checklist

- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/config/envoyconfig/clusters.go
+++ b/config/envoyconfig/clusters.go
@@ -263,6 +263,10 @@ func (b *Builder) buildPolicyCluster(ctx context.Context, cfg *config.Config, po
 	}
 	cluster.LbPolicy = policy.LoadBalancingPolicy.Value.ToEnvoy()
 
+	if policy.IsSSHUpstream() {
+		cluster.PerConnectionBufferLimitBytes = wrapperspb.UInt32(sshConnectionBufferLimit)
+	}
+
 	// Keep last
 	if policy.UpstreamTunnel != nil {
 		if cfg.Options.IsRuntimeFlagSet(config.RuntimeFlagSSHUpstreamTunnel) {
--- a/config/envoyconfig/clusters_test.go
+++ b/config/envoyconfig/clusters_test.go
@@ -1378,3 +1378,38 @@ func Test_buildPolicyCluster(t *testing.T) {
 		assert.Equal(t, "stat-name", cluster.AltStatName)
 	})
 }
+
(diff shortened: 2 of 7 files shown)


config/envoyconfig/clusters.go
```diff
@@ -263,6 +263,10 @@ func (b *Builder) buildPolicyCluster(ctx context.Context, cfg *config.Config, po
 	}
 	cluster.LbPolicy = policy.LoadBalancingPolicy.Value.ToEnvoy()
 
+	if policy.IsSSHUpstream() {
+		cluster.PerConnectionBufferLimitBytes = wrapperspb.UInt32(sshConnectionBufferLimit)
```
kenjenkins, line 267: Could we verify this behavior in a unit test?

kenjenkins (approved):

config/envoyconfig/clusters.go
kralicky: Sure thing

kenjenkins (approved):

PR #6475, opened 2026-06-23
feat: health checks for envoy dynamic extensions

alexandreLamarre:
## Summary

Adds health probes for envoy dynamic extensions.

It uses the envoy admin endpoint `/dynamic_extensions/status` on the envoy admin address. Extensions are always reported on load, and status is not updated after they load.

The health is evaluated in a background goroutine, until all extensions are reported at which point we stop polling.  In flight configuration changes / envoy server reloads cancel the health probe and start a new one.

## Related issues

[ENG-4138](https://linear.app/pomerium/issue/ENG-4138/session-recording-dynamic-extensions-health-check-integration)

## User Explanation

Status of loaded envoy extensions is now reflected in the health probes

## AI disclosure

opus 4.8 was used for code review / PR review

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/pkg/envoy/envoy.go
+++ b/pkg/envoy/envoy.go
@@ -60,13 +60,13 @@ type Server struct {
 
 	monitorProcessCancel context.CancelFunc
 
-	mu        sync.Mutex
-	shutdownC chan error
+	mu                                sync.Mutex
+	shutdownC                         chan error
+	dynamicExtensionHealthProbeCancel context.CancelFunc
 
 	recordingServer     stdatomic.Pointer[recording.Server]
 	recordingServerErrC chan error
 }
-
 type ServerOptions struct {
 	extraEnvVars          []string
 	logLevel              config.LogLevel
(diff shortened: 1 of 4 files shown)


pkg/envoy/envoy.go
```diff
@@ -65,8 +65,9 @@ type Server struct {
 
 	recordingServer     stdatomic.Pointer[recording.Server]
 	recordingServerErrC chan error
-}
 
+	dynamicExtensionHealthProbeCancel context.CancelFunc
```
kenjenkins: If this field is protected by `mu` I'd suggest moving it right after `shutdownC` to help make that relationship clearer.

kenjenkins (approved):

pkg/envoy/envoy.go
alexandreLamarre: ah yeah it is. good idea

calebdoxsey (approved):

PR #6480, opened 2026-06-29
ci: add govulncheck workflow for Go vulnerability scanning

desimone:
## What

Adds a `make govulncheck` target and a workflow that runs [govulncheck](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) across all Go workspace modules (root + `pkg/grpc/config` + `pkg/grpc/databroker`) to scan for known vulnerabilities in stdlib and dependencies.

## Why

Adds reachable-vulnerability detection for Go, complementing the existing CodeQL and gosec (golangci-lint) checks. Runs on push to `main`, a weekly schedule, manual dispatch, and on PRs that change dependency files — so a newly disclosed third-party CVE does not block unrelated open PRs.

## AI assistance

Drafted with Claude (Opus 4.8): the workflow and Makefile target. Verified: `actionlint` clean, govulncheck green in CI across all workspace modules. Human-reviewed before merge.

### DIFF
--- a/.github/workflows/govulncheck.yaml
+++ b/.github/workflows/govulncheck.yaml
@@ -1,0 +1,33 @@
+name: govulncheck
+permissions:
+  contents: read
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
+      - "**/go.mod"
+      - "**/go.sum"
+      - go.work
+      - go.work.sum
+      - Makefile
+      - .github/workflows/govulncheck.yaml
+  schedule:
+    - cron: "17 6 * * 1"
+  workflow_dispatch: {}
+
+jobs:
+  govulncheck:
+    runs-on: ubuntu-22.04
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,35 @@
...
+    branches:
+      - main
+  pull_request:
+    paths:
+      - go.mod
+      - go.sum
+      - go.work
+      - go.work.sum
+      - pkg/grpc/*/go.mod
+      - pkg/grpc/*/go.sum
```
wasaga: paths support glob

wasaga (approved):

.github/workflows/govulncheck.yaml
desimone: Good call — collapsed to `**/go.mod` / `**/go.sum`, which also auto-covers future nested modules instead of hardcoding `pkg/grpc/*`. Done in 3bddad4fa.

PR #6482, opened 2026-06-30
chore(logs): add iterate method for log reader

calebdoxsey:
## Summary
Add an `Iterate` method to `logutil` that will read log lines and yield them as an iterator.

## Related issues
- [ENG-4185](https://linear.app/pomerium/issue/ENG-4185/core-generalize-log-reading-from-envoy)

## AI disclosure
No AI was used for this PR.

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/logutil/logutil.go
+++ b/pkg/logutil/logutil.go
@@ -1,2 +1,44 @@
 // Package logutil contains functionality for working with logs.
 package logutil
+
+import (
+	"bufio"
+	"errors"
+	"io"
+	"iter"
+	"os"
+	"strings"
+
+	"github.com/pomerium/pomerium/internal/log"
+)
+
+// IterateLines iterates over all the lines of a log reader. The log reader
+// will be closed after iteration.
+func IterateLines(rc io.ReadCloser) iter.Seq[string] {
+	return func(yield func(string) bool) {
+		defer rc.Close()
+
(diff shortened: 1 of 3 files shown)


pkg/logutil/logutil.go
```diff
@@ -1,2 +1,44 @@
...
+				if !yield(ln) {
+					return
+				}
+			}
+
+			if errors.Is(err, io.EOF) || errors.Is(err, os.ErrClosed) {
+				return
+			} else if err != nil {
+				log.Error().Err(err).Msg("logutil: unexpected error while reading log line")
+			}
```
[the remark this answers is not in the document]
calebdoxsey: I changed it to just exit on error.

pkg/logutil/logutil.go
alexandreLamarre, line 1: Is there any risk omitting the previous backoff logic? I'm not sure why we had a backoff there before

alexandreLamarre (approved):

pkg/logutil/logutil.go
calebdoxsey: Yeah there was a slight risk, so I changed it to exit. I don't think it should really happen that if there was an error we could continue reading the stream afterwards as if the error never happened, so I'm not sure the backoff ever really made sense.

PR #6499, opened 2026-07-07
tracing: fix otel resource initialization

kralicky:
## Summary

This fixes initialization of the otel tracer `Resource` to not merge with `resources.Default()`, which may have a different schema version than the one from `semconv.SchemaURL` that we import with a specific version. The default attributes that were used from resources.Default() are added manually instead (the default detectors are not exported).

In `go.opentelemetry.io/otel` v1.44.0, the default schema url was changed to 1.41.0, which was conflicting with the imported version and caused an error on startup. 

## User Explanation

N/A

## AI disclosure

none

## Checklist

- [ ] updated unit tests
- [x] ready for review

### DIFF
--- a/pkg/telemetry/trace/trace.go
+++ b/pkg/telemetry/trace/trace.go
@@ -10,6 +10,7 @@ import (
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
+	otelsdk "go.opentelemetry.io/otel/sdk"
 	"go.opentelemetry.io/otel/sdk/resource"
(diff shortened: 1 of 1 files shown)


alexandreLamarre (approved):
Should be backported to 0-33-0 branch.

alexandreLamarre:
We could also update the semconv dep to 1.41.0 and keep compatibility with resources.Default()

kralicky:
> We could also update the semconv dep to 1.41.0 and keep compatibility with resources.Default()

True but it would break the same way again the next time the default schema version changes. Based on looking at the current otel docs I don't see this pattern used anymore, so it might not be best practice.

PR #6512, opened 2026-07-12
envoyconfig: avoid quadratic policy scans during route building

desimone:
## Summary

`buildMainRouteConfiguration` previously scanned every policy for every routeable host and parsed each policy's `From` URL on each scan. With distinct policy hosts, the host-to-policy selection step therefore grew as `hosts x policies`.

This builds a policy host index once per route-configuration build, keyed by the same `urlutil.GetDomainsForURL(fromURL, true)` values used by the old `urlMatchesHost` predicate. Per-host selection becomes a map lookup, wildcard policies have a separate catch-all bucket, and every bucket preserves `GetAllPoliciesIndexed` order. That order is unchanged because it determines Envoy route precedence.

`BenchmarkBuildRouteConfigurations` uses one HTTPS policy per distinct host. Results were measured on the same Apple M3 Max with Go 1.26.5:

| Configured policies | Before | After | Allocations/op |
| ---: | ---: | ---: | ---: |
| 100 | 12.60 ms | 8.29 ms (-34%) | 297k -> 217k |
| 1,000 | 510.9 ms | 85.7 ms (-83%) | 10.17M -> 2.17M |
| 10,000 | 43.43 s | 0.843 s (about 52x) | 821.7M -> 21.72M |

The 100- and 1,000-policy rows are benchstat results from 10 runs. The 10,000-policy row is the observed median of three `-benchtime=1x` samples; three samples are not enough for a confidence interval or significance claim.

In the synthetic 10,000-policy case, route configuration construction falls below one second. This is not a complete config apply: cluster construction, validation, and the xDS push are outside the benchmark. The policy count is also not a literal Envoy route count because one policy can produce multiple routes.

A differential test reconstructs the old scans and compares all policy routes produced for an adversarial fixture with `proto.Equal` in the same order across host, port, wildcard, TCP, policy-source, and runtime-flag cases. Existing golden tests pass unchanged.

## Related issues

- None

## AI disclosure

Claude Code using the Fable model drafted the initial implementation and benchmarks. Codex revised the code and tests and reproduced the benchmark results. I reviewed and edited the final diff and description.

## Checklist

- [x] updated unit tests
- [x] add appropriate label (`enhancement`)
- [x] ready for review

### DIFF
--- a/config/envoyconfig/route_configurations.go
+++ b/config/envoyconfig/route_configurations.go
@@ -85,6 +85,16 @@ func (b *Builder) buildMainRouteConfiguration(
 		return nil, err
 	}
 
+	// index policies by host once so that adding policy routes to a virtual
+	// host doesn't require scanning every policy
+	var policyIdx policyHostIndex
+	if config.IsProxy(cfg.Options.Services) {
+		policyIdx, err = indexPoliciesByHost(cfg.Options)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	var virtualHosts []*envoy_config_route_v3.VirtualHost
 	for _, host := range allHosts {
 		vh, err := b.buildVirtualHost(cfg.Options, host, host, mcpHosts[host])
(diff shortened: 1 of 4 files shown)


kralicky (changes requested):
Not comfortable merging AI generated code here. If you open an issue and assign me I can take a look at this. There have also been some prior attempts to optimize this code but were never merged due to complexity or other issues.

Issue #6530, opened 2026-07-15
MCP `getOrRecreateSession` replays a stale upstream refresh token when the IdP rotates refresh tokens


ShakeyJ82:
## Summary

When the upstream IdP is configured to **rotate refresh tokens on use** (e.g. Okta's
default Org authorization server), the MCP gateway stores two independent copies of the
upstream refresh token — one on the `session` record and one on the long-lived
`MCPRefreshToken` record — and never reconciles them. The background identity manager
rotates the upstream RT and writes the new value **only to the session**. The
`MCPRefreshToken` record is left holding a consumed RT. Once the Pomerium session hits its
absolute lifetime and is deleted, the next downstream MCP token refresh calls
`getOrRecreateSession`, which replays the **stale** RT from the `MCPRefreshToken` record and
gets `invalid_grant "The refresh token is invalid or expired."` from the IdP.

User-visible symptom: **every MCP server shows unauthenticated after the Pomerium session
expires (≈daily), even though the client's MCP refresh token is valid for a year.** The
gateway falls back to forcing a fresh interactive sign-in.

## Version

- pomerium `+55e4a97` (main, commit `55e4a9752cc3ff2e2d65f711d812b1c6cdf5d7c6`)
- envoy `20260618180819-16ede7a56f2f`
- IdP: Okta OIDC, with **refresh-token rotation enabled** ("Rotate token after every use",
  with a 30s rotation grace). Any IdP that rotates refresh tokens on use will exhibit this;
  disabling rotation ("Use persistent token") avoids it.

## Root cause

There are two independent callers of `authenticator.Refresh()` against the IdP, persisting
the rotated RT to different places:

| Caller | Refresh call | Persists new upstream RT to |
|---|---|---|
| Identity manager (background timer) | `pkg/identity/manager/manager.go:233` | **session only** — `s.OauthToken` (`:250`) → `updateSession` (`:270`) |
| MCP token/refresh (client on open) | `internal/mcp/handler_token.go:576` | `MCPRefreshToken.UpstreamRefreshToken` (`:443`) |

`MCPRefreshToken.UpstreamRefreshToken` is assigned in exactly **one** place in the codebase
(`internal/mcp/handler_token.go:443`), plus initial seeding at creation (`:217`). The
identity manager never updates it. So:

1. Sign-in seeds `MCPRefreshToken.UpstreamRefreshToken = RT0` (`handler_token.go:217`);
   session also holds RT0.
[...]

PR #6551, opened 2026-07-21
fix(mcp): clean up no longer used refresh tokens

ititandev:
## Summary

On every MCP `grant_type=refresh_token` rotation, `handleRefreshTokenGrant` stores a new
`MCPRefreshToken` and marks the old one `Revoked = true` (re-`Put`) instead of deleting it. Nothing
evicts these records — `MCPRefreshToken` has no TTL and the databroker has no eviction — so each
refresh leaks one record. With the default in-memory backend, a client that refreshes on a loop
grows Pomerium's heap without bound until the host OOMs.

The fix calls the already-present `DeleteMCPRefreshToken` (`internal/mcp/storage.go`) to remove the
superseded token. A replay of the deleted token still fails cleanly: the lookup returns not-found and
the handler returns the same `400 invalid_grant` it returned on the revoked path, so behavior for the
caller is unchanged. This matches the sibling one-time records, which already delete on use
(`DeleteAuthorizationRequest` in the auth-code path).

This does not remove a security property: nothing in the codebase reads `Revoked` for rotation-reuse
detection (no token-family/lineage field exists), and the reject-on-`Revoked` branch is left in place
as defense-in-depth. If RFC 9700 rotation-reuse detection is added later it would need retained
records plus a family id — a future feature, not a regression here.

**Repro:** point an MCP client that refreshes frequently at an `mcp:` route → databroker
`MCPRefreshToken` count and RSS climb monotonically with zero deletes; an idle instance stays flat.
A deterministic regression test (`TestRefreshTokenRotationDoesNotLeak`) drives 50 rotations and fails
without the fix (50/50 records retained), passes with it (0 retained); it also replays a rotated
token to confirm reuse is still rejected (`400 invalid_grant`).

**Follow-up (separate PR):** the same path also orphans one `session.Session` per refresh —
`getOrRecreateSession` mints a new session id each time and never deletes the old. Fixing that needs
a `session_id` on `MCPRefreshToken` plus a `DeleteSession`, so it is out of scope here.

## Related issues

- No open issue tracks this leak — this PR is the report.
- #6530 is an adjacent bug in the same handler (stale upstream refresh token on IdP rotation);
  separate, not addressed here.

## User Explanation

Fixes a memory leak in the MCP OAuth gateway: Pomerium no longer keeps one databroker record per
token refresh, so instances serving MCP traffic hold bounded memory. No config or API change.

## AI disclosure

Claude Code (Anthropic) drafted this change and description while analyzing the source and a
production OOM incident. The change itself is small and self-contained — swapping the revoke-and-retain
for the existing `DeleteMCPRefreshToken`, with matching updates to the rotation tests (plus a new
regression test) and `DESIGN.md`. I reviewed the change, ran the `internal/mcp` suite and `go vet`
(both pass), and understand the behavior before opening this PR.

## Checklist

- [x] updated unit tests
- [ ] add appropriate label (`bug`; `docs` for the DESIGN.md update)
- [ ] ready for review

### DIFF
--- a/internal/databroker/server_backend.go
+++ b/internal/databroker/server_backend.go
@@ -748,9 +748,8 @@ func (srv *backendServer) newBackendAndSetupLocked(ctx context.Context) (storage
 }
 
 func (srv *backendServer) setupRequiredIndex(ctx context.Context, backend storage.Backend) error {
-	reqCap := uint64(50000)
 	if err := backend.SetOptions(ctx, "type.googleapis.com/session.SessionBindingRequest", &databrokerpb.Options{
-		Capacity:        &reqCap,
+		Capacity:        new(uint64(50000)),
 		IndexableFields: []string{"key"},
 		Ttl:             durationpb.New(15 * time.Minute),
 	}); err != nil {
@@ -788,6 +787,12 @@ func (srv *backendServer) setupRequiredIndex(ctx context.Context, backend storag
 			"state_id",
 		},
 		Ttl: durationpb.New(15 * time.Minute),
+	}); err != nil {
+		return err
+	}
+
+	if err := backend.SetOptions(ctx, "type.googleapis.com/oauth21.MCPRefreshToken", &databrokerpb.Options{
+		Capacity: new(uint64(50000)),
 	}); err != nil {
 		return err
 	}
--- a/authorize/authorize.go
+++ b/authorize/authorize.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pomerium/pomerium/authorize/internal/store"
 	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/internal/mcp"
 	"github.com/pomerium/pomerium/internal/recording"
 	"github.com/pomerium/pomerium/internal/telemetry/metrics"
 	"github.com/pomerium/pomerium/pkg/cryptutil"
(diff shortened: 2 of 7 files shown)


alexandreLamarre (commented):
Thanks for opening this issue report/ PR! Adding some comments for clarification on the edits:
We already have some existing mechanisms to evict / bound temporary records in storage.

In-memory storage may be the **default** path, but it is not the recommended path for production deployments. Deleting the tokens in the handler doesn't limit storage for existing users using persistent storage.

We want to keep Refresh tokens around for particular authorization decisions as is recommended by the OAuth 2.1 spec & security best practices (notably RFC 9700). The linked issue with respect to token inconsistent rotation is being looked at independently. And additional improvements with respect to RFC 9700 are being looked into.

internal/databroker/server_backend.go
```diff
@@ -792,6 +791,12 @@ func (srv *backendServer) setupRequiredIndex(ctx context.Context, backend storag
 		return err
 	}
 
+	if err := backend.SetOptions(ctx, "type.googleapis.com/oauth21.MCPRefreshToken", &databrokerpb.Options{
+		Capacity: new(uint64(50000)),
+	}); err != nil {
```
alexandreLamarre, line 796: Now that I think about it, this is also probably less than ideal. A databroker syncer handler might make sense here.

alexandreLamarre:
sorry @ititandev I will close this one in favor of my PR, since a couple of in-flight improvements depend on some changes here, which makes it a little difficult to continue working from this branch.

PR #6568, opened 2026-07-22
ci: update 0-33-0 dependencies

apparitor:
This PR updates dependencies not managed by dependabot.

### DIFF
--- a/pkg/grpc/config/config.openapi.yaml
+++ b/pkg/grpc/config/config.openapi.yaml
@@ -1,6 +1,7 @@
 openapi: 3.1.0
 info:
   version: v0.33.0
+  version: v0.33.0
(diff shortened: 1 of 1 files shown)


wasaga:
Closing: the only change here over `0-33-0` was a duplicate `info.version` key in `pkg/grpc/config/config.openapi.yaml` — the root-cause was a non-idempotent version injection in `pkg/grpc/protoc.bash`. That's now fixed on `main` (#6595) and backported to `0-33-0` (#6596). Rebasing this branch onto the updated `0-33-0` and regenerating leaves an empty diff, so there's nothing left to merge. The dependency-update bot will open a fresh PR with the corrected tooling on its next run.

PR #6582, opened 2026-07-23
config: secrets bindings block and route reference validation

wasaga:
Wires secrets into configuration. Adds `headertemplate.References` (discovering references without rendering), the `secrets:` YAML block and its `Settings.secrets` proto counterpart, and two-phase validation driven from `Options.Validate`: the binding table is validated against the provider registry, then every route's `set_request_headers` value is checked so each `${secret.ID}` resolves. A stub-render residue rule rejects dynamic selection such as `${secret.${pomerium.email}}`; response-header surfaces reject secret references in v1. The proto round-trips both ways, so settings-sync deployments can carry bindings from day one.


wasaga:
@greptileai review

wasaga:
@greptileai review

wasaga:
@greptileai review

wasaga:
@greptileai review

wasaga:
@greptileai review

PR #6608, opened 2026-07-27
ssh: refactor auth logic to use updated api

kralicky:
## Summary

Several of the internal SSH auth api messages were changed in https://github.com/pomerium/envoy-custom/pull/257. This PR updates the pomerium-side auth logic to use the updated messages. 

The most important change here is the way it handles tracking the state of the auth flow. Previously there were two messages in our api named `PublicKeyAuthMethodResponse` and `KeyboardInteractiveAuthMethodResponse` which were not related to the actual SSH protocol messages but were intended to hold public key state and oauth state, respectively. These messages were replaced with a single `AuthContext` message which holds the combined state. The `Auth` type was refactored to use the `AuthContext` for state tracking instead of the two other messages. 

This lets us (mostly) decouple the protocol auth methods from the parts of the auth context that are obtained during each method. For example a successful public key auth flow will add public key information to the auth context, but in some cases it can also add session info to the auth context as well, like if there is an existing valid session bound to that public key (and that session is still authorized). 

In addition, the keyboard-interactive method is currently used to facilitate the oauth flow, but we can also use it to implement new auth patterns such as two-person rule, or other features we might want to add in the future that would make use of interactive messages or user input.

This PR also contains improvements to the existing integration test suite and lots of new integration tests which cover more of the auth code. The unit tests in auth_test.go cover similar areas of the code but those tests mock out the evaluator responses and databroker client (mostly to test internal failures or other edge cases) so they are not sufficient for testing the whole auth flow like the integration tests are.

Note that there are a few locations with placeholder code (auth.go:114, auth.go:363-375); this is for the upcoming two-person rule feature, since support for that auth flow was used to gauge what refactors would be needed. In that example the keyboard-interactive flow can be used for two different purposes, either to obtain session context or (if there is already session context, and another flag is set) to trigger the two person flow. This does rely on using the same `HandleKeyboardInteractiveMethodRequest` for both, and inspecting the current state to decide what to do next, which is maybe not perfect but I didn't want to make too many changes yet to the `AuthInterface` outside of splitting up the stream info/auth info args. In the future it may be useful to combine the two methods into one, but there wasn't an immediately obvious way to do so.

## Related issues

https://github.com/pomerium/envoy-custom/pull/257

## User Explanation

None

## AI disclosure

None

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/config/policy.go
+++ b/config/policy.go
@@ -823,6 +823,10 @@ func (p *Policy) Validate() error {
 			return fmt.Errorf("config: mcp authorization_server_url must be https, got %q", u.Scheme)
 		}
 	}
+
+	if p.IsSSH() && p.AllowPublicUnauthenticatedAccess {
+		return fmt.Errorf("config: allow_public_unauthenticated_access cannot be used with ssh routes")
+	}
 	return nil
 }
 
--- a/config/policy.go
+++ b/config/policy.go
@@ -830,6 +830,10 @@ func (p *Policy) Validate() error {
 			return fmt.Errorf("config: mcp authorization_server_url must be https, got %q", u.Scheme)
 		}
 	}
+
+	if p.IsSSH() && p.AllowPublicUnauthenticatedAccess {
+		return fmt.Errorf("config: allow_public_unauthenticated_access cannot be used with ssh routes")
+	}
 	return nil
 }
 
--- a/pkg/ssh/auth.go
+++ b/pkg/ssh/auth.go
@@ -49,7 +49,10 @@ const (
 
 //nolint:revive
 type SSHEvaluator interface {
+	// Evaluates whether a user can connect to this route
 	EvaluateSSH(ctx context.Context, streamID uint64, req AuthRequest, initialAuthComplete bool) (*evaluator.Result, error)
+
+	// Evaluates whether a user can attach an upstream tunnel to this route
 	EvaluateUpstreamTunnel(ctx context.Context, req AuthRequest, route *config.Policy) (*evaluator.Result, error)
 }
 
@@ -108,6 +111,7 @@ func NewAuth(
 	currentConfig *atomic.Pointer[config.Config],
 	tracerProvider oteltrace.TracerProvider,
 	codeIssuer code.Issuer,
+	_ any, // temporary placeholder
 	opts ...Option,
 ) *Auth {
 	options := Options{
@@ -137,18 +141,22 @@ func (a *Auth) GetDataBrokerServiceClient() databroker.DataBrokerServiceClient {
 
 func (a *Auth) HandlePublicKeyMethodRequest(
 	ctx context.Context,
-	info StreamAuthInfo,
+	streamInfo StreamInfo,
+	authInfo StreamAuthInfo,
 	user api.UserRequest,
 	req *extensions_ssh.PublicKeyMethodRequest,
-) (PublicKeyAuthMethodResponse, error) {
+) (AuthMethodResponse, error) {
 	ctx, span := a.tracer.Start(ctx, "authorize.ssh.HandlePublicKeyMethodRequest")
 	defer span.End()
-	resp, err := a.handlePublicKeyMethodRequest(ctx, info, user, req)
+	resp, err := a.handlePublicKeyMethodRequest(ctx, streamInfo, authInfo, user, req)
 	if err != nil {
 		log.Ctx(ctx).Error().Err(err).Msg("ssh publickey auth request error")
-		span.SetStatus(otelcode.Error, "internal error")
-		return resp, status.Error(codes.Internal, "internal error")
+		span.SetStatus(otelcode.Error, err.Error())
+		// TODO: handle ShowErrorDetails here
+		return resp, status.Error(codes.PermissionDenied, "permission denied")
 	}
+	resp.Validate()
+	span.SetStatus(otelcode.Ok, resp.String())
 	return resp, err
 }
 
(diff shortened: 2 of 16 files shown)


config/policy.go
```diff
@@ -823,6 +823,10 @@ func (p *Policy) Validate() error {
 			return fmt.Errorf("config: mcp authorization_server_url must be https, got %q", u.Scheme)
 		}
 	}
+
+	if p.IsSSH() && p.AllowPublicUnauthenticatedAccess {
```
kenjenkins, line 834: This looks like a change in behavior. I think this PR would be easier to review if we split off any behavior changes to a separate PR.

pkg/ssh/stream.go
```diff
@@ -45,25 +50,66 @@ type KeyboardInteractiveQuerier interface {
...
 }
 
-type (
-	PublicKeyAuthMethodResponse           = AuthMethodResponse[extensions_ssh.PublicKeyAllowResponse]
-	KeyboardInteractiveAuthMethodResponse = AuthMethodResponse[extensions_ssh.KeyboardInteractiveAllowResponse]
-)
+func (r *AuthMethodResponse) Validate() {
+	if r.AllowMethod {
+		if r.NextRequiredAuthMethod == "" && !r.NoFurtherMethodsRequired {
+			panic("invalid AuthMethodResponse: either NextRequiredAuthMethod or NoFurtherMethodsRequired must be set if AllowMethod is true")
```
kenjenkins, line 63: Ordinarily I would expect a Validate() method to return an error rather than panicking. Is there a reason than panic() is better here?

pkg/ssh/ssh_int_test.go
kenjenkins, line 1: Would it better to split off the integration test changes into a separate PR as well? If we merge any test improvements first I think that would help to demonstrate that the refactoring changes are sound.

kralicky: Yes that is a good idea.

kralicky: https://github.com/pomerium/pomerium/pull/6732

pkg/ssh/stream.go
kralicky: This is a failsafe to try to catch future programmer error. If any of these panics are hit, it means there is either a serious bug in the auth logic, or something was changed in a way that breaks an existing invariant and additional code changes are required.

config/policy.go
kralicky: This didn't work before, it would function the same as allow_any_authenticated_user, which was technically a bug. After these changes it still won't work, but the user will be denied instead and an error will be logged. A policy with this option will never be usable so this just gives a better error message for this case. There are still other ways to construct an invalid policy but this is easy to validate compared to arbitrary ppl.

pkg/ssh/auth.go
```diff
@@ -108,6 +111,7 @@ func NewAuth(
 	currentConfig *atomic.Pointer[config.Config],
 	tracerProvider oteltrace.TracerProvider,
 	codeIssuer code.Issuer,
+	_ any, // temporary placeholder
```
kenjenkins, line 114: What is this for?

pkg/ssh/auth.go
```diff
@@ -479,12 +604,21 @@ func (a *Auth) reportLoginCodeFailure(
 
 var errAccessDenied = status.Error(codes.PermissionDenied, "access denied")
 
-func (a *Auth) EvaluateDelayed(ctx context.Context, info StreamAuthInfo, user api.UserRequest) error {
-	req, err := a.sshRequestFromStreamAuthInfo(ctx, info, user)
+func (a *Auth) EvaluateDelayed(ctx context.Context, streamInfo StreamInfo, authInfo StreamAuthInfo, user api.UserRequest) error {
+	if user.Hostname() == "" {
+		panic("bug: EvaluateDelayed called with empty hostname")
```
kenjenkins, line 609: It looks like a panic here will cause Pomerium to exit. Can we return an error instead?

pkg/ssh/auth.go
kralicky: It's replaced here https://github.com/pomerium/pomerium/pull/6736/changes#diff-dfa698e1898e30a60acd64f3a596c2bdb7e42b257fa046d2f6382f03185e0f9c in a follow up. Changes to the number of arguments in this method affect mock usage and iirc was causing some annoying conflicts since there are a few other prs that depend on this one. It may or may not be still important. This code was split from the above linked pr at some point.

pkg/ssh/auth.go
kralicky: No, this is an invariant check. It should be impossible to hit unless the api is used incorrectly.

pkg/ssh/stream.go
```diff
@@ -253,7 +287,12 @@ func (sh *StreamHandler) OnClusterHealthUpdate(_ context.Context, event *datav3.
 }
 
 func (sh *StreamHandler) Terminate(err error) {
```
kenjenkins, line 290: Is it possible for Terminate() to be called multiple times? Do we need any check here in case it was already called?

kenjenkins (approved):

pkg/ssh/stream.go
kralicky: Hmm, maybe. If the context was already canceled then canceling it again would be a no-op, but then terminateErr might end up being something different than the cancel cause. But the only time terminateErr is ever checked is in a situation where terminateFunc is nil. Perhaps that is too confusing.

alexandreLamarre (approved):
Some TODOs that look leftover. Might be addressed in subsequent PRs, but worth taking a second look at before merging

PR #6626, opened 2026-07-31
feat/idp atomic token refresh

alexandreLamarre:
## Summary

With the introduction of MCP, and later on agentic identities, we need to atomically refresh tokens when multiple internal pomerium resources may be interested in the same oauth refresh token.

For instance, the MCP refresh token must refresh the token every type the token grant is requested from MCP clients, but it is also refreshed in the background on the IDP session manager.
In addition, the MCP refresh token outlives sessions, so we need a way to keep track of the last known token in the case where the Pomerium session expires. This is done in a follow PR, that uses a databroker syncer to update the latest MCP refresh token with the current session.Session token.

## Related issues

Related to https://github.com/pomerium/pomerium/issues/6530

## Checklist

- [ ] updated unit tests
- [ ] ready for review

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/authorize/authorize.go
+++ b/authorize/authorize.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pomerium/pomerium/authorize/internal/store"
 	"github.com/pomerium/pomerium/config"
 	"github.com/pomerium/pomerium/internal/log"
+	"github.com/pomerium/pomerium/internal/mcp"
 	"github.com/pomerium/pomerium/internal/recording"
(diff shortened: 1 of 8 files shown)


alexandreLamarre:
Superseded by a centralized implementation like the prototype in https://github.com/pomerium/pomerium/pull/6665

PR #6650, opened 2026-08-10
test(acceptance): observability e2e suite for logging, metrics and OTEL tracing

ssveta7ak:
## Summary

Adds a browser-driven end-to-end suite that exercises Pomerium's **observability surface** — access
and authorize logging, the Prometheus metrics endpoint, and OTEL tracing. It automates the QA plans
*Notion → QA → Test Plans → Core →* **Core.Logging**, **Core.Metrics** and **Core.OTEL Tracing**.

Everything runs in containers via [Testcontainers](https://node.testcontainers.org/) against the
**official `pomerium/pomerium:main` image**, reusing the shared Keycloak realm fixture. A real
Chromium signs in through the IdP and opens the protected route, then the tests read back what
Pomerium emitted about itself: JSON log lines on stdout, Prometheus text on the metrics port, and
spans queried from a real Jaeger collector.

**24 tests across 3 signals, all passing.** No production code changes — this is a test-only
addition under `internal/acceptance/suites/`, plus the four wiring files every new harness touches.

## Coverage

| Setting(s) under test | Spec | Cases |
|---|---|---|
| `access_log_fields` | `logging-access.spec.ts` | 4 — the 12 defaults exactly, a single field suppressing the rest, `headers.<Name>`, and an empty list logging the envelope only |
| `authorize_log_fields` | `logging-authorize.spec.ts` | 4 — the defaults for a signed-in request, `headers.Cookie`, a single field, and the full sign-in decision trail |
| `metrics_address`, `metrics_basic_auth` | `metrics.spec.ts` | 4 — listener absent when unset, aggregated `/metrics` vs raw `/metrics/envoy`, basic auth, malformed value rejected at load |
| `otel_traces_exporter`, `otel_exporter_otlp_endpoint`, `OTEL_SDK_DISABLED` | `tracing-enablement.spec.ts` | 6 — the enablement matrix, including both endpoint forms and the env kill switch |
| `otel_traces_sampler_arg` | `tracing-spans.spec.ts` | 2 — Envoy+Pomerium trace correlation, and sampling at 0.0 |
| (no new settings) | `tracing-journeys.spec.ts` | 4 — the same signals read back through the **Jaeger UI in the browser**: sign in → browse → find and open the trace; the sign-in's own stitched trace; the `traceparent` the upstream received; a policy-refused 403 |

Two of these are regression guards for named issues:

- **TC-MET-05** — malformed `metrics_basic_auth` must be *rejected at config load*. It used to
  silently disable authentication ([ENG-4311](https://linear.app/pomerium/issue/ENG-4311)); this
  pins the fix.
- **TC-TRC-01** — the *generic* `otel_exporter_otlp_endpoint` must enable **Envoy** tracing, not
  only Pomerium's own spans ([ENG-1960](https://linear.app/pomerium/issue/ENG-1960)).

## How it works

- **Topology** — one Docker network booted once per run: Keycloak (IdP), `pomerium/verify` (the
  protected upstream), and Jaeger 2.20.0 (OTLP collector + query API) — the one third-party image
  this suite introduces, **digest-pinned** like the verify image (official CNCF project image,
  non-root, static binary on alpine-minirootfs; no open advisories against Jaeger v2 core at time
  of writing). Pomerium itself is
  started **per test** with a freshly generated config. `*.localhost.pomerium.io` resolves to
  `127.0.0.1` and doubles as a network alias, with identical host/container ports, so a URL means the
  same thing to the browser and inside the network — which is what makes the OIDC issuer check pass
  front-channel and back-channel without rewriting.
- **Config injection** — `generateConfig()` writes each variant to `.gen/<name>.yaml` (JSON, which
  is valid YAML) and it is bind-mounted read-only at `/pomerium/config.yaml`. Env vars are the second
  path, used where a setting has no config-file form (`OTEL_SDK_DISABLED`).
- **Certificates** — one `mkcert` wildcard leaf for `*.localhost.pomerium.io`, covering both the
  route and the authenticate host. `mkcert -install` is deliberately never run (it needs sudo); the
  browser and request contexts set `ignoreHTTPSErrors`. Nothing is committed — `.certs/` and `.gen/`
  are gitignored.
- **Reading the signals** — logs come from the container's captured stdout, metrics from a scrape of
  the published port, traces from Jaeger's query API. Each has its own helper module and its own
  timing hazard to absorb. Log field configuration is asserted by rejecting any key outside the
  configured set plus the always-on envelope — a complement check, so it cannot drift against
  `pkg/logfields` the way a mirrored field list would.
- **The journeys assert through the UI instead** — `tracing-journeys.spec.ts` does what a person
  does: the same Chromium that signed in opens the Jaeger UI, searches, clicks into a trace and
  reads it. The Jaeger interaction is written out in each test body rather than behind a helper, so
  the HTML report and retained trace read as the journey. Locators were captured against the pinned
  image, so bumping `JAEGER_IMAGE` means re-capturing them — noted at the pin in
  `setup/containers.ts`.

## Notable behaviors this documents

One finding is worth flagging for review, since the assertions encode it:

- **Signing in *on* a path produces two traces for it, not one.** The trace id rides the sign-in
  redirect as `pomerium_traceparent`, so the refused request's trace absorbs the whole chain
  (`/.pomerium/sign_in`, `/oauth2/callback`, `/.pomerium/callback/`) and the request that follows
  is a trace of its own. A third row also appears on the Linux CI runners: after the callback
  Pomerium redirects once more to strip the `pomerium_*` params it added, and whether that second
  successful request joins the previous trace is timing-dependent. TC-TRC-21 therefore asserts
  both *kinds* of trace exist rather than an exact count.
- **`/metrics/envoy` is not behind `metrics_basic_auth`.** Basic auth is Go middleware wrapping only
  the aggregated handler; the Envoy path is a separate route to the admin cluster with no auth
  filter. TC-MET-04 asserts the full stats come back **without credentials**, so the gap is at least
  covered behavior rather than a surprise. Worth deciding whether to close it or document it.

## Coverage gap worth a decision

`otel_traces_sampler_arg` **clamping** (`max(0, min(1, x))`) is not covered. Two e2e cases used to
assert it and I removed both: `1.0` is the unset default (`RandomSampling` is only set when the
option is present), so one asserted the default that every other positive already proves, and the
other's two halves were observationally identical to the remaining `0.0` case — for two container
boots each. `config/envoyconfig/listeners_test.go` covers `sampler_arg: 1.0` through a golden file,
but nothing exercises the boundaries, so a `5.0` silently becoming 500% would ship green.

A table test over `RandomSampling.Value` for `{-1, 0, 0.5, 1, 5}` in `config/envoyconfig` is the
right home — it pins the arithmetic directly instead of inferring it from whether spans appeared, and
runs on every PR unlike this Docker-dependent suite. Say the word and I'll add it here rather than
following up.

## Verification

```bash
cd internal/acceptance
make deps-observability     # workspace deps, Chromium, refresh pomerium:main
make test-observability     # 24 tests, ~2.3 min
```

Requires `mkcert` and Docker on the host. The suite binds fixed host ports (8443, 8080, 9902,
16686), so it runs with `workers: 1` and cannot run alongside `make up` or the sibling container
suites. Note the published ports bind all interfaces for the duration of a run (documented in the
README — everything in the stack is synthetic; prefer trusted networks or CI). `OBS_E2E_LOGS=1`
streams container logs; the Jaeger UI at `http://127.0.0.1:16686` is the fastest way to see why a
trace assertion missed.

Run locally several times, including from a clean `.certs/` and `.gen/`, and under
`--repeat-each=2`, with no flakes observed.

## Related issues

- Regression guard for [ENG-4311](https://linear.app/pomerium/issue/ENG-4311) (malformed
  `metrics_basic_auth` silently disabling auth)
- Regression guard for [ENG-1960](https://linear.app/pomerium/issue/ENG-1960) (generic OTLP endpoint
  not enabling Envoy tracing)
- #6648 (upstream TLS/mTLS) merged first and touched the same four wiring files
  (`suites/package.json` + lockfile, `internal/acceptance/Makefile`,
  `.github/workflows/acceptance.yaml`). `main` is merged into this branch and the conflicts are
  resolved additively: both suites keep their workspace entry, Make targets and CI job, and
  `summary.needs` lists all five. The lockfile was regenerated rather than hand-merged.

## User Explanation

No user-facing change. This is test-only coverage for existing behavior.

## AI disclosure

Claude Code (Opus 5) assisted throughout: it drafted the harness scaffolding, the helper modules and
the spec files from the QA plans, and applied a self-review pass for reuse and simplification. I
directed the design decisions (one combined suite rather than three, which cases to automate first,
dropping upstream mTLS from scope once #6648 covered it), verified every behavioral claim against the
running stack and the Go source, and reviewed the final diff. Several assertions were corrected
against real container output rather than the plan's expectations — the `server-name` envelope
field, the 13-of-18 authorize defaults, and the outcome-dependent reason arrays.

The Jaeger UI journeys were added the same way. Every claim they encode was measured against the
running stack before it was written, which is how the design survived contact: the sign-in chain
turned out to be one stitched trace rather than several, `/json` turned out not to echo
`traceparent` (`/headers` does), and a plain proxied request carries only Envoy + Control Plane +
Authorize — an earlier draft asserting five Pomerium services would have been wrong. The Jaeger
selectors were captured from the pinned image rather than guessed. The TC-TRC-21 count fix was
verified by forcing the CI shape locally and confirming the old assertion failed there with the same
`Expected: 2, Received: 3` before confirming the new one passed.

## Checklist

- [x] updated unit tests 
- [x] add appropriate label (`enhancement`, `bug`, `breaking`, `dependencies`, `ci`) — suggest `ci`
- [x] ready for review

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/acceptance/suites/observability/tests/tracing-enablement.spec.ts
alexandreLamarre, line 1: just BTW I'm not opposed to duplicating tests, but we have tests that cover this functionality in `internal/testenv/selftests/tracing_test.go` that is a little more precise than the expected behavior here (This file and `tracing-span.spec.ts`). Might be useful to reference?

ssveta7ak: Yeah, understood. The main purpose of the Playwright E2E tests is to simulate a real environment with a real IdP (Keycloak), a real browser, Jaeger for monitoring traces, and an actual end-user flow.

PR #6651, opened 2026-08-10
feat(file): singleflight 

alexandreLamarre:
## Summary

Adds a file based implementation of a `DoTransaction` callback that implements a single-flight like mechanism for the storage layer.

The callback to `storage.DoTransaction` is a function that itself is allowed to do storage operations like `Get`, `Put`, `Patch`, `Query`.

## Related issues

[RFC-Databroker-singleflight](https://app.notion.com/p/pomerium/RFC-Databroker-singleflight-3b158e156c55802b820fffb6a3bdd39f?d=3b858e156c5580faa38a001c7f8c206e#3b258e156c55806e9142ea92c8822e34)

## User Explanation

N/A

## AI disclosure

none

## Checklist

- [X] updated unit tests
- [X] ready for review

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/pkg/grpc/databroker/databroker.proto
+++ b/pkg/grpc/databroker/databroker.proto
@@ -243,8 +243,6 @@ message TransactionResponse {
   }
 }
 
-
-
 message Checkpoint {
   uint64 server_version = 1;
   uint64 record_version = 2;
--- a/pkg/grpc/databroker/databroker.proto
+++ b/pkg/grpc/databroker/databroker.proto
@@ -223,6 +223,28 @@ service DataBrokerService {
   rpc SyncLatest(SyncLatestRequest) returns (stream SyncLatestResponse);
 }
 
+message TransactionRequest {
+  string key = 1;
+  oneof operation {
+    GetRequest get = 2;
+    PutRequest put = 3;
+    PatchRequest patch = 4;
+    QueryRequest query = 5;
+  }
+}
+
+message TransactionResponse {
+  string key = 1;
+  oneof operation {
+    GetResponse get = 2;
+    PutResponse put = 3;
+    PatchResponse patch = 4;
+    QueryResponse query = 5;
+  }
+}
+
+
+
 message Checkpoint {
   uint64 server_version = 1;
   uint64 record_version = 2;
--- a/pkg/storage/file/backend.go
+++ b/pkg/storage/file/backend.go
@@ -398,8 +398,7 @@ func (backend *Backend) Versions(
 func (backend *Backend) DoTransaction(
 	ctx context.Context,
 	key string,
-	fn func(tx storage.Transaction,
-	) error,
+	fn func(tx storage.Transaction) error,
 ) (changed []*databrokerpb.Record, shared bool, err error) {
 	if err := backend.init(); err != nil {
 		return nil, false, fmt.Errorf("pebble : error initializing : %w", err)
--- a/pkg/storage/file/backend.go
+++ b/pkg/storage/file/backend.go
@@ -15,6 +15,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 	oteltrace "go.opentelemetry.io/otel/trace"
+	"golang.org/x/sync/singleflight"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
@@ -59,6 +60,9 @@ type Backend struct {
 	closeErr  error
 	closeCtx  context.Context
 	close     context.CancelFunc
+
+	// primitives used for tracking transactions
+	txGroup singleflight.Group
 }
 
 // Option configures the backend instance.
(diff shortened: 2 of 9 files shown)


pkg/storage/storagetest/storagetest.go
alexandreLamarre, line 1: There is a flaky test to address here (usually hit with -count>1000),  possibly can use sync-test there.

pkg/storage/file/backend.go
```diff
@@ -393,6 +395,39 @@ func (backend *Backend) Versions(
 	return serverVersion, earliestRecordVersion, latestRecordVersion, nil
 }
 
+func (backend *Backend) DoTransaction(
+	ctx context.Context,
+	key string,
+	fn func(tx storage.Transaction,
+	) error,
```
calebdoxsey:
```suggestion
	fn func(tx storage.Transaction) error,
```

pkg/storage/storage.go
```diff
@@ -62,6 +62,28 @@ type Backend interface {
...
+//
+// Only persistent indices are updated, so do not submit operations on records
+// that should update an in-memory index.
+type Transaction interface {
+	// Submit adds an operation to the transaction. Intermediary responses from operations
+	// return information.
+	Submit(*databroker.TransactionRequest) (*databroker.TransactionResponse, error)
+	// commit applies the submitted operations. Concurrent calls for the same key
+	// block until the in-flight commit completes.
+	Commit() error
```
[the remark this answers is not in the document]
alexandreLamarre: This would be an incorrect use of the backend interfaces. I would have Commit be a private function but that is not possible with our current module structure.

pkg/grpc/databroker/databroker.proto
```diff
@@ -223,6 +223,28 @@ service DataBrokerService {
...
+  string key = 1;
+  oneof operation {
+    GetResponse get = 2;
+    PutResponse put = 3;
+    PatchResponse patch = 4;
+    QueryResponse query = 5;
+  }
+}
+
+
```
kenjenkins: formatting nit: looks like there's a few extra blank lines here?

pkg/storage/storagetest/storagetest.go
```diff
@@ -1402,3 +1405,528 @@ func truncateTimestamps(ts ...*timestamppb.Timestamp) {
...
+				results[0].changed, results[0].shared, results[0].err = backend.DoTransaction(context.Background(), "k", func(tx storage.Transaction) error {
+					callbacks.Add(1)
+					close(started)
+					<-release
+					_, err := tx.Submit(putRequest(newTestRecord(t, "r1", nil)))
+					return err
+				})
+			})
+
+			requireReceive(t, started, "first transaction did not start")
```
kenjenkins: It looks like requireReceive() uses a timeout. But because this test case uses synctest.Test() I wouldn't think we need any timeouts?

If we move the synctest.Wait() call from line 1585 up to here — before the rest of the transactions start — would that ensure the first transaction is blocked on the `release` receive?

(And if we do that, would that make the `started` channel unnecessary?)

alexandreLamarre: ~~yes, probably. I remembering refactoring some tests with timing issues to use synctest, so it is possible I missed this one, or there is a good reason synctest doesn't work.~~

~~In the follow up PR for the postgres implementation, it unfortunately isn't possible to use synctest because pgx dep library has a background goroutine that never durably blocks in the way synctest expects~~

alexandreLamarre: ~~It's not great but I reused `requireReceive` in tests where they were timing dependent (checking that all shared callers of singleflight are waiting for the result) and timing independent (sequential control flow). But did move timing dependent ones to use synctest to prevent flakes.~~

~~(Also, postgres implementation can't use `synctest` because there's a background goroutine that never blocks in pgx)~~

~~I can probably change all these tests to use `<-done` (or similar) instead of `requireReceive` but if there's a legitimate bug (or an invariant that's unhandled) the tests could hang and might be harder to pinpoint the expected behavior that failed? Its kinda tricky.~~

~~The silver lining is that I fixed all the test flakes by running `go test -race -count=10000` over and over for file storage, so I'm fairly confident only a critical failure like a failed syscall (futex) would cause a test flake for file storage.~~

WDYT @kenjenkins?

alexandreLamarre:
```
nice -n 19 go test -run="TestTransactions" -count=10000 ./...
ok      github.com/pomerium/pomerium/pkg/storage/file   42.489s
```
Verified again that this is safe under cpu burn (near 100% cpu saturation and scheduler de-prioritization)

alexandreLamarre: Oh wait I think I misread the original suggestion, I think that will work.

kenjenkins: I'll leave it up to you.

kenjenkins (approved):

pkg/storage/storagetest/storagetest.go
alexandreLamarre: Just rebasing the stack (since the postgres implementation made some changes to the storagetest implementation). 
Good catch!

PR #6659, opened 2026-08-11
testenv: reserve allocated ports so upstreams cannot take them from envoy

wasaga:
## Summary

Fixes a CI flake in testenv-based suites where envoy comes up without its listeners.

`netutil.AllocatePorts` binds ephemeral ports and immediately closes them, so the 13 ports a test environment claims for pomerium and envoy are reserved only by convention — the kernel is free to hand the same port to a later bind. The in-process test servers bound `:0` some time after that allocation and before envoy bound its share, so they could take one. When that happened envoy rejected the whole listener update:

```
listener 'https-ingress' failed to bind or apply socket options:
  cannot bind '127.0.0.1:43125': Address already in use
→ delta config for ...v3.Listener rejected
```

and requests aimed at pomerium silently landed on whichever test server won the race, surfacing as unrelated-looking failures — most recently a TLS error in `TestSSH/#03/TestDirectTcpipSession`, where the sign-in request reached the mock IDP and failed with `x509: certificate is valid for mock-idp.localhost.pomerium.io, not authenticate.localhost.pomerium.io`.

`testenv.AllocatePorts` now records what it hands out, and `testenv.Listen` picks an ephemeral port that avoids those reservations, holding each rejected listener open so the kernel offers a different port on retry. In-process servers go through it: the four upstream types, the middle envoy in the websocket test, and the bare upstream in the MCP ext_proc test.

Reservations are never released, so an environment shutting down cannot race an upstream starting up. They are not held open as live sockets because envoy enables `SO_REUSEPORT` only on linux — elsewhere its bind would fail against a placeholder, and on linux the kernel would load balance connections into it.

Not addressed here: `databroker.New` binds `127.0.0.1:0` inside the pomerium process, which sits in the same allocate-then-bind window and affects production as well as tests. Fixing that means moving the reservation into `pkg/netutil`.

## User Explanation

No user-facing changes; test infrastructure only.

## AI disclosure

Claude Code diagnosed the failure from CI logs and wrote the change; reviewed and directed by me.

## Checklist

- [x] updated unit tests
- [x] ready for review

### DIFF
--- a/internal/testenv/ports.go
+++ b/internal/testenv/ports.go
@@ -1,0 +1,80 @@
+package testenv
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"sync"
+
+	"github.com/pomerium/pomerium/pkg/netutil"
+)
+
+// reservedPorts holds every port AllocatePorts has handed out in this process.
+//
+// netutil.AllocatePorts works by binding ephemeral ports and immediately
+// closing them, so the port numbers it returns are only reserved by convention:
+// nothing stops the kernel from handing the same port to a later bind. Pomerium
+// and envoy bind their share of those ports some time afterwards, so an
+// in-process server binding to :0 in between can take one out from under them.
+// When that happens envoy rejects the whole listener update ("failed to bind or
+// apply socket options: cannot bind ...: Address already in use") and requests
+// silently land on whichever server won the race.
+//
+// Holding the reservations open instead of tracking them is not an option:
+// envoy only enables SO_REUSEPORT on linux (see config/envoyconfig.newListener),
+// so elsewhere its bind would fail against a held-open socket, and on linux the
+// kernel would load balance connections into the placeholder.
+//
+// Ports are never released, so that an environment shutting down cannot race an
+// upstream starting up. The set stays small relative to the ephemeral port
+// range: even a binary that builds dozens of environments reserves well under
+// 1000 ports out of ~28000, so listenUnreserved rarely retries at all.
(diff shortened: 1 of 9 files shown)


wasaga:
@greptileai review

internal/testenv/ports.go
```diff
@@ -0,0 +1,80 @@
...
+		li, err := (&net.ListenConfig{}).Listen(ctx, "tcp", net.JoinHostPort(host, "0"))
+		if err != nil {
+			return nil, err
+		}
+		_, port, _ := net.SplitHostPort(li.Addr().String())
+		if !isReserved(port) {
+			return li, nil
+		}
+		// Hold the listener open so the next attempt gets a different port.
+		rejected = append(rejected, li)
```
kenjenkins: Rather than keeping a separate `rejected` slice, would it be simpler to `defer li.Close()` here? i.e.

```go
func listenUnreserved(ctx context.Context, host string, isReserved func(port string) bool) (net.Listener, error) {
	for range listenMaxAttempts {
		li, err := (&net.ListenConfig{}).Listen(ctx, "tcp", net.JoinHostPort(host, "0"))
		if err != nil {
			return nil, err
		}
		_, port, _ := net.SplitHostPort(li.Addr().String())
		if !isReserved(port) {
			return li, nil
		}
		// Hold the listener open so the next attempt gets a different port.
		defer li.Close()
	}
	return nil, fmt.Errorf("no unreserved port available on %s after %d attempts", host, listenMaxAttempts)
}

kenjenkins (approved):

PR #6665, opened 2026-08-13
idpsession: one owner for each user's upstream IdP session

wasaga:
## Summary

Adds `internal/identity/idpsession`, which keeps one canonical databroker record per `(user_id, idp_id)` holding that user's upstream IdP session, and moves the three components that refresh upstream tokens onto it:

- `internal/identity/idpsession` — the store, and the sole presenter of a user's upstream refresh token.
- `pkg/identity/manager` — browser sessions refresh through the store instead of each rotating their own copy of the same grant.
- `internal/mcp` — the token endpoint recreates sessions from the same record.
- `internal/authenticateflow` — a login installs the canonical record.

### The two invariants

Everything here exists to hold two properties:

1. **A refresh token is never presented twice.** Every presentation is preceded by a committed write-ahead intent. An intent whose outcome never committed is resolved *without* contacting the IdP, because re-presenting a possibly-consumed token is what trips reuse detection and revokes the user's whole grant family. The one exception is a family every completed refresh has shown does not rotate, where presenting the same token again is idempotent.
2. **Death is absolute.** A record that dies stays dead. No copy of the token any consumer holds revives it, whatever killed it, because every such copy belongs to the grant that died. Only `Supersede` revives a record, and only the login flow calls it.

Both hold only while the token family has a single presenter, which is why the identity manager migration is part of this change rather than a follow-up. With the manager still rotating its own copy, two independent presenters would share one family and neither invariant would survive contact with a rotating provider.

### How it works

```mermaid
flowchart LR
    subgraph consumers ["Consumers (any replica)"]
        MGR["identity manager<br/>(leased, in databroker svc)"]
        MCP["MCP /token handler<br/>(per proxy replica)"]
        LOGIN["login flow<br/>(authenticate service)"]
    end
    STORE["idpsession.Store<br/>Register / EnsureLive / Supersede"]
    DO["databrokerutil.Do(key)<br/>winner runs · losers adopt<br/>leader-serialized, cluster-wide"]
    REC[("UpstreamIdPSession<br/>one per (user, idp)<br/>sole owner of refresh token")]
    IDP(["IdP token endpoint<br/>≤1 presentation per question"])
    PROJ["projections<br/>Session.oauth_token · MCPRefreshToken"]

    MGR --> STORE
    MCP --> STORE
    LOGIN --> STORE
    STORE --> DO
    DO --> REC
    DO -- "refresh inside flight" --> IDP
    REC -- "one-way sync<br/>(manager patch cycle)" --> PROJ
    PROJ -. "Register seeds<br/>(bootstrap only)" .-> STORE
```

`EnsureLive` proves the user's upstream session is still alive and returns a snapshot. The refresh token is presented at most once cluster-wide per refresh window, enforced with the databroker singleflight: a caller commits an intent before presenting, and concurrent callers on any replica adopt the outcome that attempt commits. The IdP call runs outside every transaction on a detached context, as does the claim flight, since the server commits the intent before the acknowledgement reaches the client. Outcomes are committed as record state, never signalled through a flight error.

A caller waits at most `callerBudget` (8 seconds) across every phase — polling, flights, and the IdP call it may itself have started. Work that has committed an intent runs to completion on a detached context bounded by `refreshTimeout + commitTimeout`, and whoever asks next adopts what it committed. The MCP refresh grant caps its whole recreate sequence at 12 seconds, so the endpoint answers with its own 503 and a `Retry-After` rather than being replaced by the gateway's bodyless 504 at 15.

Every age the store measures is bounded by the databroker's own timestamp for the record, so no replica can date a write into the future and stretch what every other replica sees.

### Timing, and the tests that pin the invariants

- The detached attempt is bounded at 90s — `refreshTimeout+commitTimeout=60s` covering the claim flight and the IdP call, plus a detached `commitTimeout=30s` for the outcome commit, which therefore always has budget. Against `settleDelay=180s` that leaves 90s of margin. `writeAnchor = min(self-stamp, ModifiedAt)`: a fast writer's future stamp is clamped by server time and costs nothing; a slow writer costs δ_slow; a reader ahead of the databroker costs δ_reader; an intent is never resolved before `anchor + 180s − δ_slow − δ_reader`, sound while `δ_slow + δ_reader < 90s`.

  That margin protects the resolve path unconditionally: early settle under strict rotation can only retire a record, never present a token. The probe path's safety under skew rests on the same never-starts-rotating assumption the probe already documents.
- `Supersede` carries an idempotency key. Without it, a retry after a lost acknowledgement could clear an intent committed in between and authorize a second presentation of the same token; with it the replay bound under lost acknowledgements is 0. `TestSupersedeRetryAfterLostAckIsIdempotent` fails without the guard.

The invariants are pinned by tests that run the real store against a reuse-detecting mock provider. `TestAtMostOncePerToken` records every presented token value across three replicas; `TestReuseDetectionFiresOnAReplay` proves that detector actually fires, so those assertions mean something; and `TestFamilyCostIsOnePresentationPerWindow` measures the cost — 180 liveness questions across 3 replicas produce one presentation per debounce window, where the pre-store behavior produced one per consumer per grant.

### The ID token's lifecycle

The record's ID token is seeded by the login that installs it, advanced by any refresh that carries one, and blanked only when a provider configured to overwrite it reports that there is none. Most providers omit `id_token` on refresh, and treating that silence as a withdrawal would blank the session's ID token on the first refresh — breaking the route ID-token headers, sign-out's `id_token_hint`, and refreshing at ID-token expiry.

That last one is why a scheduled refresh asks for freshness rather than accepting any unexpired access token. With `refresh_session_at_id_token_expiration` on and a provider whose ID token expires before its access token, being served a stored token would leave the session's ID token stale, and the scheduler would keep computing a past-due time from it and re-running every cool-off period — ten seconds — for the life of the session.

This fixes that loop for providers that return `id_token` on refresh. A provider that never returns one cannot advance the session's ID token at all, so with that flag enabled the rescheduling persists; the store bounds it to one presentation per debounce window instead of one per cool-off period, roughly a twelvefold reduction, but the manager still wakes up on the same cadence. Turning the flag off, or a provider that returns `id_token`, avoids it entirely.

### Recovery

Signing in again heals. The auth-code exchange produces a token no consumer could have been holding since before the record died, so `PersistSession` installs it through `Supersede`, which starts a new epoch and discards what was learned about the old family. That is the only revival path; it is best-effort and bounded, so a failure costs a lazy re-seed rather than the login. `Register` seeds a record that is absent and refuses every tombstone.

Refresh failures are classified so only an `invalid_grant` refusal ends the session. Transport failures that may or may not have been processed keep the intent; failures that never reached the IdP clear it. A failure Pomerium caused itself, such as `invalid_client`, is logged and keeps every session rather than signing out all users over a rotated client secret. Repeated failures are paced by an escalating `Retry-After` rather than a flat window, so a provider whose access tokens are shorter than the debounce interval is not refused for the rest of every window after one hiccup.

### Record collection

The store registers a 90 day TTL for the record type. Every refresh outcome rewrites the record, so one in active use never expires; a `DEAD` record is collected one TTL after it died, and a `LIVE` record nothing has used for 90 days is collected too, making the TTL an idle-grant expiry.

### MCP `/token` error mapping

| Outcome | Response |
| --- | --- |
| Upstream session is dead | `400 invalid_grant` |
| No upstream session and no token to seed one with | `400 invalid_grant` |
| IdP outage, another replica mid-refresh, refresh debounced | `503 temporarily_unavailable` + `Retry-After` |
| Databroker failure (`Unavailable`, `Aborted`, `DeadlineExceeded`, `Internal`, `FailedPrecondition`) | `503 temporarily_unavailable` + `Retry-After` |
| Caller budget spent while the attempt runs on | `503 temporarily_unavailable` + `Retry-After` |
| Stored record does not parse | `500 server_error`, no `Retry-After` |

The authorization endpoint refuses too, with `401` and a message telling the user to sign in again, when the canonical record is already a tombstone. Issuing a code there would mint tokens the very next refresh grant refuses, so the client would authorize and be refused in turn until something else repaired the record.

`invalid_grant` makes a client discard its refresh token, which only a new login undoes, so it is reserved for a grant that is genuinely over. `databrokerutil` reports failures as typed gRPC status errors that do not implement `Temporary()`, so the store wraps them at its boundary rather than letting them read as a refusal. An unreadable record is an operator problem: nothing may overwrite a record it cannot read, and the way out is a login, which supersedes it.

**OIDC.** `oidc.Provider.GetOauthConfig` now caches the `*oauth2.Config` for the provider's lifetime. An `oauth2.Config` records which client-auth style the token endpoint accepted; rebuilt per call, x/oauth2 re-probes on every token request and re-POSTs a failed request with the other style, presenting a one-time-use refresh token twice on a single failed refresh. For that cache to pay off the provider itself has to be reused, so both places that construct the store hand it a memoized authenticator getter (`idpsession.CacheAuthenticators`).

## User Explanation

A user's browser sessions and MCP clients no longer refresh independently against the identity provider; one refresh covers all of them. An identity provider outage or a momentary network failure no longer signs the user out everywhere, and MCP clients are told to retry rather than to re-authorize. When an upstream session does end — the user signed out at the provider, or the grant was revoked — signing in again restores every session and MCP client at once.

## Deliberately out of scope

- **Projections still carry a copy of the upstream token.** Session records keep the token they were built from. Nothing presents it any more, since the manager and MCP both go through the store, but slimming the projections is separate work.
- **`updateUserInfo`'s ticker does not go through the store.** It calls the provider with the session's cached access token rather than asking for a fresh one, so it is outside the single-presenter path. It presents no refresh token, so it cannot affect either invariant.
- **`Store.Revoke` has no callers.** Sign-out keeps its existing meaning; see the first accepted delta below for what was decided and why.
- **The MCP authorization grant only seeds.** It records the canonical session if the user has none, for a user who authorized an MCP client before any login went through the store. It cannot revive a dead record; authorization against a tombstone is refused so the user is sent through the sign-in that does revive it.

## Accepted differences from the pre-store behavior

These are the deltas this design accepts rather than removes. The first three were considered explicitly and decided.

1. **Sign-out keeps its per-device meaning, and its blast radius is now provider-dependent.** `RevokeSession` still revokes the session's *access* token upstream and `Store.Revoke` stays uncalled, so signing out of one device does not end the shared grant. The cost is that under one grant per user, an upstream revocation can — per RFC 7009's related-token behavior, which providers implement differently — end the grant for every session and MCP client of that user; and until then the canonical record stays LIVE and serves the revoked access token for up to a debounce window. Wiring `Store.Revoke` for global sign-out was considered and rejected: it would silently turn per-device sign-out into sign-out-everywhere.
2. **An unrecognized IdP error keeps the session, with backoff.** Only `invalid_grant` ends a session. Where main deleted sessions on any `RetrieveError`, an IdP that signals revocation through a nonstandard code now leaves sessions alive until they expire naturally. The direction is deliberate: a rotated client secret, a malformed request, or a provider bug must not sign out every user at once, and the escalating `Retry-After` bounds the cost of being wrong.
3. **The probe assumes a non-rotating family stays non-rotating.** Stated in full under the invariants above; providers with expiry-conditional rotation void it on that one path.
4. **Bounded shared fate.** A user's sessions and MCP clients share one record, so a failure that ends it affects all of them until the user signs in again. In exchange one login restores all of them, and a rotating provider can no longer have two presenters racing on one grant.
5. **Revival requires the login to return a refresh token.** `Supersede` is the only way back, and it needs the auth-code exchange to carry one. A provider configured without offline access, or one that issues no refresh token to this client, leaves a tombstone in place until its TTL expires.
6. **A pod restart mid-attempt costs a re-login on rotating families.** An attempt that dies between presenting and committing leaves an outcome nobody observed, which is resolved by retiring the record. The window is the duration of one attempt, typically well under a second, and the recovery is a sign-in.
7. **Residual clock-skew sensitivity.** Ages are clamped by the databroker's timestamp, so a replica cannot date a write into the future. One whose clock runs behind still ages records early; the settle window carries 120 seconds of margin for it.
8. **A finite rolling-upgrade window.** While both old and new code run, a replica without the write-ahead intent can present concurrently with one that has it, in either direction. Version gating is not attempted.
9. **Non-standard bodies on 503 and 500 at the token endpoint.** OAuth defines error bodies for 4xx; these carry the same JSON shape with a 5xx status. The HTTP status is the retry signal.
10. **Losers of a contended refresh get a 503 rather than making their own call.** A suppressed caller waits up to the caller budget and is then told to retry; its retry adopts whatever the winning attempt committed.
11. **After a user's last sign-out the canonical record keeps a live upstream refresh token.** `Store.Revoke` is deliberately uncalled, per the sign-out decision above, so nothing retires the record when the last session goes away; it holds a working token until the 90 day TTL collects it. Before this change the only copy lived on the session and went with it. The token is reachable only by something that can already read the databroker.
12. **The detached login write is exempt from a never-sample traceparent.** Recording the canonical session outlives the request that triggered it, so it runs on its own context and reports independently of that request's sampling decision. `internal/testenv/selftests` allows it by name, alongside the other request-independent traces.
13. **Provider caches live for one configuration generation.** The store gets a memoized authenticator getter, so a provider and the client-auth style its oauth2 config learned are reused across refreshes. Both are discarded on a configuration change, so the first refresh after a reload re-runs discovery.

## Checklist

- [x] updated unit tests
- [ ] ready for review


internal/identity/idpsession/store.go
```diff
@@ -0,0 +1,1774 @@
...
+	res, err := tx.Query(ctx, req)
+	if err != nil {
+		return record{}, err
+	}
+	records := res.GetRecords()
+	if len(records) == 0 {
+		return record{}, errRecordNotFound
+	}
+	return decodeRecord(records[0])
+}
```
alexandreLamarre, line 1678: This is hallucinated. storage operations return their own errors without failing the transaction. use `tx.Get` if getting by record ID and type

internal/identity/idpsession/store.go
```diff
@@ -0,0 +1,1774 @@
...
+func retryableFlightError(ctx context.Context, err error) bool {
+	if ctx.Err() != nil {
+		return false // the caller really is gone
+	}
+	switch status.Code(err) {
+	case codes.Canceled, codes.Aborted, codes.Unavailable:
+		return true
+	}
+	return false
+}
```
alexandreLamarre, line 1432: also `codes.DeadlineExceeded` . Can probably be a helper in databrokerutil to prevent differences in other places.

pkg/identity/manager/manager.go
```diff
@@ -230,29 +232,38 @@ func (mgr *Manager) refreshSession(ctx context.Context, sessionID string) {
...
-		return
+	if store := mgr.cfg.Load().idpStore; store != nil {
+		// Refresh through the shared upstream IdP session, so this user's browser
+		// sessions and MCP sessions collapse to one IdP call and one owner of the
+		// upstream refresh token.
+		if !mgr.refreshViaStore(ctx, sessionID, s, store) {
+			return
+		}
+	} else {
+		newToken, err := authenticator.Refresh(ctx, FromOAuthToken(s.OauthToken), NewSessionUnmarshaler(s))
```
alexandreLamarre, line 243: It's not clear to me if this PR really resolves the race with consume-on-use token refresh.

This needs to be locked with databrokerutil.Do as well as when it's called by `handler_token.go` in MCP.

alexandreLamarre: NVM I see it's handled differently now with the token response token derived from session id, version, and expires at.

internal/identity/idpsession/store.go
```diff
@@ -0,0 +1,1774 @@
...
+	rec, err := s.get(budgetCtx, id)
+	if errors.Is(err, errRecordNotFound) {
+		return nil, ErrNoUpstreamSession
+	} else if err != nil {
+		return nil, err
+	}
+
+	waiting := "no refresh attempt could be started"
+	retryAfter := DefaultRetryAfter
+	for attempt := range maxEnsureAttempts {
```
alexandreLamarre, line 720: can we use the `github.com/cenkalti/backoff/` library for retries here? It tends to be more concise and readable

internal/identity/idpsession/store.go
```diff
@@ -0,0 +1,1774 @@
...
+	defer cancel()
+
+	var out outcome
+	// Reported after the flight for the same reason as in Revoke: the callback
+	// can run more than once.
+	var retired record
+	var changed []*databroker.Record
+	var shared bool
+	var err error
+	for attempt := range maxWriteAttempts {
```
alexandreLamarre, line 1058: Consider `github.com/cenkalti/backoff/` if it makes sense.

internal/identity/idpsession/store.go
```diff
@@ -0,0 +1,1774 @@
...
+	rec, err := changedRecord(changed, id)
+	committed := err == nil
+	if errors.Is(err, errRecordNotFound) {
+		rec, err = s.get(ctx, id)
+	}
+	if errors.Is(err, errRecordNotFound) {
+		return outcome{kind: outcomeAbsent}, nil
+	} else if err != nil {
+		return outcome{}, storeError(err)
+	}
```
alexandreLamarre, line 1135: It's safe to assume that the changed records reflect storage as long as the transaction didn't return an error. `shared bool` is returned mostly to match `singleflight.Do` and use to evaluate some test correctness. It's useful to maybe add an info or debug log indicating this operation was evaluated by someone else, but at this point its safe to proceed if it err == nil.

internal/identity/idpsession/store.go
```diff
@@ -0,0 +1,1774 @@
...
+	if err != nil {
+		return nil, capture, fmt.Errorf("%w: get authenticator for %q: %w", errNotSent, rec.GetIdpId(), err)
+	}
+	if auth == nil {
+		return nil, capture, fmt.Errorf("%w: no authenticator for idp %q", errNotSent, rec.GetIdpId())
+	}
+
+	tok, err := auth.Refresh(refreshCtx, &oauth2.Token{RefreshToken: rec.GetUpstreamRefreshToken()}, capture)
+	return tok, capture, err
+}
```
alexandreLamarre, line 1346: This is exactly what should be called inside a single-flight.  It's unsafe for calls between processes /replicas.

Tokens can be consumed-on-use when refreshed. It doesn't matter if it blocks on the order of seconds

internal/identity/idpsession/store.go
```diff
@@ -0,0 +1,1774 @@
...
+	if out.kind == outcomeDead {
+		s.recordRetired(ctx, out.rec, out.rec.GetDeadReason(),
+			"idpsession: the identity provider refused this grant")
+	}
+	return out, nil
+}
+
+// commitAttempt writes the attempt's outcome into rec and reports it. rec is
+// always written back, because errors cross a flight as text: an outcome that
+// exists only as a flight error is invisible to every other process.
```
alexandreLamarre, line 1231: Comment doesn't make sense to me

internal/identity/idpsession/store.go
```diff
@@ -0,0 +1,1774 @@
...
+		out = s.commitAttempt(ctx, rec, claimed, tok, capture, refreshErr, class, at)
+		return txPut(commitCtx, tx, rec)
+	})
+	if err != nil {
+		return outcome{}, err
+	}
+	if out.kind == outcomeDead {
+		s.recordRetired(ctx, out.rec, out.rec.GetDeadReason(),
+			"idpsession: the identity provider refused this grant")
+	}
```
alexandreLamarre, line 1225: unsafe usage across process boundaries. only the transaction that was run will have this result. This looks incorrect as used in the calling functions.

internal/identity/idpsession/store.go
```diff
@@ -0,0 +1,1774 @@
...
+		out = outcome{}
+		changed, shared, err = databrokerutil.Do(flightCtx, s.client, id, func(tx databrokerutil.TX) error {
+			rec, err := txGet(flightCtx, tx, id)
+			if errors.Is(err, errRecordNotFound) {
+				out = outcome{kind: outcomeAbsent}
+				return nil
+			} else if err != nil {
+				return err
+			}
+			out = s.decide(rec, opt)
```
alexandreLamarre, line 1068: bad usage of captured outcome from singleflight across callbacks. Read the outcome from the returned changed records if the transaction succeeds, which is what the shared=true branch already seems to do.

alexandreLamarre (changes requested):
Looks like improper usage of databrokerutil.Do in several places, I think some special care should be taken to the control flow in several places

PR #6669, opened 2026-08-14
fix(portal): derive proxied TCP/UDP route names from target domain

### LINKED ISSUE #5555: Pomerium Route Portal: TCP/UDP Route names wrong, when using Pomerium on non-443 Port
## What happened?

When looking over my Route Portal, I noticed weird rout names for my tcp/udp routes. They were all the same and suffixed with a number (auth, auth(1), auth(2), auth(3) etc.)

On my other pomerium instance, the route names correctly derive from the route's domain name

## How'd it happen?

The instance in question of this issue is an instance, which _address_ is NOT _:443_.
That leads to my routes being configured like this:

```
- from: tcp+https://auth.<domain>:4443/ssh-db1-live.<domain>:22
```

I assume, that in this case, the name on the Route Portal is ALWAYS derived from _auth_, it should be, however, derived from ssh-db1-live

This also leads to indistinguishible route names when importing to Pomerium Desktop.

ystkfujii:
## Summary

Fix route name generation for proxied TCP and UDP routes.

For extended route URLs, use the target domain in the URL path instead of the Pomerium proxy hostname when generating route names. This prevents multiple routes from being displayed as `auth`, `auth (2)`, etc.

## Related issues

- Closes: https://github.com/pomerium/pomerium/issues/5555

## User Explanation

TCP and UDP routes using a Pomerium proxy on a non-standard port now get names derived from their target domain, making them easier to distinguish in the Routes Portal and Pomerium Desktop.

## AI disclosure

I used ChatGPT to investigate the issue, discuss possible implementation approaches, and help draft this PR description.
I wrote and reviewed the implementation myself.

## Checklist

- [x] updated unit tests
- [ ] add appropriate label ( `bug`)
- [x] ready for review

### DIFF
--- a/proxy/portal/portal_test.go
+++ b/proxy/portal/portal_test.go
@@ -50,6 +50,13 @@ func TestRouteFromConfigRoute(t *testing.T) {
 			ConnectCommand: "pomerium-cli udp dns.example.com:53",
 		},
 		{
+			ID:             "7b2efd6bc00173d7",
+			Name:           "production-database",
+			Type:           portal.RouteTypeTCP,
+			From:           "tcp+https://proxy.corp.example.com:8443/postgres.internal.example.com:5432",
+			ConnectCommand: "pomerium-cli tcp tcp+https://proxy.corp.example.com:8443/postgres.internal.example.com:5432",
+		},
+		{
 			ID:             "8544b096d71c5dfe",
 			Name:           "redis",
 			Type:           portal.RouteTypeTCP,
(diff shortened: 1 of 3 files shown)


proxy/portal/portal_test.go
```diff
@@ -83,7 +90,11 @@ func TestRouteFromConfigRoute(t *testing.T) {
 			To:   to2,
 		},
 		{
-			Name: "redis",
+			Name: "production-database",
+			From: "tcp+https://proxy.corp.example.com:8443/postgres.internal.example.com:5432",
+			To:   to3,
+		},
```
ystkfujii, line 96: Removed the existing `Name: "redis"` because the route name can now be generated correctly without an explicit `Name`.

Added the `production-database` case separately to verify that an explicit `Name` still takes precedence over the generated name.

ystkfujii:
@calebdoxsey 

When you have a chance, could you please review it?

PR #6688, opened 2026-08-24
authorize: only redirect navigations to the IdP, per Fetch Metadata

alynes:
## Summary

`ShouldRedirect` chooses between a 302 to the identity provider and a 401 by sniffing `Accept`. That cannot classify a bare `fetch()` (`Accept: */*`) and it redirects `text/event-stream`, so an XHR, an `EventSource` or a WebSocket handshake issued after a session expires follows the redirect, dies cross-origin at the IdP, and retries forever while the page looks healthy. We observed an expired session produce 45 socket.io handshakes in 75 seconds, all bounced.

This makes [Fetch Metadata](https://www.w3.org/TR/fetch-metadata/) authoritative when the browser sends it: only a navigation (`Sec-Fetch-Mode: navigate` — a top-level document or a frame) can complete the OAuth redirect; everything else gets the 401 client code can act on. Every modern browser sets `Sec-Fetch-Mode` on every request and page script cannot set it. Accept-sniffing stays as the fallback for clients that omit it, so non-browser behaviour is unchanged.

## Test plan

- `go test ./authorize/ -run TestShouldRedirect` — new `sec-fetch-mode` subtest: navigate/document and navigate/iframe → redirect; cors + `*/*` (bare fetch), cors + `text/event-stream` (EventSource), websocket → 401; no Fetch Metadata → Accept fallback unchanged.
- Existing `TestShouldRedirect` and `TestAuthorize_*` unaffected.


alynes:
> @alynes Please file an issue for this, and please make sure the description is as concise and easily understandable as possible.

Will do, thanks!

alynes:
Closing PR, opening issue for this first.

Issue #6692, opened 2026-08-25
MCP CIMD clients using `private_key_jwt` fail at `/authorize` with "unsupported token endpoint auth method"


nickytonline:
## What happened?

Connecting a ChatGPT plugin that is a Pomerium-fronted remote MCP server fails with:

```
400 Bad Request
unsupported token endpoint auth method: private_key_jwt
```

ChatGPT uses a URL `client_id` (client ID metadata document / CIMD). Its document declares:

```json
{
  "token_endpoint_auth_method": "private_key_jwt",
  "token_endpoint_auth_signing_alg": "RS256",
  "jwks_uri": "https://chatgpt.com/oauth/jwks.json"
}
```

Pomerium's CIMD validation only *blocklists* secret-based methods, so `private_key_jwt` passes the fetch step, is stored on the client registration, and then fails at `/authorize` on a hardcoded 3-value switch. The failure happens before ChatGPT can present its signed `client_assertion`.

This is spec-conformant client behavior. [`draft-ietf-oauth-client-id-metadata-document`](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document):

- [§4.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document#section-4.1): `token_endpoint_auth_method` "MUST NOT include `client_secret_post`, `client_secret_basic`, `client_secret_jwt`, or any other method based around a shared symmetric secret" — this is the exact clause Pomerium's blocklist implements.
- [§8.2](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document#section-8.2): clients capable of holding key material "SHOULD" authenticate; the draft's own example is `"token_endpoint_auth_method": "private_key_jwt"` with `jwks_uri`.
- [§8.2.6](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document#section-8.2.6): "When a client declares `token_endpoint_auth_method` as `private_key_jwt`, the authorization server MUST require client authentication according to [[RFC7523]](https://www.rfc-editor.org/rfc/rfc7523)."

[...]

PR #6700, opened 2026-08-28
fix(mcp/CIMD): negotiate supported token auth endpoint methods based on Oauth2 server metadata

alexandreLamarre:
## Summary

Our CIMD flow fetches client side metadata and was using the default advertised `token_endpoint_auth_method` metadata.

As per RFC 8414, Oauth 2 may also include an optional list of `token_endpoint_auth_methods_supported`, which include a list of supported methods.

ChatGPT exposes `private_key_jwt` as the default token endpoint auth method, which Pomerium currently does not support. It also allows for the `none` method which Pomerium does support (It's actually the only method we currently support for this downstream CIMD flow).

This PR adds a negotiation layer for supported auth methods, to future proof this downstream client flow, and returns a bad request error with relevant information about the mismatch between auth methods, so that the failure is clear.

I plan to look into the `private_key_jwt` mechanism in more detail and see if we can implement it in Pomerium.

## Related issues

Addresses: https://github.com/pomerium/pomerium/issues/6692

## User Explanation

MCP Clients could break if they advertised default token auth methods like `private_key_jwt` that Pomerium doesn't support in its flow.

## AI disclosure

I used claude Opus 5 to draft the log messages to fit the existing log messages in `internal/mcp`. The fix, approach to the problem and code is my own.

## Checklist

- [X] updated unit tests
- [X] ready for review

### DIFF
--- a/internal/mcp/client_id_metadata.go
+++ b/internal/mcp/client_id_metadata.go
@@ -43,6 +43,12 @@ type ClientIDMetadataDocument struct {
 	// Per draft, MUST NOT be client_secret_basic, client_secret_post, client_secret_jwt.
 	TokenEndpointAuthMethod string `json:"token_endpoint_auth_method,omitempty"`
 
+	// RFC8414 Oauth 2.0 Server Metadata
+	// OPTIONAL. JSON array containing a list of client authentication
+	// methods supported by this token endpoint.  Client authentication
+	// method values are used in the "token_endpoint_auth_method". See above.
+	TokendEndpointAuthMethodsSupported []string `json:"token_endpoint_auth_methods_supported,omitempty"`
+
 	// Scope is OPTIONAL.
 	Scope string `json:"scope,omitempty"`
 
--- a/internal/mcp/handler_authorization.go
+++ b/internal/mcp/handler_authorization.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"slices"
 	"time"
 
 	"buf.build/go/protovalidate"
@@ -19,6 +20,10 @@ import (
 	oauth21proto "github.com/pomerium/pomerium/internal/oauth21/gen"
 	rfc7591v1 "github.com/pomerium/pomerium/internal/rfc7591"
 )
+
+var supportedTokenAuthMethodsForCIMD = []string{
+	rfc7591v1.TokenEndpointAuthMethodNone,
+}
 
 // Authorize handles the /authorize endpoint.
 func (srv *Handler) Authorize(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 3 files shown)


internal/mcp/client_id_metadata.go
```diff
@@ -154,6 +160,8 @@ var ErrClientMetadataFetch = errors.New("failed to fetch client metadata")
 // ErrClientMetadataValidation represents a validation error for client metadata.
 var ErrClientMetadataValidation = errors.New("client metadata validation failed")
 
+var ErrClientMetadataTokenAuthMethodNotSupported = errors.New("no mutually supported token endpoint auth method")
```
kenjenkins: (totally optional)

What do you think of reusing ErrClientMetadataValidation for this case? That is: constructing the error something like `fmt.Errorf("%w: no mutually supported token endpoint auth method", ErrClientMetadataValidation)`.

That way we wouldn't need to tag another `|| errors.Is()` check onto the conditional in handler_authorization.go.

internal/mcp/handler_authorization.go
```diff
@@ -354,8 +355,66 @@ func (srv *Handler) getOrFetchClient(ctx context.Context, clientID string) (*rfc
 		if err != nil {
 			return nil, err
 		}
+
+		if err := srv.negotiateTokenEndpointAuthMethod(ctx, doc); err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
alexandreLamarre: @greptileai

internal/mcp/handler_authorization.go
```diff
@@ -354,8 +355,73 @@ func (srv *Handler) getOrFetchClient(ctx context.Context, clientID string) (*rfc
...
 	}
 
 	return srv.storage.GetClient(ctx, clientID)
 }
+
+func (srv *Handler) negotiateTokenEndpointAuthMethod(ctx context.Context, doc *ClientIDMetadataDocument) error {
+	if doc.TokenEndpointAuthMethod == "" {
+		log.Ctx(ctx).Info().
+			Str("client-id", doc.ClientID).
+			Msg("mcp/authorize: client did not specify a token endpoint auth method, defaulting to none")
```
kenjenkins, line 377: I think technically the default if omitted is `client_secret_basic`? From https://openid.net/specs/openid-connect-registration-1_0.html, section 2:

> token_endpoint_auth_method
OPTIONAL. Requested Client Authentication method for the Token Endpoint. The options are client_secret_post, client_secret_basic, client_secret_jwt, private_key_jwt, and none, as described in Section 9 of [OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-registration-1_0.html#OpenID.Core) [OpenID.Core]. Other authentication methods MAY be defined by extensions. If omitted, the default is client_secret_basic -- the HTTP Basic Authentication Scheme specified in Section 2.3.1 of [OAuth 2.0](https://openid.net/specs/openid-connect-registration-1_0.html#RFC6749) [RFC6749].

But it sounds like maybe the other existing logic also applies `none` as the default?

internal/mcp/handler_authorization.go
```diff
@@ -354,8 +355,73 @@ func (srv *Handler) getOrFetchClient(ctx context.Context, clientID string) (*rfc
...
+	return nil
+}
+
+// SupportedCIMDTokenEndpointAuthMethods returns the valid token endpoint auth methods
+// the MCP handler supports for downstream CIMD.
+// The returned slice is ordered by which methods are preferred.
+func (srv *Handler) supportedTokenAuthMethodsForCIMD() []string {
+	return []string{
+		rfc7591v1.TokenEndpointAuthMethodNone,
+	}
```
kenjenkins: nit: should we extract this to a file-level `var` to avoid allocating a slice each time?

internal/mcp/handler_authorization.go
alexandreLamarre: Good point.

I thought I read somewhere in the CIMD RFC that "none" should be preferred, although I can't find it.

The issue with defaulting to the connect registration default is that it might break servers not settings this field, I don't know if that's an issue we've run into before. Let me look some more into this

alexandreLamarre: Ah yeah I conflated the precendence in the DCR RFC 7591, with what we should prefer here:
```
token_endpoint_auth_method
      String indicator of the requested authentication method for the
      token endpoint.  Values defined by this specification are:

      *  "none": The client is a public client as defined in OAuth 2.0,
         [Section 2.1](https://datatracker.ietf.org/doc/html/rfc7591#section-2.1), and does not have a client secret.

      *  "client_secret_post": The client uses the HTTP POST parameters
         as defined in OAuth 2.0, [Section 2.3.1](https://datatracker.ietf.org/doc/html/rfc7591#section-2.3.1).

      *  "client_secret_basic": The client uses HTTP Basic as defined in
         OAuth 2.0, [Section 2.3.1](https://datatracker.ietf.org/doc/html/rfc7591#section-2.3.1).
```
Let me look into changing the default. Good catch!

kenjenkins (approved):

PR #6701, opened 2026-08-31
feat(mcp): CIMD private_key_jwt token auth method

alexandreLamarre:
- **proto: token request changes**
- **feat(mcp): support CIMD private_key_jwt method**

## Related issues

Follow to #6692

## Checklist

- [ ] reference any related issues
- [ ] updated unit tests
- [ ] add appropriate label (`enhancement`, `bug`, `breaking`, `dependencies`, `ci`)
- [ ] disclosed AI usage (or wrote "none") per AI_POLICY.md
- [ ] ready for review

PR #6711, opened 2026-09-01
fix(mcp): harden private_key_jwt client authentication

wasaga:
## Summary

Adversarial review of #6701 against the MCP draft authorization spec, the OAuth Client ID Metadata Document draft, RFC 7521/7523 and RFC 8414. Six fixes, each with a regression test that fails without it.

**Client assertions were durable credentials.** `exp` had no upper bound and there is no `jti` replay store, so an assertion minted with a far-future expiry stayed valid indefinitely — weaker than the `client_secret` this method replaces. Assertions must now expire within 5 minutes (plus the existing 1m skew). Replay inside that window is still possible; a `jti` store would be the complete fix.

**The client-controlled `jwks_uri` was fetched on every token request.** A fresh `RemoteKeySet` was built per verification, and go-oidc caches per instance. The fetch also happened before claims validation, so an unauthenticated request carrying any syntactically valid JWS paid for it. Key sets are now reused from a bounded LRU.

**The expected assertion audience came from the `Host` header.** `/.pomerium/*` is served from Envoy's catch-all virtual host, so the header is not guaranteed to name a configured route — a caller could name any audience it liked and replay an assertion minted for a different authorization server. The hostname is now validated against configured routes first, matching what `handler_cimd.go` and `handler_authorization.go` already do.

**`private_key_jwt` was advertised for DCR but rejected at `/register`.** The `rfc7591` metadata enum permits only `none`, `client_secret_post` and `client_secret_basic`, and the JWKS fetcher is scoped to the CIMD domain allowlist. It stays advertised on the CIMD path, where it works.

**A public client publishing a `jwks_uri` was rejected outright.** Validation required `jwks_uri` to imply `private_key_jwt`. Nothing in CIMD or RFC 7591 couples them — a `jwks_uri` is equally the key source for signed request objects. The check is now one-directional.

**Negotiation could pick an unusable method.** `private_key_jwt` is first by preference, so a client advertising both it and `none` without a `jwks_uri` was rejected entirely instead of falling back. Negotiation now skips candidates the document cannot satisfy; a method the client explicitly asked for is still validated strictly, so a misconfiguration stays an error rather than becoming a silent downgrade.

## Related issues

Follow-up to #6701.

## User Explanation

Clients authenticating to the MCP authorization server with `private_key_jwt` must now present assertions expiring within 5 minutes. Client key sets are cached rather than re-fetched per request. `private_key_jwt` is no longer advertised for dynamically registered clients, which never supported it.

## AI disclosure

Claude Code performed the spec review and wrote the fixes and tests; the maximum assertion lifetime and the DCR resolution were chosen by me.

## Checklist

- [x] updated unit tests
- [ ] ready for review


internal/mcp/handler_token.go
```diff
@@ -798,6 +809,9 @@ func validateClientAssertionClaims(claims jwtutil.Claims, clientID string, expec
 	if expiry.Before(now.Add(-skew)) {
 		return fmt.Errorf("expired")
 	}
+	if expiry.After(now.Add(maxClientAssertionLifetime + skew)) {
+		return fmt.Errorf("expires more than %v in the future", maxClientAssertionLifetime)
```
wasaga, line 813: **The main security fix.** `exp` previously had no upper bound, and there is no `jti` replay store, so an assertion minted with a far-future expiry stayed valid forever. That makes `private_key_jwt` *weaker* than the `client_secret` it replaces: one captured assertion is a permanent credential.

RFC 7523 §3 item 4 exists precisely to limit the window ("an `exp` claim that limits the time window during which the JWT can be used") and explicitly permits rejecting an `exp` "unreasonably far in the future". OIDC Core §9 — the spec that actually defines `private_key_jwt`, which we now advertise — goes further: `jti` is REQUIRED and "these tokens MUST only be used once".

5 minutes is the industry norm. Note this does **not** close replay *within* the window; that needs a `jti` store, which is a follow-up. Bounding the lifetime is the stateless half, and it's the half that turns a permanent credential back into a short-lived proof.

This changes behavior for clients: the existing tests minted 1-hour assertions and were updated.

internal/mcp/handler_token.go
```diff
@@ -342,11 +342,17 @@ func (srv *Handler) getTokenRequest(
...
 }
 
-// issuer or token endpoint
-func tokenEndpointAudiences(r *http.Request, prefix string) []string {
+// tokenEndpointAudiences returns the audience values an assertion may name: the
+// issuer or the token endpoint. The /.pomerium/ prefix is served from Envoy's
+// catch-all virtual host, so the Host header is validated against the configured
+// routes before it is trusted to name an audience.
+func (srv *Handler) tokenEndpointAudiences(r *http.Request) []string {
+	if _, ok := srv.hosts.GetServerHostInfo(stripPort(r.Host)); !ok {
```
wasaga, line 350: The expected audience was previously derived from `r.Host` with no validation. That matters more than it looks: `/.pomerium/*` is registered on Envoy's **catch-all** virtual host (`buildPomeriumHTTPRoutes` adds the `/.pomerium/` prefix route to every vhost, including `"catch-all", "*"`), and only the `.well-known` metadata routes are `isMCPHost`-gated. So the Host header reaching this handler is not guaranteed to name a configured route.

Consequence: `Host: some-other-as.example` made us accept an assertion minted for a different authorization server — exactly what RFC 7523 §3 item 3 forbids ("MUST reject any JWT that does not contain its own identity as the intended audience").

Only the hostname is validated; the URL is still built from the untouched `r.Host` so the port is preserved and the value stays byte-identical to the issuer advertised by `getAuthorizationServerMetadata`. An unknown host returns nil, and `validateClientAssertionClaims` already fails closed against an empty audience set.

**Known gap, deliberately not fixed here:** this is now the fifth inline copy of `stripPort(r.Host)` + `GetServerHostInfo` in this package (see `handler_authorization.go`, `handler_connect.go`, `upstream_auth.go`, `handler_cimd.go`). The deeper fix is middleware on the MCP mux plus a `HostInfo.IsKnownHost` helper — but the metadata handlers are free functions registered on the controlplane router with no access to `HostInfo`, so unifying them needs wiring changes in another subsystem. Worth its own issue. Related: `getAuthorizationServerMetadata` still advertises an issuer built from unvalidated `r.Host`, so we currently advertise one issuer and validate another.

internal/mcp/client_id_metadata.go
```diff
@@ -370,9 +398,18 @@ func (j *JWKSFetcher) KeySet(ctx context.Context, jwksURI string) (*go_oidc.Remo
...
+		return ks, nil
+	}
+
+	// A RemoteKeySet caches keys and coalesces in-flight fetches internally, but
+	// only per instance, so it has to outlive the request that created it. It
+	// retains the context it is built from, so that must not be the request's:
+	// that would pin request-scoped values for as long as the entry is cached.
 	ks := go_oidc.NewRemoteKeySet(
-		go_oidc.ClientContext(ctx, j.httpClient),
+		go_oidc.ClientContext(context.Background(), j.httpClient),
```
wasaga, line 412: Two separate problems fixed here.

**Per-request fetching.** `NewRemoteKeySet` was called on every verification, and go-oidc caches keys *per instance* — its own doc says "Reuse a single remote key set rather than creating a new one for each verification." Worse, the fetch happened *before* claims validation, so an unauthenticated `POST /token` carrying any syntactically valid JWS triggered an outbound fetch of up to 1 MB to a client-named host. That is reflected amplification plus self-DoS. Also a direct SHOULD violation of the MCP spec's "SHOULD cache metadata respecting HTTP cache headers".

**Context capture.** The obvious fix — caching a key set built from the request context — quietly makes it worse: `RemoteKeySet` retains the context it is constructed from, so each cache entry would pin that request's `CachingQuerier`/databroker-client snapshot and per-request logger fields for the entry's lifetime, keeping a pre-reload config epoch alive. `context.WithoutCancel` does *not* help — it drops cancellation, not values — and go-oidc already applies it internally. Hence `context.Background()`, with `ClientContext` carrying the only value that is actually needed.

No new dependency: `expirable` is a subpackage of the `hashicorp/golang-lru/v2` module already in go.mod.

internal/mcp/client_id_metadata.go
```diff
@@ -340,19 +356,31 @@ const MaxJWKSSize = 1024 * 1024 // 1MB
...
 var ErrJWKSFetch = errors.New("failed to fetch client JWKS")
 
+// JWKSKeySetCacheSize bounds how many distinct jwks_uri key sets are retained.
+// jwks_uri comes from client-controlled metadata, so this must stay bounded.
+const JWKSKeySetCacheSize = 256
+
+// JWKSKeySetCacheTTL bounds how long a withdrawn jwks_uri keeps serving from
+// cache. Rotation to a new key needs no TTL: a key set refetches on an unknown
+// kid.
+const JWKSKeySetCacheTTL = 30 * time.Minute
```
wasaga, line 366: The TTL is about *withdrawal*, not rotation. Rotation to a new key needs no TTL — `RemoteKeySet` refetches whenever it sees an unknown `kid`. But a plain LRU would keep serving a **revoked** key indefinitely if fewer than 256 distinct URIs are ever seen, since nothing would evict it.

The bound also has to be a bound: `jwks_uri` is client-controlled, and each entry can hold up to `MaxJWKSSize` (1 MB), so 256 entries is a ~256 MB worst case that a plain LRU would never release. 30 minutes caps both staleness and idle retention while still removing essentially all of the per-request fetching.

internal/mcp/client_id_metadata.go
```diff
@@ -211,6 +213,23 @@ func (doc *ClientIDMetadataDocument) Validate() error {
 	return nil
 }
 
+// canAuthenticateWith reports whether the document carries the material a given
+// token endpoint auth method needs. It is consulted only when negotiating among
+// the alternatives a client advertises; a method the client explicitly asked for
+// is still validated strictly, so a misconfiguration stays an error rather than
+// becoming a silent downgrade. Methods added to supportedTokenAuthMethodsForCIMD
+// need a case here, hence the closed switch.
+func (doc *ClientIDMetadataDocument) canAuthenticateWith(method string) bool {
```
wasaga, line 222: Negotiation orders `private_key_jwt` ahead of `none`, and it ran *before* `Validate()`. So a client advertising both, without a `jwks_uri`, had `private_key_jwt` selected and was then rejected outright — even though `none` was mutually supported and would have worked.

The predicate is consulted **only** when picking among advertised alternatives. A method the client explicitly asked for still goes through strict validation, so a client that asks for `private_key_jwt` with no `jwks_uri` still gets a clear error rather than a silent downgrade to `none`. That asymmetry is the point of the split.

Written as a closed switch rather than `if method == private_key_jwt { ... }; return true` on purpose: with a default of `true`, adding a method to `supportedTokenAuthMethodsForCIMD` and forgetting this function would fail *open* — we'd offer a method whose required material the document does not carry. Failing closed turns that into a visible "method never negotiated" bug instead.

internal/mcp/client_id_metadata.go
```diff
@@ -220,9 +239,6 @@ func (doc *ClientIDMetadataDocument) validateKeySet() error {
 		if u.Scheme != "https" {
 			return fmt.Errorf("jwks_uri must use the https scheme")
 		}
-		if doc.TokenEndpointAuthMethod != rfc7591v1.TokenEndpointAuthMethodPrivateKeyJWT {
-			return fmt.Errorf("jwks_uri must be configured with \"private_key_jwt\", not : %q", doc.TokenEndpointAuthMethod)
-		}
 	}
 
 	if doc.TokenEndpointAuthMethod == rfc7591v1.TokenEndpointAuthMethodPrivateKeyJWT && doc.JWKSURI == "" {
```
wasaga, line 244: The removed check made the relationship biconditional: any `jwks_uri` forced `token_endpoint_auth_method == private_key_jwt`. Nothing in CIMD or RFC 7591 couples them. RFC 7591 §2 states the only coupling there is ("MUST NOT be present if `jwks` is present"), and a `jwks_uri` is equally the key source for signed request objects (RFC 9101) or other flows.

Concretely, a valid `{"token_endpoint_auth_method": "none", "jwks_uri": ...}` document that worked before this feature landed was being rejected at `/authorize`. The implication we do need — `private_key_jwt` requires a `jwks_uri` — is kept just below.

**Related gap, not fixed here:** `jwks` (the embedded key set) still isn't supported, so `{"token_endpoint_auth_method": "private_key_jwt", "jwks": {...}}` fails as "requires jwks_uri". The proto already models it (`rfc7591_types.proto`, with the mutual-exclusion rule), and it's arguably the form worth supporting first since it needs no second fetch and has no SSRF surface.

internal/mcp/handler_metadata.go
```diff
@@ -174,7 +174,7 @@ func getAuthorizationServerMetadata(r *http.Request, prefix string, dcrEnabled b
 	}
 	if dcrEnabled {
 		md.RegistrationEndpoint = P(path.Join(prefix, registerEndpoint))
-		md.TokenEndpointAuthMethodsSupported = []string{"private_key_jwt", "client_secret_basic", "none"}
+		md.TokenEndpointAuthMethodsSupported = []string{"client_secret_basic", "none"}
```
wasaga: `private_key_jwt` was advertised for DCR but cannot work there, on two independent counts:

1. `rfc7591_types.proto` constrains `token_endpoint_auth_method` to `in: ["none", "client_secret_post", "client_secret_basic"]`, so `/register` rejects it via protovalidate. A conforming client that reads our metadata, picks the first advertised method, and registers gets `invalid_client_metadata` and dead-ends.
2. Even past registration, `jwksFetcher` is built with the **CIMD** allowlist (`mcp_allowed_client_id_domains`), so a registered client's `jwks_uri` host would not be in it.

It stays advertised on the CIMD path above, where it genuinely works. Actually supporting it for DCR means extending the proto enum, skipping secret generation, and deciding a second domain policy — a feature, not a fix, and better as its own PR.

**Note for whoever picks that up:** this literal and the protovalidate enum have to be kept in sync by hand and nothing enforces it — they already differ, since `client_secret_post` is registration-legal but not advertised (pre-existing). A single source of truth in `internal/rfc7591` referenced from both would stop the drift.

wasaga (commented):
Rationale for each change, inline. Findings came from an adversarial pass over the MCP draft authorization spec, draft-ietf-oauth-client-id-metadata-document-00, RFC 7521/7523 and RFC 8414; every fix has a test that fails without it.

PR #6736, opened 2026-09-08
ssh: implement two-person auth

kralicky:
## Summary

This implements a mechanism for two-person approval for SSH routes. Enabling this feature for a route is done using new PPL criteria, as follows:

```yaml
# as an allow rule
routes:
- from: ssh://foo
  to: ssh://bar
  policy:
  - allow:
      and:
      - email:
          is: "employee@company.org"
      - ssh_access_request_approved: {}

# alternatively, as a deny rule:
routes:
- from: ssh://foo
  to: ssh://bar
  policy:
  - allow:
      and:
      - email:
          is: "employee@company.org"
  - deny:
      or:
      - ssh_access_request_not_approved: {}

```

When one of these criteria are present, and the access request approval is the only remaining reason the user would be denied after going through the regular auth flow, a new `StreamAccessRequest` databroker record is created for the stream, containing request metadata. Approving or denying the request can be done by editing the `state` field of the record out-of-band (not implemented here). If the request is approved, then the user is allowed access, otherwise they are denied. Each new request to such a route will require a new access request.

There is a default request timeout of 5 minutes, which can be adjusted using the `ssh_two_person_approval_request_timeout` config option. 

## AI disclosure

None

## Checklist

- [ ] updated unit tests
- [ ] ready for review

### DIFF
--- a/pkg/grpc/session/session.proto
+++ b/pkg/grpc/session/session.proto
@@ -32,9 +32,9 @@ message Session {
     }
   }
 
-  string version = 1;
-  string id      = 2;
-  string user_id = 3;
+  string                                 version            = 1;
+  string                                 id                 = 2;
+  string                                 user_id            = 3;
   repeated DeviceCredential              device_credentials = 17;
   google.protobuf.Timestamp              issued_at          = 14;
   google.protobuf.Timestamp              expires_at         = 4;
@@ -89,6 +89,29 @@ message IdentityBinding {
   string idp_id   = 3;
   // Protocol level details associated with session creation.
   map<string, string> details = 4;
+}
+
+// A StreamAccessRequest represents a pending access request for an SSH client
+// to access a protected route.
+message StreamAccessRequest {
+  enum State {
+    Pending  = 0;
+    Approved = 1;
+    Denied   = 2;
+  }
+  StreamAccessRequestParams params     = 1;
+  google.protobuf.Timestamp created_at = 2;
+  google.protobuf.Timestamp expires_at = 3;
+  State                     state      = 4;
+  map<string, string>       metadata   = 5;
+}
+
+message StreamAccessRequestParams {
+  string protocol   = 1;
+  string session_id = 2;
+  string user_id    = 3;
+  uint64 stream_id  = 4;
+  string cluster_id = 5;
 }
 
 // A Handle is a reference to a session or a service account.
--- a/pkg/policy/criteria/ssh_access_request.go
+++ b/pkg/policy/criteria/ssh_access_request.go
@@ -1,0 +1,70 @@
+package criteria
+
+import (
+	"github.com/open-policy-agent/opa/ast"
+
+	"github.com/pomerium/pomerium/pkg/policy/generator"
+	"github.com/pomerium/pomerium/pkg/policy/parser"
+)
+
+var accessRequestApprovedBody = ast.Body{
+	ast.MustParseExpr(`is_boolean(input.ssh.access_request_approved)`),
+	ast.MustParseExpr(`input.ssh.access_request_approved`),
+}
+
+type sshAccessRequestNotApprovedCriterion struct {
+	g *Generator
+}
+
+func (sshAccessRequestNotApprovedCriterion) DataType() generator.CriterionDataType {
+	return generator.CriterionDataTypeUnused
+}
+
+func (sshAccessRequestNotApprovedCriterion) Name() string {
+	return "ssh_access_request_not_approved"
+}
+
+func (c sshAccessRequestNotApprovedCriterion) GenerateRule(_ string, _ parser.Value) (*ast.Rule, []*ast.Rule, error) {
+	r1 := c.g.NewRule(c.Name())
+	r1.Head.Value = NewCriterionTerm(false, ReasonSSHAccessRequestOK)
+	r1.Body = accessRequestApprovedBody
+	r1.Else = &ast.Rule{
+		Head: generator.NewHead("", NewCriterionTerm(true, ReasonSSHAccessRequestRequired)),
+	}
+	return r1, nil, nil
+}
+
+func SSHAccessRequestNotApprovedCriterion(generator *Generator) Criterion {
+	return sshAccessRequestNotApprovedCriterion{g: generator}
+}
+
+type sshAccessRequestApprovedCriterion struct {
+	g *Generator
+}
+
+func (sshAccessRequestApprovedCriterion) DataType() generator.CriterionDataType {
+	return generator.CriterionDataTypeUnused
+}
+
+func (sshAccessRequestApprovedCriterion) Name() string {
+	return "ssh_access_request_approved"
+}
+
+func (c sshAccessRequestApprovedCriterion) GenerateRule(_ string, _ parser.Value) (*ast.Rule, []*ast.Rule, error) {
+	r1 := c.g.NewRule(c.Name())
+	r1.Head.Value = NewCriterionTerm(true, ReasonSSHAccessRequestOK)
+	r1.Body = accessRequestApprovedBody
+	r1.Else = &ast.Rule{
+		Head: generator.NewHead("", NewCriterionTerm(false, ReasonSSHAccessRequestRequired)),
+	}
+	return r1, nil, nil
+}
+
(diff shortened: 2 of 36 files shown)


pkg/policy/criteria/ssh_access_request.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/pomerium/pomerium/pkg/policy/generator"
+	"github.com/pomerium/pomerium/pkg/policy/parser"
+)
+
+var accessRequestApprovedBody = ast.Body{
+	ast.MustParseExpr(`is_boolean(input.ssh.access_request_approved)`),
+	ast.MustParseExpr(`input.ssh.access_request_approved`),
+}
+
+type sshAccessRequestNotApprovedCriterion struct {
```
kenjenkins: I'm not sure I understand why there should be both an `ssh_access_request_approved` and `ssh_access_request_not_approved`.

Is `ssh_access_request_not_approved` equivalent to using `ssh_access_request_approved` with a `not` operation?

kralicky: Hmm, I suppose that is correct. I didn't realize `not` worked that way. All the tests still pass when rewriting them (except for a couple that tested certain contradictions that can't be formed anymore)

pkg/grpc/session/session.proto
```diff
@@ -91,6 +91,29 @@ message IdentityBinding {
   map<string, string> details = 4;
 }
 
+// A StreamAccessRequest represents a pending access request for an SSH client
+// to access a protected route.
+message StreamAccessRequest {
+  enum State {
+    Pending  = 0;
```
kenjenkins, line 98: It looks like we are mostly using UPPER_CASE for proto enum values.

pkg/grpc/session/session.proto
```diff
@@ -91,6 +91,29 @@ message IdentityBinding {
   map<string, string> details = 4;
 }
 
+// A StreamAccessRequest represents a pending access request for an SSH client
+// to access a protected route.
+message StreamAccessRequest {
+  enum State {
```
kenjenkins, line 97: Should we include an UNKNOWN = 0 element, in case we ever want to expand the set of possible states?

pkg/policy/criteria/ssh_access_request.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (c sshAccessRequestApprovedCriterion) GenerateRule(_ string, _ parser.Value) (*ast.Rule, []*ast.Rule, error) {
+	r1 := c.g.NewRule(c.Name())
+	r1.Head.Value = NewCriterionTerm(true, ReasonSSHAccessRequestOK)
+	r1.Body = accessRequestApprovedBody
+	r1.Else = &ast.Rule{
+		Head: generator.NewHead("", NewCriterionTerm(false, ReasonSSHAccessRequestRequired)),
+	}
+	return r1, nil, nil
```
kenjenkins, line 34: I think this can use the NewCriterionRule() helper method? Is this equivalent?

```suggestion
	return NewCriterionRule(c.g, c.Name(), ReasonSSHAccessRequestOK, ReasonSSHAccessRequestRequired, accessRequestApprovedBody), nil, nil
```

pkg/ssh/auth.go
```diff
@@ -404,25 +425,108 @@ func processSessionEvaluateResult(
...
 }
 
+func twoPersonAuthRequired(res *evaluator.Result) bool {
+	// Note that the same "ssh-access-request-required" reason is checked in both
+	// allow and deny reasons (contrary to the way ssh_publickey and source_ip
+	// reasons are handled). This is because you can use the
+	// ssh_access_request_approved criteria in an allow block using and/or, or in
+	// a deny block using not/nor. The failure reason is always
+	// "ssh-access-request-required" regardless because not/nor will swap the
+	// success and failure reasons for criteria used in those blocks.
```
kenjenkins, line 466: I don't think this comment is accurate? (Arguably `not` and `nor` _should_ swap success and failure reasons, but I don't think they currently do.)

pkg/ssh/auth.go
```diff
@@ -404,25 +425,108 @@ func processSessionEvaluateResult(
...
+
+	if res.Allow.Value == res.Deny.Value {
```
kenjenkins, line 468: I'm finding this logic hard to reason about. Would it be any clearer if we extracted some of the repeated expressions to local variables, maybe like this?

```go
	onlyAllowReason := len(res.Allow.Reasons) == 1 &&
		res.Allow.Reasons.Has(criteria.ReasonSSHAccessRequestRequired)
	onlyDenyReason := len(res.Deny.Reasons) == 1 &&
		res.Deny.Reasons.Has(criteria.ReasonSSHAccessRequestRequired)
```

pkg/ssh/auth.go
```diff
@@ -404,25 +425,108 @@ func processSessionEvaluateResult(
...
+		// are any other unrelated failure reasons then the request is denied.
+		if res.Allow.Value && len(res.Deny.Reasons) == 1 &&
+			res.Deny.Reasons.Has(criteria.ReasonSSHAccessRequestRequired) {
+			return true
+		} else if !res.Deny.Value && len(res.Allow.Reasons) == 1 &&
+			res.Allow.Reasons.Has(criteria.ReasonSSHAccessRequestRequired) {
+			return true
+		}
+	} else if !res.Allow.Value {
+		// This handles the unusual case of redundant criteria, as in:
```
kenjenkins, line 481: Should we expand on the unit tests in auth_test.go to exercise all of these cases?

pkg/grpc/session/session.proto
kralicky: Yeah that's probably better, it would match with the other api in envoy as well.

pkg/ssh/auth.go
kralicky: These are covered in the integration tests, but I'll add extra unit tests for them as well.

pkg/grpc/session/session.proto
```diff
@@ -95,9 +95,10 @@ message IdentityBinding {
 // to access a protected route.
 message StreamAccessRequest {
   enum State {
-    Pending  = 0;
-    Approved = 1;
-    Denied   = 2;
+    NONE     = 0;
```
kenjenkins: I think UNKNOWN is conventional, because the proto semantics are that any value outside the known values would read as the 0 value. Does that match your understanding?

pkg/ssh/auth.go
```diff
@@ -458,26 +490,39 @@ func processSessionEvaluateResult(
...
+
+	allowed := res.Allow.Value
+	denied := res.Deny.Value
+	isRemainingAllowReason := len(res.Allow.Reasons) == 1 &&
+		res.Allow.Reasons.Has(criteria.ReasonSSHAccessRequestRequired)
+	isRemainingDenyReason := len(res.Deny.Reasons) == 1 &&
+		res.Deny.Reasons.Has(criteria.ReasonSSHAccessRequestRequired)
+
+	// there are 9 possibilities here:
+	// 1. allowed && !denied (false)
```
kenjenkins: This method is never called in this case, is it?

kenjenkins (commented):
Would you also mind updating the PR description now that there is no longer a separate `ssh_access_request_not_approved` criterion?

pkg/grpc/session/session.proto
kralicky: I named it None to match https://github.com/pomerium/envoy-custom/blob/36777435573c1d58c819b32770d9a21c5115e930/api/extensions/filters/network/ssh/ssh.proto#L418, which is named that way because if there wasn't an access request for a session, its state _is_ known.

According to the docs, Go will always decode out-of-range enum values as the actual value, though C++ will not (but only when the enum is defined in a proto2 file)

pkg/ssh/auth.go
kralicky: currently no, but if it was it would return the correct value.

kenjenkins (approved):

pkg/grpc/session/session.proto
kenjenkins: OK thanks, TIL! Looks like this is the documentation? https://protobuf.dev/programming-guides/enum/#go.

Issue #6730, opened 2026-09-04
MCP route returns 502 permanently after an outbound-connection reload (ext_proc: error getting upstream token)


ptota-coder:
## What happened?

We put an MCP server behind Pomerium (route annotated `mcp_server: "true"`, no upstream OAuth2 configured). It works fine until Pomerium logs `outbound client connection has changed meaningfully, reloading`. After that line, every request to that route returns 502 and it never recovers on its own — the only fix is restarting the Pomerium process.

The 502s come from ext_proc failing to resolve an upstream token:

```
ext_proc: error getting upstream token
  error="failed to get upstream MCP token: rpc error: code = Canceled desc = grpc: the client connection is closing"
```

So it looks like the MCP ext_proc handler holds onto the old outbound gRPC client after the reload recreates it. Every `GetUpstreamOAuth2Token` call after that hits the closed connection, gets `Canceled`, and ext_proc fails closed with a 502. Plain proxy and authorize routes keep working through the same reload, so whatever they do to pick up the new connection, the MCP path isn't doing.

Two things stand out. First, this route has no `mcp_server_upstream_oauth2_*` config, so we don't understand why it does an upstream-token lookup at all — there's no upstream OAuth to broker. Second, it looks like a sibling of [#6101](<https://github.com/pomerium/pomerium/issues/6101>) (the deadlock in `CachedOutboundGRPClientConn.Get()` on config change, fixed in v0.32.2, which is in our build): same connection object, opposite symptom — a stale handle instead of a deadlock.

The trigger in our case is mundane and frequent: redeploying the upstream backend. Its Service Endpoints change, that fires the reload, and the route is dead until we bounce Pomerium. So any normal backend rollout, node drain, or scale event breaks it.

## What did you expect to happen?

After the outbound connection is recreated on a config change, the MCP handler should pick up the new connection the way authorize and proxy evidently do (those routes were unaffected through the same reload). Alternatively, for a route with no upstream OAuth2 configured and failed/absent discovery, ext_proc shouldn't need to consult the databroker for an upstream token at all.

## How'd it happen?

1. Deploy v0.33.1 all-in-one (Postgres databroker) with a route annotated `mcp_server: "true"`, no upstream OAuth2 config, upstream a plain-HTTP in-cluster Service.
2. Complete downstream MCP OAuth and confirm requests proxy (200/202).
[...]

PR #6738, opened 2026-09-08
grpc: identify owner and changed options when the outbound connection reloads

wasaga:
## Summary

The outbound gRPC connection cache logged "outbound client connection has changed meaningfully, reloading" without saying which service's connection reloaded or what changed, and logged the same message for the very first dial. Each cache now carries an owner name (controlplane, proxy, authorize, authenticate, databroker-config, registry), the first dial logs "creating outbound client connection", and a reload lists each changed option with its before and after value. The signed JWT key is reported by name only.

## Related issues

- #6730

## User Explanation

Log lines about outbound gRPC connection reloads now name the owning component and the configuration values that changed, which makes connection churn traceable to its cause.

## AI disclosure

Claude Code. The maintainer directed the change; Claude Code implemented it.

## Checklist

- [x] updated unit tests
- [x] disclosed AI usage
- [ ] ready for review

### DIFF
--- a/internal/databroker/config_source.go
+++ b/internal/databroker/config_source.go
@@ -80,7 +80,7 @@ func NewConfigSource(
 		dbConfigs:              map[string]dbConfig{},
 		dbVersionedConfigs:     map[string]dbConfig{},
 		bundle:                 NewConfigBundle(),
-		outboundGRPCConnection: new(grpc.CachedOutboundGRPClientConn),
+		outboundGRPCConnection: &grpc.CachedOutboundGRPClientConn{Name: "databroker-config"},
 	}
 	for _, li := range listeners {
 		src.OnConfigChange(ctx, li)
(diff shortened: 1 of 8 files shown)


internal/databroker/config_source.go
```diff
@@ -80,7 +80,7 @@ func NewConfigSource(
 		dbConfigs:              map[string]dbConfig{},
 		dbVersionedConfigs:     map[string]dbConfig{},
 		bundle:                 NewConfigBundle(),
-		outboundGRPCConnection: new(grpc.CachedOutboundGRPClientConn),
+		outboundGRPCConnection: &grpc.CachedOutboundGRPClientConn{Name: "databroker-config"},
```
kenjenkins, line 83: Should the name here be just "databroker"? It looks like there is no other outbound gRPC connection used in the databroker package.

kenjenkins (approved):

internal/databroker/config_source.go
wasaga: databroker is very common term in logs already i think it's better to keep more unique strings - it won't be coming from the databroker _server_ but rather from databroker-originating config

PR #6762, opened 2026-09-13
fix(mcp): use upstream scheme from route config for PRM URL (#6698)

### LINKED ISSUE #6698: MCP upstream OAuth: upstream URL is built from the downstream `:scheme`, breaking PRM validation behind a TLS-terminating proxy
## Summary

When Pomerium runs behind a proxy that terminates TLS and forwards plaintext (the standard
Kubernetes ingress setup), every **new** MCP upstream OAuth authorization fails with a 502.

`internal/mcp/extproc/server.go` builds the URL used for RFC 9728 protected-resource validation by
taking the **host from route config** but the **scheme from the downstream request's `:scheme`
pseudo-header**. Those describe two unrelated connections: `:scheme` is how the *client* reached
Pomerium, not how Pomerium will reach the upstream. Behind a TLS-terminating ingress `:scheme` is
`http`, so a route with `to: https://upstream.example.com` yields an upstream URL of
`http://upstream.example.com/mcp`.

`checkResourceAllowed` then compares scheme first, so a correct PRM document declaring
`"resource": "https://upstream.example.com"` is rejected:

```
upstream OAuth setup: running discovery: PRM resource "https://upstream.example.com"
does not match upstream server "http://upstream.example.com/mcp"
```

The user gets a 502.

## Why this is easy to miss

**Only new authorizations are affected.** Any user with a stored upstream token never re-enters
discovery, so an endpoint can look completely healthy — pods ready, routes serving, existing users
working normally — while being unable to onboard anyone new. In our case it went unnoticed until
someone needed fresh consent; every monitoring signal we had was green throughout.

## Versions

Reproduced on a `main` build (`575769f`). The relevant code is **byte-identical** in `v0.33.1` and
in current `main`:

- `internal/mcp/extproc/server.go`
- `internal/mcp/upstream_oauth_setup.go`

## Code path

`internal/mcp/extproc/server.go` (~line 132) — note that the code deliberately switches the *host*
from downstream to upstream via route metadata, but not the scheme:

```go
downstreamHost = getHeaderValue(v.RequestHeaders.GetHeaders(), ":authority")
upstreamHost := downstreamHost
if routeCtx != nil && routeCtx.UpstreamHost != "" {
    upstreamHost = routeCtx.UpstreamHost          // host: from route config
}
scheme := getHeaderValue(v.RequestHeaders.GetHeaders(), ":scheme")   // scheme: from the CLIENT
if scheme != "http" && scheme != "https" {
    scheme = "https"
}
u := &url.URL{Scheme: scheme, Host: upstreamHost}
```

`internal/mcp/upstream_oauth_setup.go` (~line 647) — scheme is compared first, before any path
logic, so the mismatch short-circuits:

```go
if !strings.EqualFold(upstream.Scheme, resource.Scheme) {
    return false, nil
}
```

Note the *path* handling is fine — matching is path-prefix and an empty path normalizes to `/`,
which correctly prefixes `/mcp`. Only the scheme fails.

## Reproduction

1. Run Pomerium behind any TLS-terminating reverse proxy that forwards plaintext to it (e.g.
   ingress-nginx with `nginx.ingress.kubernetes.io/backend-protocol: HTTP`).
2. Configure an MCP route with `upstream_oauth2` and **no** static `endpoint` block, so discovery
   runs:

```yaml
- from: https://gateway.example.com
  to: https://upstream.example.com
  name: mcp-upstream
  mcp:
    server:
      upstream_oauth2:
        client_id: "..."
        client_secret: "..."
        scopes: [...]
```

3. Ensure the upstream serves RFC 9728 metadata whose `resource` is its `https://` origin.
4. Connect a client that has no stored upstream token.

**Expected:** discovery succeeds and the user is redirected to the upstream authorization server.
**Actual:** `PRM resource ... does not match upstream server ...`, 502.

## Suggested fix

Take the scheme from the same place as the host — the route's `to:` configuration — falling back to
`:scheme` only where no route context exists.

`internal/mcp/host_info.go` already does exactly this when building `ServerHostInfo.UpstreamURL`:

```go
toURL := p.To[0].URL
upstreamURL := &url.URL{Scheme: toURL.Scheme, Host: toURL.Host, Path: toURL.Path}
```

so the correct value is already available; ext_proc could carry `UpstreamScheme` in route metadata
alongside `UpstreamHost`.

Please **don't** fix this by defaulting the scheme to `https` — plaintext in-cluster upstreams
(`to: http://service.namespace.svc.cluster.local:8080`) are legitimate and would break. The scheme
has to come per-route from config.

This is the same class of bug other proxies have hit behind TLS-terminating ingress; LiteLLM
addresses it with an explicit `PROXY_BASE_URL` rather than inferring from the inbound request.

## Workaround for anyone else hitting this

Pin the upstream endpoints so discovery is skipped entirely (`upstream_oauth_setup.go` takes a
static branch when both are set):

```yaml
        upstream_oauth2:
          client_id: "..."
          client_secret: "..."
          endpoint:
            auth_url: https://provider.example.com/oauth/authorize
            token_url: https://provider.example.com/oauth/token
```

Two notes for anyone applying this:

- The YAML key is `endpoint`, not the protobuf name `oauth2_endpoint`. The file-config path decodes
  via mapstructure (`config/policy.go`), and since viper does not set `ErrorUnused` an incorrect key
  is **silently ignored** — the config loads and the workaround is simply inert.
- Confirm it took effect with the log line
  `ext_proc: using static upstream OAuth2 endpoints (skipping discovery)`.

Happy to open a PR for the fix if the maintainers agree with the approach.

Tyagiquamar:
Fixes #6698


alexandreLamarre:
https://github.com/pomerium/pomerium/issues/6698#issuecomment-5454140519

PR #6752, opened 2026-09-09
[WIP]feat: identity manager v2 mcp bindings

alexandreLamarre:
## Checklist

- [ ] reference any related issues
- [ ] updated unit tests
- [ ] add appropriate label (`enhancement`, `bug`, `breaking`, `dependencies`, `ci`)
- [ ] disclosed AI usage (or wrote "none") per AI_POLICY.md
- [ ] ready for review

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #6724, opened 2026-09-03
[WIP] feat: identity manager v2

alexandreLamarre:
## Checklist

- [ ] reference any related issues
- [ ] updated unit tests
- [ ] add appropriate label (`enhancement`, `bug`, `breaking`, `dependencies`, `ci`)
- [ ] disclosed AI usage (or wrote "none") per AI_POLICY.md
- [ ] ready for review

PR #6770, opened 2026-09-15
idpsession: record lookups, validity helpers, and claim preservation on propagate

wasaga:
## Summary

Adds record accessors for the centralized IdP session, and fixes a case where propagation destroys claims on a bound record.

`GetIDPSession` and `GetBinding` return the record with the gRPC status untouched, so a caller can test for `codes.NotFound` rather than string-matching. `GetValidIDPSession` and `GetActiveBinding` add the state check on top: an IdP session whose state is `UPSTREAM_IDP_SESSION_STATE_INVALID`, or a binding whose state is `REVOKED`, is reported as `codes.NotFound`. A consumer that dereferences a binding to mint a dependent credential therefore has one refusal branch instead of three, and cannot accidentally treat a revoked binding as usable by forgetting a state check. Validity is deliberately state-only — an expired copy of the upstream access token does not make the session unusable, because the identity manager refreshes it out of band. `RevokeBinding` and `RevokeIDPSession` now go through the same lookups instead of inlining `Get` with hardcoded type URLs, and `IssueSession` stamps `accessed_at` alongside `issued_at`.

The second commit fixes claim loss during propagation. `ApplyToSession` merges the IdP's claims into the dependent session rather than assigning them, so a claim the reconciler did not author is meant to survive — `TestApplySessionClaims` pins that at the applier. The guarantee does not hold end to end: `constructPatchedBoundRecord` synthesizes a fresh `session.Session` and `claims` is in the session field mask, so the merge happens against a blank map and the stored claim map is replaced wholesale on every propagation. The reconciler got the merge for free when it patched from the record it had read; now it reads the stored session back and folds its claims in under the IdP's before patching, at a cost of one `Get` per bound session per propagation. A record that cannot be read is left alone — if it is gone, the patch reports it missing and the binding is scheduled for revocation.

`TestPropagatePreservesSessionOwnClaims` drives the real identity manager against a databroker and asserts all three halves: tokens propagate onto the bound session, claims the session owns survive, and a claim the IdP also asserts is refreshed from the IdP.

## Related issues

Stacked on #6752.

## User Explanation

No user-facing change on its own. It removes a silent data-loss path: any claim written onto a session bound to an IdP session was erased the next time that IdP session was propagated.

## AI disclosure

Claude Code, substantially. It wrote the helpers, the reconciler fix and both test files under my direction; I reviewed every line and verified the fix is load-bearing by reverting it and confirming the regression test fails.

## Checklist

- [x] updated unit tests
- [ ] ready for review


internal/idpsession/changeset.go
```diff
@@ -566,6 +571,46 @@ func (a *changeSetApplier) applyChangeSet(ctx context.Context, changeOps recordO
...
+// constructPatchedBoundRecord, so there is nothing to merge into: with "claims"
+// in the session field mask the patch replaces the stored claim map wholesale,
+// dropping any claim the session owns rather than inherits from the IdP. Reading
+// the stored record back restores the merge the applier intends, which is what
+// the reconciler got for free when it patched from the record it had read; the
+// cost is one Get per bound session per propagation.
+//
+// A record that cannot be read is left alone: if it is gone, the patch reports
+// it as missing and the binding is scheduled for revocation.
+func (a *changeSetApplier) preserveStoredSessionClaims(ctx context.Context, records []*databroker.Record) {
```
alexandreLamarre, line 588: oh in the agentic run are we patching additional claims onto the sessions ad-hoc? 

I've really tried to avoid doing extra work in the reconciler, I think I have an idea how we can do this more cleanly (and sometimes avoid this extra lookup)

wasaga: yes in the agentic run we initially merge user claims with additional `act.*` claims that carry identity of the workload executing it. 

that way i.e. you may write a policy that would restrict users accessing some MCPs on their laptops and only thru a centralized workload. or vice versa, etc. 

and currently the reconciler would've wipe those additional claims on update. 

but the logic here needs to be more complicated because we need to actually replace all user claims, leaving those additional claims intact.

alexandreLamarre:
Cherry-picked onto https://github.com/pomerium/pomerium/pull/6724

PR #6771, opened 2026-09-15
mcp: bind client credentials to the centralized IdP session

wasaga:
## Summary

Replaces the per-client `oauth21.MCPRefreshToken` record with the model browser sessions already use: one `session.Session` per consent, plus the `idpsession.Binding` that ties it to the user's `idpsession.IDPSession`. The identity manager becomes the only thing that refreshes an MCP client's upstream credentials, and the refresh token stops being a record.

Consent writes the session and its MCP-protocol binding in one `Put`. The session is issued from the user's IDPSession, so the manager keeps its tokens and claims fresh and deletes it when the binding is revoked or the IdP session dies. Its lifetime is the refresh token's — 365 days, sliding on each refresh — because the manager garbage-collects expired sessions and then revokes their orphaned binding, so the session has to outlive every refresh token minted against it. `refresh_disabled` is left unset, as for a browser session.

The refresh token is now stateless: a sealed session id with the client id as AEAD associated data, plus an `issued_at` equal to the session's `issued_at` at mint. Refresh dereferences Binding to IDPSession to session, refuses a token whose `issued_at` no longer matches the stored session, re-issues the session alone, and never rewrites the binding. That gives rotation with replay detection without a record per token: a replayed token is rejected because the session it names has moved on. Revocation is immediate through either half — revoking the binding or invalidating the IdP session both make every outstanding refresh token for that client undereferenceable.

The access token keeps its own lifetime (`cookie_expire`; the handler option is renamed `WithAccessTokenTTL`) and now references the MCP session and its record version rather than the initiating browser session. The auth-code exchange no longer loads the browser session at all — `AuthorizationRequest.user_id` resolves the IDPSession and `session_id` is reserved — so the `sid` plumbing is gone from `/authorize` and `/connect`. `oauth21.MCPRefreshToken` and its storage methods are deleted, along with the reconciler's special case for that type. The session list shows an MCP grant's real expiry instead of a placeholder.

Behavior change for clients: refresh tokens minted by earlier builds fail with `invalid_grant`, and those clients re-consent once.

Note for the base branch — `internal/authenticateflow TestPersistSession` and `authenticate TestForceLoginResponseWriter` are red on #6752 for reasons unrelated to this PR (stale `idpsession.IDPSessionBinding` type URL and IDPSession id format in the first, a missing `max_age=0` expectation in the second). Not fixed here; happy to send them separately.

## Related issues

Stacked on #6770.

## User Explanation

MCP client grants now appear in the user's session list with their real expiry and can be revoked there, and revoking a grant or ending the user's IdP session immediately invalidates that client's outstanding refresh tokens. Existing MCP clients have to authorize once more after this upgrade.

## AI disclosure

Claude Code, substantially. It implemented the token flows and rewrote the affected tests under my direction; I designed the credential model, reviewed every line, and verified the flows end to end against a live deployment.

## Checklist

- [x] updated unit tests
- [ ] ready for review


internal/mcp/handler_token.go
```diff
@@ -358,257 +240,150 @@ func (srv *Handler) handleRefreshTokenGrant(w http.ResponseWriter, r *http.Reque
...
-		return nil, 0, err
+// refreshMCPSession re-issues the MCP client session a refresh token refers to.
+//
+// The refresh token is bound to the session's Binding: revoking the binding (from
+// the user's session page, or by the identity manager when the IdP session dies)
+// stops refresh at the first check, and the binding is never rewritten here. The
+// session record is then re-issued from the IDPSession with a new issued_at,
+// which is what rotates the refresh token: a token still carrying the previous
+// issued_at is a replayed old generation and is refused.
+func (srv *Handler) refreshMCPSession(ctx context.Context, payload *opaquetoken.Payload, now time.Time) (*session.Session, uint64, error) {
```
alexandreLamarre, line 291: The actual implementation of this refreshMCPSession is hard to follow...
Can we use the session record version? or maybe IDPSession record version if required to be atomically increasing?
Do we have to recreate with the same sessionID? minting new sessions when the old doesn't exist is fine

alexandreLamarre: I feel like there might be a more effective way to issue refresh tokens

alexandreLamarre: I think some of the weirdness here and https://github.com/pomerium/pomerium/pull/6770/changes#r4020718531 can be split out with a dedicated MCPSession
with
```proto
message MCPSession {
   // can be something unique but consistently identifiable
   // like idpSessionID(userID) + clientID(from discovery)
   string id  = 1;
   // fields synced from idpsession
   XXX oauth_token = ...;
   XXX idp_claims = ...;
   //...
   
   // fields not owned by idpsession
   
   // for guarding against refresh token replay without depending on record version directly
   // which seems brittle (a patch from the identity reconciler happening between an mcp client refresh will increment the record version beyond this token check)
   uint64 tokenRefreshGeneration  = ...;
   
   // any claims that need to be set dynamically (by agentic?)
   XXX additional_claims = ...;
}
```

It needs a matching evaluator implementation to be consistent with how session.Session is evaluated today, which might be hard. I think if we implement a helper like `func (s *MCPSession) GetClaims() -> returns merged idp_claims + additional_claims` would work in the places where we check claims in the evaluators maybe without too much rework

internal/oauth21/proto/mcp_refresh_token.proto
alexandreLamarre, line 1: In my own setups, I've noticed that there can be a lot MCPRefreshToken issued for each client.  Along with removing this proto definition we should remove any existing MCPRefreshTokens on startup. We can do it by doing something like `databroker.InitialSync` with `typeURL = type.googleapis.com/oauth21.MCPRefreshToken` are deleted on startup

"Storage leak" initially reported by https://github.com/pomerium/pomerium/pull/6551

PR #6773, opened 2026-09-16
databroker: optimistic locking for Put

wasaga:
## Summary

Currently, databroker's `Put` has no way to detect a concurrent writer. 

Add optional `if_match_version` to `PutRequest`. When set, each record's `version` must equal the stored version, or 0 for a record that does not exist yet. If any record differs, nothing in the batch is written and `Put` fails with `ABORTED` and reason `RECORD_VERSION_MISMATCH`. A batch that names the same record twice is `INVALID_ARGUMENT`, since two assertions about one record cannot both hold. Without the flag, `Put` ignores `version` as before.

The file backend checks the whole batch before touching any record. 

Postgres runs the batch in one transaction and makes each write a compare-and-swap: the upsert and delete carry a `version = $expected` predicate. Postgres evaluates it against the latest committed row after waiting for concurrent writers, so unconditional writers, which take no locks, cannot race a conditional create. The unconditional `Put` path is unchanged.

## Related issues

Prerequisite for #6771 (see https://github.com/pomerium/pomerium/pull/6771#discussion_r4020264427) 

## User Explanation

None. Internal databroker API.

## AI disclosure

Claude Code drafted the Postgres compare-and-swap and the race test from Greptile's findings; I reviewed and directed the design.

## Checklist

- [x] updated unit tests
- [x] add appropriate label (`enhancement`)
- [x] ready for review


calebdoxsey (approved):

pkg/grpc/databroker/errors.go
```diff
@@ -11,15 +12,33 @@ import (
 // known errors
 var (
 	ErrClusterHasNoLeader         = newError(codes.Unavailable, "CLUSTER_HAS_NO_LEADER", "cluster has no leader")
+	ErrDuplicateRecord            = newError(codes.InvalidArgument, "DUPLICATE_RECORD", "a conditional put names the same record more than once")
 	ErrNoClusterLeaderGRPCAddress = newError(codes.FailedPrecondition, "NO_CLUSTER_LEADER_GRPC_ADDRESS", "cluster leader has no grpc_address")
 	ErrNoClusterNodeID            = newError(codes.FailedPrecondition, "NO_CLUSTER_NODE_ID", "databroker_cluster_node_id is required but not set")
 	ErrNoClusterNodes             = newError(codes.FailedPrecondition, "NO_CLUSTER_NODES", "databroker_cluster_nodes is required but not set")
 	ErrNodeIsNotLeader            = newError(codes.FailedPrecondition, "NODE_IS_NOT_LEADER", "request cannot be handled because the node is not the leader")
 	ErrNotInitialized             = newError(codes.Unavailable, "NOT_INITIALIZED", "not initialized")
+	ErrRecordVersionMismatch      = newError(codes.Aborted, reasonRecordVersionMismatch, "record version does not match the stored version")
```
alexandreLamarre: `codes.FailedPrecondition`? In most cases you can't retry a record version mismatch with the same input.

`codes.Aborted` is also shared by clustered databroker resetting a forwarded connection

alexandreLamarre (approved):

PR #6786, opened 2026-09-18
mcp: report token endpoint client authentication failures as invalid_client

wasaga:
## Summary

The MCP token endpoint reported every client authentication failure as `invalid_request`. OAuth 2.1 §3.2.3.1 reserves `invalid_client` for an unknown client or failed client authentication, and MCP clients use that code to decide a persisted dynamic client registration is dead and they must register again.

An unknown client, a missing or expired client secret, a secret mismatch, and an unsupported token endpoint auth method are now answered with `invalid_client`. A client that attempted to authenticate through the `Authorization` header gets a 401 carrying a `WWW-Authenticate` challenge, per §3.2.4. Request parsing failures are unchanged and still return `invalid_request`.

A failure to *look the client up* is deliberately not `invalid_client`. Only a NotFound from the databroker, an invalid client metadata document, or an excluded `client_id` domain mean the client is genuinely unknown; a databroker outage or a client metadata host returning 502 now answers 500. Reporting `invalid_client` there would tell every client in the fleet to discard a good registration and re-consent over a transient blip.

Client authentication failures now carry `error_description`, with the same text whether the client is unknown or its secret was rejected, so the response cannot be used to enumerate registered clients.

Reproduction: connect a client to an MCP route, clear the databroker storage, then reconnect. The client replays its stored `client_id`, the registration is gone, and the `invalid_request` answer gives it no way to tell that re-registering is what it needs to do.

### Client support

`invalid_client` only triggers re-registration in clients implementing the MCP SDK's optional `invalidateCredentials` provider hook. The TypeScript SDK calls it on `invalid_client` and then re-registers; the HTTP status does not matter, since the SDK dispatches on the error body. The MCP Inspector does not implement the hook, so it surfaces the error instead of recovering automatically — clearing its stored credentials there is still manual.

## User Explanation

An MCP client whose dynamic client registration no longer exists server-side — after the databroker storage is cleared, for example — can now detect that its registration is gone and register again, instead of retrying a `client_id` that will never work. Clients that do not implement `invalidateCredentials` will report the error rather than recover on their own.

## AI disclosure

Claude Code (Opus 5) wrote the tests and the fix under my direction, TDD red/green, then ran cleanup and adversarial review passes that I reviewed.

## Checklist

- [x] updated unit tests
- [ ] ready for review


internal/mcp/e2e/mcp_conformance_test.go
```diff
@@ -255,12 +255,35 @@ func runMCPConformance(t *testing.T, mode registrationMode) {
 					"code_verifier": {newCodeVerifier},
 				}
 				resp, result := doTokenRequest(t, params, &[2]string{clientID, "wrong-secret"})
-				assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
-				assert.Equal(t, "invalid_request", result["error"])
+				// OAuth 2.1 3.2.4: the client authenticated via the Authorization
+				// header, so the failure MUST be a 401 carrying WWW-Authenticate.
+				assert.Equal(t, http.StatusUnauthorized, resp.StatusCode)
+				assert.Equal(t, "invalid_client", result["error"])
+				assert.Equal(t, `Basic realm="pomerium"`, resp.Header.Get("WWW-Authenticate"))
```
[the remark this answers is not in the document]
wasaga: Correct that `client_secret_basic` is unverified on this branch, but that is deliberate rather than an oversight: this PR is the first rung of stack #6788 and is scoped to the OAuth error *code*. The secret verification, and un-skipping these two subtests, is #6787 stacked directly on top.

They are separate because they are separate changes for users: this PR is a pure bug fix, while verifying the secret is a breaking change on upgrade (an unspecified `token_endpoint_auth_method` defaults to `client_secret_basic`, so already-registered clients that never sent their secret have been working and will start failing). Splitting them keeps the error-code fix backportable on its own.

The 401 + `WWW-Authenticate` behaviour is still covered here, by an unknown `client_id` presented over Basic, which is reachable without any secret verification. Your suggested assertions match what #6787 asserts, including the 401.

internal/mcp/handler_token.go
```diff
@@ -29,6 +30,71 @@ const (
...
+func (e *clientAuthError) status() (code int, challenge string) {
+	if e.challengeBasic {
+		return http.StatusUnauthorized, `Basic realm="` + tokenEndpointRealm + `"`
+	}
+	return http.StatusBadRequest, ""
+}
+
+// errServerFault marks a token request that could not be resolved because
+// Pomerium itself failed, not because anything was wrong with the request.
+var errServerFault = errors.New("server fault")
```
kenjenkins, line 75: I know RFC 6749 defines a `server_error` error code for some of the OAuth endpoints (e.g. https://www.rfc-editor.org/info/rfc6749/#section-4.1.2.1).

Would that be appropriate to use here?

internal/mcp/handler_token.go
```diff
@@ -49,7 +115,19 @@ func (srv *Handler) Token(w http.ResponseWriter, r *http.Request) {
...
+		var authErr *clientAuthError
+		switch {
+		case errors.As(err, &authErr):
+			code, challenge := authErr.status()
+			if challenge != "" {
+				w.Header().Set("WWW-Authenticate", challenge)
+			}
+			oauth21.ErrorResponseWithDescription(w, code, oauth21.InvalidClient, authErr.description)
+		case errors.Is(err, errServerFault):
+			http.Error(w, "internal error", http.StatusInternalServerError)
```
alexandreLamarre, line 127: Follow up: we do return 500s in a couple places I think where we encounter storage failures or other state failures unrelated to the flows, could make sense to wrap these with the `server_error` oauth code like Ken mentioned

alexandreLamarre (approved):

wasaga:
@kenjenkins @alexandreLamarre, following up on the `server_error` threads ([1](https://github.com/pomerium/pomerium/pull/6786#discussion_r4065686536), [2](https://github.com/pomerium/pomerium/pull/6786#discussion_r4087502655)). I checked the MCP spec (2026-07-28 and draft) and the TypeScript SDK.

The MCP spec defines no authorization-server error codes of its own and defers to OAuth 2.1 draft-13, which only defines `server_error` for the **authorization endpoint** redirect (§4.1.2.1), "because a 500 Internal Server Error HTTP status code cannot be returned to the client via an HTTP redirect." The token endpoint's code list (§3.2.4, same as RFC 6749 §5.2) doesn't include it; there the 500 status is the signal. The TS SDK already maps a non-JSON 500 to `ServerError`. Its reference AS sends `500` with `{"error":"server_error"}` anyway, so we'll do the same for readability, keeping the 500.

Tracked as ENG-4417.

